PR #1, opened 2018-01-29
Merge edgex go monorepo

feclare:


feclare:
It will not be merged now.

PR #3, opened 2018-02-26
Remove export/client not compiling test

feclare:


JPWKU:
recheck

JPWKU:
@feclare Looks like the zeromq issue is resolved.  The amd verify job looks like it has code test failures.  Let me know if you think it's CI related.

feclare:
@JPWKU 
The job name should be edgex-go-arm64-verify, not amd. And right now, because we are using packages that uses .so, it can't be cross compiled from pc. I suggest to remove that check until it could be run in real arm64 hardware or we remove that packages from edgex-go. Anyway, the tests can only be run in real hw.

JPWKU:
@feclare Understood.  The plan as I understood was that we would cross-compile here and test on the hardware the Cavium is offering.  Lets talk about this further via mail list.  I have patched the templates to say arm64 instead of amd, not sure why I typed amd :).

JPWKU:
recheck

feclare:
recheck

PR #5, opened 2018-02-26
Use the same version for all commands

feclare:
Use the same version for all the components. 
Moved the version file to the root.


feclare:
Improved a bit more the version handling. I added a version.go file in the root directory which each package can access to read the version.

Maybe the string "master" could be manually changed to the correct version when a tag is created.

JPWKU:
recheck

JPWKU:
recheck

feclare:
recheck

feclare:
recheck

feclare:
recheck

feclare:
recheck

feclare:
recheck

feclare:
recheck

PR #6, opened 2018-02-27
Fix consul

feclare:
This is https://github.com/edgexfoundry/export-go/pull/26 copied to edgex-go

The only modification are some logs when connecting to consul.


JPWKU:
recheck

feclare:
recheck

jpwhitemn:
recheck

PR #7, opened 2018-02-28
Core Services Move to Mono Repo

jpwhitemn:
updated all core go file packages based on Mono repo, updated makefile for core services, added cmd folders for core services, moved Fede C fixes found in test repo


core/metadata/metadata.go
```diff
@@ -1,55 +1,44 @@
...
- * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
- * or implied. See the License for the specific language governing permissions and limitations under
- * the License.
- *
- * @microservice: core-metadata-go service
- * @author: Spencer Bull & Ryan Comer, Dell
- * @version: 0.5.0
- *******************************************************************************/
-package main
+//
```
feclare: License header should not be changed in this case. At least it should not lost the dell copyright.

cmd/core-command/main.go
```diff
@@ -0,0 +1,60 @@
+/*******************************************************************************
+ * Copyright 2018 Dell Inc.
```
drasko: I propose [SPDX](https://spdx.org/) style headers, to keep consistency across the repo

cmd/core-data/main.go
```diff
@@ -0,0 +1,88 @@
...
+
+var loggingClient logger.LoggingClient
+
+// Read the configuration file and update configuration struct
+func readConfigurationFile(path string) (*data.ConfigurationStruct, error) {
+	var configuration data.ConfigurationStruct
+	// Read the configuration file
+	contents, err := ioutil.ReadFile(path)
+	if err != nil {
+		fmt.Println("Error reading configuration file: " + err.Error())
```
drasko: Avoid printing on the screen - use logger

cmd/core-data/main.go
```diff
@@ -0,0 +1,88 @@
...
+		return nil, err
+	}
+
+	// Decode the configuration as JSON
+	err = json.Unmarshal(contents, &configuration)
+	if err != nil {
+		fmt.Println("Error reading configuration file: " + err.Error())
```
drasko: No stdout prints - log

cmd/core-data/main.go
```diff
@@ -0,0 +1,88 @@
...
+	return &configuration, nil
+}
+
+func main() {
+	start := time.Now()
+
+	// Load configuration data
+	configuration, err := readConfigurationFile(configFile)
+	if err != nil {
+		fmt.Println("Could not read config file: " + configFile)
```
drasko: Should be logged

cmd/core-metadata/main.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func main() {
+	var loggingClient = logger.NewClient(metadata.METADATASERVICENAME, "")
+
+	// Load configuration data
+	configuration, err := readConfigurationFile(CONFIG)
+	if err != nil {
+		loggingClient.Error("Could not read configuration file(" + CONFIG + "): " + err.Error())
+		os.Exit(1)
```
drasko: Do we need to exit like this, or simple `return` from main will do?

core/command/const.go
```diff
@@ -15,7 +15,7 @@
  * @author: Spencer Bull, Dell
  * @version: 0.5.0
  *******************************************************************************/
-package main
+package command
 
 // ConfigurationStruct : Struct used to pase the JSON configuration file
 type ConfigurationStruct struct {
```
drasko: No need to call it `ConfigurationStruct` - we see that it is `struct`. Go idiomatic way of naming variables suggest using very short and simple names.

JPWKU:
recheck

JPWKU:
recheck

core/command/command.go
```diff
@@ -0,0 +1,63 @@
...
+	"time"
+
+	consulclient "github.com/edgexfoundry/edgex-go/support/consul-client"
+	logger "github.com/edgexfoundry/edgex-go/support/logging-client"
+)
+
+var loggingClient logger.LoggingClient
+
+func heartbeat() {
+	for true {
```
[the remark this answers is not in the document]
drasko: Possibly better solution to heartbeat is using `time` Ticker functionality and listening on a channel: https://stackoverflow.com/questions/16466320/is-there-a-way-to-do-repetitive-tasks-at-intervals-in-golang

Something like this: https://github.com/mixer/redutil/blob/master/heartbeat/simple_heart.go#L91-L103

core/command/command.go
```diff
@@ -0,0 +1,63 @@
...
+		ServiceName:    configuration.ServiceName,
+		ServicePort:    configuration.ServerPort,
+		ServiceAddress: configuration.ServiceAddress,
+		CheckAddress:   configuration.ConsulCheckAddress,
+		CheckInterval:  configuration.CheckInterval,
+		ConsulAddress:  configuration.ConsulHost,
+		ConsulPort:     configuration.ConsulPort,
+	})
+
+	if err == nil { // Update configuration data from Consul
```
drasko: Drop the unnecessary comment

core/data/init.go
```diff
@@ -0,0 +1,89 @@
...
+var dbc clients.DBClient
+var loggingClient logger.LoggingClient
+var ep *messaging.EventPublisher
+var mdc metadataclients.DeviceClient
+var msc metadataclients.ServiceClient
+
+// Heartbeat for the data microservice - send a message to logging service
+func heartbeat() {
+	// Loop forever
+	for true {
```
drasko: See comment above for Go idiomatic infinite loops: https://tour.golang.org/flowcontrol/4

core/data/init.go
```diff
@@ -0,0 +1,89 @@
...
+		ServiceName:    configuration.Servicename,
+		ServicePort:    configuration.Serverport,
+		ServiceAddress: configuration.Serviceaddress,
+		CheckAddress:   configuration.Consulcheckaddress,
+		CheckInterval:  configuration.Checkinterval,
+		ConsulAddress:  configuration.Consulhost,
+		ConsulPort:     configuration.Consulport,
+	})
+
+	if err != nil {
```
drasko: Do not use `if - else`. If `err` then log the err and return it (I see that your function should return error)

core/metadata/metadata.go
```diff
@@ -1,55 +1,44 @@
...
-	start := time.Now()
-
-	// Load configuration data
-	err := readConfigurationFile(CONFIG)
-	if err != nil {
-		fmt.Printf("Could not read configuration file(%s): %#v\n", CONFIG, err)
-		os.Exit(1)
+// Heartbeat for the metadata microservice - send a message to logging service
+func heartbeat() {
+	// Loop forever
```
drasko: Comment bloats the code. It's obvious what code does.

core/metadata/metadata.go
```diff
@@ -1,55 +1,44 @@
...
+	for true {
```
drasko: see comment above about Go idiomatic infinite loops.

drasko:
@tsconn23 do you really approve this code, even though I listed a lot of remarks and things that should be corrected?

You are maintainer, it's your call, but usually the code that goes in the repo, stays there :). So I prefer that we filter it out here more meticulously and  improve the code quality.

cmd/core-command/main.go
jpwhitemn: We addressed the format of the header before with LF.  Keeping it the same for now per agreement of community

cmd/core-data/main.go
jpwhitemn: fixed

core/command/const.go
jpwhitemn: leaving for now as it is more cosmetic, but will address as I do more refactoring going forward.

cmd/core-command/main.go
drasko: @jpwhitemn LF proposed SPDX (while I was proposing no comments at all). So we moved to SPDX with `export` services.

This is less important as a remark, I have no objection that comments are kept by a preference Just that I suggested code style aligning.

core/command/const.go
drasko: Agreed.

drasko:
@jpwhitemn as you will not address all the remarks for now, please let us know when your changes are done, and PR is ready.

tsconn23:
@drasko Drasko, Jim asked me directly to approve it so I did

core/metadata/metadata.go
jpwhitemn: Drasko - let us just get the code moved over from the multiple repos before we start trying to clean up the world under one PR!  We agreed we just wanted to get mono repo up with existing code.  Let's get everything working before we burden the PRs with more requests.

JPWKU:
recheck

drasko:
@tsconn23 np, just wanted to check with you guys should I go further with the review. In any case - no functional issues found from my side, just these details, so we can proceed we ca proceed with the merge.

core/clients/coredataclients/coredata-clients.go
```diff
@@ -72,6 +77,19 @@ func decodeValueDescriptorSlice(resp *http.Response) ([]models.ValueDescriptor,
...
 }
 
+// Helper method to decode a device and return the device
+func (v *ValueDescriptorClient) decodeValueDescriptor(resp *http.Response) (models.ValueDescriptor, error) {
+	dec := json.NewDecoder(resp.Body)
+	vdr := models.ValueDescriptor{}
+
+	err := dec.Decode(&vdr)
+	if err != nil {
+		fmt.Println(err)
```
drasko: Printing on the stdout should be avoided. Use logs instead.

drasko: You can also retun immediately here, with `return nil, err` - no need returning a struct below, when we have an error.

core/clients/coredataclients/coredata-clients.go
```diff
@@ -103,5 +121,346 @@ func (v *ValueDescriptorClient) ValueDescriptors() ([]models.ValueDescriptor, er
...
 
-	return decodeValueDescriptorSlice(resp)
+	return v.decodeValueDescriptorSlice(resp)
+}
+
+// Get the value descriptor by id
+func (v *ValueDescriptorClient) ValueDescriptor(id string) (models.ValueDescriptor, error) {
+	req, err := http.NewRequest("GET", v.url+"/"+id, nil)
+	if err != nil {
+		fmt.Println(err)
```
drasko: Should not print on stdout. Use logs.

core/clients/coredataclients/coredata-clients.go
```diff
@@ -103,5 +121,346 @@ func (v *ValueDescriptorClient) ValueDescriptors() ([]models.ValueDescriptor, er
...
+		return models.ValueDescriptor{}, err
```
drasko: No need to return empty struct on error. Return `nil`.

feclare:
recheck

PR #8, opened 2018-02-28
Remove logs from consul client

feclare:
Libraries should not log directly to stdout. As it is always returning the error, caller should manage the error an log as needed.


feclare:
Yes, logs should be created. But not from a library, it reports the error with the needed information and the caller of the functions can create the appropriate log. Using zap in export services and support/logging-client in core services.

drasko:
@feclare what do we do then? We leave these errors without logs for now, or we add Zap logging?

feclare:
Of ll the callers to CheckKeyValuePairs:
core/command/StartCommand.go:51:	if err := consulclient.CheckKeyValuePairs(&configuration, configuration.ApplicationName, strings.Split(configuration.ConsulProfilesActive, ";")); err != nil {
core/data/main.go:96:	if err := consulclient.CheckKeyValuePairs(&configuration, configuration.Servicename, strings.Split(configuration.Consulprofilesactive, ";")); err != nil {
core/metadata/StartMetadata.go:67:	consulclient.CheckKeyValuePairs(&configuration, configuration.ApplicationName, strings.Split(configuration.ConsulProfilesActive, ";"))
export/client/server.go:70:		if err := consulclient.CheckKeyValuePairs(&cfg, applicationName, consulProfiles); err != nil {
export/distro/registrations.go:251:		if err := consulclient.CheckKeyValuePairs(&cfg, applicationName, consulProfiles); err != nil {

Only the call in metadata was not logging the error and I added the log.

JPWKU:
recheck

feclare:
recheck

jpwhitemn:
recheck

PR #9, opened 2018-02-28
[WIP] support-logging implementation

feclare:
Work in progress of support-logging microservice implementation, not should be merged yet.  Created the PR to receive comments about the design and implementation.

Issues currently not solved:

- Need to solve logs from support-logging itself
- No mongo backend
- No limit parameter control
- No res/configuration.json file
- File log storage does not rotate logs nor limit log file size
- Missing makefile entries


support/domain/domain.go
```diff
@@ -19,18 +19,15 @@ package support_domain
...
 	Message       string   `json:"message"`
 	Created       int64    `json:"created"`
 }
 
-// Log Level enum
-type LogLevel string
-
 const (
-	TRACE LogLevel = "TRACE"
+	TRACE = "TRACE"
```
drasko: Align the code (cosmetic remark)

support/logging/file.go
```diff
@@ -0,0 +1,122 @@
...
+	filename string
+	out      io.WriteCloser
+}
+
+func (fl *fileLog) add(le support_domain.LogEntry) {
+	if fl.out == nil {
+		var err error
+		fl.out, err = os.OpenFile(fl.filename, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			//fmt.Println("Error opening log file: ", fl.filename, err)
```
drasko: Remove comments

support/logging/file.go
```diff
@@ -0,0 +1,122 @@
...
+			return
+		}
+	}
+
+	res, err := json.Marshal(le)
+	if err != nil {
+		return
+	}
+	fl.out.Write(res)
+	fl.out.Write([]byte("\n"))
```
drasko: Why do we null-terminate here? If it needs to be added to the string, maybe it is better to do it before marshaling.

support/logging/file.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (fl *fileLog) remove(criteria matchCriteria) int {
+	tmpFilename := fl.filename + rmFileSuffix
+	tmpFile, err := os.OpenFile(tmpFilename, os.O_TRUNC|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		//fmt.Println("Error creating tmp log file: ", tmpFilename, err)
```
drasko: Remove comments

support/logging/file.go
```diff
@@ -0,0 +1,122 @@
...
+		return 0
+	}
+
+	defer os.Remove(tmpFilename)
+
+	f, err := os.Open(fl.filename)
+	if err != nil {
+		fmt.Println("Error opening log file: ", fl.filename, err)
```
drasko: Log, rather than printing to stdout

support/logging/file.go
```diff
@@ -0,0 +1,122 @@
...
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		var le support_domain.LogEntry
+
+		line := scanner.Bytes()
+		err := json.Unmarshal(line, &le)
+		if err == nil {
+			if !criteria.match(le) {
+				tmpFile.Write(line)
+				tmpFile.Write([]byte("\n"))
```
drasko: Again, adding this null-terminator separately looks strange to me.

support/logging/file.go
```diff
@@ -0,0 +1,122 @@
...
+			} else {
+				count += 1
+			}
+		}
+	}
+
+	tmpFile.Close()
+	err = os.Rename(tmpFilename, fl.filename)
+	if err != nil {
+		//fmt.Printf("Error renaming %s to %s: %v", tmpFilename, fl.filename, err)
```
drasko: Remove comments

support/logging/file.go
```diff
@@ -0,0 +1,122 @@
...
+	fl.out.Close()
+	fl.out = nil
+	return count
+}
+
+func (fl *fileLog) find(criteria matchCriteria) []support_domain.LogEntry {
+	var logs []support_domain.LogEntry
+	f, err := os.Open(fl.filename)
+	if err != nil {
+		//fmt.Println("Error opening log file: ", fl.filename, err)
```
drasko: Remove comments

drasko:
As a convention, in the future we can replace `[Not to be merged]` with more standard `[WIP]` and give a signal for maintainers to wait and ask before merge.

support/logging/file.go
feclare: All the comments are logs that I think are interesting to have, but I did not solved yet how to log them. It is one of the open issues.

My idea is to have a private function to generate log entries, to process them like the logs from other services. But is not done yet.

support/logging/file.go
feclare: Several reasons:
- This way the file log could be read by humans, having a line between each json make the file 'somewhat' readable
- The \n is used by bufio scanner as delimiter to parse each entry in the logs to filter it later. I'm sure there is a way to read all the json without any delimiter between them, but it seemed like a good enough solution to start with.

feclare:
@drasko Thanks for comments.

JPWKU:
recheck

jpwhitemn:
recheck

PR #12, opened 2018-03-01
Fixes #10, added docker creation for core services

### LINKED ISSUE #10: Need to update the Makefile or build purposes
The Makefile needs an update to build/test all of the services in the edgex-go repo and associated docker images.  This is the only way to ensure we have repo wide coverage.
In addition, if we ever add another service there will need to be a corresponding addition to the Makefile to build/test it.

jpwhitemn:
Signed-off-by: edgex <jpwhite_mn@yahoo.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,15 @@ prepare:
 
 docker: $(DOCKERS)
 
+docker_core_metadata:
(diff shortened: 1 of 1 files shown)


feclare:
This is correct and should be merged.
But it doesn't build the dockers, it only added rules to generete them.

Tomorrow will fix the docker files and will create new PR for them.

PR #14, opened 2018-03-02
Fix #13 changed content type to text plain for export services ping r…

### LINKED ISSUE #13: Wrong content type for export services ping API
Hitting the export client or servier PING API (192.168.99.100:48071/api/v1/ping) results in the browser asking if you want to save the response versus just display it. Check the content type returned - and see how core data does this.

jpwhitemn:
…esponse.  Also added make clean to Makefile

Signed-off-by: edgex <jpwhite_mn@yahoo.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,13 +5,16 @@
 #
 
 
-.PHONY: build test docker
+.PHONY: build clean test docker
 
 GO=CGO_ENABLED=0 go
 GOCGO=CGO_ENABLED=1 go
 
 DOCKERS=
 .PHONY: $(DOCKERS)
+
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -37,6 +40,23 @@ cmd/export-client/export-client:
 cmd/export-distro/export-distro:
 	$(GOCGO) build $(GOFLAGS) -o $@ ./cmd/export-distro
 
+clean: $(EXECUTABLES)
+
```
feclare: Remove EXECUTABLES and clean_XXXXX rules from the Makefile and make clean something like
clean:
    rm -f $(MICROSERVICES)

PR #15, opened 2018-03-02
Configurable Logging -- Remote vs Local

tsconn23:
In core Go services (data, metadata, command) there is no configuration option to allow logging to occur locally versus going to the logging service. There should be a config option that allows the remote logging service to be bypassed entirely if desired.


tsconn23:
One of the five commits isn't signed and it doesn't appear there's an easy way for me to fix this at this point. Will address in the morning.

PR #16, opened 2018-03-02
Fixing panic when config file not found in core-data

feclare:
Also removed double logging of the error.

### DIFF
--- a/core/data/init.go
+++ b/core/data/init.go
@@ -28,6 +28,10 @@ import (
 	"github.com/edgexfoundry/edgex-go/support/logging-client"
 )
 
+const (
+	ServiceName string = "core-data"
+)
+
 // Global variables
 var dbc clients.DBClient
 var loggingClient logger.LoggingClient
--- a/cmd/core-data/main.go
+++ b/cmd/core-data/main.go
@@ -19,7 +19,6 @@ package main
 
 import (
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"net/http"
 	"strconv"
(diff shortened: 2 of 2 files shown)


tsconn23:
I addressed this as well in all 3 core services with PR #17. Note that in PR #17 rather than removing the logging altogether, logs are written to the console if a config file can't be found. Also, "file not found" error isn't logged twice, but delineated between file being missing and json not being parse-able.

https://github.com/edgexfoundry/edgex-go/pull/17

Would suggest closing this PR and going with the one above.

core/data/init.go
```diff
@@ -28,6 +28,10 @@ import (
 	"github.com/edgexfoundry/edgex-go/support/logging-client"
 )
 
+const (
```
tsconn23: This would be better served being added to core/data/const.go. That's where I put it in https://github.com/edgexfoundry/edgex-go/pull/17

feclare: I agree

feclare:
Did you tried to run without a configuration.json file in your branch?
This is the result:

```
feclare@serrano:~/projects/gopath/src/github.com/edgexfoundry/edgex-go$ go run cmd/core-data/main.go 
Error reading configuration file: open ./res/configuration.json: no such file or directory
ERROR: 2018/03/02 18:06:26 logger.go:135: Could not read configuration file(./res/configuration.json): open ./res/configuration.json: no such file or directory
```
There is double logging of the same problem. The printf calls inside the function can be removed as the error is always printed when there is an error.

tsconn23:
Extra logging was removed from PR 17 per Fede's comment above. Therefore closing this PR.

PR #17, opened 2018-03-02
Adding configurable logging -- remote vs local file.

tsconn23:

In core Go services (data, metadata, command) there is no configuration option to allow logging to occur locally versus going to the logging service. There should be a config option that allows the remote logging service to be bypassed entirely if desired.


cmd/core-command/main.go
```diff
@@ -23,37 +23,55 @@ import (
...
 }
 
-func readConfigurationFile(path string) error {
+func readConfigurationFile(path string) (*command.ConfigurationStruct, error) {
+	var configuration command.ConfigurationStruct
 	// Read the configuration file
 	contents, err := ioutil.ReadFile(path)
 	if err != nil {
-		return err
+		fmt.Println("Error reading configuration file: " + err.Error())
```
feclare: Remove printf. It is already logged in line 41

cmd/core-command/main.go
```diff
@@ -23,37 +23,55 @@ import (
...
+		return nil, err
 	}
 
 	// Decode the configuration as JSON
-	return json.Unmarshal(contents, &configuration)
+	err = json.Unmarshal(contents, &configuration)
+	if err != nil {
+		fmt.Println("Error parsing configuration file: " + err.Error())
```
feclare: Remove printf. It is already logged in line 41

cmd/core-data/main.go
```diff
@@ -44,7 +44,7 @@ func readConfigurationFile(path string) (*data.ConfigurationStruct, error) {
 	// Read the configuration file
 	contents, err := ioutil.ReadFile(path)
 	if err != nil {
-		loggingClient.Error("Could not read configuration file(" + configFile + "): " + err.Error())
+		fmt.Println("Error reading configuration file: " + err.Error())
```
feclare: Remove printf. It is already logged in line 76

cmd/core-metadata/main.go
```diff
@@ -23,44 +23,55 @@ import (
...
 }
 
 // Read the configuration file and
 func readConfigurationFile(path string) (*metadata.ConfigurationStruct, error) {
 	var configuration metadata.ConfigurationStruct
-
 	// Read the configuration file
 	contents, err := ioutil.ReadFile(path)
 	if err != nil {
+		fmt.Println("Error reading configuration file: " + err.Error())
```
feclare: Remove printf. It is already logged in line 40

cmd/core-metadata/main.go
```diff
@@ -23,44 +23,55 @@ import (
...
 		return nil, err
 	}
 
 	// Decode the configuration as JSON
 	err = json.Unmarshal(contents, &configuration)
 	if err != nil {
+		fmt.Println("Error reading configuration file: " + err.Error())
```
feclare: Remove printf. It is already logged in line 40

tsconn23: Actually, no. This is different but I will change the log message accordingly. This is actually an error with parsing the JSON after the file has been found.

cmd/core-command/main.go
tsconn23: This is a different error message. The file has been loaded but JSON cannot be parsed.

feclare:
I think than better than a flag of remote logging and a parameter that changes its meaning depending on the flag it is better to have two different parameters. One filepath for local logging and a url for remote logging. If filepath is empty then the local logging will be skipped. If remoteUrl is empty remote logging will be skipped.

tsconn23:
Per Fede's input I modified the verbiage in the error statements being logged from the main.go files to clarify the specific reason why configuration could not be loaded -- missing file versus inability to parse JSON.

W/R/T the comment "One filepath for local logging and a url for remote logging. If filepath is empty then the local logging will be skipped. If remoteUrl is empty remote logging will be skipped."

That would be fine if both of the values were not specified in the configuration but as it is now, both the file path and the remote url are provided necessitating a toggle. As I said in a previous discussion, we can refactor the NewClient() parameter to be a "LoggingConfiguration" interface but because of the presence of both values, you still need to boolean EnableRemoteLogging in order to make a decision about where the content will go.

feclare:
I will not insist on the flag. But I don't understand why you don't want to log to file and to remote at the same time.

feclare:
@tsconn23 You did change the message, but there is still two logs for the same problem. I don't think they are needed.
This is what happens with your current branch:
```
feclare@serrano:~/projects/gopath/src/github.com/edgexfoundry/edgex-go$ go run cmd/core-data/main.go 
Error reading configuration file: open ./res/configuration.json: no such file or directory
ERROR: 2018/03/02 18:37:03 logger.go:135: Could not read configuration file(./res/configuration.json): open ./res/configuration.json: no such file or directory
```
The first log does not supply any information missing in the second log. I still think logs from inside the function should be removed.

tsconn23:
Logging statements considered redundant have been removed. Please approve.

PR #27, opened 2018-03-02
Add launcher script

drasko:
Adds script to launch all EdgeX Go services

Signed-off-by: Drasko DRASKOVIC <drasko.draskovic@gmail.com>


bin/edgex-launch.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
feclare: Add license header

bin/edgex-launch.sh
```diff
@@ -0,0 +1,61 @@
...
+
+###
+# Launches all EdgeX Go binaries (must be previously built).
+#
+# Expects that Consul is already installed and running.
```
feclare: Add comment about mongo, that it is also needed to be running

bin/edgex-launch.sh
drasko: Added

drasko:
@janko-isidorovic this one can be merged.

Issue #20, opened 2018-03-02
Package import links error


drasko:
```
drasko@Marx:~/go/src/github.com/edgexfoundry/edgex-go$ glide install
[INFO]	Lock file (glide.lock) does not exist. Performing update.
[INFO]	Downloading dependencies. Please wait...
[INFO]	--> Fetching updates for github.com/eclipse/paho.mqtt.golang
[INFO]	--> Fetching updates for gopkg.in/mgo.v2
[INFO]	--> Fetching updates for go.uber.org/zap
[INFO]	--> Fetching updates for github.com/pebbe/zmq4
[INFO]	--> Fetching updates for github.com/hashicorp/consul
[INFO]	--> Fetching updates for github.com/go-zoo/bone
[INFO]	--> Fetching updates for github.com/robfig/cron
[INFO]	--> Fetching updates for github.com/gorilla/mux
[INFO]	--> Fetching updates for gopkg.in/yaml.v2
[INFO]	Resolving imports
[INFO]	--> Fetching updates for go.uber.org/atomic
[INFO]	--> Detected semantic version. Setting version for go.uber.org/atomic to v1.3.1
[INFO]	--> Fetching updates for go.uber.org/multierr
[INFO]	--> Detected semantic version. Setting version for go.uber.org/multierr to v1.1.0
[ERROR]	Error scanning github.com/edgexfoundry/core-domain-go/models: cannot find package "." in:
	/home/drasko/.glide/cache/src/https-github.com-edgexfoundry-core-domain-go/models
[INFO]	--> Fetching updates for github.com/gorilla/context
[INFO]	--> Fetching updates for golang.org/x/net
[INFO]	--> Fetching updates for github.com/hashicorp/go-cleanhttp
[INFO]	--> Fetching updates for github.com/hashicorp/go-rootcerts
[INFO]	--> Fetching updates for github.com/hashicorp/serf
[INFO]	--> Fetching updates for github.com/mitchellh/go-homedir
[INFO]	--> Fetching updates for github.com/armon/go-metrics
[INFO]	--> Fetching updates for github.com/hashicorp/go-immutable-radix
[INFO]	--> Fetching updates for github.com/hashicorp/golang-lru
[ERROR]	Failed to retrieve a list of dependencies: Error resolving imports
```

drasko:
Should be fixed by https://github.com/edgexfoundry/edgex-go/pull/21

PR #28, opened 2018-03-05
Add new target to compile all go code

feclare:
Modify make build to compile all the go code in the project instead of compiling the microservices. If there is some code not used by the microservices it was not tested in CI.

For example, issue #20 was caused by #11 and it should be detected by CI.


feclare:
recheck

drasko:
@janko-isidorovic you can merge this one.

PR #30, opened 2018-03-05
Adding go vet to test makefile target

feclare:
Also fixing the issues reported.

### DIFF
--- a/core/clients/metadataclients/metadata-clients.go
+++ b/core/clients/metadataclients/metadata-clients.go
@@ -22,11 +22,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/edgexfoundry/edgex-go/core/domain/models"
 	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strconv"
+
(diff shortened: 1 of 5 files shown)


core/clients/metadataclients/metadata-clients.go
```diff
@@ -193,7 +194,7 @@ func (a *AddressableClient) AddressableForName(name string) (models.Addressable,
 	}
 	defer resp.Body.Close()
 	if err != nil {
-		fmt.Println("AddressableForName makeRequest failed: %s", err)
+		fmt.Printf("AddressableForName makeRequest failed: %v\n", err)
```
drasko: Do we need hese printf lines? I would like to remove all of these in the production and use only logs.

feclare: I agree with you. But in this case I just wanted to remove the vet warnings.

We will need to cleanup all the codebase logs.

drasko: Agreed on this.

drasko:
LGTM

PR #31, opened 2018-03-06
Modify dockerfiles to compile with monorepo

feclare:
Not working:
- version is not set in core-data docker
- arm64 not tested


drasko:
Great job here @feclare! I really like this as an example of code integrity and unification that benefits from monorepo.

Code looks more tight and higher quality, and this is great to see and gives confidence to EdgeX users and developers.

PR #36, opened 2018-03-07
Refactoring core-data

tsconn23:
Service cleanup -- restructuring of core-data initialization, including cmd line flag for Consul(Y/N) and Profile (default vs Docker). Also used service name constants in log init message instead of hard-coded strings.

### DIFF
--- a/cmd/core-data/main.go
+++ b/cmd/core-data/main.go
@@ -19,6 +19,8 @@ package main
 
 import (
 	"encoding/json"
+	"flag"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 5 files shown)


cmd/core-data/main.go
```diff
@@ -29,42 +31,91 @@ import (
...
+
+func logBeforeTermination(err error) {
+	loggingClient = logger.NewClient(data.COREDATASERVICENAME, false, "")
+	loggingClient.Error(err.Error())
+}
+
+func determineConfigFile(profile string) string {
+	switch profile {
+		case "docker":
+			return "./res/configuration-docker.json"
```
drasko: Please do not hard code strings. Use `const` instead.

cmd/core-data/main.go
```diff
@@ -29,42 +31,91 @@ import (
...
 	// Time it took to start service
 	loggingClient.Info("Service started in: "+time.Since(start).String(), "")
 	loggingClient.Info("Listening on port: " + strconv.Itoa(configuration.Serverport))
-
 	loggingClient.Error(http.ListenAndServe(":"+strconv.Itoa(configuration.Serverport), r).Error())
 }
 
+func startHeartbeat(msg string, interval int) {
+	chBeats := make(chan string)
+	go data.Heartbeat(msg, interval, chBeats)
```
feclare: Why creating another goroutine just for sending the log?

PR #39, opened 2018-03-08
add a new item into .gitignore file to filter glide.lock

yanghua:
glide command will create a `glide.lock` we should filter it from the project source code, enviroment :  
OS: Mac os x 10.13.2
IDE: GoLand
package tool : homebrew and so on...


yanghua:
@drasko I add the DCO sign-off.

PR #40, opened 2018-03-08
improve README document about installing pkgconfig on mac os x

yanghua:
add some supplementary for the README document about installing `pkgconfig`, if we don't install it, the `make build` command would occur a error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,14 @@ For example, in the case of Debian Linux system you can:
 sudo apt-get install libzmq3-dev
 ```
 
+And if you want to build project from source code, make sure that you have installed "pkg-config" tool before you build project.
+
+For Mac os x, you could use [Homebrew](https://docs.brew.sh/) to install it:
+
+```
+brew install pkgconfig
+```
+
 To fetch the code and compile the microservices execute:
 
 ```


README.md
```diff
@@ -17,6 +17,14 @@ For example, in the case of Debian Linux system you can:
 sudo apt-get install libzmq3-dev
 ```
 
+And if you want to build project from source code, make sure that you have installed "pkd-config" tool before you build project.
```
feclare: There is a typo in "pkd-config"

tonyespy:
@yanghua 

In general, it's better to amend a commit and then force push to your branch, vs. submitting a commit with a generic message like "fixing a typo issue".

yanghua:
@tonyespy OK, thanks for your suggestion. I will accept this idea next time.

feclare:
@yanghua The commits are missing the DCO:
https://wiki.edgexfoundry.org/display/FA/Contributor%27s+Guide+-+Go+Lang#Contributor'sGuide-GoLang-SubmissionDCO

yanghua:
@feclare I havee updated. Does it compliance with requirements?

feclare:
@yanghua 
The commits are what needs to have the signoff, not the comments.

yanghua:
@feclare sorry, I have no idea about the DCO before, I just changed, it is right?

feclare:
You can find some more information in:
https://wiki.edgexfoundry.org/display/FA/Committing+Code+Guidelines

yanghua:
@feclare thanks a lot, should I do something others for this PR?

drasko:
I would avoid putting unnecessary and redundant info in this README. Not having basic development tools is a problem of a developer. We can not list all the possible OS dependencies in the README, only necessary minimum to get developers going.

If really needed, this kind of information belongs in the technical documentation, not README.

I am against this PR.

tonyespy:
I don't think mentioning that pkg-config (the Ubuntu package name) is redundant or useless, as it's non-obvious that its required to build this project.  Ideally, we'd use auto tools or similar with a configure script to deal with this kind of dependency, but in the long run the best solution is to just get rid of the C code as we've all been discussing.  In this case a simple note that it's needed (vs. install instructions for a single OS) would work for me.

yanghua:
Hi @drasko and @tonyespy , I use mac os x as work environment for a long time, it really would occur a error if we do not install pkg-config before invoke `make build`. In view of Edgex grows at a faster rate and go lang is also not more mature than other language, such as java. I think we should make other developers easy to practice and install the project. We should gather the errors and failure at all kinds of os. In future, we can split it into a Individual section named "install" or a new install document.

yanghua:
Hi guys, can we create a new installation file(named `install.md`) which  describes the full installation steps for the mainstream OS, I can claim mac os x and someone can describe others, such as Windows. How about this idea?

drasko:
No, documentation will not go into sources repo. There is a separate repo for doc.

Here we put just a bare minimum in README

drasko:
Can be addressed after refactoring README and publishing technical documentation. Closing for now.

PR #41, opened 2018-03-09
Standardize Heartbeat Implementation in Core Services

tsconn23:
In order to standardize the heartbeat implementation in the core services:

1.) Changed LoggingClient to an interface
2.) What was previously known as "LoggingClient" is now EdgeXLogger impl type
3.) Moved heartbeat logic invoked when service starts into support/heartbeat package
4.) Implemented unit test for above
5.) Standardized configuration property names across all 3 services.


drasko:
Good work here @tsconn23 

I am pleased to see that code starts to get this "Go" smell - simple, easy to read, variable naming respecting Go style, etc... Really nice.

support/logging-client/logger.go
```diff
@@ -32,7 +32,15 @@ import (
...
-type LoggingClient struct {
+type LoggingClient  interface {
+	Debug(msg string, labels ...string) error
+	Error(msg string, labels ...string) error
+	Info(msg string, labels ...string) error
+	Trace(msg string, labels ...string) error
+	Warn(msg string, labels ...string) error
+}
+
+type EdgeXLogger struct {
```
feclare: Make EdgeXLogger private

support/logging-client/logger.go
```diff
@@ -32,7 +32,15 @@ import (
 	"strings"
 )
 
-type LoggingClient struct {
+type LoggingClient  interface {
+	Debug(msg string, labels ...string) error
```
feclare: Why this definition in the interface? Why not copying the real functions signature (logLevel string, msg string, labels []string)?

support/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+		close(chBeats)
+	}()
+}
+
+// Executes the basic heartbeat for all servicves. Writes entries to the supplied channel.
+func sendBeats(heartbeatMsg string, interval int, beats chan<- string) {
+	// Loop forever
+	for true {
+		beats <- heartbeatMsg
```
feclare: Why having two different goroutines and a channel when with only one gorutine it works?

Just replacing the:
beats <- heartbeatMsg
to logger.Info

tsconn23: I'll look at this. I used the two goroutines in the previous version in order to remove the Logging Client parameter and brought the code over into the new PR. Since the logic is now in one heartbeat package, it can probably be slimmed to one goroutine.

support/logging-client/logger.go
tsconn23: Because the method names ARE the logLevel. The interface provides a discrete contract for what those log levels are so that calling logic doesn't pass in an unsupported value.

Anyway, calling logic throughout the application doesn't call the LoggingClient as
loggingClient.log("Info",fmt.Sprintf("Starting %s %s ", command.COMMANDSERVICENAME, edgex.Version))

"log" is an internal method on the EdgeXLogger implementation class. Everywhere the use of LoggingClient is done via:
loggingClient.Info(fmt.Sprintf("Starting %s %s ", command.COMMANDSERVICENAME, edgex.Version))

Thus the interface is defined accordingly.

feclare: Sorry,  you are right. I misunderstand the code because of the 'labels ...string'.  Still too much go to learn.

PR #42, opened 2018-03-09
Added event validation. Enabled by configuration parameter Validatecheck

chenchix:
I opened a discussion (email list and issue https://github.com/edgexfoundry/export-go/issues/19)  about this commit and the place to put the validation, as nobody commented, I understand that this place is ok.


tsconn23:
If adding a new entry to the configuration.json files, doesn't this also need a new entry in the config-seed service to be found in Consul? Otherwise when run with Consul "Validatecheck" will always be false (verified locally).

Also, needs tests on the new functions in core/data/validate.go

tsconn23:
FYI, PR is failing to build on this line:
# github.com/edgexfoundry/edgex-go/core/data
core/data/validate_test.go:233:16: undefined: isValidValueDescriptor_private

chenchix:
Added validate.go missing in the last commit, and some changes suggested by @feclare

PR #47, opened 2018-03-10
add some ignore items to filter built binary and log files

yanghua:
current if user build and the project will create some binary and log files we should filter them.

Signed-off-by: vinoyang <vinoyang@tencent.com>


yanghua:
@drasko would you please review this PR? Thanks!

PR #49, opened 2018-03-13
Docker improvements

feclare:
Added docker generation to docker file.
Added script to launch go microservices dockers.

@tsconn23 @jpwhitemn 
I changed how core-data and export-distro are compiled in the dockerfile. Changed it to be compiled with the makefile and added the zeromq library to the final docker. I don't know why you were linking static the zeromq before, but now it is dynamic. Any know issue?


docker/docker-compose.yml
```diff
@@ -0,0 +1,371 @@
...
+      - log-data:/edgex/logs
+      - consul-config:/consul/config
+      - consul-data:/consul/data
+    depends_on:
+      - data
+      - command
+      - scheduler
+      - notifications
+
+#   device-bluetooth:
```
drasko: I do not like the idea of dragging this dead code around. It's in the Git, if someone needs it later, can be always checked.

feclare: I will remove them.

PR #51, opened 2018-03-13
#50 add test for support notifications client

yanghua:

hi @drasko and @feclare , I add a test for support notification client because I find this call : 
https://github.com/edgexfoundry/edgex-go/blob/master/support/notifications-client/client.go#L91

will not trigger a real remote request if it run in a new `goroutine`.

Please review this test, I will try to refactor the notification client and this test code. Thanks!

### DIFF
--- a/support/notifications-client/client_test.go
+++ b/support/notifications-client/client_test.go
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2017 Tencent
+// Copyright (c) 2018 Tencent
 //
 // SPDX-License-Identifier: Apache-2.0
 //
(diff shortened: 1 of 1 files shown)


support/notifications-client/client_test.go
```diff
@@ -0,0 +1,112 @@
+//
+// Copyright (c) 2017 Tencent
```
drasko: We're in 2018.

yanghua: I will change this, I just follow the pre-existing style before.

yanghua: @drasko changed it.

yanghua:
hi @tsconn23 , if you have time, would you please merge this PR? Because, I will refactor the notification client and use this test case to verify the function. Thanks!

PR #60, opened 2018-03-15
support-logging improvements

feclare:
Adding:
- docker
- Makefile microservice compilation
- Added to launch scripts


feclare:
recheck

feclare:
recheck

Issue #57, opened 2018-03-14
Implement support scheduler server


yanghua:
assignee : @yanghua

jpwhitemn:
On the roadmap to do - hopefully by California release.

yanghua:
Hi @Jim White please be assured that I am doing this job. Thanks. On 2018-03-22 07:39 , Jim White Wrote: On the roadmap to do - hopefully by California release. — You are receiving this because you authored the thread. Reply to this email directly, view it on GitHub, or mute the thread.

jpwhitemn:
Hi Yanghua - that would be great if you (and any team you have) would like to take the work to move the Scheduler service from Java to Go.  What would your time line be for this work?

yanghua:
hi @jpwhitemn , As planned, this weekend I will open a PR about this issue, however it's a big PR, the code review may take a longer time. Hopefully, next weekend this issue could be closed.

yanghua:
@jpwhitemn  is this time plan suitable？

drasko:
@yanghua if this issue is in progress, can you please move the card in "In Progress" state here: https://github.com/edgexfoundry/edgex-go/projects/2?add_cards_query=is%3Aopen

yanghua:
@drasko I tried. But I don't konw how to do this operation?  Does this operation need some permission?

drasko:
@yanghua yes, probably. Np, I did it.

tsconn23:
@jpwhitemn Should we move this tag out to "delhi"?

yanghua:
@tsconn23  I think it is not belongs California Milestone so you can mark as "delhi".

PR #61, opened 2018-03-15
refactor support notifications client

yanghua:


yanghua:
hi @drasko please review this PR , thanks!

yanghua:
@drasko thanks, I just find a build error. so I make a empty commit to trigger rebuild.

drasko:
I think `recheck` is a magic word for Jenkins - @JPWKU can explain this more

feclare:
The edgex-go-arm-verify build error is not caused by your code. The board doing the arm64 compilations is down today. Sorry about that. 

But the errors in edgex-go-verify are.

You can reproduce the jenkins compilation in your environment with 
```
make test  
```
in edge-go/

feclare:
@yanghua 
You are changing the semantics of the function, running the http petition in the same thread instead of creating a new one. Did you check if this change breaks metadata?

yanghua:
@feclare You are right, there are two errors. The reason is I changed the notification client's `RemoteUrl` field and replaced with two fields : `NotificationServiceHost`, `NotificationServicePort`. I will fix this issue. thanks!

yanghua:
cc @feclare Have fixed the errors.

feclare:
@yanghua 
You changed the semantics of the function, doing the http request in the same thread, and I don't know metadata enough to evaluate if this change could be problematic for metadata.

yanghua:
@feclare Yes, I changed this and removed the `go` keyword, the reason is the `go` keyword will run the function in a new goroutine(means async invoke), then the method return directly. This style will not trigger expected http request. I have tested the old code. I found this behavior when I wrote the scheduler's client.

yanghua:
hi @feclare , about this issue you could run my notification client's test case to verify I said before. The old code could not send a request to remote notification's microservice, because it did not block until the goroutine finished. Actually, it's a client (not a server) , there is no need to send just one request asynchronously.

drasko:
@yanghua please resolve the conflicts

yanghua:
@drasko  OK, please wait a moment, thank you.

yanghua:
@drasko please hold on a few days(at most a week). I am working for #57 , the code is confused in my local branch about this issue currently. So please wait, thanks.

JPWKU:
@yanghua this will need to be rebased against the master branch.

yanghua:
cc @drasko

cmd/core-metadata/res/configuration-docker.json
```diff
@@ -0,0 +1,37 @@
+{
```
feclare: Remove old json configurations files, toml files are used now.

cmd/support-scheduler/logs/edgex-support-scheduler.log
```diff
@@ -0,0 +1,6 @@
+INFO: 2018/04/02 23:54:06 logger.go:96: bypassing Consul configuration...
```
feclare: Remove log file

support/notifications-client/client.go
```diff
@@ -68,38 +85,42 @@ type Notification struct {
...
+	req, err := http.NewRequest(HttpPostMethod, url, binaryReqBody)
+	req.Header.Add(ContentType, ContentTypeJsonVal)
+
 	if err != nil {
-		fmt.Println(err.Error())
 		return err
 	}
 
-	// Asynchronous call
-	go makeRequest(client, req)
```
feclare: You are changing the call to be made asynchronous. This will change the performance of the function on the caller, to complete mostly in no time to make a complete http request in place.
This should not be approved without the approval from some of the core microservices developers.

yanghua: @feclare It's not about performance, the real thing is : this code style in my test could not trigger a real http request to server. You can run my test with http mock server to see the result. This file shows the author is `Ryan Comer, Dell`, who can contact with he? The code style can not get http error, just return a `nil`. This function has no test, has wrong method name. @drasko can you arbitrate this or run my test case?

support/notifications-client/client.go
```diff
@@ -68,38 +82,42 @@ type Notification struct {
...
-
-	return nil
+	return sendRequest(client, req)
 }
 
 // Function to actually make the HTTP request
-func makeRequest(client *http.Client, req *http.Request) {
+func sendRequest(client *http.Client, req *http.Request) error {
 	resp, err := client.Do(req)
 	if err == nil {
```
drasko: Strange construct here - rather check `if err != nil`

support/notifications-client/client.go
```diff
@@ -68,38 +82,42 @@ type Notification struct {
...
-	// Asynchronous call
-	go makeRequest(client, req)
-
-	return nil
+	return sendRequest(client, req)
 }
 
 // Function to actually make the HTTP request
-func makeRequest(client *http.Client, req *http.Request) {
+func sendRequest(client *http.Client, req *http.Request) error {
```
drasko: Returning an error is a good thing here.

support/notifications-client/client.go
drasko: No sense in spawning a goroutine if sucess/failure and/or response can not be communicated back to the calling code. I would go with the synchronous call here.

I do not think that `makeRequest` is a wrong name, however (I have seen this pattern Go code already - example InfluxDB's Telegraf: https://github.com/influxdata/telegraf/blob/master/plugins/inputs/httpjson/httpjson.go#L64).

drasko: It is possible that author wanted to achieve this: https://medium.com/dev-bits/making-concurrent-http-requests-in-go-programming-language-823b51bb1dc2 - but without communicating result via channel, current call does not make much sense.

yanghua: @drasko 

* to your first comment:
 Actually, I want to say the original `ReceiveNotification` is a wrong name (not `makeRequest`). I just want to indicate the exists and original code is not completely right.

* to your second comment:
I totally agree with you, actually the old code style did not use a channel to wait the http request finish and can not get the response.

support/notifications-client/client.go
yanghua: will change this mode by follow the offical : https://golang.org/pkg/net/http/

yanghua: cc @drasko changed this.

core/metadata/const.go
```diff
@@ -56,7 +56,8 @@ type ConfigurationStruct struct {
 	NotificationSender                  string
 	NotificationDescription             string
 	NotificationLabel                   string
-	SupportNotificationsNotificationURL string
+	SupportNotificationsHost			string
```
drasko: Please align

yanghua: @drasko done.

yanghua:
cc @tsconn23 this PR could be merged.

feclare:
@yanghua The binary file still exists in the history, taking space. I recommend you to squash all your changes to only one commit.
I also recommend you to read this edgex guide: https://wiki.edgexfoundry.org/display/FA/Committing+Code+Guidelines
It is not useful to have 8 commits with the same commit msg ` refactor support notifications clien`

yanghua:
@feclare OK, I will squash there commits into one commit

yanghua:
cc @feclare squash these commits

drasko:
@JPWKU I think it would be useful to force commit squash in GitHub, so that we can have one commit per PR (i.e. let GitHub squash all commits before merge).

yanghua:
@feclare Can you make a test for the original code? Can you believe the fact of testing? Can you understand go routine more deeply? Can you not just consider performance (if the old code can not waiting for the http response, it means nothing!)? Can you check the support notification java client which also send a sync HTTP request? Can you???

I suggest you don't pay attention to this PR, if you'd rather not test that method. Thanks!

yanghua:
@drasko , it seems @feclare is not online recently. I'd like to merge this PR into master branch, it contains a `client_test.go` file could make the core question we discussed more clearly.

feclare:
@yanghua @drasko It is ok for me to merge this PR. You know I don't agree with the change of performance but I understand it is great to have new tests and it could be fixed later.

@yanghua Of course you can test a function with the original semantics. You can't test the original code, but with some refactoring you would be able to test it. 

And @yanghua, thanks with all the patience and the changes you made. Your collaborations are really appreciated.

tsconn23:
@feclare @drasko @jpwhitemn Hi Fede -- Yes, please merge this. There are some things in here Jim is asking me to look at, so let's go ahead and pull it in to unblock the flow.

PR #63, opened 2018-03-16
Improve core/domain/enums and tests

feclare:
There was a panic if trying to convert to string some invalid values.
Added tests to AdminState and OperatingState
protocol.go is not used anywhere, should be removed
database.go is only used by metadata, should be moved to core/metadata/

### DIFF
--- a/core/domain/enums/admin-state.go
+++ b/core/domain/enums/admin-state.go
@@ -29,13 +29,20 @@ type AdminStateType uint8
 const (
 	LOCKED AdminStateType = iota
 	UNLOCKED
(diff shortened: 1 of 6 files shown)


feclare:
recheck

feclare:
recheck

feclare:
@tsconn23 
Should I remove protocol.go and move database.go to metadata?

tsconn23:
@feclare Removing protocol.go and the related test file should be fine. Personally, I'd probably just leave database.go where it is for now.

PR #68, opened 2018-03-20
Build in docker

akbennett:
This pull request does 3 things
- adds the logging microservice to the DOCKERS targets in the Makefile
- makes end image more reproducible by putting all of the builder dependencies within the Dockerfiles, limiting the risk that a build system artifact ends up in the resulting docker image.  This also makes it more straightforward when I am building for another architecture, i.e. aarch64
- sets all Dockerfiles to build with the same builder container (all except the final build artifacts).  This speeds up image rebuilds on slower architectures / systems and establishes a 'builder' container that we may want to optimize/isolate in the future.

### DIFF
--- a/docker/Dockerfile.core-command
+++ b/docker/Dockerfile.core-command
@@ -7,9 +7,15 @@
 #
 
 FROM golang:1.9-alpine AS builder
-WORKDIR /go/src/github.com/edgexfoundry/edgex-go
-RUN apk update && apk add make
+RUN apk --update add \
+    make \
+    git \
+    glide \
+    zeromq-dev \
+    build-base
+WORKDIR $GOPATH/src/github.com/edgexfoundry/edgex-go
 COPY . .
+RUN glide install
 RUN make cmd/core-command/core-command
 
 FROM scratch
--- a/docker/Dockerfile.export-distro
+++ b/docker/Dockerfile.export-distro
@@ -7,9 +7,15 @@
 #
 
 FROM golang:1.9-alpine AS builder
-WORKDIR /go/src/github.com/edgexfoundry/edgex-go
-RUN apk update && apk add zeromq-dev libsodium-dev pkgconfig build-base
+RUN apk --update add \
+    make \
+    git \
+    glide \
+    zeromq-dev \
+    build-base
+WORKDIR $GOPATH/src/github.com/edgexfoundry/edgex-go
 COPY . .
+RUN glide install
 RUN make cmd/export-distro/export-distro
 
 FROM alpine:3.7
--- a/docker/Dockerfile.support-logging
+++ b/docker/Dockerfile.support-logging
@@ -6,9 +6,15 @@
 #
 
 FROM golang:1.9-alpine AS builder
-WORKDIR /go/src/github.com/edgexfoundry/edgex-go
-RUN apk update && apk add make
+RUN apk --update add \
+    make \
+    git \
+    glide \
+    zeromq-dev \
+    build-base
+WORKDIR $GOPATH/src/github.com/edgexfoundry/edgex-go
 COPY . .
+RUN glide install
 RUN make cmd/support-logging/support-logging
 
 FROM scratch
--- a/docker/Dockerfile.core-data
+++ b/docker/Dockerfile.core-data
@@ -6,11 +6,16 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-# Docker image for Golang Core Data micro service 
 FROM golang:1.9-alpine AS builder
-WORKDIR /go/src/github.com/edgexfoundry/edgex-go
-RUN apk update && apk add zeromq-dev libsodium-dev pkgconfig build-base
+RUN apk --update add \
+    make \
+    git \
+    glide \
+    zeromq-dev \
+    build-base
+WORKDIR $GOPATH/src/github.com/edgexfoundry/edgex-go
 COPY . .
+RUN glide install
 RUN make cmd/core-data/core-data
 
 #Next image - Copy built Go binary into new workspace
--- a/docker/Dockerfile.core-metadata
+++ b/docker/Dockerfile.core-metadata
@@ -7,9 +7,15 @@
 #
 
 FROM golang:1.9-alpine AS builder
-WORKDIR /go/src/github.com/edgexfoundry/edgex-go
-RUN apk update && apk add make
+RUN apk --update add \
+    make \
+    git \
+    glide \
+    zeromq-dev \
+    build-base
+WORKDIR $GOPATH/src/github.com/edgexfoundry/edgex-go
 COPY . .
+RUN glide install
 RUN make cmd/core-metadata/core-metadata
 
 FROM scratch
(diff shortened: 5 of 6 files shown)


docker/Dockerfile.core-command
```diff
@@ -7,9 +7,19 @@
...
+RUN apk --update add \
+    make \
+    git \
+    ca-certificates \
+    glide \
+    zeromq-dev \
+    libsodium-dev \
+    pkgconfig \
+    build-base
+RUN go get github.com/edgexfoundry/edgex-go
```
drasko: Nope, please do not do this. Why would you be getting the code from the GitHub, hen you have one locally?

This:
1. Prevents developers to build the Docker containers and test the code during the development
2. Breaks the CI - how would you build the version x.y.z with your approach? You are obviously always building them master in your Docker

akbennett: Thanks.  I've removed this

akbennett:
re: development reproducibility 
I worry when different development systems have different versions of build dependencies.  i.e. go on Mac vs Ubuntu 16.10 vs. 16.04 ...  By putting all of the dependencies within the docker builder, it guarantees that anyone building/rebuilding the containers aligns on 1 version set of dependencies, which in this case would be those in (golang:1.9-alpine).  

re: "building for another architecture, i.e. aarch64"
It is almost the same justification as above.  If we put the dependencies in the Dockerfile, you can easily rebuild on other architectures without having to do ANY local environment setup beyond docker support as the upstream golang containers now support multi-arch as of 1.9.

Image: golang:1.9-alpine
 * Manifest List: Yes
 * Supported platforms:
   - linux/amd64
   - linux/arm/v6
   - linux/arm64/v8
   - linux/386
   - linux/ppc64le
   - linux/s390x

This allows you to completely do the same operation on any supported architecture without worrying about dependencies of the build environment:

i.e. on x86 or arm64, you simply docker build  with the architecture and docker uses the architecture's images (assuming you are building with a recent docker).  Enabling only a simple docker build to produce binaries on the other architectures: 
  docker build -f docker/Dockerfile.core-command -t edgexfoundry/docker-core-command .

drasko:
@akbennett I am not sure I am getting this correctly:
- When you do not put for example `libsodium-dev` - where is this lib taken from? From your local system? I doubt - if it is not in the Docker already - build will fail. Even if you have it on your system.

So - which service is using `libsodium-dev` today and how it is built today if we do not have this lib in the Docker already.

Also - `glide` has to be installed on build system and `glide install` should be run prior to issuing docker build. This is just downloading the dependency sources - no system artifacts here.

docker/Dockerfile.core-command
```diff
@@ -7,9 +7,18 @@
 #
 
 FROM golang:1.9-alpine AS builder
-WORKDIR /go/src/github.com/edgexfoundry/edgex-go
-RUN apk update && apk add make
+RUN apk --update add \
+    make \
+    git \
```
drasko: Why do you need `git` here?

docker/Dockerfile.core-command
```diff
@@ -7,9 +7,18 @@
...
+    ca-certificates \
```
drasko: Do this service use `ca-certificates`? I would not say so.

docker/Dockerfile.core-command
```diff
@@ -7,9 +7,18 @@
...
+    glide \
```
drasko: Do not need `glide` - `glide install` is done prior to any build.

docker/Dockerfile.core-command
```diff
@@ -7,9 +7,18 @@
...
+    zeromq-dev \
```
drasko: This service do not use `ZMQ`

docker/Dockerfile.core-command
```diff
@@ -7,9 +7,18 @@
...
+    libsodium-dev \
```
drasko: Makes no sense here - this service is not using it.

docker/Dockerfile.core-command
```diff
@@ -7,9 +7,18 @@
...
+    pkgconfig \
```
drasko: Not used.

docker/Dockerfile.core-command
```diff
@@ -7,9 +7,18 @@
...
+    build-base
```
drasko: Not used.

docker/Dockerfile.core-command
```diff
@@ -7,9 +7,18 @@
...
+WORKDIR $GOPATH/src/github.com/edgexfoundry/edgex-go
 COPY . .
+RUN glide install
```
drasko: No - this should be done prior to any build.

drasko:
I would understand these changes only if you were trying to make one single Dockerfile which holds the super-set of all build dependencies, and then just pass into it a service name as an argument - this way we would have only one Dockerfile that is capable of building **all** services, in the unified manner. This would  probably be an interesting approach. But the current approach is unnecessarily bloating the containers and add confusion.

akbennett:
Hi, I restructured the git commits, hopefully this will make the intent more clear.
1. add dependencies to the Dockerfile.export-distro builder definition so that the micro-service can be built entirely by docker build (no reliance on host machine tools).
2. Use the export-distro builder definition allowing docker build on the other microservices to benefit from the builder multi-stage layers being the same.  

Note as the changes are being made only to the intermediate / builder stage of the Dockerfile, these changes should not affect the final micro-service docker images (unless contents are explicitly included, which they aren't at this time)

docker/Dockerfile.export-distro
```diff
@@ -7,9 +7,15 @@
 #
 
 FROM golang:1.9-alpine AS builder
-WORKDIR /go/src/github.com/edgexfoundry/edgex-go
-RUN apk update && apk add zeromq-dev libsodium-dev pkgconfig build-base
```
drasko: Is `libsodium-dev` finally needed or not?

docker/Dockerfile.support-logging
```diff
@@ -6,9 +6,15 @@
 #
 
 FROM golang:1.9-alpine AS builder
-WORKDIR /go/src/github.com/edgexfoundry/edgex-go
-RUN apk update && apk add make
+RUN apk --update add \
+    make \
+    git \
+    glide \
+    zeromq-dev \
```
drasko: Does logging use ZMQ?

docker/Dockerfile.support-logging
```diff
@@ -6,9 +6,15 @@
...
+    build-base
```
drasko: `build-base` needed?

drasko:
> add dependencies to the Dockerfile.export-distro builder definition so that the micro-service can be built entirely by docker build (no reliance on host machine tools).

Please add **only** used dependencies, foe each container.

>Use the export-distro builder definition allowing docker build on the other microservices to benefit from the builder multi-stage layers being the same.

Why? Makes no sense. Companies will replace reference implementation of some micoroservices  with some others - why do you force them to download dependencies for `export-distro` if they will not need them for building their microservices.

Again - please add in each container only dependencies needed for building that particular microservice.

Also - please remove `glide` and `git` from containers. Do `glide install` step before building container.

PR #69, opened 2018-03-21
Add more export distro tests

feclare:


export/distro/http_test.go
```diff
@@ -0,0 +1,107 @@
...
+	var tests = []struct {
+		name string
+		addr models.Addressable
+	}{
+		{"noMethod", models.Addressable{
+			Protocol: "http",
+			Path:     path}},
+		{"get", models.Addressable{
+			Protocol:   "http",
+			HTTPMethod: models.MethodGet,
```
tsconn23: @feclare I believe the build is failing due to the remaining models.MethodGet/Post and so on in this file. A later PR was merged before this one wherein all of the HTTP verbs were repointed. Once this edit is made, should be good to go.

feclare: Thanks @tsconn23 , it should be fixed now.

PR #73, opened 2018-03-23
Use Influxdb in Core-data

m-masataka:
Signed-off-by: Masataka Mizukoshi <m.mizukoshi.wakuwaku@gmail.com>
This code makes coredata possible to use influxdb. (re PR of #58, I accidentally erasing commit. Sorry)

Now, this implementation purely imitate client code of mongo.
I think we have to revise mongo-dependent implementation to use any other databases eventually,
(especially bson data structure)


core/data/clients/influx-client.go
```diff
@@ -0,0 +1,857 @@
...
+	e.Modified = time.Now().UnixNano() / int64(time.Millisecond)
+
+	// Delete event
+	err := ic.deleteById(EVENTS_COLLECTION, e.ID.Hex())
+	if err != nil {
+		return ErrNotFound
+	}
+
+	// Add the event
+	err = ic.addEventToDB(ic.Database, EVENTS_COLLECTION, &e)
```
feclare: You can just do `return ic.addEventToDB(ic.Database, EVENTS_COLLECTION, &e)`. 

This pattern is also found in some other functions.

drasko:
@feclare this PR looks good to me, but please review it also.

If all is good, I think we can move on and merge this PR.

drasko:
@jpwhitemn @tsconn23 this is core-related PR. Please see how it fits in core services and does additional DB option (in this case InfluxDB) fits in the system.

Code was reviewed several times (https://github.com/edgexfoundry/edgex-go/pull/58) and it looks good to me now - however I let you merge, because I can not see if it fits from overall arch perspective.

tsconn23:
@drasko OK, will take a closer look once the issues with the Jenkins build server are worked out. I couldn't merge this PR right now even if I wanted to.

Issue #26, opened 2018-03-02
Core services read configuration from hard-coded file


drasko:
Core services currently read config file from hard-coded location: https://github.com/edgexfoundry/edgex-go/blob/master/cmd/core-command/main.go#L37

Go binaries should be portable, and probably end-up in `$GOBIN` (via `go install`). Also, very often `$GOBIN` is added to system path, so that these binaries are accessible system-wide.

Because they should be capable to start from different locations, they should be capable to accept config file location via command-line parameter or ENV variable.

For the sake of simplicity, I would suggest ENV variable here (easy to pass into Docker via `docker-compose `environment` parameter)

tonyespy:
My vote would be for a command-line option, as environment variables can be set via a number of different ways, and thus can be hard to keep track of, whereas there's only one place a command-line argument can be specified, and it's easy to inspect the command-line options passed to a running process (at least it's easy on Linux, not sure about Windows).

jpwhitemn:
Based on our Go project call yesterday, I have Trevor working on this for core-data.  We'll work on a few of these options and provide something for people to look at once we think we have something in place that we are satisfied with - and get everyone's opinion.  As we discussed yesterday, it would be nice to eventually replicate the type of escalating mechanism that Java has (property file, command line, ENV, etc.).  However, we might start simple and then incorporate other options as we go (and as we need) but try to facilitate that type of extension.

drasko:
@tonyespy no need to open command line for passing just one single option: location of the file. this makes code more complicated - you are adding dependencies to parse command line options, send help to command line, etc - this SW will not be run by hand, it will be management service that will be running the services.

Using command line here in my opinion goes against programming principles of simplicity and producing tighter code with less binaries. Parsing command is not only useless but probably adds additional security risks, while all ENV variables can (and probably will be) stacked in Vault: https://www.vaultproject.io/intro/use-cases.html#general-secret-storage - _At a bare minimum, Vault can be used for the storage of any secrets. For example, Vault would be a fantastic way to store sensitive **environment variables**, database credentials, API keys, etc._

tsconn23:
PR https://github.com/edgexfoundry/edgex-go/pull/75 created. Feel free to comment/approve.

tsconn23:
PR has been merged. If there are no follow-up comments/discussion in 24 hours, I'll close the issue as resolved.

tsconn23:
[...]

PR #75, opened 2018-03-25
Accept config directory from ENV variable

tsconn23:
PR for this issue: https://github.com/edgexfoundry/edgex-go/issues/26

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -22,17 +22,20 @@ import (
 	"fmt"
 
 	"github.com/BurntSushi/toml"
+	"os"
 )
 
 const (
-	configDefault = "./res/configuration.toml"
-	configDocker = "./res/configuration-docker.toml"
-	configUnitTest = "./res/configuration-test.toml"
+	configDirectory = "./res/"
+	configDefault = "configuration.toml"
+	configDocker = "configuration-docker.toml"
+	configUnitTest = "configuration-test.toml"
 )
 
 func LoadFromFile(profile string, configuration interface{}) error {
-	path := determineConfigFile(profile)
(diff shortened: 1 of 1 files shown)


pkg/config/config.go
```diff
@@ -20,19 +20,22 @@ package config
...
+	configDocker = "configuration-docker.toml"
+	configUnitTest = "configuration-test.toml"
 )
 
 func LoadFromFile(profile string, configuration interface{}) error {
-	path := determineConfigFile(profile)
-	contents, err := ioutil.ReadFile(path)
+	path := determinePath()
+	fileName := determineConfigFile(profile)
+	contents, err := ioutil.ReadFile(path + fileName)
```
drasko: I think it would be more self-documenting if we were not to force `path` ending with `/`, but rather have here something like: `path + "/" + filename`

pkg/config/config.go
```diff
@@ -56,3 +59,18 @@ func determineConfigFile(profile string) string {
...
+func determinePath() string {
+	//Assumption: one service per container means only one var is needed, set accordingly for each deployment.
+	//For local dev, do not set this variable since configs are all named the same.
+	path := os.Getenv("EDGEX_CONF_DIR")
+
+	if len(path) == 0 { //Var is not set
+		path = configDirectory
+	}
+
+	if path[len(path)-1:] != "/" {
```
drasko: Maybe no need to enforce finishing with `/`. As explained before.

pkg/config/config.go
tsconn23: Good, this is something I wanted to get feedback on. I don't have a strong preference.

PR #77, opened 2018-03-25
fix URL command GET PUT path

bullapse:
update the PUT GET block to use the correct path

Signed-off-by: bullapse <spencerbull2554@gmail.com>


bullapse:
fix requested from core-command-go repo issue and updated in monorepo @tsconn23

PR #78, opened 2018-03-26
[support-logging] Add mongodb support

chenchix:


support/logging/types.go
```diff
@@ -7,15 +7,32 @@
...
+	defaultMongoDB             = "logging"
+	defaultMongoCollection     = "logEntry"
+	defaultMongoURL            = "127.0.0.1"
+	defaultMongoPort           = 27017
+	defaultMongoConnectTimeout = 5000
+	defaultSocketTimeout       = 5000
+	defaultMongoUsername       = "logging"
+	defaultMongoPassword       = "password"
+
+	envMongoURL  = "EXPORT_CLIENT_MONGO_URL"
```
drasko: Export client? I guess it should be `SUPPORT_LOGGING`?

support/logging/types.go
```diff
@@ -38,10 +65,38 @@ type persistence interface {
...
+		MongoConnectTimeout: defaultMongoConnectTimeout,
+		MongoSocketTimeout:  defaultSocketTimeout,
 	}
+
+	MongoPortStr := env(envMongoPort, strconv.Itoa(defaultMongoPort))
+	MongoPort, err := strconv.Atoi(MongoPortStr)
+	if err == nil {
+		cfg.MongoPort = MongoPort
+	} else {
+		fmt.Println("Could not parse port", MongoPort)
```
drasko: Please do not print on stdout. If there is an error - maybe this should be returned as an indication to a calling function.

support/logging/types.go
chenchix: Right! Fixed

Issue #44, opened 2018-03-09
Makefile improvements


feclare:
I have found a way to compile all the go code in the repository and to generate all the microservices. The only problem is that the executables are stored in $GOPATH/bin

```
go install ./...
```
So we will need to move them afterwards to other location. 
We could use something like:
```
build:
    go install ./...
    mv $GOPATH/bin/core-data $DESTDIR
...
    mv $GOPATH/bin/export-distro $DESTDIR
```
Removing the build_microservices target. Opinions?

drasko:
@feclare please refer to https://github.com/edgexfoundry/edgex-go/issues/26 and also https://github.com/edgexfoundry/edgex-go/issues/25 because I expect problems here with `go install`

jpwhitemn:
Makefile issues and changes to be discussed in 3/22 Core WG meeting.

feclare:
The only remaining issue is where should the executables be generated. I know @drasko likes to have them in a different directory like `edgex-go/build`. With the configuration files sharing the same filenames each microservice needs to be in a different directory, if we want to have this I should use something like:
```
edgex-go/build/core-data/ 
edgex-go/build/core-data/core-data
edgex-go/build/core-data/res/configuration.toml
edgex-go/build/core-metadata/ 
edgex-go/build/core-metadata/core-metadata
edgex-go/build/core-metadata/res/configuration.toml
```
Any other alternative?

drasko:
@feclare I propose that we leave it for now as is. We'll see later.

If you are OK with this, please close this issue.

PR #79, opened 2018-03-26
Makefile: compile all microservices with build target. #44

feclare:
Removing build_microservices target and generate all executables as dependencies of the build target.


Makefile
```diff
@@ -20,11 +20,9 @@ VERSION=$(shell cat ./VERSION)
 
 GOFLAGS=-ldflags "-X github.com/edgexfoundry/edgex-go.Version=$(VERSION)"
 
-build:
+build: $(MICROSERVICES)
 	go build ./...
```
tsconn23: Isn't this line redundant?

tsconn23: Isn't line 24 redundant now with the $(MICROSERVICES) block above it?

feclare: No. There is a subtle difference. 
For example `support/scheduler-client/` is not used by any microservice, if we don't add the `go build ./...` we will not test if it compiles at least.

PR #80, opened 2018-03-26
Duplicated support-logging

chenchix:
Signed-off-by: chencho <smunoz@cavium.com>

### DIFF
--- a/bin/edgex-launch.sh
+++ b/bin/edgex-launch.sh
@@ -56,15 +56,6 @@ exec -a edgex-core-metadata ./core-metadata &
 cd $DIR
 
 ###
-# Support logging
-###
-printf "\n### Starting support-logging\n"
-cd $CMD/support-logging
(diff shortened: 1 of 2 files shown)


drasko:
Why is this a duplicate? Is support logging started by some other service implicitly?

chenchix:
@drasko i added it in my merge by mistake at the end of the script

drasko:
I do not see it in the end of the script, I see only one occurrence. This is why I am asking.

chenchix:
Current version  (merged in the last PR) has both support-logging script calls, https://github.com/edgexfoundry/edgex-go/blob/master/bin/edgex-launch.sh

This is why I have removed now.

feclare:
It is being launched the first and the fifth 
https://github.com/edgexfoundry/edgex-go/blob/master/bin/edgex-launch.sh

PR #88, opened 2018-03-28
Switch mirrors for alpine image apk update

JPWKU:
The default mirrors are not responsive and are
timing out the builds periodically.

### DIFF
--- a/docker/Dockerfile.core-command
+++ b/docker/Dockerfile.core-command
@@ -8,6 +8,12 @@
 
 FROM golang:1.9-alpine AS builder
 WORKDIR /go/src/github.com/edgexfoundry/edgex-go
+
+# The main mirrors are giving us timeout issues on builds periodically.
+# So we can try these.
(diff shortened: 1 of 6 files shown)


JPWKU:
There are a several alpine mirrors we can choose from.  I'm fairly certain this is the issue we are hitting.

JPWKU:
Here in this issue (which looks identical to ours) someone has listed all of the apk mirrors.
https://github.com/gliderlabs/docker-alpine/issues/279

JPWKU:
@feclare The only thing that makes sense is that since the arm machine is a static minion we probably have cached docker image layers locally.

feclare:
It makes sense

JPWKU:
We are a bit backed up on build machines.  We are in the queue along with a bunch of daily jobs.  It shouldn't take too long to clear up though.

JPWKU:
recheck

JPWKU:
Hey all, Looks like this mirror change worked.  Since @feclare approved I will merge.  @tsconn23  and others, after this is merge, do a rebase on the current PR's and this should clear out.
Jeremy

PR #95, opened 2018-03-29
Update mirror for alpine 3.7 image

JPWKU:


JPWKU:
@drasko heh I don't disagree, but as it is, we have no control over those apk repos.  Do you have another suggestion?

drasko:
Hmmm.. we can reach out to Alpine community and see with them maybe?

drasko:
Ah, OK... Well... Dirty hack then :).

PR #119, opened 2018-04-03
Don't override export-distro hostname if env set

jascott1:
Fix issue where os.hostname overrides the hostname set thru ENV.

### DIFF
--- a/cmd/export-distro/main.go
+++ b/cmd/export-distro/main.go
@@ -121,10 +121,6 @@ func loadConfig() (distro.Config, config) {
 		ConsulPort: defConsulPort,
 		Hostname:   env(envDistroHost, defHostname),
 	}
-	hostname, err := os.Hostname()
-	if err == nil {
-		cfg.Hostname = hostname
-	}
 
 	portStr := env(envConsulPort, strconv.Itoa(cfg.ConsulPort))
 	port, err := strconv.Atoi(portStr)
--- a/cmd/export-client/main.go
+++ b/cmd/export-client/main.go
@@ -145,16 +145,10 @@ func loadConfig() (*config, *client.Config) {
 		MongoSocketTimeout:  defMongoSocketTimeout,
 		ConsulHost:          env(envConsulHost, defConsulHost),
 		ConsulPort:          defConsulPort,
+		Hostname:            env(envClientHost, defHostname),
 	}
-
-	hostname, err := os.Hostname()
-	if err == nil {
-		cfg.Hostname = hostname
-	}
-	cfg.Hostname = env(envClientHost, cfg.Hostname)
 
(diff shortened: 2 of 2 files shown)


cmd/export-distro/main.go
```diff
@@ -121,9 +121,12 @@ func loadConfig() (distro.Config, config) {
 		ConsulPort: defConsulPort,
 		Hostname:   env(envDistroHost, defHostname),
 	}
-	hostname, err := os.Hostname()
-	if err == nil {
-		cfg.Hostname = hostname
+	// if using default hostname, try to use os.hostname
+	if cfg.Hostname == defHostname{
```
drasko: This is not how the algorithm is supposed to work.

Hostname is set to default here: https://github.com/edgexfoundry/edgex-go/blob/master/cmd/export-distro/main.go#L122

Actually, this whole part: https://github.com/edgexfoundry/edgex-go/blob/master/cmd/export-distro/main.go#L124-L128 does not make sense and should be removed.

`cfg.Hostname` should be either set to default, or from an ENV variable. Service should not try to guess `hostname` from underlying OS by using `os` module.

While you are at it, you can remove this one also: https://github.com/edgexfoundry/edgex-go/blob/master/cmd/export-client/main.go#L150-L153

drasko:
@feclare how did this end up in the code? Should be removed, we should pass `hostname` info either via ENV variable or `defHostname` string constant (`localhost`).

tonyespy:
Jim's close to releasing the new base service design, and this is addressed (pun intended).  Also this setting is more accurately "host", as it can either be a resolvable name or an actual IP address.

In the new design, a service's "host" and "port" are configuration settings.  If a service is started with the config/registry (aka Consul currently) enabled, then the "host" & "port" settings values read from config/registry are used.  If config/registry is not enabled, then the settings come from the local configuration[-profile].toml file.

drasko:
@tonyespy thanks. Yep - we definitely need a spec to align all microservices on this.

ENV variable occurred most appropriate for me, because I wanted to avoid creating config file just for one config (we have not so much in export services).

Trouble with config files is that you have to carry them around when you move statically compiled binaries, this is also a second point because I wanted to avoid writing separate config file for `export` at the time we migrated it.

Spec should define this, and we'll all align.

tonyespy:
Here's the latest version fo the spec, it works as described above:

https://wiki.edgexfoundry.org/download/attachments/7602423/ServiceNameDesign-v4.odt?version=1&modificationDate=1522787058000&api=v2

jascott1:
I corrected the PR to only use default and env vars. This should probably be labeled as a bug since export-distro currently does not respect the env var and overrides with os.Hostname.

drasko:
@tonyespy any updates on this? Should we close this PR?

drasko:
Closing dues to inactivity.

PR #122, opened 2018-04-06
Fixed error when updating an addressable in core-metadata

jduranf:
When updating an addressable (i.e. to change the path), the value in protocol field is lost.
Also seems to be better to remove the check because this check doesn't exist when adding an addressable and also doesn't allow use "OTHER" protocol.

### DIFF
--- a/core/metadata/mongoOps.go
+++ b/core/metadata/mongoOps.go
@@ -22,7 +22,6 @@ package metadata
 import (
 	"errors"
 	"reflect"
-	"strings"
 	"time"
 
 	"github.com/edgexfoundry/edgex-go/core/domain/models"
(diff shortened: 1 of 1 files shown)


core/metadata/mongoOps.go
```diff
@@ -649,7 +648,7 @@ func mgoUpdateAddressable(ra *models.Addressable, r *models.Addressable) error {
 	if ra.Name != "" {
 		r.Name = ra.Name
 	}
-	if strings.Compare(ra.Protocol, "HTTP") != 0 || strings.Compare(ra.Protocol, "TCP") != 0 { // TODO create ENUMS that can be unmarshalled by JSON
+	if ra.Protocol != "" {
```
drasko: Both `r` and `ra` variable names are confusing and not self-documenting.

If `ra` can take values only from a predefined set (some enum), then this comparison is not good.

tsconn23: FWIW, I found from going up the call chain that "r" is the new addressable and "ra" is the old addressable. These do need to be addressed but outside of scope for this change

tsconn23: That protocol property should really be an enum, but outside the scope of this change.

tsconn23:
@jduranf Unfortunately we can't pull this in until it's signed. Examples for how to do that are provided here:

https://wiki.edgexfoundry.org/display/FA/Committing+Code+Guidelines

You might have to create another PR.

drasko:
Also please rebase to latest master.

jduranf:
OK, I close the PR and I will try to create a new one signed
Also should be fine to update the CONTRIBUTING.md guide

PR #123, opened 2018-04-06
Fixed error when updating an addressable in core-metadata

jduranf:
New PR with signed commit which replaces #122


drasko:
@jduranf sign-off still missing.

PR #127, opened 2018-04-10
Add EdgeX reST documentation

andyf1967:
Signed-off-by: steve osselton <steve@iotechsys.com>


docs/build.sh
```diff
@@ -0,0 +1,36 @@
...
+cp ../core/data/raml/*.raml ./core/data/
+cp ../core/metadata/raml/*.raml ./core/metadata/
+cp ../core/command/raml/*.raml ./core/command/
+cp ../support/logging/raml/*.raml ./support/logging/
+cp ../export/client/raml/*.raml ./export/client/
+
+# Build image (copying in documentation sources)
+
+docker build -t doc-builder:latest -f Dockerfile.build .
+if [ -d _build ]
```
feclare: No need to test if _build exists, just run the `rm -rf`

docs/build.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+
```
feclare: Add `set -e`, if some command fails the script should stop and report error

docs/build.sh
```diff
@@ -0,0 +1,36 @@
...
+# Clean up
+
+rm ./core/data/*.raml
+rm ./core/metadata/*.raml
+rm ./core/command/*.raml
+rm ./support/logging/*.raml
+rm ./export/client/*.raml
+docker rm builder
+docker image prune -f
```
feclare: Not sure if we should prune docker images, may be some user does not want it

docs/build.sh
SteveOss: OK, Will do.

docs/build.sh
SteveOss: OK

docs/build.sh
SteveOss: Have updated logic so that run container with --rm and use a bind mount to access built documentation, so this no longer needed.

Issue #107, opened 2018-04-01
Support logging should refactor configuration reading logic


yanghua:
Now it used const hardcode, should read from config file. Can I apply this ticket to fix?

yanghua:
@drasko closing this issue, thank you and other reviewers.

PR #130, opened 2018-04-12
#107 Support logging should refactor configuration reading logic

yanghua:

cc @drasko , please review this PR , thanks!


support/logging/server.go
```diff
@@ -223,28 +219,28 @@ func httpServer() http.Handler {
...
 }
 
-func getPersistence(config Config) persistence {
+func getPersistence() persistence {
 	var retValue persistence
-	if config.Persistence == PersistenceFile {
-		retValue = &fileLog{filename: config.LogFilename}
-	} else if config.Persistence == PersistenceMongo {
-		ms, err := connectToMongo(&config)
+	if configuration.Persistence == PersistenceFile {
```
drasko: `switch` would be more readable construct here

yanghua:
cc @drasko

jduranf:
Maybe is out of scope of this PR, but parameters like Hostname or MongoURL shouldn't be ServiceAddress or MongoDBHost? To use the same names used in core services...

yanghua:
@jduranf you are right, I will create a new issue about refactoring the same semantic configuration item with the unified key. However this PR just refactored the way of reading the configuration. They are two things. Thanks!

PR #134, opened 2018-04-13
Docupdates 1

SteveOss:
Minor documentation updates an build tidy including previous review suggestions.


SteveOss:
More DCO woes !

PR #140, opened 2018-04-13
Also return an error to registration.Validate

feclare:
Modify registration.Validate to also return an error with the reason when a
registration is considered invalid.

Also added tests to Validate

### DIFF
--- a/export/distro/client.go
+++ b/export/distro/client.go
@@ -47,8 +47,10 @@ func getRegistrationsURL(url string) []export.Registration {
 
 	results := registrations[:0]
 	for _, reg := range registrations {
-		if reg.Validate() {
+		if valid, err := reg.Validate(); valid {
 			results = append(results, reg)
+		} else {
+			logger.Warn("Could not validate registration", zap.Error(err))
 		}
 	}
 	return results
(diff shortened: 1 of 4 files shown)


export/distro/client.go
```diff
@@ -47,8 +47,10 @@ func getRegistrationsURL(url string) []export.Registration {
 
 	results := registrations[:0]
 	for _, reg := range registrations {
-		if reg.Validate() {
+		if valid, err := reg.Validate(); valid {
```
drasko: Rather check if error, then drop the `else` and second `{}`

feclare: @drasko I'm sorry, but I don't understand. There is two cases, the registration is valid and it should be added to results or it is invalid and a trace is generated. What are you suggesting to be removed

drasko: Ah, I see now outer loop. I guess that something like this would work:
```
if valid, err := reg.Validate(); !valid {
   logger.Warn("Could not validate registration", zap.Error(err)
   continue
}
results = append(results, reg)
```
but not sure that it would contribute the readability now.

feclare: I find more readable the if else, than the if continue. But it is just a style issue. I will use your code if you find it better.

I tend to not use continue/break unless it make the code easier to understand. And in this case they are about the same.

drasko: @feclare no, please keep whatever you feel better. I just did not see the outer loop first time, so I thought that we can get rid of this `else` easily.

PR #145, opened 2018-04-17
 Refactored database system in export client to allow different DB types 

jduranf:
Export client database system have been refactored in order to allow different databases types, in the same way is done in core-data


jduranf:
Seems the edgex-go-arm-verify check failed because there is no space on the device...

feclare:
recheck

feclare:
@jduranf Thanks for the heads-up. We made some space on the machine

export/client/clients/database-client.go
```diff
@@ -0,0 +1,132 @@
...
+var ErrNotUnique error = errors.New("Resource already exists")
+
+// Return the dbClient interface
+func NewDBClient(config DBConfiguration) (DBClient, error) {
+	switch config.DbType {
+	case MONGO:
+		// Create the mongo client
+		mc, err := newMongoClient(config)
+		if err != nil {
+			fmt.Println("Error creating the mongo client: " + err.Error())
```
feclare: Do not print to stdout. Maybe something like:
return nil, fmt.Error(""Error creating the mongo client: " + err.Error())

jduranf:
@feclare I don't know why, but I received a mail with your comments but couldn't see in GitHub... anyway I removed print in stdout.
Also fixed an issue similar to #147

export/client/registration.go
```diff
@@ -109,6 +109,12 @@ func addReg(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if reg.Name == "" {
```
feclare: Could you move this logic to registration.Validate()? And add/modify tests following the new logic

feclare:
@jduranf I deleted a comment, it doesn't make sense after reading all the changes. My error.

export/client/clients/mongo-client.go
```diff
@@ -0,0 +1,186 @@
...
+	mongoDBDialInfo := &mgo.DialInfo{
+		Addrs:    []string{connectionString},
+		Timeout:  time.Duration(config.Timeout) * time.Millisecond,
+		Database: config.DatabaseName,
+		Username: config.Username,
+		Password: config.Password,
+	}
+	session, err := mgo.DialWithInfo(mongoDBDialInfo)
+	if err != nil {
+		fmt.Println("Error dialing the mongo server: " + err.Error())
```
feclare: @jduranf this fmt.Println also need to be removed.

export/client/registration.go
feclare: Name can't be empty when creating a registration but it can be empty when updating? Why can it be empty in that case?

jduranf: When updating you can send something like this:
`{"id":"57db5bd2add4d779d38ff066","compression":"GZIP"}`
So, there is no need to send the name

feclare: You should not call validate on update. It will set some of the missing fields to the default values( ompNone for compression) and will test that the remaining fields had a valid value.

jduranf: @feclare After recheck the code, I realized we can move the name check to registration.Validate() without more modifications. Also I have had to modify registration_test.go

export/registration_test.go
```diff
@@ -11,26 +11,30 @@ import "testing"
...
 	}{
-		{"empty", "", "", "", "", false},
-		{"valid", CompZip, FormatJSON, DestMQTT, EncAes, true},
-		{"defaultCompression", "", FormatJSON, DestMQTT, EncAes, true},
-		{"defaultEncryption", CompZip, FormatJSON, DestMQTT, "", true},
-		{"wrongCompresion", "INVALID", FormatJSON, DestMQTT, EncAes, false},
-		{"wrongFormat", CompZip, "INVALID", DestMQTT, EncAes, false},
-		{"wrongDestination", CompZip, FormatJSON, "INVALID", EncAes, false},
-		{"wrongEncryption", CompZip, FormatJSON, DestMQTT, "INVALID", false},
+		{"empty", "reg", "", "", "", "", false},
```
feclare: Change the name to empty string in the "empty" case. If not the case with an empty name is not tested.

jduranf:
Seems edgex-go-verify check failed because connection with nl.alpinelinux.org failed...

export/client/registration.go
feclare: Yes, you are right. After inspecting the code again the validate when updating a registration also make sense.

export/registration_test.go
jduranf: I also added case "withoutName" which checks the case of empty name, so already is tested. But if you want, I can remove this new case and modify "empty" case

feclare: I missed your new entry. Sorry.
But I still think that is better to have the "empty" entry with a default empty registration.

And it will also launch new jenkins jobs.

feclare: I think my comment was not clear. It is better to keep both in the tests. Just remove the name in the empty registration.

feclare:
@JPWKU it looks like downloading docker images is failing again.

JPWKU:
recheck

JPWKU:
@feclare I've not seen this error yet, I'm retrying and also reading up on it.

JPWKU:
@feclare Is all good now, looks like that could have been a network blip and intermittent.

JPWKU:
@feclare This PR does need a rebase against master, it's out of date.

export/client/clients/database-client.go
```diff
@@ -0,0 +1,131 @@
...
+func (db DatabaseType) String() string {
+	if db >= INVALID && db <= MOCK {
+		return databaseArr[db]
+	}
+	return invalidStr
+}
+
+// Return enum value of the Database Type
+func GetDatabaseType(db string) DatabaseType {
+	if mongoStr == db {
```
drasko: `switch` would be a better construct here

jduranf: Done!

drasko:
@feclare please review

drasko:
Merged! Thanks @jduranf - great contribution here!

PR #146, opened 2018-04-17
Refactor notification-client usage to interface

tsconn23:
Includes factory pattern to initialize and obtain latest instance.
Includes handling change for configuration allowing client to be
         utilized and customized from any service.


feclare:
I think that interfaces are great and that they should be used when needed. But I don't think this is the case. We have only one client type (http) and this will not change anytime soon. We are making the code more complex without any immediate gain.

I should propose to remove the interface and that the notification_client package only exports a function like:
```
 func SendNotification(host string, port int, n Notification) error {
```
This will also remove the need in the package using this to have a notifications_client instance to be able to call the function (or having a global instance in the package)

feclare:
Ok, I understand.
What it still does not make sense to me is to have a global configuration in the package and a global instance ( support/notifications-client/client.go:87). 
Why not having a function like:
```
func NewNotificationsHttpClient(host string, port int) NotificationClient
```

tsconn23:
@feclare Well, in effect what we have right now in core-metadata is a global instance -- not even an interface but an instance.

core/metadata/init.go: line 34
`var notificationsClient = notifications.NotificationsClient{}`

The host and port are then set on lines 73 and 74 of that same file, equivalent to passing them when initializing the type as you show above. My refactoring effects this same thing, initializing the settings in a notifications config that is encapsulated within the notification_client package (allowing for on the fly updates later BTW) and providing a public way to access the current instance of the NotificationClient throughout the application -- whether it be a mock or an actual implementation.

I separated the config into its own type because I considered the question of whether or not some other service may want to utilize the Notification service. This could be an unnecessary consideration but stems from the fact that the name "Notification Service" is very general. What kind of notification? From whom to where? Any service could conceivably need to send a "notification" of some kind.

It sounds to me like your main concern is the new signature is too verbose, going from this:
`notificationsClient.SendNotification(notification)`

to this:
`notifications.GetNotificationsClient().SendNotification(notification)`

I'll concede that, and if you have suggestions for how to make the signature more terse without sacrificing readability, let me know. You will see this pattern all over the place in the holding area I linked to above.

jpwhitemn:
@feclare - thanks for your good review and continued input on this.  Let's discuss this in an upcoming working group meeting as a potential global issue that we need to address (I am in the process of redoing Notification in Go right now and will have to address there too).  Let's develop the issue and get that into the system separate from this PR.

Issue #147, opened 2018-04-17
Name for addressable can be 'null' and also stored


tsconn23:
I have verified this issue exists in the Go reference implementation as well. Original issue report below.

[Java core-metadata issue #31](https://github.com/edgexfoundry/core-metadata/issues/31)

tsconn23:
Resolved for this particular issue. Other similar cases may exist and we need to review.

PR #150, opened 2018-04-18
Name can't be empty when adding addressable (Issue #147)

tsconn23:
https://github.com/edgexfoundry/edgex-go/issues/147


tsconn23:
Dang merge commit. Something got messed up 👎

jduranf:
@tsconn23 This issue affects other objects in core-metadata (I tested a couple), maybe should be fine to fix all

Issue #96, opened 2018-03-29
Add "help" argument support to core services


tsconn23:
Provide verbiage w/r/t supported command line flags and their usage via ./service --help 

See 
https://stackoverflow.com/questions/23725924/can-gos-flag-package-print-usage or this: https://github.com/MainfluxLabs/fluxmq/blob/master/main.go

drasko:
@tsconn23 yes, probably the best golang CLI library. We are using it in Mainflux CLI: https://github.com/mainflux/bashflux - you can take a look here: https://github.com/mainflux/bashflux/blob/master/bashflux.go

But for this solution it would be an overkill :).

PR #153, opened 2018-04-20
Adding --help cmd line arg support

tsconn23:
Issue https://github.com/edgexfoundry/edgex-go/issues/96


drasko:
@tsconn23 yes, looks good. Just that I thought that `loggingClient` (https://github.com/edgexfoundry/edgex-go/pull/153/files#diff-d04c63c44dd4a2c836de5382b23a4c10R58) is creted earlier and that we can report these error into it. But I see not that you need Consul data in order to instantiate the logging client - which then opens the question where do we log these errors that happen prior to fetching data from Consul.

tsconn23:
@drasko Right now, they're just written to stdOut but there's no reason that configuration.LoggingFile couldn't be provided to write to the file as well. That kind of change will need to be in all four of the services, so will create an issue. Can you approve and merge this PR?

PR #155, opened 2018-04-22
Add VERSION to docker tag

JPWKU:
This will lessen the chance of error
when using docker images built locally,
and those built in the CI env.  This
also adds docker_support_logging to
DOCKERS.


feclare:
We should also update docker-compose.yml to use the new tags.

PR #156, opened 2018-04-23
Implement a memory based db interface in export client instead of a mock

feclare:
This could be useful in development to be able to run without any database, and
to test the remaining export client (not done in this commit).
The memory database always starts empty. Any change is lost when the
microservice is stopped


feclare:
@jduranf I modified the mock to be a complete db implementation. Could you review it?

drasko:
@feclare maybe I am missing something, but what is different comparing to @jduranf 's implementation, except that you renamed `mock` to `mem`?

feclare:
@drasko This is a full db implementation, not just a mock. So you can actually run export-client with it as backend.

export/client/clients/memory-database.go
```diff
@@ -0,0 +1,97 @@
...
+	reg.ID = bson.NewObjectId()
+
+	mc.regs = append(mc.regs, *reg)
+
+	return reg.ID, nil
+}
+
+func (mc *memDB) UpdateRegistration(reg export.Registration) error {
+	for i, r := range mc.regs {
+		fmt.Printf("updating %v <-> %v\n", r.ID, reg.ID)
```
jduranf: Maybe this print should be removed...

feclare: Completely. Thanks

drasko:
Merged! Thanks @feclare

PR #172, opened 2018-04-26
Test toml configuration files

feclare:
Added tests to each microservice in cmd that verify that the toml files in res/ can be parsed.

I modified how a profile is mapped to a config file, but generate the same files for the empty and docker profiles.


pkg/config/config.go
```diff
@@ -18,49 +18,47 @@
...
 }
 
 func determineConfigFile(profile string) string {
-	switch profile {
-	case "docker":
-		return configDocker
-	case "unit-test":
-		return configUnitTest
-	default:
+	if profile == "" {
```
tsconn23: This causes a bug. By default, the value of profile is "default" -- thus the switch case from the previous version. As this code is written, when the services start the following error is thrown by "make run":

ERROR: 2018/04/26 10:33:47 could not load configuration file (./res/configuration-default.toml): open ./res/configuration-default.toml: no such file or directory

tsconn23:
Is there possibly some other way to organize the tests such that we don't require an additional file at the root of each service? It's clean to just have the main.go in the cmd/<service_name> directory.

feclare:
Maybe I can move the test to cmd/toml_test.go and test all configuration files from there. Do you like it best?

pkg/config/config.go
feclare: I can think of two different solutions:
 - test also for the string "default"
 - we can set the default profile to the empty string.

I prefer the second option as there is no 'magic' strings, just the default go value for a string. But I'm ok with both.

feclare:
Go doesn't like having a directory with tests and no go code, so having the test in cmd is not an option.

pkg/config/config.go
tsconn23: Either is fine. In doing the former, you don't have to change the flags logic in the main.go of all the services. If the latter, then you do have to change it. It depends on your preference for limiting the scope of change in this PR.

tsconn23: Ah, I see you already decided to do the latter ;-)

tsconn23:
@feclare Have a look at resolving the conflicts and then I'll do another review/merge.

pkg/config/config.go
feclare: Yes, I like it best. But I can change to to "default" if you think it is better.

feclare:
Fixed the conflicts.

PR #180, opened 2018-04-27
Modifying makefile

chuyval:
Adding latest tag to the docker images. Adding 'docker-' to the name of the images. Formatting docker build commands for legibility

Signed-off-by: Jesus Valenzuela <chuyval@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,19 +61,43 @@ run_docker:
 docker: $(DOCKERS)
 
 docker_core_metadata:
-	docker build -f docker/Dockerfile.core-metadata -t edgexfoundry/core-metadata-go:$(VERSION) .
+	docker build \
+		-f docker/Dockerfile.core-metadata \
+		-t edgexfoundry/docker-core-metadata-go:latest \
+		-t edgexfoundry/docker-core-metadata-go:$(VERSION) \
+		.
 
 docker_core_data:
-	docker build -f docker/Dockerfile.core-data -t edgexfoundry/core-data-go:$(VERSION) .
+	docker build \
+		-f docker/Dockerfile.core-data \
+		-t edgexfoundry/docker-core-data-go:latest \
+		-t edgexfoundry/docker-core-data-go:$(VERSION) \
+		.
 
 docker_core_command:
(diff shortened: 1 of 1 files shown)


SteveOss:
Not wishing to be a nay sayer here, but isn't adding "docker" to the name of a docker image,
completely redundant ?

SteveOss:
@JPWKU I agree is harmless, but dislike stuff that doesn't "make sense". IMHO should consistently strip docker prefix, as per your initial patch.

JPWKU:
@tsconn23 @chuyval Regarding my earlier message about interpolating the short git-sha into the tag; I'm not sure this will actually be helpful to folks as one will not be able to automate the pulling of such images.  Better to add a tag which appends "-dev" in the tag name like so:
"docker tag <imageid> edgexfoundry/<servicename>:${VERSION}-dev".  I can then publish the dev tags for daily automated consumption.  Thoughts?

JPWKU:
@jpwhitemn Thoughts on @SteveOss comments on the "docker-" prepend?

JPWKU:
recheck

chuyval:
@SteveOss, I was just keeping things consistent with the most recent docker image naming convention in the docker repo. I agree with removing the 'docker-' from the name for same reason, redundancy. 

@JPWKU, Are you suggesting having two tags being '$(VERSION)-dev' and 'latest', removing the current $(VERSION) tag? I'd be ok with this since it would indicate to users that the tag thats versioned with '-dev' is still under works and the 'latest' is just so users are able to pull a docker container from repository without specifying a version in their yml file, or else user would have to check repo and specify the version they want to download.

JPWKU:
@chuyval You could just leave the tags you have in place and add the '$(VERSION)-dev' tag as a third.  It is not expensive to add a tag as they are just pointers to the image digest.

drasko:
Also - this attaches `latest` tag to newly produced images, but how this tag is removed later, when this version is not the `latest`. I think it is better to produce 2 different images with 2 tags and upload both to `DockerHub`, and in this way `latest` will always be overwritten.

Or am I getting something wrong?

Issue #187, opened 2018-05-04
Improve core-data add event performance


jduranf:
In order to improve the add event performance in core-data, which should be called several times, there are different things which can be done:

1. As said in #128, why use Device property as dual purposed? I think should be better to only allow to use device name

```
// Get device from metadata
deviceFound := true
// Try by ID
d, err := mdc.Device(e.Device)
if err != nil {
	// Try by name
	d, err = mdc.DeviceForName(e.Device)
	if err != nil {
		deviceFound = false
	}
}
```

2. If MetaDataCheck isn't enabled, why get Device name in metadata? This requires some time which should be avoided if MetaDataCheck isn't enabled
```
// See if metadata checking is enabled
if configuration.MetaDataCheck && !deviceFound {
	loggingClient.Error("Device not found for event: "+err.Error(), "")
	http.Error(w, err.Error(), http.StatusNotFound)
	return
}
```
3. Now ValueDescriptorByName is called if ValidateCheck is enabled (inside isValidValueDescriptor) and also if PersistData is enabled. This second check should be removed
```
if configuration.ValidateCheck {
	loggingClient.Debug("Validation enabled, parsing events")
	for reading := range e.Readings {
		valid, err := isValidValueDescriptor(e.Readings[reading], e)
		if !valid {
			loggingClient.Error("Validation failed: %s", err.Error())
			return
		}
	}
}

// Add the readings to the database
if configuration.PersistData {
	for i, reading := range e.Readings {
		// Check value descriptor
		_, err := dbc.ValueDescriptorByName(reading.Name)
```

4. AddReading should be called inside of AddEvent. This will allow to use transactions or similar in data base, which increases considerably the performance (I performed some tests using InfluxDB and other databases, with an important reduction of time)
```
	// Add the reading
	id, err := dbc.AddReading(reading)
	if err != nil {
		http.Error(w, err.Error(), http.StatusServiceUnavailable)
		loggingClient.Error(err.Error())
		return
	}

	e.Readings[i].Id = id // Set the ID for referencing later
}

// Add the event to the database
id, err := dbc.AddEvent(&e)
if err != nil {
	http.Error(w, err.Error(), http.StatusServiceUnavailable)
	loggingClient.Error(err.Error())
	return
}
```
5. There are several logs which can be removed (or changed to debug)

All this issues are still present in the refactored branch. If you want, I can help in order to solve them.

tsconn23:
Regarding your points:

[...]

Issue #128, opened 2018-04-10
Readings pushed to Core Data trigger Invalid Object ID error messages from Core Metadata


tonyespy:
While testing my edgex core snap (version: 0.5.1+cali-20180322) on an Ubuntu 16.04 LTS Desktop system (amd64), I noticed the following error messages every time a new event is pushed from device-virtual to core-data output on the terminal used to start edgex:

_INFO: 2018/04/10 09:39:18 Posting Event: {"id":"","pushed":0,"device":"KMC.BAC-121036\
CE01","created":0,"modified":0,"origin":1523367558014,"schedule":null,"event":null,"r\
eadings":[{"id":"","pushed":0,"created":0,"origin":1523367558014,"modified":0,"device\
":"KMC.BAC-121036CE01","name":"AnalogValue_20","value":"41.76"}]}

ERROR: 2018/04/10 09:39:18 mgoGetDeviceById Invalid Object ID KMC.BAC-121036CE01_

The Error message is being generated by the function mgoGetDeviceById() (src/core/metadata/mongoOps.go), in response to a call from Core Data's eventHandler function (src/core/data/event.go):

```
     // Get device from metadata
     deviceFound := true
     // Try by ID
     d, err := mdc.Device(e.Device)
     if err != nil {
       // Try by name
       d, err = mdc.DeviceForName(e.Device)
       if err != nil {
         deviceFound = false
       }
     }
     // Make sure the identifier is the device name
     if deviceFound {
       e.Device = d.Name
     }

     // See if metadata checking is enabled
     if configuration.MetaDataCheck && !deviceFound {
       loggingClient.Error("Device not found for event: "+err.Error(), "")
       http.Error(w, err.Error(), http.StatusNotFound)
       return
     }

```
There are a few problems with this code:

 * it does a lookup by deviceId, however the event from device-virtual doesn't include a deviceId, it just includes a devicName, hence this call **will always fail**
 * the code checks after the lookups to see whether a MetadataCheck was required, thus the code makes two REST API calls, neither of which may have been required
 * the REST call returns the entire device, however it's then just discarded.  Ideally, Core Metadata should expose DeviceExists endpoints that just return bool results.  In lieu of this change, the metadata client could define similar functions. Passing `models.Device` structs around by value, especially when they're not being used is just wasted compute cycles.
[...]

PR #177, opened 2018-04-26
#131 Replacing the fmt.Println code with log framework api for the whole project.

yanghua:
Signed-off-by: yanghua <yanghua1127@gmail.com>

cc @drasko

Issue #131, opened 2018-04-12
Replacing the fmt.Println code with log framework api for the whole project.


yanghua:
If agree this ticket, I could fix it.

drasko:
OK for me, but let @jpwhitemn or @tsconn23 confirm.

yanghua:
@jpwhitemn and @tsconn23 , what's your opinion?

tsconn23:
Can you provide a specific example of the usage you want to replace and the specific proposed change? We may want to handle this as more of a style guideline with issues sent to the specific teams responsible for their portion of EdgeX.

yanghua:
@tsconn23 there are many examples, such as [print error](https://github.com/edgexfoundry/edgex-go/blob/master/core/clients/metadataclients/metadata-clients.go#L169), [print debug](https://github.com/edgexfoundry/edgex-go/blob/master/core/clients/metadataclients/metadata-clients.go#L270), [print error](https://github.com/edgexfoundry/edgex-go/blob/master/core/data/init.go#L80), [print info](https://github.com/edgexfoundry/edgex-go/blob/master/cmd/support-logging/main.go#L54) and so on. I agree with you, the log level should be decided by the specific teams(maintainer). However, currently we could do some common refactor. For example we could replace this mode : 

```go
if err != nil {
  fmt.Println(err.Error())
}
```
with : 

```go
if err != nil {
  logger.Error(err.Error())
}
```

The reason about replacing the print code is that we should unified log output format and code style.

tsconn23:
@drasko Just to clarify, are you agreeing with me or with yanghua's approach to introduce the logger everywhere including packages where it may not be referenced yet?

yanghua:
@tsconn23 consider a scenario, if a module which used `fmt.println` to print a error message to std out.  But the log framework configuration tends to log the message into a file. In this case, the system's log is incomplete and would stop developer making diagnosis about issues.

I think import a logger package is not the core problem. Maybe it do not need a logger (for example just print to std out now), but it would need logger, especially in complex production and distributed environment.

drasko:
@tsconn23 - I would go tith @yanghua's proposition for now - looks simpler to me (pure change)

drasko:
@yanghua @tsconn23 yep, this "stdout logger" does not mean anything - in production logs would be sent directly from logger to `syslog`, some file or - most probably - a remote machine. This is why I insisted from the start on removing pritfs and using logging.

[...]

PR #199, opened 2018-05-08
Eliminate Redundant Device Lookups Across Service Boundaries by ID/Name

tsconn23:
This change refers to a couple issues:
https://github.com/edgexfoundry/edgex-go/issues/128
https://github.com/edgexfoundry/edgex-go/issues/187 (point 1)

Please see the commit notes for specifics but the gist is two-fold:
1.) Preference is given to device name when lookups are performed, so the new CheckForDevice() method assumes it has been given a name. This eliminates call paths which were always calling by ID first.
2.) If we have to perform 2 lookups -- name then ID -- this is now done solely within the metadata service. It does not require two round trips from core-data to metadata.

I did not remove the old Device and DeviceByName methods yet. Also, these changes have all been verified successfully against the suite of core-data blackbox tests -- events, readings and value_descriptors.


core/clients/metadataclients/metadata-clients.go
```diff
@@ -301,6 +302,46 @@ func (d *DeviceRestClient) decodeDevice(resp *http.Response) (models.Device, err
 	return dev, err
 }
 
+//Use the models.Event.Device property for the supplied token parameter.
+//The above property is currently double-purposed and needs to be refactored.
+//This call replaces the previous two calls necessary to lookup a device by id followed by name.
+func (d *DeviceRestClient) CheckForDevice(token string) (models.Device, error) {
+	req, err := http.NewRequest(http.MethodGet, d.url+"/check/"+token, nil)
+	if err != nil {
+		fmt.Println(err)
```
falkerson: with this https://github.com/edgexfoundry/edgex-go/pull/177 fmt is going to be replaced with logging-client. Also you already use this in code below. So could we have same approach everywhere?

core/data/event.go
```diff
@@ -219,23 +191,22 @@ func eventHandler(w http.ResponseWriter, r *http.Request) {
 		loggingClient.Info("Posting Event: " + e.String())
 
 		// Get device from metadata
-		deviceFound := true
-		// Try by ID
-		d, err := mdc.Device(e.Device)
+		d, err := mdc.CheckForDevice(e.Device)
```
jduranf: I think this check should be performed only if MetaDataCheck or DeviceUpdateLastConnected or ServiceUpdateLastConnected are enabled. Otherwise we are performing a useless check...
Another option is remove MetaDataCheck.

core/data/event.go
```diff
@@ -328,25 +299,26 @@ func eventHandler(w http.ResponseWriter, r *http.Request) {
 
 		// Update the fields
 		if from.Device != "" {
-			deviceFound := true
-			d, err := mdc.Device(from.Device)
+			d, err := mdc.CheckForDevice(from.Device)
```
jduranf: As before, only perform this check if MetaDataCheck is enabled. Otherwise, remove MetaDataCheck

core/data/event.go
```diff
@@ -446,16 +417,16 @@ func eventCountByDeviceIdHandler(w http.ResponseWriter, r *http.Request) {
...
-		d, err := mdc.Device(id)
+		d, err := mdc.CheckForDevice(id)
 		if err != nil {
-			// Try by Name
-			d, err = mdc.DeviceForName(id)
-			if err != nil {
-				loggingClient.Error("Device not found for event: "+err.Error(), "")
+			loggingClient.Error(fmt.Sprintf("error checking device %s %v", id, err))
+			switch err := err.(type) {
+			case types.ErrNotFound://ignore
```
jduranf: Remove ignore comment

core/data/event.go
```diff
@@ -569,23 +540,18 @@ func getEventByDeviceHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Get the device
-	deviceFound := true
-	// Try by ID
-	d, err := mdc.Device(deviceId)
+	d, err := mdc.CheckForDevice(deviceId)
```
jduranf: Only perform this check if MetaDataCheck is enabled. There are more cases like this one.

core/data/event.go
```diff
@@ -219,23 +191,22 @@ func eventHandler(w http.ResponseWriter, r *http.Request) {
...
 		if err != nil {
-			// Try by name
-			d, err = mdc.DeviceForName(e.Device)
-			if err != nil {
-				deviceFound = false
+			loggingClient.Error(fmt.Sprintf("error checking device %s %v", e.Device, err))
+			switch err := err.(type) {
+				case types.ErrNotFound://ignore
+			default: //return an error on everything else.
+			    loggingClient.Debug(fmt.Sprintf("error of type %v", err))
```
jduranf: This log can be removed, we already logged 4 lines before

falkerson:
Please add unit tests

core/clients/metadataclients/metadata-clients.go
tsconn23: I have no intention of accepting PR 177 because I'm not in agreement with it. See this thread
https://github.com/edgexfoundry/edgex-go/issues/131

Therefore I'm not taking those code changes into account. If someone else accepts the PR, so be it.

core/data/event.go
tsconn23: removed with above changes

core/data/event.go
tsconn23: making changes

core/metadata/raml/core-metadata.raml
```diff
@@ -47,6 +47,24 @@ schemas:
...
+            required: true
+            repeat: false
+    get:
+        description: Obtain a device by either name (preferred) or id. Id must be a valid BSON identifier.
+        responses:
+            "200":
+                description: JSON payload representing the device
+            "500":
+                description: unexpected server error
+            "404"
```
jduranf: Missing dots `:`

core/data/reading.go
```diff
@@ -299,16 +290,18 @@ func readingByDeviceHandler(w http.ResponseWriter, r *http.Request) {
 		// Try to get device
 		// First check by name
 		var d models.Device
-		d, err := mdc.DeviceForName(deviceId)
+		d, err := mdc.CheckForDevice(deviceId)
```
jduranf: This check can be avoided using MetaDataCheck

core/data/event.go
```diff
@@ -47,44 +48,21 @@ func updateDeviceLastReportedConnected(device string) {
 		return
 	}
 
-	t := time.Now().UnixNano() / int64(time.Millisecond)
-
 	// Get the device by name
-	d, err := mdc.DeviceForName(device)
+	d, err := mdc.CheckForDevice(device)
```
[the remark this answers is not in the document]
jduranf: ok

core/metadata/raml/core-metadata.raml
tsconn23: I hate this hand-written RAML crap

core/data/reading.go
tsconn23: done

PR #208, opened 2018-05-09
Add git_sha label during build of images

JPWKU:
It is easier for me to filter on a label
and less error prone.  Additionally, labels are
internal to an image and therefore better suited
in most cases.


JPWKU:
@jpwhitemn I did a rebase and pushed again; it removed your approval.

Issue #216, opened 2018-05-12
Drop attribution lines from support service files


yanghua:
cc @tsconn23 : 
shall we do this refactor like core service #213 ?
If need, I claim it.

tsconn23:
@yanghua Sure, go ahead.

yanghua:
thanks @SteveOss closing this...

PR #217, opened 2018-05-13
#216 Drop attribution lines from support service files

yanghua:

cc @tsconn23


yanghua:
the reason about the CI build error is not because of the PR's changed code

tsconn23:
recheck

Issue #223, opened 2018-05-13
Metadata /api/v1/command Put method error


weichou1229:
run `http://localhost:48081/api/v1/command` with put method will occur error

api request body
```
{
    "id": "57f396f8e4b060e2071a5c2d",
    "name": "current",
    "put": {
      "path": "/current",
      "response": {
        "code": "200",
        "errorDescription": "not in amps",
        "expectedValues": "amps"
      }
    }
}
```
test data in mongo
```
{
  "_id": {
    "$oid": "57f396f8e4b060e2071a5c2d"
  },
  "_class": "org.edgexfoundry.domain.meta.Command",
  "name": "current",
  "get": {
    "path": "/current",
    "response": {
      "code": "200",
      "errorDescription": "not a float",
      "expectedValues": "amps"
    }
  },
  "put": {
    "path": "/current"
  },
  "created": 1475581688051,
  "modified": 1475581688051,
  "origin": 1471806386919
}
```

error log
```
2018/05/13 13:35:21 http: panic serving 172.18.0.1:37982: runtime error: invalid memory address or nil pointer dereference
goroutine 11153 [running]:
net/http.(*conn).serve.func1(0xc42017c0a0)
        /usr/lib/go/src/net/http/server.go:1697 +0xd2
panic(0x55987ae19320, 0x55987b01ff20)
        /usr/lib/go/src/runtime/panic.go:491 +0x287
github.com/edgexfoundry/edgex-go/core/metadata.mgoUpdateCommand(0xc4201ca0f0, 0xc4201ca190, 0x0, 0x0)
        /go/src/github.com/edgexfoundry/edgex-go/core/metadata/mongoOps.go:1092 +0x312
github.com/edgexfoundry/edgex-go/core/metadata.updateCommand(0xc4201ca0f0, 0xc4201ca190, 0x0, 0x0)
        /go/src/github.com/edgexfoundry/edgex-go/core/metadata/dbOps.go:562 +0x44
github.com/edgexfoundry/edgex-go/core/metadata.restUpdateCommand(0x55987b030440, 0xc42033c000, 0xc420190300)
        /go/src/github.com/edgexfoundry/edgex-go/core/metadata/rest_command.go:113 +0x32c
net/http.HandlerFunc.ServeHTTP(0x55987ae8ccf8, 0x55987b030440, 0xc42033c000, 0xc420190300)
        /usr/lib/go/src/net/http/server.go:1918 +0x46
[...]

PR #224, opened 2018-05-14
Addressing panic from issue #223

tsconn23:
The approach here was to utilize UnmarshalJSON() on the Command domain
model in order to ensure that both Get and Put are initialized. The
bug was caused because the sample request did not have a GET so when
Mongo tried to call Get.String() it panicked due to NIL reference.

I wanted a solution that would NOT entail adding nil checks throughout the code base. We should be able to rely on whether or not a given property is initialized when deserializing.

#223 

### DIFF
--- a/core/domain/models/command.go
+++ b/core/domain/models/command.go
@@ -70,6 +70,27 @@ func (c Command) String() string {
 	return string(out)
 }
 
+func(c *Command) UnmarshalJSON(b []byte) error {
+	type Alias Command
+	alias := &struct {
+		*Alias
+	} {
+		Alias: (*Alias)(c),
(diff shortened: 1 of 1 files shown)


core/domain/models/command.go
```diff
@@ -70,6 +70,27 @@ func (c Command) String() string {
 	return string(out)
 }
 
+func(c *Command) UnmarshalJSON(b []byte) error {
```
tsconn23: The approach below was taken from the following article and is utilized in order to prevent an endless loop when calling json.Unmarshal().

http://choly.ca/post/go-json-marshalling/

Issue #173, opened 2018-04-26
Refactor the go constant with the unified mixed caps


yanghua:
suggested : `MaxLength`
Not suggested : `MAXLENGTH`, `MAX_LENGTH`
discuss refer to : [Go naming conventions for const
](https://stackoverflow.com/questions/22688906/go-naming-conventions-for-const)

claim it.

yanghua:
@drasko do you agree this?

yanghua:
@jpwhitemn @tsconn23 any idea?

drasko:
Yes.

drasko:
@yanghua here you go ;)

yanghua:
hi @jpwhitemn and @tsconn23 is issue will bring a big change, should it waiting for the refactor you are doing?

tsconn23:
@yanghua Don't start on this yet please.

PR #227, opened 2018-05-15
Relocated constants for service naming/keys

tsconn23:
Partially related to #173.

This will be used in the work for #182. Some global location holding the keys to services registered in Consul has to be available for all services. Instead of having the URL in a config file, the key will be in Consul and a dependent service has to know what to ask for. In other words, core-data needs to know core-metadata's key without referencing /core/metadata.


xmlviking:
Merging in issue_173 onto main.

Issue #225, opened 2018-05-15
ThingsBoard IoT export connector


jduranf:
We are starting to work in a export connector to allow EdgeX to export device data to [ThingsBoard](https://thingsboard.io/) using MQTT and a custom JSON format.
As MQTT is already implemented this connector will only imply add a new format to export-distro, maybe called FormatThingsBoardJSON.
If everything goes fine, we will have this connector implemented in just few days.
Please feel free to add any comment.

jpwhitemn:
This is great Joan.
 Going forward, we want to make the export distro more modular and scalable so that connectors like this will be easier to put together and also configurable for inclusion or exclusion at deployment time.  But for now, this helps extends EgdeX's reach to more cloud and enterprise platforms.

    On Tuesday, May 15, 2018, 6:50:43 AM CDT, Joan Duran <notifications@github.com> wrote:  
 
 
We are starting to work in a export connector to allow EdgeX to export device data to ThingsBoard using MQTT and a custom JSON format.
As MQTT is already implemented this connector will only imply add a new format to export-distro, maybe called FormatThingsBoardJSON.
If everything goes fine, we will have this connector implemented in just few days.
Please feel free to add any comment.

jpwhitemn:
Putting this in the Delhi Milestone, but if it is available for California - more better

Wanderzhd:
Hi Joan,

Thanks for adding in Thingsboard connector.

As to the format, if we want to send data to thingsboard, what's the format we need to set? thanks.
format: JSON or XML are your options at this time

jduranf:
As a format you must use `THINGSBOARD_JSON`

Wanderzhd:
Thank you Joan.

With below config, still failed to get any msg in ThingsBoard server side.  Any suggestion?
curl -X PUT -d '{
    "name":"1",
    "addressable":{
        "name":"ThingsBoard",
        "protocol":"tcp",
        "address":"52.80.40.46",
        "port":9000,
        "user":"boxing",
        "publisher":"EdgeXExportPublisher",
        "topic":"v1/devices/me/telemetry"
    },
    "format":"THINGSBOARD_JSON",
    "filter":{
        "deviceIdentifiers":["Random-Integer-Generator01"]
    },
    "enable":true,
    "destination":"MQTT_TOPIC"
}'

Wanderzhd:
After changed the topic to v1/gateway/telemetry, the issue resolved.

jduranf:
[...]

PR #238, opened 2018-05-18
Added export ThingsBoard JSON format

jduranf:
Related with issue #225 

More information about this JSON format:
https://thingsboard.io/docs/reference/gateway-mqtt-api/#telemetry-upload-api

### DIFF
--- a/export/distro/format.go
+++ b/export/distro/format.go
@@ -38,3 +38,34 @@ func (xmlTr xmlFormater) Format(event *models.Event) []byte {
 	}
 	return b
 }
+
+type thingsboardJSONFormater struct {
+}
+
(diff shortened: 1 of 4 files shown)


export/distro/format.go
```diff
@@ -38,3 +38,34 @@ func (xmlTr xmlFormater) Format(event *models.Event) []byte {
 	}
 	return b
 }
+
+type thingsboardJSONFormater struct {
```
tsconn23: Minor typo -- should have two "tt"s. thingsboardJSONFormatter

jduranf: Completely agree. But, should I fix all other "formater"?

feclare:
recheck

PR #239, opened 2018-05-18
DeviceClient Initialization and Monitoring via Service Discovery

tsconn23:
Part 1 of #182 

Here's a summary because I know there's a lot here. I'm putting this in now before it gets too big and because what is here is functionally equivalent to the current code. There is more work to do in extending the initialization and monitoring pattern to the other service clients, as well as handling exceptions at the point of making a call.

1.) Removed redundancy of pkg naming in core/clients packages: metadataclients --> metadata
2.) Revised initialization of DeviceClient via ServiceParams, moving us toward using host/port instead of full URL (it's not there yet, but this is where it will go)
3.) Async goroutines in NewDeviceClient() monitor Consul and poll the latest endpoint every 15 seconds, keeping the data current.

Comments are welcome.


xmlviking:
We can update some of our comments later :)

tsconn23:
recheck

Issue #235, opened 2018-05-17
Change device_test from integration test to unit test


yanghua:
file location: core/clients/metadataclients/device_test.go

cc @tsconn23 and @drasko 

claim it~

tsconn23:
not at this time

PR #240, opened 2018-05-19
#235 Change device_test from integration test to unit test

yanghua:

cc @tsconn23 and @drasko

### DIFF
--- a/core/clients/metadataclients/device_test.go
+++ b/core/clients/metadataclients/device_test.go
@@ -61,8 +61,7 @@ func TestAddDevice(t *testing.T) {
 
 	receivedDeviceId, err := dc.Add(&d)
 	if err != nil {
-		t.Log(err.Error())
-		t.FailNow()
+		t.Error(err.Error())
 	}
 
 	if receivedDeviceId != addingDeviceId {
--- a/core/clients/metadataclients/device_test.go
+++ b/core/clients/metadataclients/device_test.go
@@ -14,88 +14,58 @@
 package metadataclients
 
 import (
-	"fmt"
 	"github.com/edgexfoundry/edgex-go/core/domain/models"
-	"gopkg.in/mgo.v2/bson"
-	"os"
+	"net/http"
+	"net/http/httptest"
 	"testing"
 )
 
 const (
-	deviceUrl        = "http://localhost:48081/api/v1/device"
-	addressableUrl   = "http://localhost:48081/api/v1/addressable"
-	deviceServiceUrl = "http://localhost:48081/api/v1/deviceservice"
-	deviceProfileUrl = "http://localhost:48081/api/v1/deviceprofile"
+	DeviceUriPath = "/api/v1/device"
 )
 
 // Test adding a device using the device client
 func TestAddDevice(t *testing.T) {
 	d := models.Device{
-		Addressable:    a,
+		Id:             "1234",
+		Addressable:    models.Addressable{},
 		AdminState:     "UNLOCKED",
 		Name:           "Test name for device",
 		OperatingState: "ENABLED",
-		Profile:        dp,
-		Service:        ds,
+		Profile:        models.DeviceProfile{},
+		Service:        models.DeviceService{},
 	}
 
-	_, err := dc.Add(&d)
+	addingDeviceId := d.Id.Hex()
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+
+		if r.Method != http.MethodPost {
+			t.Errorf("expected http method is %s, active http method is : %s", http.MethodPost, r.Method)
+		}
+
+		if r.URL.EscapedPath() != DeviceUriPath {
+			t.Errorf("expected uri path is %s, actual uri path is %s", DeviceUriPath, r.URL.EscapedPath())
+		}
+
+		w.Write([]byte(addingDeviceId))
+
+	}))
+
+	defer ts.Close()
+
+	url := ts.URL + DeviceUriPath
+
+	dc := NewDeviceClient(url)
+
+	receivedDeviceId, err := dc.Add(&d)
 	if err != nil {
 		t.Log(err.Error())
 		t.FailNow()
 	}
-}
 
-var dc DeviceClient
-var a models.Addressable
-var ds models.DeviceService
-var dp models.DeviceProfile
-
-// Main method for the tests
-func TestMain(m *testing.M) {
-	dc = NewDeviceClient(deviceUrl)
-	ac := NewAddressableClient(addressableUrl)
-	dsc := NewServiceClient(deviceServiceUrl)
-	dpc := NewDeviceProfileClient(deviceProfileUrl)
-
-	a = models.Addressable{
-		Address: "http://localhost",
-		Name:    "Test Addressable",
-		Port:    3000,
+	if receivedDeviceId != addingDeviceId {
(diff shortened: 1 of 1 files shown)


core/clients/metadataclients/device_test.go
```diff
@@ -14,88 +14,58 @@
...
-	id, err = dpc.Add(&dp)
+	}))
+
+	defer ts.Close()
+
+	url := ts.URL + DeviceUriPath
+
+	dc := NewDeviceClient(url)
+
+	receivedDeviceId, err := dc.Add(&d)
```
falkerson: Add method should return predictable output:
* it return device_id and err == nil
* return err

In other words it should be mocked and not depend on Add() logic because in such case it integration not unit test
In other words it should be mocked and does not depend on Add() logic because in such case it integration not unit test

core/clients/metadataclients/device_test.go
```diff
@@ -14,88 +14,58 @@
...
 	if err != nil {
-		fmt.Println("Error posting new device profile: " + err.Error())
-		return
+		t.Log(err.Error())
```
falkerson: Pls remove Log from unit test

core/clients/metadataclients/device_test.go
yanghua: @falkerson  you mean I should mock every return branch to cover the test coverage? Actually, I tested rest client. Because it is not necessary to test the client itself. I just want test the full http rest flow.

falkerson: @yanghua ideally yes. Unit testing tests piece of codes on some logic with most granular and isolated level and it shouldn't have any external dependencies like: open file descriptors, network connects etc. Also it shouldn't depend on output of some another function because behavior of unit test became unpredictable. For me if we have 'if' statements in unit test means thats smth wrong with test itself and if couldn't easily mock some internal functions smth wrong with code design. And full flow testing is kind of integration test. But it depends on what "unit" means for us (https://martinfowler.com/articles/mocksArentStubs.html#TestIsolation)

yanghua: @falkerson I agree the unit test case should promote the test coverage, it's best practice. But I don't agree the unit test should not contain "if statements", here you can consider it as `assert` `assertEquals` in JUnit (the most popular unit test framework for Java), the different is `assert` encapsulates the "if statements". About the external dependencies, I think it depends on the context of code. Just like here, the rest client must make a http request to the microservice, it's the part of the business logic. Consider, if it can not make a http request and get a response, Can it return successfully? Of cause not. So for test it, I mocked a local http server used Golang's httptest framework and let the `Add` method passthrough normally. 

If there is another method use `DeviceClient` we could implement a `MockDeviceRestClient` and let the `Add` method do not make a real http request (for example, just return a const). But here we need test a rest client itself, so we need mock a http server.

So I think we should promote the test coverage, I will do it if I have enough time. This issue's purpose is remove the dependency to  real microservices.

yanghua:
cc @tsconn23

core/clients/metadataclients/device_test.go
falkerson: @yanghua about 'if' statement I mean:

receivedDeviceId, err := dc.Add(&d)

So in one case err == nil and in another not. What case we have in this test case function? dc.Add(&d) should be **idempotent** and return same result every time test runs. If err will not be nil test ends with t.Error(err.Error()) and if receivedDeviceId != addingDeviceId doesn't make sense. In another case if err !=nil {} is redundant because err always be nil.

tsconn23:
I'm not convinced unit testing the clients as they're currently implemented provides any value. There's no business logic in the clients. They a pass-through layer facilitating a service call.

If you look at PR 182, there is a test for the DeviceClient that specifically tests the initialization logic of the client itself with regard to service discovery. That is a responsibility encapsulated within the client that needs to be verified.

But otherwise I don't see much point in unit testing a pass through layer that proxies and relies on an actual service to do anything useful.

core/clients/metadataclients/device_test.go
yanghua: @falkerson you are caught in this function's current implementation, `Add` method returns error also may cause by the `json.Marshal`, I said when I have enough time I would give more input data case to promote the test coverage. If I build a invalidated Device model, it will return error.

yanghua:
@tsconn23 this PR used to fixed a test error, if there are not some real microservice about device and so on. 

Actually, I think the rest client need a unit test. And current rest client's business logic is not completely. There missed some validate、check task about business rule and security. We can not trust user's input data. Just think about in a web page, you give user a form, user fill the form, when sent a request to server, you must validate in the client end.

So I think the unit test is necessary in everywhere, it could validate the change later is right. See a problem cannot look only at present. The test case should be enhanced later and the rest client's validation also need be enhanced.

@jpwhitemn and @drasko what's your opinion?

drasko:
@yanghua I am not against accepting this PR just because we are still in the process of heavy dev, and you are not touching the main code. However - I do agree with @falkerson - functions you call must be mocked and have predictive and idempotent responses. You should test all cases and you should be capable to predicatively produce all these cases.

yanghua:
@drasko yes, you and @falkerson are right. Moreover, as I said before, I would enhance the test case if I have time in another issue. But this issue is used to fix the test error, when I was replacing the `fmt.Println` to logger, I invoked `go test` then found it was a integration test case, so I did this.  And based on `one PR , one thing`, this PR just for repairing.

drasko:
@yanghua understood. Just one more question - if the tests are failing without this fix, how come that CI passed before?

yanghua:
@drasko I don't know. Maybe the CI started the specific dependency microservices?  You can see this PR's code changes. The old code depends on some microservices. If just run `go test` without started those local services, would report http error, because it makes a real http request, this is the reason why I used httptest framework to build a mock server.

Actually, in a user's local, run `go test` should not contains other dependencies over the project.

yanghua:
@drasko it shows this : 

```
Post http://localhost:48081/api/v1/addressable: dial tcp [::1]:48081: connect: connection refused
Error posting addressable: Post http://localhost:48081/api/v1/addressable: dial tcp [::1]:48081: connect: connection refused
ok      github.com/edgexfoundry/edgex-go/core/clients/metadataclients   0.018s
```

drasko:
Would that mean that something is wrong with actual CI?

If that is the case, it would be good that @JPWKU take a look at it.

feclare:
@yanghua Do you reproduce the original unit test problem with ```make test``` ?

yanghua:
@drasko @feclare I never used `make test` before. 
But I used `make test` just now, it shows **OK** : 

```
ok      github.com/edgexfoundry/edgex-go/core/clients/metadataclients   0.033s
```

I always use `go test` in specific sub modules. Just now, when I removed the specific dependencies code, just leave device testing code, but also let it make a real http request, it failed : 

```
metadataclients vinoyang$ go test
Post http://localhost:48081/api/v1/device: dial tcp [::1]:48081: connect: connection refused
Post http://localhost:48081/api/v1/device: dial tcp [::1]:48081: connect: connection refused
--- FAIL: TestAddDevice (0.00s)
        device_test.go:42: Post http://localhost:48081/api/v1/device: dial tcp [::1]:48081: connect: connection refused
FAIL
exit status 1
FAIL    github.com/edgexfoundry/edgex-go/core/clients/metadataclients   0.018s
```

I think it shows `ok` before, because of the `addressable`'s http exception interrupts the real test code.

JPWKU:
Just saw this ping...I don't understand the issue yet, I'll have to dig in to see if this is a CI env issue.

yanghua:
@drasko merge this?

drasko:
Merged. Thanks!

Issue #245, opened 2018-05-20
Microservice Naming -- Format edgex-<layer>-<name>


tsconn23:
As found in the ServiceNameDesign document:

https://wiki.edgexfoundry.org/display/FA/Architecture+Issues+and+Decisions#ArchitectureIssuesandDecisions-ServiceIdentifiers

(Page 2)
b.	The proposed standard of the service API set name (aka service name) is edgex-<layer>-<name>. Examples:  edgex-core-data, edgex-device-modbus, etc.  TSC must govern the names of services (agreeing to the names of any new service API sets).

PR #246, opened 2018-05-20
Core/Support service naming per design standard 

tsconn23:
#245


tsconn23:
recheck

tsconn23:
@jduranf Yeah, I thought about that but left them in just to be safe, or in case someone brought up a concern about removing them during their review. If no one else comments to the contrary then I'll remove them.

tsconn23:
@feclare @jduranf Historically I've left those alone as the ownership of those services belongs to the Mainflux team. But due to the specific request, I'll make the changes here.

internal/constants.go
```diff
@@ -16,5 +16,7 @@ package internal
 
 const CommandServiceKey            = "edgex-core-command"
 const CoreDataServiceKey           = "edgex-core-data"
+const ExportClientServiceKey       = "edgex-export-client"
+const ExportDistroServiceKey       = "edgex-export-distro"
 const MetaDataServiceKey           = "edgex-core-metadata"
```
jduranf: Maybe the layer name can be added in "CommandServiceKey" and "MetaDataServiceKey", so it will be "CoreCommandServiceKey" and "CoreMetaDataServiceKey". This is how it's done with other service keys.

But anyway, looks good for me.

tsconn23:
I don't understand why all of those commits are listed as part of this PR. Most of those are committed to trunk already. Makes this PR look much larger than it is.

tsconn23:
recheck

tsconn23:
Due to merge conflicts I think the commit history on this PR has gotten messed up. I'm going to recreate the PR from a fresh branch.

PR #247, opened 2018-05-21
Reviewed core data checks

jduranf:
Reviewed all core data checks related with device name and value descriptor.
Also refactored value descriptor validation checks. Now they are performed by reading and not by event, because an event can have different readings with different value descriptors.


jduranf:
recheck

tsconn23:
Please hold off on any more merges until tomorrow May, 22nd CST hours.

tsconn23:
recheck

tsconn23:
Related #187

PR #248, opened 2018-05-21
Adding a new in memory database to core-data

feclare:
This PR adds a new database type to core-data that does not use any backend, only keeps the data in memory while it is running and data is lost when core-data stops. It is usefull while developing or demos.

It also add some generic unittest for DBClient, that could be used to test other implementation. Something similar to what is done in https://github.com/edgexfoundry/edgex-go/blob/master/export/client/clients/mongo-client_test.go


tsconn23:
recheck

tsconn23:
Please hold off on any more merges until tomorrow May, 22nd CST hours.

core/data/clients/memory-database_test.go
```diff
@@ -0,0 +1,612 @@
...
+	}
+}
+
+func testDB(t *testing.T, db DBClient) {
+	testDBReadings(t, db)
+	testDBEvents(t, db)
+	testDBValueDescriptors(t, db)
+
+	db.CloseSession()
+	db.CloseSession()
```
jduranf: This line should be removed

feclare: I'm calling the ```CloseSession``` a second time to test that there is no panic when closing a db already closed. I will add a comment.

jduranf: ok! thanks for the clarification

tsconn23:
recheck

PR #256, opened 2018-05-24
EF-255 - Replace Glide with Dep

### LINKED ISSUE #255: Switch to Go dep
Go [dep](https://github.com/golang/dep) is an official Go vendoring tool tad became suitable for production use.

Switch from `glide` should be fairly easy.

mijicd:
This pull request resolves #255.

**Summary:**
- removed `vendor/` from ignored files
- replaced Glide with Dep

**Notes:**

`vendor/` directory should not be removed in order to achieve truly reproducible build. Yes, the size of the repository will be increased, but it can be pruned (as it is done) to contain the bare minimum needed. This will also speed up the entire build flow, since there's no need for dependencies to be pulled everytime, and, what's even more dangerous, updated without control.


drasko:
Fantastic - thanks @mijicd!

@jpwhitemn @tsconn23 I know we are close to Cali release, but can we gave "go" for this one?

CC @feclare @JPWKU @tonyespy @janko-isidorovic

falkerson:
@mijicd about build:

1. If there is no need to fetch dependencies every time, it needs to fetch original repo with dependencies in vendor which is equivalent.
2. About updated without control: for such case we should have https://golang.github.io/dep/docs/Gopkg.lock.html or with Glide https://glide.readthedocs.io/en/latest/glide.lock/

Also could u please describe critical cases where godep overcome glide?

cc: @drasko

mijicd:
@falkerson not exactly. Presence of `vendor` would allow reproducible builds, and the same library dependencies. As you are probably aware, most of the packages listed in the dependencies are either missing an explicit version, or they don't have an explicit release number at all (this is a nasty part of some Go libraries). Letting glide (or dep) install the dependencies on each build would eventually (rather sooner than later) cause failures due to incompatibility between pulled libraries and the code itself. Moreover, each developer could end up using its own set of versions during the development.

drasko:
> Also could u please describe critical cases where godep overcome glide?

It is a project from Go team, no critical issues here - just maintainability and converging towards officially supported technology.

@mijicd I totally agree for `vendor`. Disappearing GitHub repos and broken dependencies are not something that is unusual for open-source projects. However, I think that explanation from LF was that they do not/can not host other projects (because of the licensing issues) in their repos.

@jpwhitemn can give more information. Maybe @JPWKU also.

drasko:
Again, I would strongly support adding `vendor` into Go monorepo, and if possible and needed I would like to revise this subject on TSC/WG meeting.

JPWKU:
Lets discuss this on today's Core WG call if possible.

mijicd:
@falkerson 

@mijicd

> sorry, maybe I've explain it wrong. This will also speed up the entire build flow, about equivalent I mean execution time for pulling sources: time to pull (repo + vendor) == time to pull (repo without vendor) + install with dep/glide.

Yes, at least they're comparable.

> glide allow to pin versions with glide.lock(same with dep). If there is no lock file it's our mistake but not glide approach.

I did not say Glide is evil, bad or having significant deficiencies, and I am more than well aware of its capabilities. The current codebase does not have any lock files, hence it is prone to all of the aforementioned problems.

> Currently I can't see advantages of this approach before glide. But I can be wrong.
> @drasko I wrote about dep here #255. It's not official tool but only candidate.

Yes, the dep is still an "official prototype", and it is not an official tool. However, please note that even Glide's authors suggest ["migrate if possible"](https://github.com/Masterminds/glide#golang-dep) approach.

drasko:
>  I wrote about dep here #255. It's not official tool but only candidate.

@falkerson so what? Glide's not even a candidate.

What's important is that it is obvious that Go community is converging towards this tool and that the tool is mature enough (it has not been at the time we were starting with the Go port).

feclare:
Makefile need fixing, removing the glide install from the prepare rule.

mijicd:
@feclare done, see commit message for explanation.

drasko:
Thanks @mijicd. As I said - this looks great to me, I would really like that @jpwhitemn greenlights this.

I do not see any risk in merging this one, on the contrary.

Only thing that needs to be checked is `vendor` presence and LF rules, but I really do not see why we would not have open-source dependencies in our repo, especially that it would guarantee availability and reproducibility of EdgeX product.

jpwhitemn:
All, really appreciate the work and commentary on this.  This requires more of the community to weigh in and provide feedback.  Also, we don't want to do this before the California release. I ask that no one merge this PR until we have larger community input and agreement.  Topic for upcoming F2F.

mijicd:
@jpwhitemn This is somewhat understandable, though I am not sure whether it is meaningful to have a time slot for debate about something that is Go community recommendation (direction?), and it took ~10 minutes to implement.

Having said that I propose this pull request to be either merged or closed together with related issue.

xmlviking:
I should not have merged this pull request in prior to California as per Jim .  This change should be reverted until after the California release.

mijicd:
Too much noise over nothing. Jim clearly stated in the comments here that he does not want this to be merged prior to community agreement. Since it was merged, I would suggest that it either stay as it is (it's a 10-minute fix as I said, no need for debating) or revert the merge and reject this PR for good.

JPWKU:
This was reverted last night, we will revisit it as soon as California is released.

PR #257, opened 2018-05-24
Improved adding an event with readings performance in core-data

jduranf:
With this modification there is an important performance increase when adding an event with readings. 
I made a small benchmark adding events with 5 readings, and this are the results with the current code:

```
MongoDB - 392290 ns/op
InfluxDB - 15792943 ns/op
```

And the results with the new code are:

```
MongoDB - 162425 ns/op
InfluxDB - 3382725 ns/op
```

Related with #187 


tsconn23:
recheck

feclare:
@jduranf Do you think it is interesting to add the benchmark?

feclare:
@jduranf Great!
I have also seen some mongodb degradation over time.

Issue #198, opened 2018-05-08
Standard project layout


falkerson:
It's a good practice to have some standard project layout for go projects. With current approach all services stored in a root of project directory which could provide mess with time. Propose to take a look https://github.com/golang-standards/project-layout.

yanghua:
+1, looks good to me cc @drasko what's your opinion?

drasko:
My opinion is that premature optimization is the root of all evil.

jpwhitemn:
@tsconn23 - Trevor, please have a look at the above standard project layout.  How does this line up with our core service refactors?  Along with Drasko, I don't want to re-org if it does not offer some compelling assistance/improvements.  We need to understand how pervasive is this "standard" and is it adopted by many go projects and if it seems to fit/meet our needs.  If not, we need to close this issue.  If so, we need to prioritize for work in Delhi.

yanghua:
@drasko seems it is not a optimization (mostly not change the code), it's refactor about the project's structure. Also we should consider whether the 'standard' is a popular convention or not just like jim said.

falkerson:
@drasko @jpwhitemn it's not only about standard in that link. Project should have some well known layout. So it will be easy for every new contributor look through project structure and find interesting part. Example: for me as some new person without onboarding it was little not obviously to understand that directories in root related to services. If we won't such layout we need document somehow that structure

tsconn23:
FYI, I have a need when addressing Issue #182 whereby it would be useful to have the /internal directory with our current /pkg moved inside of that. The packages in /pkg are not intended for external use and so will be moved to /internal/pkg. See the following:
https://github.com/golang-standards/project-layout/tree/master/internal

Will be submitting PR shortly.

tsconn23:
Closing this issue as it isn't a specific actionable item, but has informed several PRs since its creation and will be taken under advisement during Delhi refactoring efforts.

PR #260, opened 2018-05-24
Create clients for readings and events

dylanhantula:


tsconn23:
@dylanhantula Please create an issue summarizing the goal of these changes and associate it to this PR.

A few comments based on my understanding that this is intended to be a client for external applications written in Go that would need to access the core-data service. Your summary on the issue above may clarify.
1.) I think a better location for these would be in a /pkg directory off the root. Please see #198 created by @falkerson
2.) If these are to be usable by external clients, we can't expose the internal domain model of EdgeX for use. Thus the import to "github.com/edgexfoundry/edgex-go/core/domain/models" will need to be removed. Usually, this would mean a library of externally facing immutable value objects that your client would map the internal domain into as return values. However, given Go's duck-typing preference, maybe this doesn't make sense. Perhaps a client like what you've got here that standardizes the behavior of the endpoint is enough, and we put the onus of deserialization on the client application by reference to either RAML or Swagger.

feclare:
recheck

tsconn23:
@dylanhantula @tonyespy OK, I'm willing to pull this in prior to California deadline given Tony's need. However the clients in this PR need to be updated:

1.) The New() functions need parameters of the types.EndpointParams and clients.Endpointer.
2.) The clients need an internal member "endpoint clients.Endpointer"
3.) Each client needs an init() method which activates the polling of Consul for the current endpoint
4.) I would like for each new client to have a test file with at least one test to cover the initialization with Consul enabled. See the "TestNewValueDescriptorClientWithConsul" test for example.

You can readily see examples of all the above in the current master branch, so it should just be a matter of following the pattern.

PR #262, opened 2018-05-28
Added MongoDB test in core-data

jduranf:
Added MongoDB test in core-data, also fixed two issues in MongoDB detected with the test:
- Readings using limits doesn't include start and end date
- Fixed issue in ValueDescriptorsByName

Along with adding MongoDB test, also added ScrubAllValueDescriptors to make tests more reproducible.

Finally, removed core-data mock database, as this is useless because we have memory database thanks to @feclare in #248.


feclare:
@jduranf did you run ```make test```. There are some test failing because of the db mock removal.

jduranf:
Thank you @feclare, I'm checking the issue. Is it work keeping mock client? Maybe it's better to just modify event_test.go... 
What do you think?

feclare:
I should remove it, memorydb could be used for the unittest instead of the mock, as you suggested. But lets @tsconn23 decide what should be done. There are tests that can be done with a mock and can't be done with the memorydb as it is implemented (like testing db returning errors)

jduranf:
Fixed the issue with the tests.

@tsconn23, what do you think about remove the mock database to use memory database?

tsconn23:
LGTM

PR #267, opened 2018-05-30
Removed unused NewDeviceClient err return type

tsconn23:
If you see the old code only ever returned a literal "nil" for error from NewDeviceClient. No sense in having it in the function signature.


xmlviking:
recheck

feclare:
Tests need to be fixed:
```
go test ./...
?   	github.com/edgexfoundry/edgex-go	[no test files]
# github.com/edgexfoundry/edgex-go/core/command
core/command/init.go:71:5: undefined: err
core/command/init.go:72:23: undefined: err
```

PR #268, opened 2018-05-30
Consul service lookup for metadata command

tsconn23:
Continuing #182 


xmlviking:
recheck

PR #269, opened 2018-05-30
Removed unused metadata device_profile client

tsconn23:
Continuing #182 

This client isn't used anywhere.


xmlviking:
💯

PR #271, opened 2018-05-31
Fix export-client when there is no db connection

feclare:
Making export-client exit when there no connection to the database


JPWKU:
recheck

JPWKU:
Looks like an intermittent network blip that caused that last failure.

feclare:
@JPWKU Yes, I seems we have some problem with the board connectivity. Sometimes we have some packet lost, but we haven't found the root cause yet.

Issue #278, opened 2018-06-04
BUG: Reinstate Deleted Clients -- Command, Device Profile, Addressable


tsconn23:
@tonyespy Has made it explicitly clear to me that I've broken his hard work on the device services due to the removal of the above clients. Even though these clients aren't used internally by the EdgeX platform and thus their absence doesn't affect the operation of the microservices, the clients are used by external applications such as the Device Service.

We have discussed reinstating these clients as they were for now so that Tony's Device Service work is restored, and then moving forward the clients will be moved to a directory indicating their availability for external use.

PR #279, opened 2018-06-05
Reinstate Deleted Clients -- Command, Device Profile, Addressable

tsconn23:
#278


tonyespy:
+1 from me, thanks @tsconn23 for the quick turnaround!

Issue #280, opened 2018-06-06
Integrate go report card service to give a check and add a badges for the project 


yanghua:
here is a report about edgex-go : https://goreportcard.com/report/github.com/edgexfoundry/edgex-go

cc @jpwhitemn @drasko

tsconn23:
I don't particularly see an action item here. It's a nice tool that we can run on the fly but this looks to me like an FYI rather than an issue to be solved. Could be shared via EdgeX-GoLang mailing list or, alternatively, RocketChat.

drasko:
@yanghua @tsconn23 I would like that we add this badge to our GitHub README.

@yanghua we need too do the linting first and push this to A+.

tsconn23:
I do not want a series of tiny code formatting changes interleaving with PRs to support the Delhi roadmap.

drasko:
@tsconn23 are you proposing that we whould reject this task and close the issue?

Our code is not linted: https://goreportcard.com/report/github.com/edgexfoundry/edgex-go#golint and having a badge on the GitHub page that shows A+ would be more attractive for contributors.

How do you propose that we do these code improvements, which are not big issues/milestones for Delhi? We just reject them and never improve code quality?

tsconn23:
@drasko Taking examples from #194 and #198 PRs have been submitted toward California goals which include refactoring that moves us toward the goals outlined in these recommendations. Why can't this issue be addressed the same way? 

OR, if you like, why not close this umbrella topic (which I maintain would be more relevant as a topic on a mailing list or Rocket Chat) and create an issue for each file called out in the linting step on the report card site and address a specific PR for each file being cleaned up? I do not want 30 PRs tied to this one issue and I do not want one massive PR addressing all of the report card edits.

yanghua:
I think open a issue is also a way of discussing. 

Most popular Go projects have this badge, such as : 

* [openfaas](https://github.com/openfaas/faas)
* [influxdb](https://github.com/influxdata/influxdb)
* [beego](https://github.com/astaxie/beego)
* ....

Does this a disincentive problem need to be discussed? Actually, I think it is an obvious improvement for EdgeX.
As @drasko said, this is good for EdgeX's code quality. If a stranger did not know about EdgeX. Some badges would be the primary impression, those badges would give the potential contributors confidence.

The first step, I just want to integrate with the go report card and add a badge for the project, not related to refactor the code, it's future work.

[...]

PR #281, opened 2018-06-07
#280 Integrate go report card service to give a check and add a badges for the project(part 1)

yanghua:

cc @drasko


yanghua:
cc @drasko updated

feclare:
@drasko We have some intermittent networks problems that we have not been able to solve yet. The arm64 build is not required yet. Feel free to merge PR with a failing arm64 build because of networks issues.

PR #297, opened 2018-06-13
EF-100 - Azure IoT Hub export connector

### LINKED ISSUE #100: Azure IoT Hub export connector
Duplicate functionality found in the Java export client and export distro services to allow for export to Azure IoT Hub

dusanb94:
This pull request resolves #100.

Signed-off-by: Dušan Borovčanin <borovcanindusan1@gmail.com>


export/distro/azure.go
```diff
@@ -0,0 +1,84 @@
+//
```
feclare: This file is almost the same as mqtt.go. Why duplicate the file instead of using only one?

dusanb94: I agree. Since there is no suitable Azure SDK for devices written in the Go, sending MQTT will work. Someday when proper SDK is released, we can update this feature.

export/distro/format.go
```diff
@@ -9,14 +9,27 @@ package distro
 import (
```
drasko: If adding Mainflux contributions, please add Mainflux to copyright header.

export/distro/registrations.go
```diff
@@ -85,7 +85,7 @@ func (reg *registrationInfo) update(newReg export.Registration) bool {
 	case export.DestIotCoreMQTT:
 		reg.sender = NewIoTCoreSender(newReg.Addressable)
 	case export.DestAzureMQTT:
```
jduranf: DestAzureMQTT can be removed, as there is DestMQTT with the same functionality

PR #309, opened 2018-06-19
Add snap packaging

tonyespy:
This commit adds a new directory called /snap which contains
the packaging for the edgexfoundry snap.  The snap builds and
includes all of the microservices from the mono-repo, as well
as the two Java-based supports services and device-virtual.
The snap also includes consul, mongodb, and the Java-based
core-config-seed.


tsconn23:
@tonyespy Sign-off is missing

tonyespy:
Added to the commit, and re-pushed...

tsconn23:
@drasko But, to be fair, we have a docker folder off the root and Tony's just emulating that. I think we can refactor toward a /platform/docker and /platform/snap post California. @tonyespy What do you think of that? Otherwise, I'm good with this.

tonyespy:
A top-level packaging directory is a pretty standard approach (eg. Debian, Docker, ...).  At this time, snapcraft only works with a top-level /snap directory, so moving our packing directories under /platform won't work.

tonyespy:
And by the way, the generated binary snap files will not be checked into repo, and I made sure to include *.snap in .gitignore.  The snap packaging will be used by CI to generate architecture snap packages which then will be published to the snap store.

PR #327, opened 2018-06-24
Add InfuxDB support in export distro

davidmsh:
- Change the Send function of Sender interface to pass in the original event as we don't need the formatted message but the readings in the event

- Add a "NOOP" formatter which, as its name suggests, does nothing. This formatter is used when we do not used the formatted form of the event

- Add a new "INFLUXDB_ENDPOINT" destination type and implement NewInfluxDBSender
to handle it by sending the reading to influxdb


cmd/core-metadata/res/configuration.toml
```diff
@@ -1,7 +1,7 @@
 DBType = 'mongodb'
 MongoDatabaseName = 'metadata'
-MongoDBUserName = ''
-MongoDBPassword = ''
+MongoDBUserName = 'meta'
```
drasko: Why is this needed? Why not leaving empty strings?

cmd/export-client/res/configuration.toml
```diff
@@ -2,8 +2,8 @@ Hostname = '127.0.0.1'
 Port = 48071
 DBType = 'mongodb'
 MongoURL = '0.0.0.0'
-MongoUsername = ''
-MongoPassword = ''
+MongoUsername = 'core'
+MongoPassword = 'password'
 MongoDatabase = 'coredata'
```
drasko: Looks like in other `toml` we are using `MongoDatabaseName`. Please align.

export/distro/format.go
```diff
@@ -167,3 +168,11 @@ func (af azureFormatter) Format(event *models.Event) []byte {
 	}
 	return msg
 }
+
+type noopFormatter struct {
+}
+
+func (noopTr noopFormatter) Format(event *models.Event) []byte {
+	var ret []byte
+	return ret
```
drasko: No need for declaring a var. You can `ret []byte` directly.

BTW. why is this a separate function? When you need empty byte array in the code, can't you just declare the one over there?

export/distro/format.go
```diff
@@ -167,3 +168,11 @@ func (af azureFormatter) Format(event *models.Event) []byte {
 	}
 	return msg
 }
+
+type noopFormatter struct {
```
drasko: I do not see use in declaring an empty struct as a separate type...

export/distro/influxdb.go
```diff
@@ -0,0 +1,107 @@
...
+	httpInfo client.HTTPConfig
+	database string
+}
+
+func NewInfluxDBSender(addr models.Addressable) Sender {
+	connStr := "http://" + addr.Address + ":" + strconv.Itoa(addr.Port)
+
+	influxdbHTTPInfo := client.HTTPConfig{
+		Addr:     connStr,
+		Timeout:  time.Duration(60000) * time.Millisecond,
```
drasko: Do not hardcode numeric constants. No magic numbers, please declare a const.

export/distro/format.go
SteveOss: Slightly disagree here. The type name does impart some additional information i.e the intended usage. In C/C++ quite often see string types, typedef'd to something more meaningful.

export/distro/format.go
feclare: @drasko They just want to create a type that satisfies the ```Formatter``` interface. It does nothing because they only use the ```Sender``` interface of the distro pipeline. 
How do you recommend to do that without defining a type?

feclare:
@davidmsh Tests need to be fixed. 

Please run ```make test``` before creating a PR.

export/distro/format.go
drasko: Especially C++ produces a lot of bloat that go philosophy tries to avoid. Feel free to return empty arrays or strings directly from the code, leave a short comment if necessary, but oplease do not bloat the code.

Every code line is potential bug introduced, keep them at minimum.

export/distro/format.go
davidmsh: Looks like "return []byte" is not legal, I will change it to "return []byte{}"
As @feclare said, the separate function is to match the Formatter interface we already have for formatter. I return a byte array to make the function in the interface. It's empty because this formatter is meant to do nothing, that why it's called noop.

export/distro/format.go
davidmsh: The empty struct looked strange to me when I first saw the code as a new golang developer. But it's pretty much like a class to implement certain interface, just golang does require you list all the function in the interface in the implementing class.

export/distro/influxdb.go
davidmsh: Will do.

cmd/core-metadata/res/configuration.toml
davidmsh: I thought it's more convenient to leave some default values here. I can change it back.

davidmsh:
@drasko I have addressed your requested changes. Can you check again if you are ok with it ?

feclare:
@davidmsh Can you add a test for the added formatter in format_test.go? Just test that it outputs empty arrays.

davidmsh:
@feclare Done adding the test case for noop formatter

drasko:
@feclare and @tsconn23 please review this one, because it introduces `event` in `Send()` signature.

feclare:
Who is updating the configurations files in config-seed? Because there is a field change in this PR.

davidmsh:
@feclare You suggested to align the name MongoDatabase with others as MongoDatabaseName.

davidmsh:
@feclare Sorry, it's @drasko who made such suggestion

tsconn23:
@davidmsh Please rebase and commit your branch to clear the "out of date" message above then we can pull this in.

Issue #324, opened 2018-06-23
Add more methods in device profile client


jduranf:
Device service need more device profile client methods (core/clients/metadata/device_profile.go), specially query device profiles. Currently only Add method is implemented.
If there is no problem, I will work on it

tsconn23:
@jduranf Can you please make the title of this issue more specific? Also, can you enumerate the methods you plan to add and provide a summary of what they're for?

jduranf:
@tsconn23 I'm helping @tonyespy in gxds. When the device service starts, it needs to retrieve the device profiles used by the devices which use the device service. So the idea it's to modify the device profile in order to have those methods:
```
type DeviceProfileClient interface {
	Add(dp *models.DeviceProfile) (string, error)
	Delete(id string) error
	DeleteByName(name string) error
	DeviceProfile(id string) (models.DeviceProfile, error)
	DeviceProfileForName(name string) (models.DeviceProfile, error)
	DeviceProfiles() ([]models.DeviceProfile, error)
}
```

PR #329, opened 2018-06-26
Added methods in device profile client

jduranf:
Related to #324 

### DIFF
--- a/core/clients/command/command.go
+++ b/core/clients/command/command.go
@@ -38,25 +38,25 @@ type CommandClient interface {
 }
 
 type CommandRestClient struct {
-	url string
+	url      string
 	endpoint clients.Endpointer
 }
 
 // NewCommandClient : Create an instance of CommandClient
 func NewCommandClient(params types.EndpointParams, m clients.Endpointer) CommandClient {
-	c := CommandRestClient{endpoint:m}
+	c := CommandRestClient{endpoint: m}
 	c.init(params)
 	return &c
 }
 
-func(c *CommandRestClient) init(params types.EndpointParams) {
+func (c *CommandRestClient) init(params types.EndpointParams) {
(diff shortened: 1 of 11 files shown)


jduranf:
@drasko fixed all http magic numbers and also infinite loops

tonyespy:
@jduranf Your second commit also touches a lot more code than just the device profile client.  If you're going to do this type of cleanup across multiple services, it probably deserves its own PR.

jduranf:
@tonyespy I just tried to fix the source of my errors, and all the changes are related to /core/clients, which I used in order to add methods in the device profile client.
But you are right, probably should be better to perform this kind of clean up in its own PR

JPWKU:
recheck

tonyespy:
Although the infinite loops might now be correct from a language standpoint, IMHOP there's no place for infinite loops in our services.  It looks to me like these loops are using when initializing a client used to talk to a dependent service.  At some point after a certain number of retries the service should give up and exit (after logging an error).

tsconn23:
@tonyespy The loops you're referring to are how the service clients know whether a given endpoint has changed in Consul. They are only used when a service is passed the --consul flag. Their iteration pauses on this construct

```
select {
	case url := <-ch:
		d.url = url
}
```

whereby we wait to read a message out of the channel. Messages are published via this channel every 15 seconds (soon to be configurable) so this isn't a tight loop. When deploying in a multi-tenant scenario using Consul for service discovery, the endpoint of a service can change at any time so we use this to keep the client's endpoint up to date.

Also, note that this loop is happening in a go-routine so it's non-blocking for the execution flow. There will be other functionality that will use this pattern, like the service dependency resolution on bootstrap that you mention above. We can certainly have a loop like this check every second, for example, and then die after some period.

PR #331, opened 2018-06-26
[snap] Switch to core-config-seed-go

tonyespy:
This PR updates the snap to build and run the core-config-seed-go.

### DIFF
--- a/snap/bin/start-config-seed.sh
+++ b/snap/bin/start-config-seed.sh
@@ -32,7 +32,7 @@ done
 if [ $CONSUL_RUNNING != "[]" ] ; then
     cd $SNAP/config/core-config-seed-go
 
-    $SNAP/bin/core-config-seed-go -consul &
+    $SNAP/bin/core-config-seed-go --consul &
 fi
 
     
--- a/snap/bin/start-config-seed.sh
+++ b/snap/bin/start-config-seed.sh
@@ -10,8 +10,6 @@ else
     exit 1
 fi
 
-JAVA="$SNAP/usr/lib/jvm/java-8-openjdk-$ARCH/jre/bin/java"
-
 MAX_TRIES=10
 
 while [ "$MAX_TRIES" -gt 0 ] ; do
@@ -32,12 +30,9 @@ done
 #
 # TODO: this success check could be improved...
 if [ $CONSUL_RUNNING != "[]" ] ; then
-    cd $SNAP/jar/config-seed/
+    cd $SNAP/config/core-config-seed-go
 
-    $JAVA -jar -Djava.security.egd=file:/dev/urandom -Xmx100M \
-        $SNAP/jar/config-seed/core-config-seed.jar
-
-
+    $SNAP/bin/core-config-seed-go -consul &
 fi
 
     
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -132,64 +132,67 @@ parts:
       - make
       - zip
   config-seed:
-    source: https://github.com/edgexfoundry/core-config-seed.git
-    source-commit: 79faf75
-    plugin: maven
-    override-build: |
-      snapcraftctl build
+    source: https://github.com/edgexfoundry/core-config-seed-go.git
+    plugin: shell
+    after:
+      - glide
+      - go
+    shell: bash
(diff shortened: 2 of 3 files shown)


snap/snapcraft.yaml
```diff
@@ -132,64 +132,67 @@ parts:
...
-      echo "Installing config files"
+    source: https://github.com/edgexfoundry/core-config-seed-go.git
+    plugin: shell
+    after:
+      - glide
+      - go
+    shell: bash
+    shell-flags: ['-eux', '-o', 'pipefail']
+    shell-command: |
+      export GOROOT="$SNAPDIR/parts/go/install"
```
anonymouse64: I learned a better trick for this as SNAPDIR went away for me for some reason:
```
export GOROOT=$("$SNAPCRAFT_STAGE/bin/go env GOROOT")
```

anonymouse64: I think we should also switch all the other instances of using SNAPDIR with the go part to this too

snap/bin/start-config-seed.sh
```diff
@@ -32,12 +30,9 @@ done
...
 # TODO: this success check could be improved...
 if [ $CONSUL_RUNNING != "[]" ] ; then
-    cd $SNAP/jar/config-seed/
-
-    $JAVA -jar -Djava.security.egd=file:/dev/urandom -Xmx100M \
-        $SNAP/jar/config-seed/core-config-seed.jar
-
+    cd $SNAP/config/core-config-seed-go
 
+    $SNAP/bin/core-config-seed-go -consul &
```
tsconn23: @tonyespy This line might be incorrect. The flag should be "--consul". Two dashes. Alternately, you could use "-c". Please confirm you've seen this work.

tonyespy:
Nice catch @tsconn23.  Updated the start script to use "--consul".

@anonymouse64 Addressed your comments as well.

snap/snapcraft.yaml
```diff
@@ -140,7 +140,7 @@ parts:
     shell: bash
     shell-flags: ['-eux', '-o', 'pipefail']
     shell-command: |
-      export GOROOT="$SNAPDIR/parts/go/install"
+      export GOROOT=$("$SNAPCRAFT_STAGE/bin/go env GOROOT")
```
anonymouse64: This is my typo, unfortunately this needs to be :
```
export GOROOT=$("$SNAPCRAFT_STAGE/bin/go" env GOROOT)
```
with quotes around just the `$SNAPCRAFT_STAGE/bin/go`, not around the whole subshell.

snap/snapcraft.yaml
```diff
@@ -150,7 +150,7 @@ parts:
 
       mkdir -p .gopath/src/github.com/edgexfoundry
       export GOPATH="$PWD/.gopath"
-      export PATH="$GOPATH/bin:$PATH:$SNAPDIR/parts/glide/install"
+      export PATH="$GOPATH/bin:$PATH:$SNAPCRAFT_STAGE"
```
anonymouse64: I don't think this is correct as is with the `glide` part because the glide part is installed into `$SNAPCRAFT_PART_INSTALL` which is something like what it was before `$SNAPDIR/parts/glide/install`, and not `$SNAPCRAFT_STAGE` which is something like `$SNAPDIR/stage`. I would be fine with a change to the glide part that makes glide installed into `$SNAPCRAFT_STAGE` and then was filtered out from the final snap in the glide part with prime: 
```
  prime:
    - -*
```

tsconn23:
recheck

tsconn23:
@JPWKU glide install error happening during build after recheck

JPWKU:
recheck

JPWKU:
recheck

tonyespy:
Just re-pushed and fixed the consul flag in start-config-seed.sh. Although it's not clear to me why you'd run core-config-seed-go without this flag, as its whole purpose is to populate consul with configuration settings?

tsconn23:
@tonyespy Because it is assumed you're running config-seed-go in order to populate Consul (or eventually your configuration provider of choice). That's its job, so the flag is redundant.

JPWKU:
Rebase against upstream/master?

PR #344, opened 2018-07-01
Fixes benchmark and some fixes to test

itamarhaber:
This includes (I hope) the contents of https://github.com/edgexfoundry/edgex-go/pull/302.

More importantly, it addresses an issue in the benchmark that's manifestated twice - in events and in readings. The original behavior was to test AddReading (and AddEvent) to establish the measurement, and then continue with the rest of the benchmarks. However, the faster the database is actually doing the adds, more readings (or events) are created and passed to subsequent benchmarks. Then, doing a full fetch of all objects as in Readings (or Events) will take considerably longer for a faster database (IO capped) than a slower one.

The proposed solution is scrubbing the db after benching the addition, and populating it with a predetermined known static sized (e.g. 1K) dataset.

Signed-off-by: Itamar Haber <itamar@redislabs.com>

### DIFF
--- a/core/data/clients/memory-database_test.go
+++ b/core/data/clients/memory-database_test.go
@@ -20,9 +20,10 @@ func populateDbReadings(db DBClient, count int) (bson.ObjectId, error) {
 	var id bson.ObjectId
 	for i := 0; i < count; i++ {
 		name := fmt.Sprintf("name%d", i)
+		device := fmt.Sprintf("device" + strconv.Itoa(i/100))
 		r := models.Reading{}
 		r.Name = name
-		r.Device = name
+		r.Device = device
 		r.Value = name
 		var err error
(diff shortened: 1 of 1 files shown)


feclare:
@itamarhaber Why did you changed the devices to ```i / 100```?

itamarhaber:
@feclare my reasoning for that is creating several (10) devices that have events (and readings), so ReadingsByDeviceID would mean something. The way the benchmark was done, we have N devices and N readings.

feclare:
Thanks @itamarhaber

itamarhaber:
Sure - working on fixing the tests, because this breaks them and I didn't bother running `go test` before committing :)

itamarhaber:
Closing this to reintroduce after refractoring.

PR #345, opened 2018-07-02
snap/snapcraft.yaml: use govendor install instead of sync

anonymouse64:
Consul uses govendor for their vendor dependency management and currently we are doing `govendor sync`, which only will update with locally available commits, etc. that are already in the vendor folder, when we really need to do `govendor install`, which will download/install from external places all necessary vendor dependencies. This should fix builds such as : https://jenkins.edgexfoundry.org/view/Snap/job/edgex-go-snap-master-stage-snap/14/console


tsconn23:
@tonyespy @anonymouse64 Please rebase the branch for the PR. Otherwise I will click the button.

PR #356, opened 2018-07-04
export-client: Fix getting no registrations from memorydb

feclare:
It was returning null instead of empty array when a registration was never added.


feclare:
I'm not sure, but I should tell that it is not the case. In data y metadata I used a different approach when getting all the items in one table.

jduranf:
@feclare, I checked this issue it's present in core-data. Should be fine to fix it in another PR

feclare:
@jduranf Thanks for the test. 
I think I will add the fix to this PR as it is the same error.

tsconn23:
@feclare @jduranf Do we have an issue for what's being addressed here? If so, can you link it in the PR comment above?

PR #357, opened 2018-07-04
core-metadata: mongo error was used instead of core/db error

feclare:
Fixing issue added in #298
Metadata is checking if the error is ErrNotFound using the mongo error instead of the error defined in core/db which is returned.
Also added tests to verify that db.ErrNotFound is returned.


feclare:
It breaks core/db/mongo. I will fix it and open it again.

PR #360, opened 2018-07-05
Modified delete methods in core-metadata

jduranf:
Data base delete methods have been modified in core-metadata in order to work in the same way as in core-data.


tsconn23:
@jduranf Can you please create an issue describing the scope and then reference it in the PR summary? Going forward, I'd like for us to be more diligent about creating an issue then the PR.

PR #363, opened 2018-07-05
Updated support-logging to include log of port it started on.

brandonforster:
#361 

### DIFF
--- a/cmd/support-logging/main.go
+++ b/cmd/support-logging/main.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"strconv"
 	"syscall"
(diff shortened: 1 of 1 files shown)


cmd/support-logging/main.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/pkg/usage"
 	"github.com/edgexfoundry/edgex-go/support/logging"
 	"github.com/edgexfoundry/edgex-go/support/logging-client"
+	"strconv"
```
tsconn23: Since strconv is a package from the std library, please include it with the other std library packages above in alphabetical order.

Issue #366, opened 2018-07-06
core-metadata doesn't handle properly the not found error


jduranf:
In core-metadata, MongoDB  uses `mgo.ErrNotFound` for not found error, while MemoryDB uses `db.ErrNotFound`. The best option should be to use `db.ErrNotFound`

PR #367, opened 2018-07-06
Fixed issue while handling not found error in core-metadata

jduranf:
Related to #366 


core/db/mongo/metadata.go
```diff
@@ -99,7 +100,9 @@ func (m *MongoClient) GetScheduleEvent(se *models.ScheduleEvent, q bson.M) error
 	var mse mongoScheduleEvent
 
 	err := col.Find(q).One(&mse)
-	if err != nil {
+	if err == mgo.ErrNotFound {
```
tsconn23: Question -- Why do some of these have an "else if" null check but others don't?

Could this be streamlined by creating a core/db/mongo/utils.go containing the following function?

```
func errorMap(e error) error {
    if e == mgo.ErrNotFound {
        e = db.ErrNotFound
    }
   return e
}
```

The caller would then be:
```
err := col.Find(q).One(&mse)
if err != nil {
    return errorMap(err)
}
```

feclare:
@jduranf I'm working on fixing this on #357. But I will not be able to work on it for a while, I will sent you my work in progress and you can merge them as you want.

jduranf:
@feclare sorry, I didn't remember you were already working in this issue. Anyway, this PR should fix the issue without the mongo side effects.
@tsconn23 added errorMap() as you suggested

Issue #368, opened 2018-07-06
Refactoring to strictly delineate internal code vs. externally usable code.


tsconn23:
**The Problem**
Currently the EdgeX GoLang implementation has no separation between its internal domain model and the models used to pass data out of the system. They are one and the same. Likewise, there is no visibility as to which service client is used by an external application versus solely used internal to the EdgeX platform. This means we are at risk of making changes to either our data models or our service client code that break any applications which have imported those items. We have no way to insulate our consumers from changes we make to the application based on an internal need, nor do we have the ability to hide internal/private details of our data model since our data model is also part of our client contract.

**The Recommendation**
In order to facilitate the separation of internal/external domain models and logic, we realized that we would need to restructure the application packages. At a high level, the changes we decided to make are as follows:

/edgex-go/core
•	will be re-purposed to contain those models and libraries which are usable by external applications
•	will include the following:
    o	clients package – service clients usable by external applications 
    o	models package – data models usable by external applications
             	NOTE: These will be identical to the current property bags found today in /edgex-go/core/domain/models

/edgex-go/internal
•	This package will be extended to include all code that we do not want external applications explicitly relying on as recommended by the [golang-standards](https://github.com/golang-standards/project-layout) project layout
•	That means the following packages will be moved into here in some fashion
    o	/edgex-go/core/command
    o	/edgex-go/core/core-data
    o	/edgex-go/core/core-metadata
    o	/edgex-go/export
    o	/edgex-go/support
        	Consul-client, domain, logging
        	Service client packages will most likely go to core above
•	Domain models included herein may start off as identical to those found today in /edgex-go/core/domain/models but have the flexibility to change and move around over time according to the internal needs of the platform.
                         
We are looking at making this change early in the Delhi development cycle, so please provide feedback in a timely fashion if you have it. Also please note that any external applications which are directly linking to the current package structure within the master branch of EdgeX will break when we do this. That is unavoidable. We are doing this now so that the pill isn’t too bitter to swallow later.

jduranf:
Maybe will be also fine to change all `Id bson.ObjectId` in domain/models to `ID string`. The change from Id to ID is a linter recommendation. The change from `bson.ObjectId` to `string` will make the code less dependent from MongoDB, and also will simplify the code (there are lots of conversions from `ObjectId` to `string` and viceversa).

tsconn23:
[...]

PR #369, opened 2018-07-07
Refactoring toward external/internal packages -- Part A

tsconn23:
#368 

This PR is mainly focused on reorganizing the code.  It does NOT yet address the separation of the domain model into internal/external. I think this is enough in one shot. Please see the comment I just left on the issue above for information about where this is heading.

- Moved the service directories out of edgex-go/core and moved them to edgex-go/internal/core.
- Moved edgex-go/core/db to edgex-go/internal/pkg/db
- Updated any import statements where necessary
- Validated all changes via "make test" and local blackbox testing
- No functional changes have been made at all. This is a re-organization only.


tsconn23:
@JPWKU Docs-verify is failing because of the error below. The path doesn't exist anymore so we need to adjust the script. Let's discuss on Monday.

```
[edgex-go-docs] $ /bin/sh -xe /tmp/jenkins6799755749349068922.sh
+ cd docs/
+ ./build.sh
cp: cannot stat â€˜../core/data/raml/*.ramlâ€™: No such file or directory
Build step 'Execute shell' marked build as failure
```

JPWKU:
@tsconn23 Looks like we will need to adjust `docs/build.sh`.  The `cp` commands on the raml files are failing.  If we updated those to the new location I think we will be good.

tsconn23:
@JPWKU Updated PR. The location of the RAML files that used to be in /edgex-go/core as of this PR will be edgex-go/external/raml. That means the three in this list:

- edgex-go/core/command/raml/core-command.raml
- edgex-go/core/data/raml/core-data.raml
- edgex-go/core/metadata/raml/core-metadata.raml

Please modify the `docs/build.sh` script

JPWKU:
recheck

Issue #315, opened 2018-06-21
export distro won't mark an event as pushed after it's sent to its destination


davidmsh:
export distro in Go won't mark an event as pushed after it's sent to its destination as its Java counterpart does. We had some client need this pushed time for some benchmarking purposes. I think we should probably add it in go version of edgex. Also we probably want to make this feature configurable to avoid unnecessary service invocation and database updates when pushed are not really useful in someone's deployment. 
I can pick up this one if we are gonna implement it.

cloudxxx8:
@davidmsh the PRs are merged.  Please close this issue.

cloudxxx8:
https://github.com/edgexfoundry/edgex-go/pull/421
https://github.com/edgexfoundry/edgex-go/pull/454

Issue #316, opened 2018-06-21
PUT requests to /api/v1/event/id/{id} will only update pushed timestamp for event not its readings


davidmsh:
Put method of /api/v1/event/id/{id} behave differently than its Java counterpart  in that it won't update pushed timestamp for readings belong to an event, only the ts on the event itself will be updated. 
markPushed() method in src/main/java/org/edgexfoundry/controller/impl/EventControllerImpl.java will iterate through each reading and update its pushed ts. 
From core-data.raml, it states "Update the event to be pushed" Looks like it will only evnet's pushed ts and it works as designed. Just use this issue to confirm if this is what we want.
From the data model of core data point of view, do we really have cases that readings belonging to an event have different pushed ts than the event ? From current export distro implementation, we export by events, when an event is exported all its readings get exported together. 
Or we want future flexibility that, event may have no pushed ts while each reading has its own ts. If event itself has pushed ts, all its readings can be viewed as having the ts even they do have have pushed field set.

yanghua:
@tsconn23  Can I claim it? Did the California version's branch been cutting?

cloudxxx8:
it's the real issue and the behavior is not accordance with RAML
event is not totally immutable.  event contains some field for marking.
the immutable one is reading value.  Please reopen this issue.

cloudxxx8:
If the Event will be changed to immutable, the fields like modified and pushed should be removed as well as all the update API

jduranf:
@tsconn23 I think it's very important to know if an event have been pushed or not. If an event haven't been pushed (maybe there is a connectivity error), then can be requested later. Also we have the Pushed attribute in the Event structure, which is used for this purpose, and there are several functions in core-data, defined in the RAML and already implemented (PUT in `api/v1/event/id/{id}` or DELETE in `api/v1/event/scrub`) , which handle the Pushed attribute.

I agree an Event it's immutable, and currently there is no way no modify it, except the pushed attribute, which happens later.

cloudxxx8:
@tsconn23 , we are not asking to add a new feature to mark the pushed timestamp via REST API.  This is the current design, and this API exists right now with defect.  We are asking to fix the defect of this API.  Core WG could discuss whether we should remove this API and other event update API in the next API version, such as /api/v2/...., but we had better make the current and existing API work properly.

tsconn23:
Changes have been requested on PR #370. Re-opening issue until PR is merged.

jduranf:
[...]

PR #370, opened 2018-07-07
If enabled, mark pushed time after an event is sent in export distro

davidmsh:
As with java version, mark pushed time after an event is sent in export distro
*1 Add MarkPushed in export distro config to optionally enable this feature
*2 Add MarkPushed() function in coredata client to invoke corresponding rest api
*3 Invoke MarkPushed() at the end of processEvent() if configured to do so

### DIFF
--- a/export/distro/const.go
+++ b/export/distro/const.go
@@ -16,7 +16,7 @@ type ConfigurationStruct struct {
 	DistroHost           string
 	ClientHost           string
 	DataHost             string
-	DataPort			 int
+	DataPort             int
 	ConsulHost           string
 	ConsulPort           int
 	ConsulProfilesActive string


tsconn23:
Is this a PR targeting Issue #316 which was closed some time ago? It looks like it. Please see my comments there. I stand by them. This PR is on hold.

jduranf:
@davidmsh this implementation doesn't take in account if the event have been sent or not. Maybe the send interface should be modified to something like this:
```
type Sender interface {
	Send(data []byte, event *models.Event) bool
}
```
And then in registration.go do something like this:
```
res := reg.sender.Send(encrypted, event)
if configuration.MarkPushed && res {
    ...
}
```
Anyway, first should be fine to know if this PR will be accepted or not. And then do this modification in this PR or in another one.

davidmsh:
@tsconn23 This PR is targeting Issue #316 but related to Issue #315 in the way pushed timestamp is updated in an event. This PR mark an event's push ts using existing rest api from core-data. 
Regarding Issue #315, we all agree that the content of an event should be immutable but it's valuable to update some of its metadata to reflect its current state (and we store them in the same event) and I believe edgex api doesn't allow event content modification. This PR comes from a PoC from the field which should be valuable to the project. Your thought ?

davidmsh:
@jduranf that's a good idea, let's do that after this PR gets accepted.

cloudxxx8:
@tsconn23, this PR is totally different from #316 .  #316 is the bug of an existing API in core-data, and #315  is the bug of export-distro doesn't follow the original design to notify core-data when the event is pushed.  This PR is to fix #315 only.

tsconn23:
@jpwhitemn Would like us to pull this in. @davidmsh Please rebase the branch to get rid of the out-of-date message. Meanwhile, will review.

export/distro/registrations.go
```diff
@@ -158,6 +158,17 @@ func (reg registrationInfo) processEvent(event *models.Event) {
 	}
 
 	reg.sender.Send(encrypted, event)
+
+	if configuration.MarkPushed {
+		go func(id string) {
```
tsconn23: Why not utilize a goroutine on line 183 of registrations.go instead?
```
case event := <-reg.chEvent:
reg.processEvent(event)
```
Pass event in as a variable to an anonymous function invoked via goroutine....?

tsconn23:
@drasko @janko-isidorovic @feclare It sounds like from talking to @jpwhitemn that this is a bug and will prevent events from being cleared out of the database based on the "pushed" timestamp. If that's so and if we can get this PR merged, we will need this fix for the 0.6.1 update at the end of June. Can I please get one of you gentlemen to oversee getting this into the California branch for inclusion in 0.6.1?

feclare:
There is one goroutine for each registration. If you have several registrations running this code will call MarkPushed several times(one for each registration) for each event.

A better approach is to have a new goroutine accepting ids in a channel and calling to core-data markpush once for each id. If it have an small id cache (just several seconds) of the pushed id we can only make the call once for id.

PR #388, opened 2018-07-11
Update Docker-Compose Integration with Make in the 'california' branch

QKaseman:
Issue #287 has been resolved in master but still isn't in `california`.

As `tsconn23/issue_287` has a lot of commits not in `california` I went ahead and cherry picked the commits @tsconn23 made (`b45e041` and `6a0bde1`) and submitted a new pull request rather than another pull request from `tsconn23/issue_287`.

`make run_docker` now pulls the `v0.6.0` `docker-compose.yml` file.


tsconn23:
@JPWKU This is an interesting case where someone has created a PR by cherry picking two commits that are signed by me. I'm guessing that's why the DCO is passing.

@QKaseman If you aren't already familiar with the process of signing commits, you can see how to set that up here:
https://wiki.edgexfoundry.org/display/FA/Secrets+of+GitHub+Signing%2C+Keys+and+Commits+Setup

I'm not sure if we'll pull this in because we hadn't intended for this work to go into California as we were mostly thinking of having this fix for those working from master. In this case folks who just want to take the platform for a spin will pull from master, execute a "make run_docker" and download the California containers (v 0.6.0). Once the 0.6.1 release goes out at the end of July, then we plan to update the code you found here to pull the 0.6.1 images and so on.

QKaseman:
Interesting. I wasn't sure exactly what the process would be as I only was cherry picking commits. I'll make sure I have the GPG key setup properly for the future. 

(Though if the cherry-pick only pull request ever comes up again, what is the correct process to use?)

Ah, okay. Part of the confusion on my part is `california` is at version `0.6.0`, there is a `0.6.0` `docker-compose.yml` file but the `docker-compose.yml` file in `california` is version `0.5.0`.

drasko:
@QKaseman please sync the branch

tsconn23:
I propose closing this PR per my explanation above.

PR #390, opened 2018-07-12
Refactoring toward external/internal packages -- Part B

tsconn23:
#368 

Support clients have been moved into edgex-go/pkg/clients
Support implementations have been moved into edgex-go/internal/support

Tested support-logging API via Postman as well as service-to-service call
Tested both persistence in Mongo and to file system

Now only export remains.


tsconn23:
@tonyespy Will need help from either you or @anonymouse64 w/r/t what may be wrong with the snapcraft.yaml for this PR. I did make one change to repoint the Attribution.txt files in the core services, but that didn't seem to help.

tsconn23:
recheck

tsconn23:
FYI @tonyespy was able to build this PR locally and has submitted subsequent PRs to possibly address a memory issue with Snappy on the build agents. Snap verifications are not required to pass for this PR.

internal/support/logging/models/log_entry.go
```diff
@@ -11,7 +11,7 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  *******************************************************************************/
-package support_domain
```
feclare: Why having a models folder for just a type? It could be moved to internal/support/logging

tsconn23: It's not really duplicated. I mean, yes right now its form is the same however the scope is different. One difference has to do with the bson serialization support on the support/logging/models type. If you look at pkg/models, there are no bson specs. This is because at the point where an external client is using the model to send a log entry, the client doesn't care how we persist the data. Internally to the support-logging service though, it's using Mongo for its back-end persistence, so the bson is there. Another way we'll use this is when we remove the type bson.ObjectId from our pkg/models type keys (like event.go). Externally, these should be strings or something else that is not database provider specific.

Also, our domain models are currently just property bags so they're going to look very similar to start. Once we get the separation in place, then we can make them a little more intelligent or diverse as required.

feclare: I dont fully understand what you are trying to do, but from the design point of view we should never duplicate types.

tsconn23: Sure, I understand. You're thinking of the scenario where you define a type once, in a common library, and that single type definition is used by all services. This is also known as the "DRY" principle (for "Don't Repeat Yourself"). It doesn't apply in a microservice architecture though because it leads to coupling.
* See p. 59 of "Building Microservices" by Sam Newman
* Google "microservice DRY" and you'll see plenty of articles where they talk about how it's perfectly alright to define a type more than once across multiple services that use it. For example: https://www.oreilly.com/ideas/microservices-antipatterns-and-pitfalls (use your browser to find "DRY" in the article)

So why is this OK? It preserves the context of each microservice so that each service can potentially vary the type if it needs to. It may never have to do that, but it can if necessary and no other service has to change. Further, within a service, there can be multiple representations of a type according to the layer at which the type functions. In this case specifically, we have a client contract logEntry model and  an internal, support domain model. They have the same properties right now, but they don't have to. The external model only defines json mappings for marshaling, whereas the domain model defines both json AND bson. This alone makes them functionally different. A logEntry type defined in an externally available pkg as a contract entity has a different context altogether than a logEntry type defined in an internal pkg.

tsconn23: @feclare As a result of some other work I was doing, I noticed the following type in the db/mongo package

```
type mongoEvent struct {
	models.Event
}
```

and I think I see where you're coming from. Perhaps wrapping types like this is a way to avoid outright duplication of definitions. I'm not thrilled about how it refers all the way up to the external pkg/models package, but I admit I don't have a better suggestion right now. I'll take this into account going forward.

Issue #416, opened 2018-07-23
Retry Service Dependencies on Startup -- Metadata


tsconn23:
This issue is for making changes to the core-metadata service so that it doesn't rely so much on an explicit start order when deployed to an environment. We should be able to start core-metadata independently of other services, and expect that it will start correctly. Upon startup, core-metadata should retry its connection attempts to other dependencies such as Mongo and Consul rather than exit right away if those platforms aren't up yet. The overall retry window should be configurable.

PR #418, opened 2018-07-23
Core-metadata -- Handle bootstrap dependency timing

tsconn23:
#416 

Submitting as a pattern proposal for solving the above issue. I have an open question in my mind as to whether or not some of this code can make its way into a common internal/pkg implementation, but that refactoring will come later as I see the effect on subsequent services.


tsconn23:
Closing for now. Realized I have more work to do in order to handle the case where none of the dependencies comes up. We need to log that condition and then the service should exit.

PR #421, opened 2018-07-24
 If enabled, mark pushed time after an event is sent in export distro

davidmsh:
This PR is a recreation of my previous one of same purpose since repos of my previous commits is changed to private. I also addressed some of the review comments.


pkg/clients/coredata/event.go
```diff
@@ -469,3 +470,38 @@ func (e *EventRestClient) DeleteOld(age int) error {
 
 	return nil
 }
+
+// Mark event as pushed
+func (e *EventRestClient) MarkPushed(id string) error {
```
tsconn23: Please provide a unit test in event_test.go for this new method.

tsconn23:
@janko-isidorovic @drasko Would appreciate your review on this PR which replaces https://github.com/edgexfoundry/edgex-go/pull/370. Also, as I noted there, this work should also be pulled into the California branch for 0.6.1 release.

export/distro/registrations.go
```diff
@@ -157,7 +157,17 @@ func (reg registrationInfo) processEvent(event *models.Event) {
 		encrypted = reg.encrypt.Transform(compressed)
 	}
 
-	reg.sender.Send(encrypted, event)
+	if reg.sender.Send(encrypted, event) && configuration.MarkPushed {
+		go func() {
```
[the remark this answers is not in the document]
davidmsh: Looks like I misused the anonymous goroutine here. I will change the markPushed call to be inline

pkg/clients/coredata/event.go
davidmsh: Will do

pkg/clients/coredata/event.go
```diff
@@ -469,3 +470,38 @@ func (e *EventRestClient) DeleteOld(age int) error {
...
+	req, err := http.NewRequest(http.MethodPut, e.url+"/id/"+id, nil)
+	if err != nil {
+		fmt.Println(err)
```
drasko: Please do not print on console. Log instead.

davidmsh: @drasko looks like all other methods print on console in this source file. Do you want to change them all ?

pkg/clients/coredata/event.go
davidmsh: @tsconn23 I have pushed couple commits adding a test case and change make code in goroutine in line. Let me know if you have further concerns except for @drasko 's below.

pkg/clients/coredata/event.go
davidmsh: @drasko  And since we always return the errors to the client, do we really need to log the errors ? We probably can let the client decide what to do with the errors. In some cases, the client may not want it logged.

cloudxxx8: @drasko 
Please see the content of this file
https://github.com/edgexfoundry/edgex-go/blob/master/pkg/clients/coredata/event.go

There are lots of places using fmt.Println in this file, so it is another issue.  David just followed the same convention in this file.  We should create another issue and PR to fix this.  Please help merge this PR first.  Thanks.

@SteveOss  please help comment.  Thanks.

tsconn23:
recheck

Issue #410, opened 2018-07-20
HTTP 503 -> 500 in support-logging service


tsconn23:
Refactor support-logging API implementation and associated RAML so that all cases where HTTP 503 (Service Unavailable) is returned are replaced with the more appropriate HTTP 500 (Internal Server Error).

tsconn23:
Associated with #284

brandonforster:
Closing this issue as the PR has been merged.

PR #423, opened 2018-07-25
Attempting to make RAML and code for support-logging agree

brandonforster:
#410 

This is my first strike at the problem. I removed a lot of error messages that don't seem possible since there doesn't appear to be error handling for those execution paths, would we rather add error handling for those?

### DIFF
--- a/api/raml/support-logging.raml
+++ b/api/raml/support-logging.raml
@@ -58,7 +58,9 @@ schemas:
           }
     responses:
       202:
-        description: return 202 Accepted to clients with timestamp being accepted.
+        description: accepted to clients with timestamp being accepted.
+      503:
+        description: creation request is malformed
   /{limit}:
     uriParameters:
       limit:
@@ -67,10 +69,8 @@ schemas:
         example: 10
     get:
       responses:
-        413: 
-          description: if the number of events exceeds the current max limit
-        503: 
-          description: for unknown or unanticipated issues.
+        400: 
+          description: request is malformed
         200:
           description: list a collection of LogEntry - limited in size by the limit parameter
           body:
@@ -117,7 +117,7 @@ schemas:
       responses:
         413: 
           description: if the number of events exceeds the current max limit
-        503: 
+        500: 
           description: for unknown or unanticipated issues.
         200:
           description: list a collection of LogEntry created between the specified start and end dates - limited in size by the limit parameter
@@ -169,7 +169,7 @@ schemas:
       responses:
         413: 
           description: if the number of events exceeds the current max limit
-        503: 
+        500: 
           description: for unknown or unanticipated issues.
         200:
           description: list a collection of LogEntry matching any of the specified labels and being created between the specified start and end dates - limited in size by the limit parameter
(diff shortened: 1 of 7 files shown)


tsconn23:
recheck

api/raml/support-logging.raml
```diff
@@ -58,7 +58,9 @@ schemas:
           }
     responses:
       202:
-        description: return 202 Accepted to clients with timestamp being accepted.
+        description: accepted to clients with timestamp being accepted.
+      503:
```
tsconn23: I think you may have meant for this to be a 400...?

api/raml/support-logging.raml
```diff
@@ -67,10 +69,8 @@ schemas:
         example: 10
     get:
       responses:
-        413: 
-          description: if the number of events exceeds the current max limit
-        503: 
-          description: for unknown or unanticipated issues.
+        400:
```
tsconn23: You'll want to keep the 413 here. It's indicating that there is more data to return than what the client asked for. Since that's determined on the server and not specifically by the request, 413 is appropriate. However the DB query should cap the number of records returned, so I don't think this will be frequent.

api/raml/support-logging.raml
```diff
@@ -67,10 +69,8 @@ schemas:
         example: 10
     get:
       responses:
-        413: 
-          description: if the number of events exceeds the current max limit
-        503:
```
tsconn23: This should be a 500. Looks to be removed altogether from the new file.

tsconn23:
@brandonforster FYI, build is failing due code changes breaking unit tests. You can execute unit tests locally by running "make test". From the logs:

```
2018/07/26 13:46:39 http: multiple response.WriteHeader calls
--- FAIL: TestGetLogs (0.01s)
    --- FAIL: TestGetLogs/limitToLow (0.00s)
    	server_test.go:204: Returned status 200, should be 413
2018/07/26 13:46:39 http: multiple response.WriteHeader calls
2018/07/26 13:46:39 http: multiple response.WriteHeader calls
2018/07/26 13:46:39 http: multiple response.WriteHeader calls
2018/07/26 13:46:39 http: multiple response.WriteHeader calls
2018/07/26 13:46:39 http: multiple response.WriteHeader calls
2018/07/26 13:46:39 http: multiple response.WriteHeader calls
2018/07/26 13:46:39 http: multiple response.WriteHeader calls
2018/07/26 13:46:39 http: multiple response.WriteHeader calls
2018/07/26 13:46:39 http: multiple response.WriteHeader calls
FAIL
coverage: 69.2% of statements
FAIL	github.com/edgexfoundry/edgex-go/internal/support/logging	0.0
```

tsconn23:
@brandonforster Still need 500 responses restored on delete endpoints. Also, as discussed Friday afternoon, need to not bury the errors in mongo.go in order to return those 500s. See remove() and find() methods.

tsconn23:
@brandonforster This looks pretty good although there are a couple delete endpoints in the RAML only returning 200s and a 500 response should be included. These are the endpoints:
```
/logLevels/{logLevels}/originServices/{originServices}/labels/{labels}/{start}/{end}:
/originServices/{originServices}/{start}/{end}:
```

Also do a `git pull --rebase upstream master` and then push to origin to clear the "out of date" message above.

tsconn23:
recheck

PR #428, opened 2018-07-27
Notifications into monorepo

akramtexas:
- The goal of the work (for which this PR is being submitted) is to bring in the "support-notifications" Golang package into the EdgeX monorepo.


cmd/support-notifications/main.go
```diff
@@ -85,7 +84,7 @@ func main() {
 		return
 	}
 
-	r := notifications.LoadRestRoutes()
+	//r := notifications.LoadRestRoutes()
```
tsconn23: This line commented out intentionally. It's redundant with content in the startHttpServer() function. I should have removed the line before commit. Perhaps @akramtexas can remove it.

akramtexas:
`Related Blackbox tests continue to pass 100%.`

akramtexas:
👍

cmd/support-notifications/main.go
akramtexas: Good point @tsconn23 (Done, earlier. At this time, adding a comment here for acknowledgment and closure).

support/notifications/transmission.go
```diff
@@ -0,0 +1,297 @@
...
+
+func transmissionByEndHandler(w http.ResponseWriter, r *http.Request) {
+
+	if r.Body != nil {
+		defer r.Body.Close()
+	}
+
+	vars := mux.Vars(r)
+	end, err := strconv.ParseInt(vars["end"], 10, 64)
+	// Problem converting start
```
drasko: Confusing comment.

PR #448, opened 2018-08-02
Use internal.SupportNotificationsServiceKey and change type of CleanupDefaultAge to int.

akramtexas:
#449

Signed-off-by: Akram Ahmad <sftwr2020@gmail.com>


support/notifications/cleanup.go
```diff
@@ -61,7 +61,7 @@ func cleanupAgeHandler(w http.ResponseWriter, r *http.Request) {
 	case http.MethodDelete:
 		loggingClient.Info("Cleaning up of old notifications and transmissions")
 
-		if err = dbc.CleanupOld(age); err != nil {
+		if err = dbc.CleanupOld(int(age)); err != nil {
```
tsconn23: @akramtexas Instead of the cast here, try using strconv.Atoi which will return a native int.

akramtexas: @tsconn23 Excellent suggestion. Done (Pushed update to repo).

Issue #412, opened 2018-07-20
HTTP 403 -> 400 in core-command


tsconn23:
Operations that attempt to load a specified entry from the database and do not find any results currently return HTTP 403 (Status Forbidden). These should simply be 404 (Not Found). Example from internal/core/command/device.go:

```
func commandByDeviceID(did string, cid string, b string, p bool) (string, int) {
	d, err := mdc.Device(did)

	if err != nil {
		loggingClient.Error(err.Error(), "")
		// send 403: no device exists by the id provided
		return "", http.StatusForbidden
	}
...
}
```

tsconn23:
Associated with #284

Issue #284, opened 2018-06-07
RAML Audit for Current Service Operations


tsconn23:
During the Java --> Go conversion, the definition and behavior of certain endpoints within the EdgeX microservices changed but the RAML spec was not updated. An audit needs to be performed in order to ensure the implementation conforms to the spec. Coordination with blackbox testing team will be necessary as tests are based on the RAML and, for example, some tests currently fail even though the API operation is correct because the RAML is out of date.

During the Palo Alto F2F it was decided that the API spec (RAML in this case) would be the source of truth and not the code. This facilitates implementation of the spec in multiple languages.

NOTE: This needs to be done before we consider moving to Swagger.

PR #450, opened 2018-08-02
HTTP 403 -> 400 in core-command

brandonforster:
Issues #412 and #284


tsconn23:
@brandonforster We need to discuss.

PR #453, opened 2018-08-03
Update link to core-data.raml file

mhall119:
Update link to core-data.raml file to point to the correct location


tsconn23:
@mhall119 Sign-off is missing. 

https://wiki.edgexfoundry.org/display/FA/Contributor+Process.

Scroll down to the yellow section.

mhall119:
I made the edit via Github's web UI, is there a way to do the equivalent step through the same UI?

tsconn23:
@mhall119 Nope.

mhall119:
@tsconn23 better now?

PR #454, opened 2018-08-03
If enabled, mark pushed time after an event is sent in export distro (port from master)

davidmsh:
This is porting https://github.com/edgexfoundry/edgex-go/pull/421 from master into california.
Done some local testing, marking pushed works fine if enabled.


jpwhitemn:
LGTM.

Issue #465, opened 2018-08-08
core-data's api “/event/scrub” block long time when delete events's item more 200000+


unanao:
software:
california

Enviromnet:
The mongo have more 200000+ events ane 200000+ readmings

Get all events costs about 3 minutes
Execute more than 30 minutes only delete 100000+ events and related readings.
CPU is more than 40%
![cpu](https://user-images.githubusercontent.com/2163628/43835482-b20068e8-9b44-11e8-81c5-6f32e669beea.png)

jpwhitemn:
The intention of scrub was to be done periodically by a call in scheduling service - thereby avoiding large blocking scrub.  Having said that, with Go concurrency model, we should have it non block.  Assigning to @tsconn23 and @xmlviking but probably not going to be addressed until Edinburgh release.

jpwhitemn:
@tsconn23 @jduranf - can we either close this issue out or move it to Geneva?  Is it still being worked as it has an Edinburgh label?

jduranf:
If I'm not wrong, as #1286 is closed, this issue also can be closed

PR #466, opened 2018-08-08
Bugfix for issue 465: “/event/scrub” block long time

unanao:
when delete events's item more 200000+ events

https://github.com/edgexfoundry/edgex-go/issues/465


tsconn23:
@unanao Commit needs DCO sign-off. Info is in the yellow box on this page:
https://wiki.edgexfoundry.org/display/FA/Contributor+Process

Aside from that, I don't think I prefer to solve the problem this way. We already have way too much logic in the controllers and I don't think we need to make this a configurable setting. If instead the service were to start a long-running goroutine that subscribes to a channel. the relevant controller method could publish a "Scrub" message to that channel. The goroutine would handle that message via a select statement. This would decouple the logic to perform the scrubbing from the controller, and allow the HTTP request to return after the "Scrub" message publication,

unanao:
@tsconn23 Yes, a long-running goroutine is better,  it  I will change the code this way.

But, is the" configurable setting" refer to ?
```
ScrubNumber = 2000
ScrubSleepTime = 5
```
If we read all the events once time, will coust more time and memory.  Yesterday，Get all 200 thousand events costs about 3 minutes, dete 100 thousand cost more 30 minutes. Now, my Data acquisition cycle is 100ms,  only one schedueEvent,  more than 800 thousand， in fact we are trying to acquisition more data.  

I think we shoud not delete all the data once time,  batch by batch deletion have less effective on current process and more efficient.  I test we can stably delete 2000 every 5 seconds.

If there is no "configuration", how can I batch by batch deletion.  Looking forward your advice.

feclare:
What about solving the issue adding a new function to the db interface. This kind of operation could be optimally resolved by the db itself, instead of getting all the events and removing them one by one.

Only problem with this approach is that all db packages need to be modified to add the new function.

unanao:
@feclare yeah，we should update reading's pushed when markPushed,  then we can delete event and reading based on  pushed.

unanao:
@drasko sorry, I wil make the remain work done this week, I am a bit of busy this work.

unanao:
@drasko I have change the patch as discussed.

core/data/init.go
```diff
@@ -95,9 +97,20 @@ func Init(conf ConfigurationStruct, l logger.LoggingClient, useConsul bool) erro
...
+	dataScrubInit()
 	return nil
 }
 
 func Destruct() {
 	dbc.CloseSession()
 }
+func dataScrubInit() {
+	//Create scrub pushed events long go routine and channel
+	scrubPushedCh = make(chan int64, 5)
```
drasko: Define magic number `5` as constant at the top of the file.

core/data/init.go
```diff
@@ -95,9 +97,20 @@ func Init(conf ConfigurationStruct, l logger.LoggingClient, useConsul bool) erro
...
+	go DeletePushedEvents()
+
+
+	//Create delete old events long go routine and channel
+	deleteOldCh = make(chan int64, 5)
```
drasko: Same here.

core/data/init.go
unanao: done

core/data/init.go
unanao: done

tsconn23:
@unanao As I said above, all commits need DCO sign-off. Info is in the yellow box on this page:
https://wiki.edgexfoundry.org/display/FA/Contributor+Process

unanao:
close the pull request, I  will pull a new request with signoff

PR #467, opened 2018-08-08
Update error handling in core-command

brandonforster:
Issue #412: HTTP 403 -> 400 in core-command

### DIFF
--- a/api/raml/core-command.raml
+++ b/api/raml/core-command.raml
@@ -42,7 +42,7 @@ schemas:
                 description: if the request is malformed or unparsable
             "404": 
                 description: if no device exists by the id provided
-            "422": 
+            "423": 
                 description: if the device is locked (admin state).
             "500": 
                 description: for unanticipated or unknown issues encountered.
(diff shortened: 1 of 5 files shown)


api/raml/core-command.raml
```diff
@@ -38,9 +38,11 @@ schemas:
         responses: 
             "200": 
                 description: String as returned by the device/sensor via the device service.
+            "400":
+                description: if the request is malformed or unparsable
             "404": 
                 description: if no device exists by the id provided
-            "423": 
+            "422":
```
tsconn23: @brandonforster We were talking about this in stand-up. Probably want to keep 423 here.

pkg/clients/types/errors.go
```diff
@@ -13,8 +13,25 @@
...
 
+import "fmt"
+
 type ErrNotFound struct{}
 
 func (e ErrNotFound) Error() string {
 	return "item not found"
 }
+
+type ErrClientServices struct{
```
tsconn23: @brandonforster While you're in there, I'm thinking "ErrServiceClient" is a better name for this indicating that the error is coming from one of our service client wrappers. If you agree, can you rename?

Issue #457, opened 2018-08-06
Increase unit test coverage -- core/data/event


tsconn23:
Current unit test coverage for the core/data/event package stands at

`edgex-go/internal/core/data	0.020s	coverage: 9.6% of statements`

We need to increase this coverage. In order to do so, it will be necessary to decouple the MUX Router code from the relevant business logic. The business logic should be unit testable independently of infrastructure dependencies. Right now, it is not.

I'm going to make changes applicable to this one package to illustrate a pattern and to keep the PR manageable.

tsconn23:
Assigned to @brandonforster for completion

PR #482, opened 2018-08-14
core/data/event.go Partial Refactoring for Increased Unit Test Coverage

tsconn23:
#457 

This is as far as I've been able to go on this issue prior to vacation. However, what is presented here compiles, the unit tests work and the services do come up via `make run`.

My intention here is to show work done thus far for trying to decouple our infrastructure code (such as REST routing) from actual business functionality (such as core/data/event.go). I am of the opinion that the business logic should be cleanly separated from infra so that it is easily, cleanly and completely unit testable.

Your comments/questions are welcome.

A summary of things to look for in this PR:

- ZeroMQPublisher has been put behind an interface and a mocking stub provided
- Individual error type files found in core/data/errors have been consolidated into one types.go file
- All of the above errors have "New" constructors. These are used by the controller for type checking in order to determine what HTTP response to send back. This allows decoupling of business logic from HTTP concerns.
- As stated in an earlier email to the golang list, mockery and testify are used for stubs.
- Note the introduction of domain events, particularly when adding a new event. Domain events are published to a channel which is then processed by a goroutine. Depending on the event type, a different handler in core/data/device.go will be called. In this way, the caller POSTing the event to the API does not need to wait for us to do domain-related housekeeping which involves multiple additional REST calls. (I believe that this can be further streamlined by eliminating the REST calls, but that's beyond the scope of this PR)
- The goroutine and channel related to domain event handling are initialized in core/data/init.go::Init()
- Previously, routing and business logic were all co-mingled in event.go. The approach below leaves the business logic in event.go, and moves the routing logic over to router.go.


akramtexas:
Leading up to visiting this PR, I was tackling very similar goals over the past 16 hours, looking to implement some scheme to drive the business logic for the Sys Mgmt Agent-related codebase: Decoupling infra concerns from the business (service layer) logic. Thanks for the heads up, and I'm glad to check this out! 👍

akramtexas:
- Adding that of the resources I had found thus far, for example, the following one, was about the best: [Testing Your (HTTP) Handlers in Go](https://blog.questionable.services/article/testing-http-handlers-go/). But it was, nonetheless, entangled with the routing and plumbing (aka INFRA)!

- Only [one other resource](https://github.com/gorilla/mux/issues/167) offered some glimmers of INFRA-independence, claiming: `This way—regardless of whether you use gorilla/mux or not—you're still testing that your application handles the route variables you're passing.`

internal/core/data/event_test.go
```diff
@@ -16,32 +16,185 @@ package data
...
+	"gopkg.in/mgo.v2/bson"
+	"sync"
+	"time"
 )
 
 var testEvent models.Event
 var testRoutes *mux.Router
 
+const (
+	TEST_DEVICE_NAME string = "Test Device"
```
drasko: Using all capitals is not Go idiomatic. I think that `testDeviceName` would work fine here.

internal/core/data/event_test.go
```diff
@@ -16,32 +16,185 @@ package data
...
+		return
+	}
+
+	if count == 0 {
+		t.Errorf("no events found")
+		return
+	}
+}
+
+func TestGetEvents(t *testing.T) {
```
drasko: I would strongly advise table-based test approach, as described here: https://medium.com/@sebdah/go-best-practices-testing-3448165a0e18 or here: https://www.youtube.com/watch?v=hVFEV-ieeew for testing different cases (and asserting different outcomes).

Also, for assertion I advise [testify](https://github.com/stretchr/testify).

internal/core/data/messaging/eventpublisher.go
```diff
@@ -22,25 +24,25 @@ import (
 const (
 	ZEROMQ int = iota
```
drasko: Do we need these constants to be capitalized? This is not Go idiomatic. `zmq` will fork fine here.

internal/core/data/router.go
```diff
@@ -72,3 +80,212 @@ func LoadRestRoutes() *mux.Router {
...
+			case types.ErrNotFound:
+				http.Error(w, err.Error(), http.StatusNotFound)
+				return
+			default: //return an error on everything else.
+				http.Error(w, err.Error(), http.StatusServiceUnavailable)
+				return
+			}
+		}
+
+		// Return result
```
drasko: Is this comment really useful?

internal/core/data/router.go
```diff
@@ -72,3 +80,212 @@ func LoadRestRoutes() *mux.Router {
...
+
+// Remove all the old events and associated readings (by age)
+// event/removeold/age/{age}
+func eventByAgeHandler(w http.ResponseWriter, r *http.Request) {
+	defer r.Body.Close()
+
+	vars := mux.Vars(r)
+	age, err := strconv.ParseInt(vars["age"], 10, 64)
+
+	// Problem converting age
```
drasko: No need for this comment.

internal/core/data/router.go
```diff
@@ -72,3 +80,212 @@ func LoadRestRoutes() *mux.Router {
...
+		if err != nil {
+			LoggingClient.Error(err.Error())
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		// Return the count
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte(strconv.Itoa(count)))
```
drasko: Is this valid JSON you are returning?

internal/core/data/router.go
```diff
@@ -72,3 +80,212 @@ func LoadRestRoutes() *mux.Router {
...
+				http.Error(w, t.Error(), http.StatusBadRequest)
+			default:
+				http.Error(w, t.Error(), http.StatusInternalServerError)
+			}
+			LoggingClient.Error(err.Error())
+			return
+		}
+
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte(newId))
```
drasko: What's the `Content-Type` here?

internal/core/data/router.go
```diff
@@ -72,3 +80,212 @@ func LoadRestRoutes() *mux.Router {
...
+			}
+			LoggingClient.Error(err.Error())
+			return
+		}
+
+		LoggingClient.Info("Updating event: " + from.ID.Hex())
+
+		// Update the fields
+		if len(from.Device) > 0 {
+			// Check device
```
drasko: No need for this comment IMHO

internal/pkg/db/influx/data.go
```diff
@@ -161,6 +165,27 @@ func (ic *InfluxClient) getEvents(q string) ([]models.Event, error) {
...
+// Get events for the passed query
+func (ic *InfluxClient) getEventsWithLimit(q string, limit int) ([]models.Event, error) {
+	events := []models.Event{}
+	query := fmt.Sprintf("SELECT * FROM %s %s LIMIT %v", db.EventsCollection, q, limit)
+	res, err := ic.queryDB(query)
+	if err != nil {
+		return events, err
+	}
+
+	if len(res) == 1 {
```
drasko: Maybe `if len(res) == 1 && len(res[0].Series) == 1 {...}`

internal/pkg/db/test/db_data.go
```diff
@@ -422,7 +424,7 @@ func testDBEvents(t *testing.T, db interfaces.DBClient) {
 		t.Fatalf("There should be 100 events, not %d", len(events))
 	}
 
-	events, err = db.EventsOlderThanAge(0)
+	events, err = db.EventsOlderThanAge(afterTime + 10)
```
drasko: Magic constant `10`. Please define it as a `constant` at the top of the file and give it a meaningful, self-descriptive name.

internal/core/data/router.go
```diff
@@ -72,3 +80,212 @@ func LoadRestRoutes() *mux.Router {
...
+/*
+Return number of events in Core Data
+/api/v1/event/count
+*/
+func eventCountHandler(w http.ResponseWriter, r *http.Request) {
+	defer r.Body.Close()
+
+	switch r.Method {
+	case http.MethodGet:
+		count, err := count()
```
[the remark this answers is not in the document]
akramtexas: - Having had the opportunity to check out the (implementation) details in this PR, I'm very pleased by the conceptual simplification offered by the approach. I like how the (seemingly inevitable) entanglement of business logic with the routing and plumbing (aka INFRA) is disentangled! 

In fact, I'm happy to report that I've now made use of this pattern already in [my System Management (SMA) work](https://github.com/akramtexas/edgex-go/commits/sys-management-agent).

akramtexas:
- Having had the opportunity to check out the (implementation) details in this PR, I'm very pleased by the conceptual simplification offered by the approach. I like how the (seemingly inevitable) entanglement of business logic with the routing and plumbing (aka INFRA) is disentangled! 

- In fact, I'm happy to report that I've now made use of this pattern already in [my System Management (SMA) work](https://github.com/akramtexas/edgex-go/commits/sys-management-agent).

internal/core/data/messaging/eventpublisher.go
```diff
@@ -22,25 +24,25 @@ import (
...
-	zmq      zeroMQEventPublisher
+// Configuration struct for PubSub
+type PubSubConfiguration struct {
+	AddressPort string
 }
 
-func NewZeroMQPublisher(configuration ZeroMQConfiguration) *EventPublisher {
-	return &EventPublisher{protocol: ZEROMQ, zmq: newZeroMQEventPublisher(configuration)}
+type EventPublisher interface {
+	SendEventMessage(e models.Event) error
```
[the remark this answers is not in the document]
tsconn23: I'm going to leave this as is for now. That's what it was called previously. If we need to change it later, we can.

internal/core/data/messaging/eventpublisher.go
```diff
@@ -22,25 +24,25 @@ import (
...
-// Send the event
-func (ep *EventPublisher) SendEventMessage(e models.Event) error {
-	// Switch based on the protocol you're using
-	switch ep.protocol {
+func NewEventPublisher(pubType int, conf PubSubConfiguration) (EventPublisher, error) {
+	switch pubType {
 	case ZEROMQ:
-		return ep.zmq.SendEventMessage(e)
+		return newZeroMQEventPublisher(conf), nil
+	case MOCK:
```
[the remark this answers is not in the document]
tsconn23: I reworked this to put the mockPublisher in event_test.go. I couldn't directly call `ep = newMockEventPublisher(conf)` due to the function not being exported from the messaging package. event_test.go is currently the only place where the mock is needed currently so this works.

internal/core/data/messaging/zeromq.go
```diff
@@ -21,22 +21,19 @@ import (
...
 )
 
-// Configuration struct for ZeroMQ
-type ZeroMQConfiguration struct {
-	AddressPort string
-}
-
 // ZeroMQ implementation of the event publisher
 type zeroMQEventPublisher struct {
+	pubType   int
```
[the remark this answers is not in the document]
tsconn23: Removed

pkg/clients/metadata/mocks/DeviceClient.go
```diff
@@ -0,0 +1,432 @@
+// Code generated by mockery v1.0.0
```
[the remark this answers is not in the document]
tsconn23: We discussed this a couple weeks ago on the Core WG call. We decided for now to check it in because it would allow the DeviceServices to also utilize the mock for unit tests there. If it's a pain, we can re-evaluate.

internal/core/data/event.go
```diff
@@ -47,81 +80,81 @@ func checkDevice(device string, w http.ResponseWriter) bool {
 	return true
 }
 
-// Put event on the message queue to be processed by the rules engine
-func putEventOnQueue(e models.Event) {
-	LoggingClient.Info("Putting event on message queue", "")
-	//	Have multiple implementations (start with ZeroMQ)
-	err := ep.SendEventMessage(e)
+func deleteByAge(age int64) (int, error) {
+	events, err := dbClient.EventsOlderThanAge(age)
```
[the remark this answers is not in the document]
tsconn23: That would require some logic rework as the `EventsOlderThanAge` call obtains the list that we subsequently loop though, deleting the events one by one. Would be great to make this a batch operation and eliminate the loop.

internal/core/data/messaging/eventpublisher.go
tsconn23: Removed these constants altogether due to Fede's comment above and subsequent rework.

internal/core/data/router.go
tsconn23: No, it's just a number. I didn't write this. There's been a discussion between me, @SteveOss and the testing team about making all responses valid JSON types but that's a later effort.

internal/core/data/event_test.go
tsconn23: Agreed, table-driven testing is something we should be more familiar with. Testify is used in the current code for mocking out certain methods and behaviors, but I'm sure it has more capabilities. If you'd care to provide a sample with some guidance, that'd be great.

tsconn23:
I realized after my last commit that I should probably hold off on adding any more to this PR. If I can get an approval on this PR then we can continue with a "Part 2" in another PR. This will at least get the pattern in place and then we can use it as an example to disseminate through the platform.

internal/core/data/router.go
drasko: @tsconn23 do you plan to set `Content-Type` header there?

tsconn23: @drasko It is currently defaulting to text/plain. There is a blackbox test that expects this value and it's passing. If I set it explicitly to "text/plain", it will be different from the other content types in the same file so you have to promise not to call that out. If I leave it as is, since it's working, then we can address as part of the later effort to ensure all requests return valid JSON.

internal/core/data/event_test.go
feclare: You can find some examples in edgex-go, like:
https://github.com/edgexfoundry/edgex-go/blob/master/internal/support/logging/criteria_test.go

internal/core/data/router.go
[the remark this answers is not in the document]
feclare: At least change ``count`` to just return ``dbClient.EventCount``.

I don't really understand how the code is decoupled just by calling ``count`` that only calls ``dbClient.EventCount``.
It makes all the sense to create ``countByDevice`` instead of adding the logic of checking the device existence to the http handler. But I don't agree with ``count``, but it is probably just personal style than a real concern.

internal/core/data/event_test.go
feclare: I created an example table driven test for getEvent: https://gist.github.com/feclare/b901d17c7d9498270e10d7baab2d40ef

feclare:
Another detail in the tests, there is no need of return after a t.Error

tsconn23:
@feclare returns have been removed. Can I get an approval?

tsconn23:
@jduranf I believe I addressed the issues you raised. We've had three pairs of eyes on this now, so please approve and merge if you think it's good.

PR #483, opened 2018-08-15
[snap] Update to support-notifications-go

tonyespy:
For 0.6.1 the go version of support-notifications
was added, so update the snap to use it.

### DIFF
--- a/snap/bin/start-edgex.sh
+++ b/snap/bin/start-edgex.sh
@@ -52,13 +52,11 @@ if [ $SUPPORT_LOGGING = "y" ] ; then
 fi
 
 if [ $SUPPORT_NOTIFICATIONS = "y" ] ; then
-    sleep 65
+    sleep 60
     echo "Starting notifications"
 
-    $JAVA -jar -Djava.security.egd=file:/dev/urandom -Xmx100M \
-               -Dspring.cloud.consul.enabled=true \
-               -Dlogging.file=$SNAP_COMMON/logs/edgex-notifications.log \
-               $SNAP/jar/support-notifications/support-notifications.jar &
+    cd $SNAP_DATA/config/support-notifications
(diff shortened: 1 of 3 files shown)


snap/bin/start-edgex.sh
```diff
@@ -52,13 +52,11 @@ if [ $SUPPORT_LOGGING = "y" ] ; then
 fi
 
 if [ $SUPPORT_NOTIFICATIONS = "y" ] ; then
-    sleep 65
+    sleep 60
```
anonymouse64: My CPU shall forever revel in these 5 seconds back...

snap/snapcraft.yaml
```diff
@@ -382,6 +386,11 @@ parts:
             -e s:'http\://localhost\:48061/api/v1/logs':: > \
        "$SNAPCRAFT_PART_INSTALL/config/support-logging/res/configuration.toml"
 
+      cat "./cmd/support-notifications/res/configuration.toml" | \
+        sed -e s:./logs/edgex-support-notifications.log:/var/snap/edgexfoundry/common/support-notifications.log: \
```
anonymouse64: I guess I never noticed we're doing this replacement here to set the log value to be in `$SNAP_COMMON` at build time. 
I think this is fine as we do it with other services for this PR, but I'd like to file an issue to change this replacement to happen during the `install` hook because on some platforms, such as Arch Linux, $SNAP_DATA and $SNAP_COMMON aren't in the same `/var/snap/edgexfoundry` directories.

tonyespy: Good point re: those cross-distro paths.  +1 on moving all of this logic into the install hook.  Note, we do this for mongo-config too.

anonymouse64: Issue filed for Delhi as https://github.com/edgexfoundry/edgex-go/issues/486

PR #487, opened 2018-08-16
[DO NOT MERGE] Up version number

anonymouse64:
Specifically adding ~pre1 so that all releases to the edge channel version with 0.6.1~pre have a lesser version number than just 0.6.1, when we finally release.

NOTE: Do not merge this until support-notifications in go has landed: #483


anonymouse64:
After discussion with @tonyespy we are now going to increase the grade to stable on california specifically, and change the release scripts to only release to `california/edge`. See also the ci-management PR: https://github.com/edgexfoundry/ci-management/issues/203

PR #490, opened 2018-08-16
snap/snapcraft.yaml: fix config file locations in config-seed-go

anonymouse64:
This should fix builds such as https://jenkins.edgexfoundry.org/view/Snap/job/edgex-go-snap-master-stage-snap/64/console which fails with:

```
10:28:01 install: cannot stat './config/core-command/configuration.toml': No such file or directory
```

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -138,68 +138,6 @@ parts:
     build-packages:
       - make
       - zip
-  config-seed:
-    source: https://github.com/edgexfoundry/core-config-seed-go.git
-    plugin: shell
-    after:
-      - glide
-      - go
-    shell: bash
-    shell-flags: ['-eux', '-o', 'pipefail']
-    shell-command: |
-      export GOROOT=$("$SNAPCRAFT_STAGE/bin/go" env GOROOT)
-      export PATH="$GOROOT/bin:$PATH"
-
-      export XC_ARCH=`dpkg --print-architecture`
-      export XC_OS="linux"
-
-      go version
-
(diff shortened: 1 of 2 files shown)


anonymouse64:
I reworked this PR to accommodate config-seed being moved into edgex-go. Now we don't build from the `core-config-seed-go` repo at all, and only use the monorepo.
This also means that we're now using the new binary that is in the monorepo here: https://github.com/edgexfoundry/edgex-go/tree/master/cmd/config-seed 

I ran into some issues with using the new version of config-seed, mainly that it segfaults if the `res/configuration.toml` file doesn't exist, and additionally there's no way to specify which services to load/not load, so for example it fails when it comes to support-notifications as we're still using the java version in the snap and not using the new go version. We can update the snap to use the go version of `support-scheduler` and install the configuration.toml file for `support-scheduler` once I receive confirmation from @tsconn23 that it's ready to use; last I heard it wasn't ready yet.

snap/snapcraft.yaml
```diff
@@ -347,7 +296,7 @@ parts:
        "$SNAPCRAFT_PART_INSTALL/config/export-distro/res/configuration.toml"
 
       cat "./cmd/support-logging/res/configuration.toml" | \
-        sed -e s:./logs/edgex-support-logging.log:/var/snap/edgexfoundry/common/support-logging.log: \
+        sed -e s:./logs/support-logging.log:/var/snap/edgexfoundry/common/support-logging.log: \
```
anonymouse64: Note that I filed an issue for this logfile being different from the rest: https://github.com/edgexfoundry/edgex-go/issues/545

PR #499, opened 2018-08-23
Use db.MakeTimestamp

jduranf:
Replaced all `time.Now().UnixNano() / int64(time.Millisecond)` by `db.MakeTimestamp()`

As this is a very simple PR, no issue it's created (but will be created if required)


drasko:
Merged, thanks @jduranf

PR #500, opened 2018-08-23
Fix some issues in database test and benchmark

jduranf:
Some issues have been fixed in database test and benchmark.
Most of this work have been done by @itamarhaber in #344, but also some other issues related with database benchmark and test have been fixed.


drasko:
Merged, thanks!

Issue #501, opened 2018-08-24
Check gofmt in all pull requests


jduranf:
In the last commits, several code have been pushed which fails in gofmt check. One month ago, gofmt check was 100% and now is 94%, as can be seen in [edgex-go report card](https://goreportcard.com/report/github.com/edgexfoundry/edgex-go).

Maybe should be a good idea to modify [edgex-go-master-verify](https://jenkins.edgexfoundry.org/job/edgex-go-master-verify-go/) in order to check if the PR pass the gofmt check or not. [Here](https://github.com/opentable/sous/blob/master/bin/check-gofmt) is a simple example in how to do it.

feclare:
I agree
Adding it to make test will make it easy for people to test locally.

tsconn23:
@JPWKU Thoughts?

jduranf:
@JPWKU any news about this suggestion?

jpwhitemn:
@JPWKU - Jeremy, I thought this was already in our CI/CD pipe? Can you comment?  If it is not, we should add to our DevOps roadmap for Edinburgh.  Doable??

andresrinivasan:
Re consideration of gometalinter. I've been using [golangci-lint](https://github.com/golangci/golangci-lint) which aggregates additional linters.

brandonforster:
closed via #1298

PR #502, opened 2018-08-24
Fixed gofmt, ineffassign and misspell issues

jduranf:
Fixed all gofmt, ineffassign and misspell issues detected by the [report card](https://goreportcard.com/report/github.com/edgexfoundry/edgex-go).

Related with #501

### DIFF
--- a/internal/core/data/event.go
+++ b/internal/core/data/event.go
@@ -131,9 +131,9 @@ func addNew(e models.Event) (string, error) {
 					return "", err
 				}
 			}
-			valid, _ := isValidValueDescriptor(vd, e.Readings[reading])
-			if !valid {
-				return "", errors.NewErrValueDescriptorInvalid(vd.Name)
+			err = isValidValueDescriptor(vd, e.Readings[reading])
+			if err != nil {
+				return "", errors.NewErrValueDescriptorInvalid(vd.Name, err)
 			}
 		}
 	}
--- a/internal/core/data/event.go
+++ b/internal/core/data/event.go
@@ -131,7 +131,7 @@ func addNew(e models.Event) (string, error) {
 					return "", err
 				}
 			}
-			valid, err := isValidValueDescriptor(vd, e.Readings[reading])
+			valid, _ := isValidValueDescriptor(vd, e.Readings[reading])
 			if !valid {
 				return "", errors.NewErrValueDescriptorInvalid(vd.Name)
 			}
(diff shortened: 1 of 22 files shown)


internal/core/data/event.go
```diff
@@ -131,7 +131,7 @@ func addNew(e models.Event) (string, error) {
 					return "", err
 				}
 			}
-			valid, err := isValidValueDescriptor(vd, e.Readings[reading])
+			valid, _ := isValidValueDescriptor(vd, e.Readings[reading])
```
feclare: I know it is not directly related to this PR, but can you change the next if to add a test for err != nil

pkg/clients/coredata/reading_test.go
```diff
@@ -17,8 +17,8 @@ package coredata
 
 import (
 	"fmt"
-	"github.com/edgexfoundry/edgex-go/pkg/clients/types"
 	"github.com/edgexfoundry/edgex-go/internal"
+	"github.com/edgexfoundry/edgex-go/pkg/clients/types"
```
feclare: Move up the net and testing imports

internal/core/data/event.go
feclare: Yes, I like your idea.

jduranf:
@feclare, modified `isValidValueDescriptor()` as commented

internal/core/data/reading.go
```diff
@@ -72,10 +73,11 @@ func readingHandler(w http.ResponseWriter, r *http.Request) {
 				return
 			}
 
-			valid, err := isValidValueDescriptor(vd, reading)
-			if !valid {
-				http.Error(w, "Validation failed", http.StatusConflict)
-				LoggingClient.Error("Validation failed")
+			err = isValidValueDescriptor(vd, reading)
+			if err != nil {
+				err = errors.NewErrValueDescriptorInvalid(vd.Name, err)
```
feclare: Maybe, instead of creating the ErrValueDescriptorInvalid here that isValidValueDescriptor return it directly?

jduranf: Done!

pkg/models/subscription.go
```diff
@@ -28,7 +28,7 @@ import (
  * Subscription struct
  */
 type Subscription struct {
-	BaseObject                                   `bson:",inline"`
+	BaseObject           `bson:",inline"`
```
drasko: Looks confusing, needs indenting and alignment to other fields.

jduranf: I agree with you, looks confusing, but otherwise gofmt fails... also there are several other structs (like Addressable or Command) with the same "confusing" structure, so I think it's better to not modify again this line.

drasko: OK, valid points.

drasko:
@jduranf please update branch and let's merge this baby ;)

PR #504, opened 2018-08-24
Add AWS IoS support in export

davidmsh:
A new topic "AWS_JSON" and new destination "AWS_TOPIC" are added
And two extra properties AWSCert, AWSKey in configuration.toml for AWS specific cert and key authentication


davidmsh:
@jduranf I have to say I agree with you. But given current limitation of Addressable, we don't have a good way to accomplish the same goal. I ever developed a Java version of AWS export "borrowing" some of the properties in Addressable. But really think specifying directory from external party is not a good idea as far as security or operatability are s concerned. So I think we take this approach before we have a complete key/cert management in place in edgex. And I really don't like borrowing properties around which would cause lots of confusion. Your thought ?

feclare:
I agree with @davidmsh 
@janko-isidorovic, does the new export services redesign take care of this issue?

davidmsh:
The other way would be sending cert/key as base64-enacoded blob in a rest call. But we need to make the endpoint to support https. Any way we have to borrow unrelated properties in Addressable which is I'm not willing to do in an official release.

jduranf:
@davidmsh I agree with you that none of the both options is good, but I prefer to borrow two variables (which are quite equivalent) than the side effects related in using the configuration file.
But this is just my opinion.

davidmsh:
If I remember correctly, the new export service design will support multiple instance of export services and local configuration w/o dynamic registration. If that part is ever implemented, key/cert properties in configuration won't become a problem.

tsconn23:
recheck

davidmsh:
@feclare Done adding AWS forst and dest to getRegList()
But it looks like there are many other types not on the list, is it on purpose ?

feclare:
@davidmsh No, it isn't on porpouse. I just realised that the recently added formaters and  destinations weren't included in the list. 
If you want you can add them, if not I will create a PR for fixing it.

davidmsh:
@feclare Added the rest missing entries to RegList

export/distro/mqtt.go
```diff
@@ -41,7 +41,7 @@ func NewMqttSender(addr models.Addressable) Sender {
 		protocol == "ssl" ||
```
drasko: Don't break the lines here, it's actually harder to read.

export/distro/mqtt.go
```diff
@@ -41,7 +41,7 @@ func NewMqttSender(addr models.Addressable) Sender {
...
 		protocol == "tls" {
```
drasko: Kill empty line.

davidmsh:
@drasko done

drasko:
@feclare @jduranf please review and let's merge this one.

jduranf:
@drasko I still disagree with the idea to have some export client parameters in the configuration, but the rest of this PR it's fine for me

davidmsh:
@feclare @drasko @jduranf 
Hey guys, Do we have an agreement on how to handle cert/key file yet ? Even we don't. May I suggest we merge this change first then open an issue to track it ? Or I have to rebase from master from time to time waiting it to be merged. It took me some time for merging and testing each time of rebasing as the export code is under some non-trivial change.

export/distro/const.go
```diff
@@ -0,0 +1,27 @@
+//
```
jduranf: This file should be removed

davidmsh:
@jduranf done. some miss during rebasing

PR #511, opened 2018-08-25
Add schedule related clients

tonyespy:
This commit adds public REST clients for Core Metadata's **schedule** and **scheduleevent** endpoints.  These are needed to allow device-sdk-go to query and add both object types.

### DIFF
--- a/pkg/clients/metadata/schedule.go
+++ b/pkg/clients/metadata/schedule.go
@@ -1,0 +1,307 @@
+/*******************************************************************************
+ * Copyright 2018 Dell Inc.
+ * Copyright (C) 2018 Canonical Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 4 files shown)


tsconn23:
@tonyespy The only comment I have on this has to do with file naming. pkg/clients/metadata/scheduleevent.go should have an underscore in it to conform with the rest of the files names. If that's not agreeable, then perhaps the ScheduleEventClient can be included in schedule.go. I believe it was you and I that were discussing the approach of defining multiple types in a single file scoped to the overall intent versus the Java style of putting one type in a respectively named file. Thoughts?

Otherwise, code looks fine and can be accepted.

anonymouse64:
Just to chime in on filenames, I agree with Tony that we should attempt to follow what's done in the models package as that seems to follow the upstream go projects naming convention.

tsconn23:
@tonyespy Will add to Core WG agenda for discussion

PR #513, opened 2018-08-27
Rev to 0.6.2 for future release

JPWKU:


tonyespy:
A couple of comments/suggestions:

 * In the future for maintenance branches, I would prefer if we updated the version on-demand once we know for sure there's going to be a dot release. We're still working on releasing a 0.6.1 snap to **california/candidate** channel, but now VERSION has been updated to 0.6.2.  Creating a branch for this seems like a heavy-weight solution at this point.

 * I checked, and it seems none of the services actually reference this file.  Is it installed in the Docker containers for each service?

 * Looks like the services all use the file[ version.go](https://github.com/edgexfoundry/edgex-go/blob/california/version.go) which wasn't changed. Perhaps we can re-open this issue, and leave it open until we actually need a 0.6.2 release, and then update version.go then, as per my original suggestion?

JPWKU:
@tonyespy 
- The release refers to the actual code and not the packaging mechanism, I was under the impression that the snap version would follow from the code version; I realize this is a sticky point since snaps contain more than 1 service.  I'm not understanding what you are trying to say about creating a branch.
- VERSION is actually used in the go compilation.  You can see it here in the Makefile https://github.com/edgexfoundry/edgex-go/blob/master/Makefile#L20-L22.  I wanted a simple VERSION file to cat in the CI process since it's used in tagging and labeling the docker images, as well as setting the go bin filenames.
- See my first point, lets discuss this soon.

SteveOss:
While we are considering what we want to do for release "process" as a minimum can I suggest that
we identify all main stakeholders (e.g. Tony for Snap support), and before doing any branch or tag
release, get the sign off from them all that they are ready for a release (also a good point to gather release 
notes). IMHO the last thing that should be done for a release (after generating/testing/reviewing all release
artefacts - images, docs etc), is to increment the (various) version numbers.

tonyespy:
@JPWKU 

 * I realize the release refers to the actual code, but as Makefiles aren't really standard Go practice, I didn't search it.  Turns out the Makefile does this:

`GOFLAGS=-ldflags "-X github.com/edgexfoundry/edgex-go.Version=$(VERSION)"
`
...which I think is defining a string or variable embedded in the binary. That said, I can't find a man page or linker reference which explains the **-ldflags -X** args, so maybe a comment is a good idea in the Makefile to explain this?

 * What I'm saying about the snap is that if were to build a snap this afternoon and publishe it in the Snap Store, this version would be 0.6.2, not 0.6.1, and from what you've said, this version is being included in the final binaries.  If we included native RPM or Debian packaging in the build, we'd have the same problem.

 * What I meant about a branch, is that if we really were going to start development on 0.6.2, but knew we still needed to release another iteration of 0.6.1, we'd create a branch to do this. My contention is that to do this is a bit heavy weight for what we're trying to do here.  So, all I was suggesting was again that for maintenance releases, we don't rev VERSION till we start working on a new VERSION.  This allows us to make packaging fixes to Dockerfiles or snapcraft.yaml without having to worry that VERSION had already changed.

 * My point about version.go is till valid though, we missed this.

PR #516, opened 2018-08-27
snap/bin/start-edgex.sh: go into a safe directory before starting

### LINKED ISSUE #515: security denial for opening original working directory from start-edgex.sh
I didn't notice this denial earlier, but this happens when running strictly confined either on a 16.04 classic system or on a 18.04 classic system. I haven't tested a UC16 system yet. If you run the `edgexfoundry.start-edgex` script as sudo from say your (non-root) normal user's $HOME directory like follows:

```bash
cd $HOME
sudo edgexfoundry.start-edgex
```

You will see the following denial show up in the system logs:

```
Aug 27 13:29:27 xenial-classic kernel: audit: type=1400 audit(1535394567.710:466): apparmor="DENIED" operation="open" profile="snap.edgexfoundry.start-edgex" name="/home/ijohnson/" pid=4060 comm="start-config-se" requested_mask="r" denied_mask="r" fsuid=0 ouid=1000
```

I was able to fix this issue by adding a `cd $SNAP_DATA` to the top of the start-edgex.sh script, though I don't understand the underlying issue as the script triggering this (`comm="start-config-se"` is almost certainly the `start-config-seed.sh` script) doesn't use cd anywhere except here: https://github.com/edgexfoundry/edgex-go/blob/california/snap/bin/start-config-seed.sh#L33

which should never trigger a `cd $HOME` or anything like that...

anonymouse64:
Currently, when running start-edgex say from one's $HOME directory as sudo, there will be a denial in the syslog about comm="start-config-se" (which I believe is the start-config-seed.sh script) attempting to open $HOME and being denied. EdgeX doesn't need to access this directory so I think this is due to some shell command attempting to read the original working directory, which would be $HOME. We can resolve this by first going into $SNAP_DATA which is a writable location and then running the rest of the scripts from there.

Fixes #515


anonymouse64:
recheck

PR #520, opened 2018-08-30
Enable parameter wasn't updated in registration export-client

### LINKED ISSUE #512: It's not possible to modify the Enable parameter in the Registration
Currently it's not possible to modify the Enable parameter in the Registration (export-client).
The code in registration.go is:
```
	if fromReg.Name != "" {
		toReg.Name = fromReg.Name
	}
	if fromReg.Addressable.Name != "" {
		toReg.Addressable = fromReg.Addressable
	}
	if fromReg.Format != "" {
		toReg.Format = fromReg.Format
	}
	if fromReg.Filter.DeviceIDs != nil {
		toReg.Filter.DeviceIDs = fromReg.Filter.DeviceIDs
	}
	if fromReg.Filter.ValueDescriptorIDs != nil {
		toReg.Filter.ValueDescriptorIDs = fromReg.Filter.ValueDescriptorIDs
	}
	if fromReg.Encryption.Algo != "" {
		toReg.Encryption = fromReg.Encryption
	}
	if fromReg.Compression != "" {
		toReg.Compression = fromReg.Compression
	}
	if fromReg.Destination != "" {
		toReg.Destination = fromReg.Destination
	}
```

So, the Enable parameter isn't updated.

jduranf:
This PR fix #512.
Maybe it's not the best way to fix the issue, but I don't know any best way without touching Registration structure.

### DIFF
--- a/export/client/registration.go
+++ b/export/client/registration.go
@@ -209,6 +209,9 @@ func updateReg(w http.ResponseWriter, r *http.Request) {
 	if fromReg.Destination != "" {
 		toReg.Destination = fromReg.Destination
 	}
+
+	// In order to know if 'enable' parameter have been sent or not, we unmarshal again
+	// the registration in a map[string] and then check if the parameter is present or not
 	var objmap map[string]*json.RawMessage
 	json.Unmarshal(data, &objmap)
 	if objmap["enable"] != nil {
--- a/export/client/registration.go
+++ b/export/client/registration.go
@@ -209,6 +209,11 @@ func updateReg(w http.ResponseWriter, r *http.Request) {
 	if fromReg.Destination != "" {
 		toReg.Destination = fromReg.Destination
 	}
+	var objmap map[string]*json.RawMessage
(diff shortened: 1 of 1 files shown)


feclare:
It is not easy to understand why this code is needed. Can you add a comment in the code explaining it?
I will approve this change as it is, but another solution is to create a new registration type with all the fields being pointers. This way it can be tested if a fields exists in the json. It also fixes the problem of not being able to set any of the strings to an empty string.
I don't like the idea of duplicating types, but it is the only alternative I can think of.

jduranf:
@feclare I adopted this solution because it's less intrusive than your proposal and doesn't require to duplicate the types, but mine it's less effective and more tricky.
If you want, I can modify the code to use the duplicate types solution, otherwise I prefer to leave as it is.

feclare:
I don't know which of both solutions I prefer, so I propose to keep it as you did.

drasko:
Merged, thanks.

PR #523, opened 2018-08-30
Add support for basic auth to HTTP export client

### LINKED ISSUE #522: Add support for basic authentication to HTTP export client
Other export client implementations already allow authentication and I was thinking it would be interesting to add support for basic authentication in the HTTP client as well.

I'll submit a PR to address this.

adrianmo:
This PR fixes #522.

Signed-off-by: Adrián Moreno <adrian@morenomartinez.com>

### DIFF
--- a/export/distro/http.go
+++ b/export/distro/http.go
@@ -19,8 +19,10 @@ import (
 )
 
 type httpSender struct {
-	url    string
-	method string
+	url      string
+	method   string
+	user     string
+	password string
 }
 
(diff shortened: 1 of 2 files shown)


drasko:
@tsconn23 @feclare @jpwhitemn @jduranf please advise.

tsconn23:
@drasko I agree. I'd think this should allow flexibility to associate Auth strategy with registration endpoint. Could be several ways of doing Auth over HTTP.

PR #530, opened 2018-09-01
EF-529 - Use empty username/pass for Mongo

### LINKED ISSUE #529: Core config should use default empty Mongo user/pass
Currently some of the core services use different `username` and `password` for MongoDB. This prevents script to easily launch services on the local machine (not in docker) during the dev phase, without modification of `.toml` files.

drasko:
Resolves #529


drasko:
@jpwhitemn situation we have today is not consistent - so it is confusing. Some of the config files keep `username` and `password` empty, a few of them don't. I do not understand why someone would put in the default template config files this user/pass when there is no way that this will correspond to anything that developers have on their machines - these username/pass from template `.toml` would have to be changed anyway. But now they only make problem, becuse I had to go and manually change 3 files in 3 different directories - all having different user/pass for the DB.

Keeping username/pass empty at least would permit developers to easily spin the system just by executing `edgex-launch.sh`.

drasko:
One more thing - people who use Docker containers will not see the change, because they should use dockers from the release, not the one from the master. Master is dedicated for developers, who should know their way around.

drasko:
@jpwhitemn are you refering to this `init` scipt: https://github.com/edgexfoundry/developer-scripts/blob/master/init_mongo.js?

Can you point me in the code when is this one run? I do not see it in Docker files.

Actually, from here: https://github.com/edgexfoundry/developer-scripts/blob/master/compose-files/docker-compose-california-0.6.1.yml#L77 I'd say that core services do not even use `mongo` container and map from there to disk, but demand of user to install MongoDB on a host and set it (at least I can not see here which other MongoDB container is used by core services).

jpwhitemn:
@drasko - agree, we should have consistency.  We also need to make sure that the scripts are consistent with the changes.  The init_mongo.js and clean_mongo.js is another (might be others - I have not done and exhaustive look).  Per the recent core working group meetings, @tsconn23 and @xmlviking have been working on the Config V2 (organizing the configuration and providing for a hierarchy at the same time).  Trevor is also working to make sure that there is only a single config file and it gets moved to config seed during the CI/CD build process.  Hopefully, these upcoming changes will mean that you won't have to change the username/password in three places.  Suggest you connect with Trevor.  If he approves this PR and it is consistent with his upcoming changes, then you can dismiss my review.

tsconn23:
@drasko There's a mongo image in that docker-compose file you linked to. The configuration-docker.toml files all point to that image via hostname, so I'm not sure what you're trying to point out here. I'm not onboard with this PR for reasons Jim has already mentioned.

drasko:
@tsconn23 I am trying to point out that it has been wrong for a very long time. Please read the explications - it is inconsistent on the project level and confusing. It is also making mess when you do not want to develop in docker. And finally - it is an arbitrary security policy that someone invented, there is no project-wide decision from Security WG.

tsconn23:
@drasko ```Currently when you execute make run half of the services start running, and half fail because of connection with MongoDB.```

This is not the case in my experience. It could be that you have not run the init_mongo.js script from the developer-scripts repo or have otherwsie changed the credentials in Mongo to differ from what's in the edgex-go repo.

drasko:
OK, @tsconn23, let me then rephrase the question: do you expect that anyone will run EdgeX in production with defult user names and passwords that we keep in GitHub?

If not - why does it matters if we put some dummy text there (which has absolutely no value and does not contribute to any security) and not keep them just empty placeholders?

Because today when you put some text, that forces developers to change MongoDB settings on their hosts (which is empty username/pass by default) or change these files by hand to make these fields empty (as I explained in the blog).

jduranf:
I'm not a MongoDB expert, but core-metadata have empty fields in MongoDBUserName and MongoDBPassword in configuration.toml, and I have no problems with core-metadata. Also taking in account that `init_mongo.js` sets a user and password for metadata...
Also exportclient isn't created by `init_mongo.js`, but works without problems.

I agree with @drasko that it's strange to have core-metadata and export-client without user/password and core-data, support-logging and support-notifications with password. I think the criteria should be unified.

drasko:
Problem is that I do not want to run `init_mongo.js` on my host (Javascript, really? Always reminds me of https://www.destroyallsoftware.com/talks/wat). This is my dev host (my machine), and not a container I will destroy later.

Setting machine correctly demands a lot of careful devops work, Mongo & Co will be probably run as a statefull replica sets in Kubernetes - so if I have to go through all this work, I'll save it for production.

For development I just need something that I can run fast and smoothly on my dev host, without too much useless friction.

tsconn23:
One other general comment -- Mongo in docker is initialized via the following script in the mongo-seed repo:
https://github.com/edgexfoundry/docker-edgex-mongo/blob/master/init_mongo.js

These logins need to remain consistent with what's in the configuration-docker.toml files (they're fine right now). It would be nice to bring the init_mongo.js script in the developer-scripts repo inline with these changes, perhaps by simply removing the accounts. The developer-scripts file is only run when testing locally, creating default collections and so forth.

drasko:
@tsconn23 great, this will probably help with issues like these: https://github.com/edgexfoundry/edgex-go/issues/533

I'll update the branch.

Regarding the config scripts - and everything related EdgeX, for example `device services` - should be included into monorepo. All of these artifacts are tightly related to project and development, and keeping them in separate repos and far away from developers eyes just make development harder (need to clone several repos) and can lead to inconsistencies and diverging code. So +1 from my side in including all config scripts here.

PR #531, opened 2018-09-02
NOISSUE - Add provisioning scrip

drasko:

### DIFF
--- a/scripts/provision/camera_monitor_profile.yml
+++ b/scripts/provision/camera_monitor_profile.yml
@@ -1,0 +1,106 @@
+#
+# Copyright (c) 2018
+# Dell
+# Mainflux
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+# File taken from https://docs.edgexfoundry.org/Ch-WalkthroughDeviceProfile.html
+
+name: "camera monitor profile"
+manufacturer: "Dell"
+model: "Cam12345"
+labels: 
+    - "camera"
+description: "Human and canine camera monitor profile"
+commands: 
+  - 
+    name: People
+    get: 
+        path: "/api/v1/devices/{deviceId}/peoplecount"
+        responses:
+          - 
+            code: "200"
+            description: "Number of people on camera"
+            expectedValues: ["humancount"]
+          -
+            code: "503"
+            description: "service unavailable"
+            expectedValues: ["cameraerror"]
+  - 
+    name: Dogs
+    get: 
+        path: "/api/v1/devices/{deviceId}/dogcount"
+        responses:
+          - 
+            code: "200"
+            description: "Number of dogs on camera"
+            expectedValues: ["caninecount"]
+          -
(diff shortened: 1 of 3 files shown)


scripts/provision/camera_monitor_profile.yml
```diff
@@ -0,0 +1,106 @@
...
+        path: "/api/v1/devices/{deviceId}/counts"
+        responses:
+          -
+            code: "200"
+            description: "Get the people and human counts"
+            expectedValues: ["count","count"]
+          -
+            code: "503"
+            description: "service unavailable"
+            expectedValues: []
```
jduranf: As defined in [Device Services functional requeriments](https://wiki.edgexfoundry.org/download/attachments/7602423/edgex-cali-delhi-device-requirements-v8.pdf?version=1&modificationDate=1533761857000&api=v2): _A DeviceProfile must at minimum define a single DeviceObject_, and this is not the case. Should be fine if this example is according to specifications.

scripts/provision/provision.sh
```diff
@@ -0,0 +1,250 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+###
+# Core Metadata
+###
+printf "> Provisioning Addressibles\n"
+
+# Addressable for the Device Service
```
jduranf: Why not use edgexfoundry/device-sdk-go/example/simple instead of creating a device service? Although is still in alpha state, this will simplify the provisioning script and add visibility to edgexfoundry/device-sdk-go

scripts/provision/provision.sh
```diff
@@ -0,0 +1,250 @@
...
+# Copyright (c) 2018
+# Mainflux
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+###
+# Core Metadata
+###
+printf "> Provisioning Addressibles\n"
```
jduranf: Addressibles → Addressables

scripts/provision/provision.sh
```diff
@@ -0,0 +1,250 @@
...
+    "user":"none",
+    "password":"none",
+    "topic":"none"
+}
+EOF
+
+###
+# Core Data
+###
+printf "\n> Provisioning Value Descriptors\n"
```
jduranf: If edgexfoundry/device-sdk-go/example/simple is used, then there is no need to add value descriptors

scripts/provision/provision.sh
```diff
@@ -0,0 +1,250 @@
...
+        "name":"camera control device service"
+    },
+    "profile":{
+        "name":"camera monitor profile"
+    }
+}
+EOF
+
+###
+# Add Events
```
jduranf: Why are added those events?

drasko:
@jduranf whoops, I saw the review just now, sorry I missed it. I'll amuse myself with it over the weekend ;). Thanks!

Issue #517, opened 2018-08-28
Merge core-config-seed-go into the edgex-go repo


tsconn23:
In preparation for the upcoming ConfigV2 enhancements, we discussed during the 23-Aug-18 Core WG call the idea of merging the core-config-seed-go repo into the edgex-go repo. There was no emphatic opposition to doing this, so I'm creating this issue to initiate the effort.

- Initial migration should cover current functionality
- Java application.properties files should be in their own folder (supported through Edinburgh)
- Go service configurations will be read from their respective cmd/res directories
- Docker image will be built from edgex-go repo (retire old CI jobs)

PR #532, opened 2018-09-04
Merge config-seed into edgex-go

tsconn23:
#517 

This PR contains work to merge the config-seed service into the edgex-go repo as has been discussed on the Core WG calls. There are several points to note w/r/t this PR:

- The PR looks bigger than it is due to all of the application.properties files that had to be brought over. This is so that we can avoid config file duplication between two repos.
- As described on the calls, the goal was to achieve functional parity with the existing core-config-seed-go service. No functionality for Config V2 has been added yet.
- Today the config-seed is not run unless one wishes to exercise EdgeX via docker containers, so I have not added it to the `make run` command. It IS present in `make build`. You can run the service via
```
cd cmd/config-seed
./config-seed
```
- I have added a docker target for creating the relevant image. I have tested this locally and verified Consul population.
- No accounting has been made for Snap deployment here. I was discussing my approach with @tonyespy last week and sent him the included Dockerfile. Feedback is TBD.
- For the legacy application.properties files, I put these in the local .config-seed/res/properties directory. This seemed the easiest to me and this directory will be removed after Edinburgh. (This was shown on last week's Core WG call)
- For the Go services, TOML is assumed. If not running the services via local cmd line, then one needs to use the "--cmd" command line property to specify where the `edgex/cmd` directory lives (this includes via GoLand IDE)
- Approval/merge of this PR is a pre-requisite for continuing on with the Config V2 work

I look forward to any comments/input you may have.

### DIFF
--- a/cmd/config-seed/main.go
+++ b/cmd/config-seed/main.go
@@ -16,27 +16,26 @@ package main
 import (
 	"flag"
 	"fmt"
-	"os"
-	"os/signal"
-	"sync"
-	"syscall"
-
 	"github.com/edgexfoundry/edgex-go/internal"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/usage"
 	"github.com/edgexfoundry/edgex-go/internal/seed/config"
 	"github.com/edgexfoundry/edgex-go/pkg/clients/logging"
+	"sync"
 )
 
 var bootTimeout int = 30000 //Once we start the V2 configuration rework, this will be config driven
 
 func main() {
 	var useProfile string
+	var dirCmd string
 	var dirProperties string
 
 	flag.StringVar(&useProfile, "profile", "", "Specify a profile other than default.")
 	flag.StringVar(&useProfile, "p", "", "Specify a profile other than default.")
-	flag.StringVar(&dirProperties, "props", "./res/properties", "Specify alternate properties location")
-	flag.StringVar(&dirProperties, "r", "./res/properties", "Specify alternate properties location")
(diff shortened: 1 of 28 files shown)


cmd/config-seed/main.go
```diff
@@ -0,0 +1,84 @@
...
+	var dirProperties string
+
+	flag.StringVar(&useProfile, "profile", "", "Specify a profile other than default.")
+	flag.StringVar(&useProfile, "p", "", "Specify a profile other than default.")
+	flag.StringVar(&dirProperties, "props", "./res/properties", "Specify alternate properties location as absolute path")
+	flag.StringVar(&dirProperties, "r", "./res/properties", "Specify alternate properties location as absolute path")
+	flag.StringVar(&dirCmd, "cmd", "../cmd", "Specify alternate cmd location as absolute path")
+	flag.StringVar(&dirCmd, "c", "../cmd", "Specify alternate cmd location as absolute path")
+
+	flag.Usage = usage.HelpCallback
```
jduranf: `usage.HelpCallback` it's not valid in this case

tsconn23:
@drasko As I said, I'm trying to achieve functional parity with the current config-seed-go which uses ";". When we get to config V2, the need to use ";" will go away. I need this PR to be accepted first.

drasko:
@tsconn23 OK then - approved.

cmd/config-seed/main.go
tsconn23: @jduranf Good catch. Config-seed specific support has been added.

JPWKU:
https://github.com/edgexfoundry/ci-management/pull/209 should be merged before this is merged.

Issue #538, opened 2018-09-07
Config V2 -- core-command


tsconn23:
Changes to support config V2 in the core-command service:

1.) Reorganize configuration file using hierarchical TOML
2.) Hydrate TOML hierarchy into composable structs rather than one flattened struct
3.) Allow for re-use of individual configuration structs by other services
4. ) When passing configuration information into initialization functions, only pass the relevant struct
5.) When full URLs are present in current config, eliminiate them. Instead use host/port to assemble URLs within the given service client.
6.) Ensure the following settings are part of configuration -- period of service bootstrap, service client refresh interval

PR #541, opened 2018-09-09
Config V2 -- Config Seed and Core-Command

tsconn23:
#538 

This PR contains the first set of changes toward the V2 configuration scheme, and will probably be the largest of the effort since it touches two services -- config-seed and core-command.

Some points to note:

- While the services are in transition to ConfigV2 it is necessary to maintain backward compatibility for services that have not yet been migrated. You will see functions with V2 in their name which are largely clones of their predecessors (such as pkg/config/LoadFromFileV2). Once the final service is migrated, the old functions will be deleted and the V2 functions will be renamed
- The "res" directory containing core-command's configuration.toml now has a docker folder with its own configuration.toml instead of a configuration-docker.toml. I plan to make this change in the rest of the services whereby the filename is standardized, with each profile having a folder.
- If you run the config-seed, it will create a config/V2 key in Consul. The hierarchical configuration structure for the given services (edgex-core-command in this case) will fall under that.
- Note the removal of metadata client URLs from configuration. As discussed, these are now being built from constituent elements such as protocol, host, port and routes defined as constants.
- Note that the particular "Info" type (such as ServiceInfo or ClientInfo) has a method to construct the requisite URL or HealthCheck, eliminating the need to do this inline each time it is needed.
- During execution of the blackbox tests I found a few methods in the command API still returning "StatusServiceUnavailable" instead of "InternalServerError". I included the fix for these in this PR.
- Following @tonyespy's lead in the Device Service SDK, I used one file to define all of the configuration Info types and their BaseConfig composition. This seemed to make sense since the types are all very small, as opposed to creating a bunch of smaller files.

When this PR goes in after review, it will set the pattern for the Config V2 refactoring in all subsequent services. I look forward to your comments.


cmd/core-command/res/configuration.toml
```diff
@@ -1,34 +1,26 @@
...
-MetaCommandPath = '/api/v1/command'
-MetaEventURL = 'http://localhost:48081/api/v1/event'
-MetaEventPath = '/api/v1/event'
-MetaScheduleURL = 'http://localhost:48081/api/v1/schedule'
-MetaSchedulePath = '/api/v1/schedule'
-MetaProvisionWatcherURL = 'http://localhost:48081/api/v1/provisionwatcher'
-MetaProvisionWatcherPath = '/api/v1/provisionwatcher'
+Host = 'localhost' #Is this used?
+Port = 48082
+Protocol = 'http' #Is this used?
```
jduranf: If not used, maybe should be better to remove it

cmd/core-command/res/configuration.toml
```diff
@@ -1,34 +1,26 @@
...
+
+[Registry]
+Host = 'localhost'
+Port = 8500
+Type = 'consul'
+
+[Logging]
+EnableRemote = false
+File = './logs/edgex-core-command.log'
+RemoteURL = 'http://localhost:48061/api/v1/logs' #Should this also be decomposed -- protocol / host / port?
```
jduranf: I think should be decomposed and /api/v1/logs moved into pkg/clients/constants.go

pkg/clients/metadata/command_test.go
```diff
@@ -18,14 +18,15 @@ import (
...
 
 func TestNewCommandClientWithConsul(t *testing.T) {
-	deviceUrl := "http://localhost:48081" + commandUriPath
+	deviceUrl := "http://localhost:48081" + clients.ApiCommandRoute
 	params := types.EndpointParams{
 		ServiceKey:  internal.CoreMetaDataServiceKey,
-		Path:        commandUriPath,
+		Path:        clients.ApiCommandRoute,
 		UseRegistry: true,
 		Url:         deviceUrl}
```
jduranf: Interval isn't initialized. There are more cases like this one.

cmd/core-command/res/configuration.toml
tsconn23: This is included in the updated PR

internal/seed/config/populate.go
```diff
@@ -216,3 +291,56 @@ func readTomlFile(filePath string) (map[string]string, error) {
...
+			kvs = append(kvs, skvs...)
+		}
+	case int:
+		kvs = append(kvs, &pair{Key: path, Value: strconv.Itoa(j.(int))})
+	case int64:
+
+		var y int = int(j.(int64))
+
+		kvs = append(kvs, &pair{Key: path, Value: strconv.Itoa(y)})
+		//kvs = append(kvs, &KV{Key: path, Value: strconv.FormatInt(j.(int64),64)})
```
drasko: Dead code.

tsconn23: I will remove this in the updated PR.

Issue #544, opened 2018-09-10
don't call device service call back with address is absent in metadata


jpwhitemn:
Don't issue callback to Device Service and get 404 error when the device service callback hasn't been provided.  See:  https://github.com/edgexfoundry/edgex-go/blob/eedc9eb0a22020749f2a2a4d87fc2a2b8f2f8a12/internal/core/metadata/rest_device.go#L1193

PR #549, opened 2018-09-11
Do not attempt callback if no URL provided

tsconn23:
#544 


tsconn23:
recheck

Issue #202, opened 2018-05-09
Brightics IoT export connector


hyogrin:
We would like to implement another export-distro which will allow EdgeX to export device data (via export distro) to Brightics IoT which is an B2B targeted enterprise IoT platform. Since Brightics IoT supports MQTT protocols, we will contribute another Brightics IoT-MQTT-export-distro in Go so that our platform users allow to collect device data via EdgeX device services. The target release version would be delhi.

 

* For your information, we have been developing an enterprise IoT platform named "Brightics IoT" for about 3years. (https://www.samsungsds.com/global/en/solutions/off/insa/brighticsIoT.html) The reason why we decide to participate EdgeX poject is that we want to extend platform capabilities via device services of EdgeX.
[1525864011977_Brightics IoT Briefing (EdgeX)_20180509.pdf](https://github.com/edgexfoundry/edgex-go/files/1987555/1525864011977_Brightics.IoT.Briefing.EdgeX._20180509.pdf)

hyogrin:
Please feel free to add any comments we are going to implement the export-distro starting next week.

hyogrin:
We will start the work from next month(Because of the another biz trip). Let's keep in touch on the rocketchat to talk about technical issues and share the architecture / design documents.

jduranf:
@hyogrin still working in this issue or we can close it?

hyogrin:
I have been testing some of test cases right now but it is ok. you can close it.

hyogrin:
@jduranf Still need some time. I have been testing some testcases right now. Would you give me a couple of weeks?

jduranf:
@hyogrin sure, I just want to know if this issue is still alive or not

jduranf:
Done via #565

PR #550, opened 2018-09-12
#202 BIoTMessage, brightics IoT biotFormatter

hyogrin:
added BIoTMessage, brightics IoT biotFormatter (BIoTMessage <-> bytepayload)
I will reuse the mqttSender to connect and sent IoT data.
Sorry for the late because of installing and testing under strong security policy;;  :-)


jduranf:
@hyogrin there are some things you should fix:
- Build fails, please execute `make` and `make test` before commit the code
- All commits need DCO sign-off. Info is in the yellow box on this [page](https://wiki.edgexfoundry.org/display/FA/Contributor+Process)
- Also run `go fmt` before commit

tsconn23:
@hyogrin There needs to be an issue associated with this PR that describes what you're trying to accomplish. I would prefer the issue be created before the PR so that it can be classified with regard to the Delhi or Edinburgh releases, or if it's a bug vs enhancement.

hyogrin:
@tsconn23 PR was this one "Brightics IoT export connector - https://github.com/edgexfoundry/edgex-go/issues/202"  jpwhitemn added enhancement  feature  export-services  labels on May 12 for Edinburgh but I think Delhi will be fine because It is just an enhancement.

PR #556, opened 2018-09-14
replaced normal error type with ErrServiceClient in pkg/clients packages

cloudxxx8:


jduranf:
Nice work @cloudxxx8, but sign is missing

cloudxxx8:
@jduranf , recommitted with sign.  thanks for the reminder

PR #559, opened 2018-09-14
security services forward-port from california

anonymouse64:
This backports the security work from california into master. Specifically PR: #464 

This was tested on Ubuntu 16.04 LTS classic and on Ubuntu 18.04 LTS classic, and has basically parity with the current california snap in the candidate channel.

Note: currently this builds the security services from california, using the shell scripts that were implemented there. I will open a follow-up PR to support the new implement in go. This is meant to provide an initial implementation for edge builds for delhi.


anonymouse64:
@tonyespy To be fair, it was a lot easier to just combine all the work from my branch on top of master than it was to just cherry-pick out the commits that add parts like kong, etc. I think for the very same reasoning as you that it's not a big deal to *have* the california version of security services in the edge snap, it's just easier for me to include it here than go through and pick out those bits for master in this PR.

However, if you still want me to remove those bits I can do that. I would still like to include things like kong, vault, etc. in this PR however.

anonymouse64:
recheck

PR #563, opened 2018-09-17
Add dockerfile to build snap

### LINKED ISSUE #557: Use docker to build the snap for CI builds
Currently, we need to spin up a 16.04 VM to do the snapcraft builds, as the snap cannot currently be built with any other OS without using some sort of a container (i.e. snapcraft cleanbuild). Instead, we can do run snapcraft inside of a docker container using the official docker image for snapcraft: https://hub.docker.com/r/snapcore/snapcraft/

anonymouse64:
This provides a simple Dockerfile (and script which builds the docker image) that is to be used by CI to build the snap. This removes the need to build the snap on Ubuntu 16.04, and the build for the snap may be run on any OS that runs docker.

To test, clone this branch and just run:

```bash
snap/build.sh
```

After a while you should see output similar to this:

```
[[ lots of other log output ]]
Staging mongo-config 
Staging mongodb 
Priming config-common 
Priming go 
Priming consul 
Priming curl 
Priming glide 
Priming edgex-go 
Priming mongo-config 
Priming mongodb 
Determining the version from the project repo (version-script).
The version has been set to '0.5.2-20180917+11f7d4b'
Snapping 'edgexfoundry' ...

Snapped edgexfoundry_0.5.2-20180917+11f7d4b_amd64.snap
```

The build is currently entirely done inside the container, and whether or not to do the release is determined by a few Jenkins specific environment variables, i.e. `JENKINS_URL` and `JOB_NAME`. If `JENKINS_URL` exists, then we copy the snap store login file into the container. Then we check if the JOB_NAME matches a stage job, and if it does then set a environment variable so the entrypoint script will attempt to release the snap on the specified snap channel.

See the associated ci-management job: https://github.com/edgexfoundry/ci-management/pull/211

This fixes #557


JPWKU:
@anonymouse64 I didn't think of this before, but we will need a docker image that is both x86 and ARM compatible.

anonymouse64:
Hmm, you are right, and the `snapcore/snapcraft` image is only published for amd64. I will ask and see if there are plans to support additional architectures for the image, if not it's not too hard for us to roll our own.

anonymouse64:
@JPWKU one thing I would like for you to take a particular look at is how the snap store login file is handled here. I need to copy it into the git root for it to be accessible during the build process for the container image. After the image is built I delete this file, and then after the container has finished running the job I delete the image locally so it doesn't persist. Let me know if this is sufficiently secure handling of the login file, because that file effectively gives full control over any snaps that the `edgexfoundry` account has access to.

anonymouse64:
Also, I just spoke with @tonyespy about this, but I think this is simple enough that we can backport this change to california builds so that we have a snap that works on arm64 for california as well.
If we agree that's what we want to do, then after this has been merged to master I will open another PR to california. 
If however we decide not to do this for california I will need to go back to the ci-management PR and undo my changes to the california jobs

JPWKU:
This looks generally good too.  I think we can copy from $HOME, but the container doesn't know what it is, so we will have to pass it in as a docker arg.

anonymouse64:
I don't think we can copy from $HOME when building, at least it didn't work when I tried it locally. The docs have this to say about it:

> Note: docker build will return a no such file or directory error if the file or directory does not exist in the uploaded context. This may happen if there is no context, or if you specify a file that is elsewhere on the Host system. The context is limited to the current directory (and its children) for security reasons, and to ensure repeatable builds on remote Docker hosts. This is also the reason why ADD ../file will not work.


(from the NOTE blurb under this section: https://docs.docker.com/engine/reference/commandline/build/#specify-a-dockerfile--f )

JPWKU:
@anonymouse64 Ah that makes sense
We use config files in other places as well; we just have to be careful about not setting a verbose flag or cat'ing the file.

anonymouse64:
@JPWKU what are your thoughts on adapting the california jobs for this?

snap/build.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# delete the login file we copied to the git root so it doesn't persist around
+rm $GIT_ROOT/edgex-snap-store-login
+
+# now run the build with the environment variables 
+docker run --rm -e "IS_RELEASE_JOB=$IS_RELEASE_JOB" -e "RELEASE=$RELEASE" -e "SNAP_CHANNEL=$SNAP_CHANNEL" edgex-snap-builder:latest
+
+# also delete the image we created so that it doesn't persist on the system as it contains credentials
+docker image rm --force edgex-snap-builder:latest
+docker image prune
```
tonyespy: Missing newline...

anonymouse64: Done

JPWKU:
@anonymouse64 I suppose since it is not an edgex-go feature, then it would be ok to port to cali.  I'd like to get others thoughts as well though.

anonymouse64:
@JPWKU I agree, it is a bit of a chicken/egg problem with the arm snap verify job, as the snap will never successfully build natively on anything other than Ubuntu 16.04, and it is currently running a modified version of 17.10 IIRC. 
One thing we could try is that I could revisit my ci-management PR and change it so that only the x86 builds are run on docker for now, merge that and if it succeeds then go back and open another ci-management PR enabled docker builds on x86. Or we could do the other way around, trying arm first and then doing x86.

anonymouse64:
recheck

anonymouse64:
recheck

JPWKU:
Looking as these verify's complete.  We may force merge one if it fails on ARM like I think it will.

PR #580, opened 2018-09-20
Moving export interfaces back to distro.

### LINKED ISSUE #579: Move internal/export/interfaces back to internal/export/distro
Those interfaces are only used by distro and should not be exported.

feclare:
Renamed some function to make them privates

Fixes #579


tsconn23:
@feclare Looks like there's a file conflict.

feclare:
@brandonforster rebased
But do not let this PR block yours, I can rebase it again if needed.

PR #583, opened 2018-09-21
Migrating core-metadata to Config V2

tsconn23:
#574 

Also updated NotificationsClient to use Endpoint Monitor, and standardized its initialization.


internal/core/metadata/const.go
```diff
@@ -13,42 +13,6 @@
...
-	NotificationLabel                   string
-	SupportNotificationsHost            string
-	SupportNotificationsPort            int
-	SupportNotificationsSubscriptionURL string
-	SupportNotificationsTransmissionURL string
-	StartupTimeout                      int
-}
-
 const (
 	MAX_LIMIT int = 1000
```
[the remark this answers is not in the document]
tsconn23: Change made.

internal/core/metadata/config.go
```diff
@@ -11,15 +11,18 @@
...
-}
-
-var clientConfig notificationConfig
+import (
+	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
+)
 
-func SetConfiguration(host string, port int) {
-	clientConfig = notificationConfig{serviceHost: host, servicePort: port}
+// Struct used to pase the JSON configuration file
```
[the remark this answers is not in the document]
tsconn23: Change made.

internal/core/metadata/init.go
```diff
@@ -108,38 +118,90 @@ func initializeConfiguration(useConsul bool, useProfile string) (*ConfigurationS
...
+		select {
+		case ex := <-errCh:
+			LoggingClient.Error(ex.Error())
+		case raw, ok := <-chConfig:
+			if ok {
+				actual, ok := raw.(*ConfigurationStruct)
+				if !ok {
+					LoggingClient.Error("listenForConfigChanges() type check failed")
+				}
+				Configuration = actual //Mutex needed?
```
feclare: Should we reconnect the db and clients here? Some of the services ip(db, logging...) could be changed.

internal/core/metadata/init.go
```diff
@@ -108,38 +118,90 @@ func initializeConfiguration(useConsul bool, useProfile string) (*ConfigurationS
...
 func connectToDatabase() error {
+	var err error
 	dbConfig := db.Configuration{
-		Host:         Configuration.MongoDBHost,
-		Port:         Configuration.MongoDBPort,
-		Timeout:      Configuration.MongoDBConnectTimeout,
-		DatabaseName: Configuration.MongoDatabaseName,
-		Username:     Configuration.MongoDBUserName,
-		Password:     Configuration.MongoDBPassword,
+		Host:         Configuration.Databases["Primary"].Host,
```
feclare: I should remove the db.Configuration type and use config.Database

internal/core/metadata/init.go
tsconn23: @feclare Well, the service clients already have the Endpointer.Monitor() loop which looks for updated Consul information on a configurable interval. But you're right that we could look at making those endpoints refresh as part of the above flow. And also the database client doesn't have a Monitor to refresh itself if the connection params change. Might be good work for an enhancement PR.

PR #604, opened 2018-09-28
Migrated support-logging to Config V2

tsconn23:
- Created api/v1/config endpoint, fixed log directory
- Removed logging-specific Mongo implementation, added to impl found
      in internal/pkg/db/mongo
- Because of the above, also removed internal/support logEntry model
     and extended pkg/models/logEntry
- Renamed persistence interface to DBClient for consistency, renamed
     methods where appropriate for inclusion into larger DBClient type
- Tested via blackbox tests


jpwhitemn:
Looks like Delete and FindLog are also new functionality?

tsconn23:
> Looks like Delete and FindLog are also new functionality?

No, they used to be named differently -- "remove" and "find" respectively. Since these method signatures had to be part of the larger Mongo DB type, they needed to be qualified to express what they're trying to remove and find -- thus "FindLog" and "DeleteLog".

internal/support/logging/criteria.go
```diff
@@ -80,3 +82,47 @@ func (criteria matchCriteria) match(le models.LogEntry) bool {
...
+func createConditions(conditions []bson.M, field string, elements []string) []bson.M {
+	keyCond := []bson.M{}
+	for _, value := range elements {
+		keyCond = append(keyCond, bson.M{field: value})
+	}
+
+	return append(conditions, bson.M{"$or": keyCond})
+}
+
+func (criteria matchCriteria) CreateQuery() map[string]interface{} {
```
feclare: There shouldn't be mongo code in support.
I don't understand why the type LogMatcher is needed, maybe circular dependencies? If it is the case, moving the matchCriteria to internal/support/logging/criteria could be cleaner to me.

internal/support/logging/server.go
```diff
@@ -149,16 +164,16 @@ func getLogs(w http.ResponseWriter, r *http.Request) {
...
 
-	logs, err := dbClient.find(*criteria)
+	logs, err := dbClient.FindLog(*criteria, criteria.Limit)
 	if err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
 
-	if criteria.Limit > 0 && len(logs) > criteria.Limit {
+	if criteria.Limit > 0 && len(logs) > Configuration.Service.ReadMaxLimit {
```
feclare: You are ignoring the limit from the query and only testing if the log count is lower than the configured ReadMaxLimit. 

You should compare the limit read in sever.go:80 and if it is greater than ReadMaxLimit set it to ReadMaxLimit in the getCriteria function.

internal/support/logging/server_test.go
```diff
@@ -114,83 +126,103 @@ func TestGetLogs(t *testing.T) {
...
 			matchCriteria{}},
 		{"limit",
 			"1000",
 			http.StatusOK,
+			100,
 			matchCriteria{Limit: 1000}},
 		{"limitToLow",
 			"1",
 			http.StatusRequestEntityTooLarge,
+			1,
```
feclare: Some tests should be added about limits. MaxReadLimit being lower than the limit in url and viceversa.

internal/support/logging/file.go
```diff
@@ -25,23 +28,19 @@ type fileLog struct {
 	out      io.WriteCloser
 }
 
-func (fl *fileLog) closeSession() {
+func (fl *fileLog) CloseSession() {
```
jduranf: I think this code should be moved into internal/core/pkg/db/file/logging.go

internal/support/logging/types.go
```diff
@@ -40,31 +52,41 @@ func (l privLogger) log(level string, msg string, labels []string) {
 			Message:       msg,
 			Created:       db.MakeTimestamp(),
 		}
-		dbClient.add(logEntry)
+		dbClient.AddLog(logEntry)
 	}
 }
 
 func (l privLogger) Debug(msg string, labels ...string) error {
+	l.stdOutLogger.SetPrefix(fmt.Sprintf("%s: ", models.DEBUG))
```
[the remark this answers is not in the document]
tsconn23: Will be included in revised PR.

internal/pkg/db/memory/logging.go
```diff
@@ -11,25 +11,27 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  *******************************************************************************/
-package models
+package memory
```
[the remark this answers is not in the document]
tsconn23: OK, this isn't really used anyway so will remove. It looks like the more proper thing to do based on the tests is to implement a file match criteria type.

internal/support/logging/criteria.go
tsconn23: I created the db.LogMatcher interface in order to move the existing "persistence" implementation (which has explicit mongo code) from support-logging to internal/pkg/db/mongo. Here's the existing code, so I don't understand why you say "There shouldn't be mongo code in support" because currently there is:
https://github.com/edgexfoundry/edgex-go/blob/140f76327c59478a9a18c932c500ece70163137a/internal/support/logging/mongo.go#L66

Perhaps you're trying to say we should have the goal of removing the mongo code from support-logging.

The LogMatcher implementation is the matchCriteria type supplied by support-logging which uses the previously imported mongo types in order to construct queries. Further, the matchCriteria type is populated from the support-logging REST controller:
https://github.com/edgexfoundry/edgex-go/blob/140f76327c59478a9a18c932c500ece70163137a/internal/support/logging/server.go#L60

So I was attempting to retain access to the matchCriteria type for the controller, while abstracting away the implementation for how the mongo query is assembled from the db layer.

Another possible approach might be to instead move the CreateQuery implementation into the internal/pkg/db/mongo, but then the question will be where does the MatchCriteria type get defined? It can't be in support-logging because that will cause a circular reference, so it will have to be in /pkg/models. But that's unfortunate because it isn't an externally usable model. Would it make sense to define it right in internal/pkg/db/types? It shouldn't be in the mongo package because the same type can be used for other databases...?

internal/support/logging/server.go
tsconn23: OK, I can do that. But then why do we ever need to prepare the StatusRequestEntityTooLarge response? It never will be too large with your approach.

internal/support/logging/server_test.go
tsconn23: See above. If we take your approach, we don't need these tests.

cmd/support-logging/res/configuration.toml
```diff
@@ -1,16 +1,31 @@
...
-LoggingFile = './logs/support-logging.log'
-MongoDB = 'logging'
-MongoCollection = 'logEntry'
-MongoURL = 'localhost'
-MongoPort = 27017
-MongoConnectTimeout = 5000
-SocketTimeout = 5000
-MongoUsername = ''
-MongoPassword = ''
+Persistence = 'database'
```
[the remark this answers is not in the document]
tsconn23: I think that mixes the usage of the Databases node in the TOML though. This would then be the only service where Databases["Primary"] could be a file instead of a platform. Additionally, the way it is now, someone can toggle between file and Mongo (for example) by changing the value of only one property if desired.

internal/pkg/db/db.go
```diff
@@ -26,9 +26,10 @@ const (
 
 	// Data
 	EventsCollection          = "event"
+	ExportCollection          = "exportConfiguration"
```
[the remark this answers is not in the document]
tsconn23: Oh, I see. I was looking at the alphabetization. Will change.

internal/support/logging/file.go
tsconn23: I'll need to look at this more closely.

tsconn23:
I'm considering redoing this PR without the data layer refactoring. Perhaps I was getting off-topic with this effort.

jduranf:
Maybe this data layer refactoring is off-topic, but should be done, and now you already have done half of the effort :)

internal/support/logging/server.go
feclare: I'm sorry, I didn't realize that you are limiting the size of the results in the findLog. It is ok as is.

internal/support/logging/file.go
```diff
@@ -99,7 +120,11 @@ func (fl *fileLog) remove(criteria matchCriteria) (int, error) {
 	return count, nil
 }
 
-func (fl *fileLog) find(criteria matchCriteria) ([]models.LogEntry, error) {
+func (fl *fileLog) FindLog(criteria db.LogMatcher, limit int) ([]models.LogEntry, error) {
```
feclare: A testPersistenceFind should be added. The limit is not working in the files backend. It is the cause of the other test returning StatusRequestEntityTooLarge.

internal/support/logging/criteria.go
feclare: Perhaps you're trying to say we should have the goal of removing the mongo code from support-logging.
Yes, that is my comment. 
What about moving it to internal/support/logging/criteria/match.go?

Issue #607, opened 2018-10-01
Update build/install instructions in top level README


andresrinivasan:

jpwhitemn:
@mhall119 - do you think this is something you can help clean up??

mhall119:
Absolutely! What specifically needs to be updated?

jpwhitemn:
Michael - perhaps if we organize it similar to what you did for documentation with pointers to the documentation (check out https://github.com/nats-io/gnatsd for a detailed readme):
-what you need to use EdgeX (install docker and docker compose)
-what you need to develop or contribute EdgeX (go, 0MQ lib, probably an editor - we use GoLand)
-how to build the project (make prepare, make build, etc.)
-how to run the project (make run)
-how to run via docker...
etc...

andresrinivasan:
This specific issue is closed.

PR #608, opened 2018-10-01
Clarify build and non-docker startup

andresrinivasan:
Addresses #607


README.md
```diff
@@ -26,6 +34,8 @@ make build
 make run
 ```
 
+**Note** You will need to have the database and Consul running before you execute `make run`. If you don't want to install them, start them via their respective Docker containers.
```
tsconn23: The services can be run without Consul. They only require Consul if they are started with the `--consul` flag which is not the default via `make run`

andresrinivasan: Thanks for reviewing. Corrected.

andresrinivasan:
Closing this to replace with simplified branch.

Issue #609, opened 2018-10-01
Simplify clients code


jduranf:
Currently the EdgeX clients code in pkg/clients are composed by lots of functions copy and pasted with minor modifications.

This can be simplified, which will make the code more easy to maintain and also will reduce the probability of any bug.

tsconn23:
@jduranf After merging the PRs shown above, is there still work to be done here or can we close this?

jduranf:
The only missing things to do are:

1. Remove the [scheduler client](https://github.com/edgexfoundry/edgex-go/tree/master/pkg/clients/scheduler), as it's not used and the API doesn't match with the API implemented in support-scheduler
2. Add pkg/clients/export/client. Currently there is a [TODO](https://github.com/edgexfoundry/edgex-go/blob/master/internal/export/distro/client.go#L19)

The first point is easy to do, but the second point I don't know if it's worth it, because I'm not sure if export will be refactored or not.

What do you think?

tsconn23:
@xmlviking Do you have any input on Joan's first point above? Are there any external applications (like Device Services, etc) that would be using this and break if we remove it?

@jduranf I'd leave the 2nd point alone for now. Fede and I were going back and forth about that but with the developing plan to replace export services, wouldn't worry about it.

jduranf:
Done!

PR #610, opened 2018-10-01
Refactored core-data clients

jduranf:
This is the first part of #609.
If everything it's fine, I will refactor the other micro service clients in the same way.


pkg/clients/coredata/event.go
```diff
@@ -72,426 +68,86 @@ func (e *EventRestClient) init(params types.EndpointParams) {
...
-		if err != nil {
-			fmt.Println(err.Error())
-			return models.Event{}, err
-		}
-
-		return models.Event{}, types.NewErrServiceClient(resp.StatusCode, bodyBytes)
-	}
-
-	return e.decodeEvent(resp)
+	return e.requestEvent(e.url + "/" + id)
```
drasko: I do not like this sting constant hardcoding. Also, I find `fmt.Sprintf()` more readable here.

pkg/clients/coredata/event.go
```diff
@@ -72,426 +68,86 @@ func (e *EventRestClient) init(params types.EndpointParams) {
...
-
-	if resp.StatusCode != http.StatusOK {
-		return 0, types.NewErrServiceClient(resp.StatusCode, bodyBytes)
-	}
-	count, err := strconv.Atoi(bodyString)
-	if err != nil {
-		return 0, err
-	}
-	return count, nil
+	return countRequest(e.url + "/count")
```
drasko: Here also - hardcoded string constant.

pkg/clients/coredata/event.go
```diff
@@ -72,426 +68,86 @@ func (e *EventRestClient) init(params types.EndpointParams) {
...
-	if resp.StatusCode != http.StatusOK {
-		return 0, types.NewErrServiceClient(resp.StatusCode, bodyBytes)
-	}
-
-	count, err := strconv.Atoi(bodyString)
-	if err != nil {
-		return 0, err
-	}
-	return count, nil
+	return countRequest(e.url + "/count/" + url.QueryEscape(deviceId))
```
drasko: And here.

pkg/clients/coredata/event.go
```diff
@@ -72,426 +68,86 @@ func (e *EventRestClient) init(params types.EndpointParams) {
...
-		bodyBytes, err := getBody(resp)
-		if err != nil {
-			fmt.Println(err.Error())
-			return []models.Event{}, err
-		}
-
-		return []models.Event{}, types.NewErrServiceClient(resp.StatusCode, bodyBytes)
-	}
-	return e.decodeEventSlice(resp)
+	return e.requestEventSlice(e.url + "/device/" + url.QueryEscape(deviceId) + "/" + strconv.Itoa(limit))
```
drasko: And here is nice example where `fmt.Sprintf()` would be really beneficial.

pkg/clients/coredata/event.go
```diff
@@ -72,426 +68,86 @@ func (e *EventRestClient) init(params types.EndpointParams) {
...
-		bodyBytes, err := getBody(resp)
-		if err != nil {
-			fmt.Println(err.Error())
-			return []models.Event{}, err
-		}
-
-		return []models.Event{}, types.NewErrServiceClient(resp.StatusCode, bodyBytes)
-	}
-	return e.decodeEventSlice(resp)
+	return e.requestEventSlice(e.url + "/" + strconv.Itoa(start) + "/" + strconv.Itoa(end) + "/" + strconv.Itoa(limit))
```
drasko: Here also.

pkg/clients/coredata/event.go
```diff
@@ -72,426 +68,86 @@ func (e *EventRestClient) init(params types.EndpointParams) {
...
-		bodyBytes, err := getBody(resp)
-		if err != nil {
-			fmt.Println(err.Error())
-			return []models.Event{}, err
-		}
-
-		return []models.Event{}, types.NewErrServiceClient(resp.StatusCode, bodyBytes)
-	}
-	return e.decodeEventSlice(resp)
+	return e.requestEventSlice(e.url + "/device/" + url.QueryEscape(deviceId) + "/valuedescriptor/" + url.QueryEscape(vd) + "/" + strconv.Itoa(limit))
```
drasko: Again, hardcoded string constants :).

jduranf:
@drasko although I agree with you that using hard-coded string constants should be avoided, the intention of this PR is not fix this issue. The hard-coded string constants where present before this PR, so I disagree that this is a good reason to reject this PR.

Anyway, if we want to define some strings like you suggested in this line:
```
return e.requestEventSlice(e.url + "/device/" + url.QueryEscape(deviceId) + "/valuedescriptor/" + url.QueryEscape(vd) + "/" + strconv.Itoa(limit))
```
We should also use the same defined strings in internal/core/data/router.go:
```
e.HandleFunc("/device/{deviceId}/valuedescriptor/{valueDescriptor}/{limit:[0-9]+}", readingByDeviceFilteredValueDescriptor).Methods(http.MethodGet)
```
But as I said, this is out of the scope of this PR.

pkg/clients/coredata/request.go
```diff
@@ -0,0 +1,191 @@
+/*******************************************************************************
+ * Copyright 2017 Dell Inc.
```
tsconn23: @jpwhitemn How should the copyright work here? This is a new file created by @jduranf but AFAIK he's working on the project himself and his employer isn't a member company. Does that mean he should assign copyright to himself? If his employer IS a member company then their name should be here, correct?

jpwhitemn: Yes - Joan, just add your own copyright below the Dell one.  See https://github.com/edgexfoundry/device-sdk-go/blob/master/config.go as an example.

Something like this:

// -*- mode: Go; indent-tabs-mode: t -*-
//
// Copyright (C) 2017 Dell Inc.
// Copyright (C) 2018 Joan Duran
//
// SPDX-License-Identifier: Apache-2.0

jduranf:
@tsconn23 fixed. Also fixed some golint issues (one day we should fix all golint issues...).

PR #614, opened 2018-10-02
Device service is notified when device addressable is updated

### LINKED ISSUE #613: Device addressable update should notify associated device service
Currently if a device addressable is updated the associated device service isn't notified.
The device service only is notified if it's addressable is updated, which in fact makes no sense because the device service addressable is created and maintained by himself.

jduranf:
This should fix #613.
Is implemented in the same way as it's done with profiles.

### DIFF
--- a/internal/core/metadata/rest_addressable.go
+++ b/internal/core/metadata/rest_addressable.go
@@ -360,13 +360,27 @@ func restGetAddressableByAddress(w http.ResponseWriter, r *http.Request) {
 	json.NewEncoder(w).Encode(res)
 }
 
-// Notify the associated device services for the addressable
+// Notify the associated device services for changes in the device addressable
 func notifyAddressableAssociates(a models.Addressable, action string) error {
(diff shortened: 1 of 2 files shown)


internal/core/metadata/rest_addressable.go
```diff
@@ -362,11 +362,25 @@ func restGetAddressableByAddress(w http.ResponseWriter, r *http.Request) {
 
 // Notify the associated device services for the addressable
```
tsconn23: @jduranf I think you'll want to update this comment to reflect your change. I see you've eliminated the notification to the Device Service itself per your comment in the issue. Just want to confirm we don't want to maintain the old behavior in addition to your change.

Once above is addressed and branch is rebased, will merge.

jduranf:
Yes @tsconn23, the device service should be notified for any change in it's device addressable, and there is no point to notify the device service for any change in it's addressable.

Fixed the comments and rebased

PR #615, opened 2018-10-02
Delhi snapcraft improvements

### LINKED ISSUE #461: All of the `snap` folder is staged into the snap
Currently, because we put all the config files, scripts, etc. that the project needs to be built as a snap inside the `snap` folder, snapcraft stages all of this into the snap itself under the `snap` folder. This is undesirable, as we already stage files that we need from there into the snap in other, more usable places and so these files end up being duplicated inside the snap, thus bloating the file size of the snap somewhat.

### LINKED ISSUE #479: Remove shell plugin for snap packaging
Now that snapcraft.yaml supports `override-build` for all plugins, we can stop using the custom shell plugin and use that instead. This will mainly reduce complexity and maintenance burden for the snap in the snapcraft.yaml and also slightly decrease build times.

### LINKED ISSUE #488: Install edgex-services-env file during install hook in snap
Currently the snap only copied the edgex env file controlling which services startup after the start-edgex.sh script is run once, which means that to change it you either have to manually copy it into $SNAP_DATA and modify it, or run start-edgex then stop-edgex and modify the file and re-run start-edgex.

Also note that this may become somewhat of a moot point with #485 as this environment file will be going away if everything is a service, but still in the meantime it might be worth it to fix this.

### LINKED ISSUE #588: Update Java services to 0.7.0 in snap
For Delhi, we should update the Java services to 0.7.0.

### LINKED ISSUE #589: Update snap support-notifications to use go
The snap for master is currently using the Java version of support-notifications, but we should update it to use the new go version.

anonymouse64:
This PR addresses the following issues:
* fixes #589 
* fixes #588 
* fixes #488 
* fixes #479 
* fixes #461.

Note that I tested this on a classic Ubuntu 16.04 and all the services started up correctly.


snap/snapcraft.yaml
```diff
@@ -368,8 +368,8 @@ parts:
       - libzmq3-dev
   support-scheduler:
     source: https://github.com/edgexfoundry/support-scheduler.git
-    # the following commit == 0.5.0 version update +1 (core-domain fix)
-    source-commit: "00707ae"
+    # the following commit == 0.7.0 version update
+    source-commit: "194094a3cf2a0a3a2cc608cec7cd3c4904ac6b5b"
```
tonyespy: Is there a tag that corresponds to this commit, or are we still forced to use a commit SHA? Also any reason you switched to using the full SHA vs. a short SHA?

snap/snapcraft.yaml
```diff
@@ -410,8 +410,8 @@ parts:
       - -usr/share/
   support-rulesengine:
     source: https://github.com/edgexfoundry/support-rulesengine.git
-    # the following commit == 0.5.0 version update
-    source-commit: "2e282f5"
+    # the following commit == 0.7.0 version update
+    source-commit: "3aca4a7bf9774ad07b9b74c7b05c45deb74412f2"
```
tonyespy: Same as above (ie. for support-scheduler).

anonymouse64: Same as above, not 0.7.0 tag

PR #617, opened 2018-10-02
Migrating support-logging to ConfigV2 -- 2nd attempt

tsconn23:
This PR replaces #604 
Issue #595 

I have eliminated the prior database refactoring work because after some of the comments, I felt like I had gone off-topic in what I submitted.

-  Created api/v1/config endpoint
-  Fix to create log file when logs directory doesn't exist
 - Tested via blackbox tests
- Updated with fix for issue #618 


internal/support/logging/server.go
```diff
@@ -74,6 +89,9 @@ func getCriteria(w http.ResponseWriter, r *http.Request) *matchCriteria {
 			io.WriteString(w, s)
 			return nil
 		}
+		if criteria.Limit > Configuration.Service.ReadMaxLimit {
```
feclare: I was wrong suggesting this. Something similar to what you had is better. Queries without explicit limit in the url will not have readmaxlimit neither

Add something like this to line 176:
```
if (criteria.Limit > 0 && len(logs) > criteria.Limit) || len(logs) > Configuration.Service.ReadMaxLimit
```

tsconn23: @feclare Help me understand why it's better to throw an error when more logs exist than have been asked for versus capping the number of logs returned with a success status?

To your second point:
>Queries without explicit limit in the url will not have readmaxlimit neither

I can fix this by adding an else condition at line 95 in `getCriteria` where if limit is 0 it's set to the ReadMaxLimit.

tsconn23: I went ahead and, in essence, took this approach. Although because of how the criteria parsing is tested, I had to make some changes. I also removed the logic for returning/testing status 413 (ResponseTooLarge) as it won't ever happen now.

Issue #581, opened 2018-09-21
Add additional EdgeX Micro Service Management (MSM) capabilities. 


akramtexas:
- Add new endpoint to the notifications micro-service as part of the MSM (Micro-Service Management) API, the System Of Record (SOR) for configuration, and which will work with the SMA service when it (SMA) needs to fetch latest configuration from a given micro-service (in this case, notifications).
- Update MSM service module to reflect the most up-to-date data resulting from processing the incoming request
- Add capability to determine, process, and load contents of the SMA's manifest file.
- Add functionality for performing operations on a micro-service by a delegated call such as to start and stop a micro-service.
- Add capability to get metrics for EdgeX micro-services, starting with notifications.

akramtexas:
- The relevant work is being done on a branch of this fork of the mono repo: [Construction Site](https://github.com/akramtexas/edgex-go/commits/msm_api)

akramtexas:
- Some relevant REST client (Postman) test data for handy reference:
[operation URL](http://localhost:48090/api/v1/operation)

> {
   "action":"stop",
   "services":[
   	  "edgex-core-metadata",
      "edgex-support-notifications",
      "edgex-core-data"
   ],
   "params":[
   	"graceful"
   	]
}

anonymouse64:
Is this for Delhi?

jpwhitemn:
Yes Ian - this is for Delhi.

anonymouse64:
Okay, I've filed an item to add this to the snap for Delhi as #624.

PR #620, opened 2018-10-02
Msm api

akramtexas:
- This PR contains the work done in accordance with: [Issue Number 581](https://github.com/edgexfoundry/edgex-go/issues/581)
- At present, I have retained a fairly extensive number of commits (with accompanying comments) to help reviewer with background notes and context of any given commit. Plan is to **squash** the commits _prior_ to merging into master branch.
- If this helps in providing insight into DEV testing, in addition to the separate unit testing, Postman was used with the following five HTTP requests:

1. Operation start ==> http://localhost:48090/api/v1/operation (POST payload contains "start")
2. Operation restart ==> http://localhost:48090/api/v1/operation (POST payload contains "restart")
3. Operation stop ==> http://localhost:48090/api/v1/operation (POST payload contains "stop")
4. Request config ==> http://localhost:48090/api/v1/config/edgex-support-notifications
5. Request metrics ==> http://localhost:48090/api/v1/metrics/edgex-support-notifications

> For (1, 2, and 3 above), the POST payload is like so, containing respectively a variation of the action **directive** (either "start", or "restart", or "stop").

`{
   "action":"<directive-goes-here>",
   "services":[
   	  "edgex-config-seed",
   	  "edgex-support-logging",
   	  "edgex-core-metadata",
      "edgex-support-notifications",
      "edgex-core-data",
      "edgex-core-command",
      "edgex-export-client",
      "edgex-export-distro"
   ],
   "params":[
   	"graceful"
   	]
}
`
> And for (4 and 5 above), the HTTP request each is the GET verb.
- A number of attempts were made (today) to reorganize the three executors (i.e. one each for: Docker, OS, Docker Compose) into an "executor" folder (under the package "agent") but ran into an import cycle issue:
`
import cycle not allowed
package github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent
        imports github.com/edgexfoundry/edgex-go/internal/system/agent
        imports github.com/edgexfoundry/edgex-go/internal/system/executor
        imports github.com/edgexfoundry/edgex-go/internal/system/agent
`
> Therefore, decided to and went ahead with consolidating those executors into a single executor named, simply, executor.go, situated at the same level (i.e. under the package "agent").

- The logging is admittedly verbose (I was going to yank it out, but decided to leave it in as I think it will help with situating the reviewer in establishing which interactions with the API correspond to which actions getting kicked off. Plus we should have a way to disable and prevent "DEBUG"-level logging from being done in Production deployments.) 
- As an example of said logging, please consider this snippet:

> DEBUG: 2018/10/02 16:11:54 The container for {edgex-core-command} has started! Some (container) details as follows:
{b83f09fe902b        edgexfoundry/docker-core-command-go:0.6.1            "/core-command --con…"   8 minutes ago       Up Less than a second   0.0.0.0:48082->48082/tcp                                                                                         edgex-core-command}
INFO: 2018/10/02 16:11:54 About to {start} the service {edgex-export-client} with parameters {[graceful]}.
DEBUG: 2018/10/02 16:11:54 Pulling latest compose file from Github edgexfoundry repository...
DEBUG: 2018/10/02 16:11:54 Determined this to be the directory we will be writing to: {/Users/akramahmad/go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent}
INFO: 2018/10/02 16:11:54 File {/Users/akramahmad/go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/docker-compose.yml} already exists! Therefore, NOT creating it.
DEBUG: 2018/10/02 16:11:56 For the micro-service {edgex-export-client}, we got this docker-compose output:
 {edgex-files is up-to-date
edgex-core-consul is up-to-date
edgex-mongo is up-to-date
edgex-config-seed is up-to-date
edgex-support-logging is up-to-date
edgex-core-data is up-to-date
Starting edgex-export-client ... done
}
- And this snippet as well:
> DEBUG: 2018/10/02 16:34:52 It is for these micro-service that their metrics data has been requested: map[services:edgex-support-notifications,edgex-core-data]
DEBUG: 2018/10/02 16:34:52 About to fetch the metrics for the service {edgex-support-notifications}!
DEBUG: 2018/10/02 16:34:52 Issuing GET command to: http://localhost:48060/api/v1/metrics
DEBUG: 2018/10/02 16:34:52 Fetched these metrics for the notifications service: {{1019192 1019192 71629048 8287 1464 6823}}:
DEBUG: 2018/10/02 16:34:52 For the {edgex-support-notifications} service, fetched these metrics: {{"Service":"edgex-support-notifications","Metrics":{"Alloc":1019192,"Frees":1464,"LiveObjects":6823,"Mallocs":8287,"Sys":71629048,"TotalAlloc":1019192}}}:
INFO: 2018/10/02 16:34:52 The micro-service {edgex-core-data} currently does not support an endpoint for providing its metrics data...


glide.yaml
```diff
@@ -34,3 +34,11 @@ import:
   version: =1.2.0
 - package: github.com/stretchr/testify
   version: =1.2.1
+- package: github.com/docker/distribution
+  version: 16128bbac47f75050e82f7e91b04df33775e0c23
```
tsconn23: Did you check to see if there are actual release versions corresponding to these SHAs? If not, they may be SHAs that represent the HEAD of a repo -- effectively between supported releases.

internal/support/notifications/config_system.go
```diff
@@ -0,0 +1,24 @@
...
+
+func configHandler(w http.ResponseWriter, r *http.Request) {
+
+	if r.Body != nil {
+		defer r.Body.Close()
+	}
+
+	// The micro-service is to be considered the System Of Record (SOR) in terms of accurate information.
+	// Fetch configuration from cache (the in-memory store which is populated when micro-service is bootstrapped!)
+	conf := &Configuration
```
tsconn23: You don't need this. Just pass Configuration to encode()

internal/support/notifications/metrics_system.go
```diff
@@ -0,0 +1,39 @@
+package notifications
```
tsconn23: Recommend changing the name of the file to metrics.go

internal/support/notifications/router.go
```diff
@@ -68,5 +68,11 @@ func LoadRestRoutes() *mux.Router {
 	// /api/v1/ping
 	b.HandleFunc("/ping", pingHandler).Methods(http.MethodGet)
 
+	// Configuration
+	b.HandleFunc("/config", configHandler).Methods(http.MethodGet)
+
+	// Metrics
+	b.HandleFunc("/metrics", metricsHandler).Methods(http.MethodGet)
+
```
tsconn23: Move these and also the /ping assignment up above line 26. Since you will be adding these routes to the services, you will also need to add them to the respective clients since the SysMgmt will call the services through a client rather than inline. For this PR, modify the /pkg/clients/notifications/client.go so that these routes can be called. You should also define constants for these routes in /pkg/clients/constants.go. The above function calls will then look like:

`b.HandleFunc(clients.ApiConfigRoute, configHandler).Methods(http.MethodGet)`

You will then use the constants on subsequent PRs when defining the routes in other services.

internal/support/notifications/config_system.go
```diff
@@ -0,0 +1,24 @@
+package notifications
```
tsconn23: Don't use a separate file for this. Just put the function at the bottom of router.go. I would tell you to call it config.go but when I come through for ConfigV2, I will create a config.go that contains the ConfigurationStruct for the service to be consistent with other services.

internal/support/notifications/const.go
```diff
@@ -56,6 +56,15 @@ type ConfigurationStruct struct {
 	DBType                        string
 }
 
+type Monitor struct {
```
tsconn23: This type will be used by every service, correct? Put it someplace in the internal package hierarchy where it can be shared.

Also, some name besides Monitor. Maybe something more like "telemetry" (definition: the process of recording and transmitting the readings of an instrument.)

internal/support/notifications/metrics_system.go
```diff
@@ -0,0 +1,39 @@
...
+
+	// Miscellaneous memory stats
+	m.Alloc = rtm.Alloc
+	m.TotalAlloc = rtm.TotalAlloc
+	m.Sys = rtm.Sys
+	m.Mallocs = rtm.Mallocs
+	m.Frees = rtm.Frees
+
+	// Live objects = Mallocs - Frees
+	m.LiveObjects = m.Mallocs - m.Frees
```
tsconn23: This should be calculated internally by the "Monitor" type, not by every service that performs the metrics operation.

cmd/sys-mgmt-agent/res/manifest.toml
```diff
@@ -0,0 +1,17 @@
...
+
+# title = 'EdgeX manifest for consumption by the SMA to bootstrap managed services.'
+
+# edgex-support-notifications
+EdgexServicesNotificationsName = 'notifications'
+EdgexServicesNotificationsHost = 'localhost'
+EdgexServicesNotificationsPort = 48060
+EdgexServicesNotificationsProtocol = 'http'
+EdgexServicesNotificationsCommandstop = 'stop'
+EdgexServicesNotificationsCommandstart = 'start'
```
tsconn23: Utilize the hierarchical capabilities of TOML to organize the content of this file, esepcially as it grows.
```
[Services]
    [Services.Notifications]
    Name = 'notifications'
    Host   = 'localhost'
    Port    = 48060
...
```
As an aside, instead of "Notifications" you could use the value of the service key "edgex-support-notifications" if you think that would be cleaner in the code. This will make the "Name" property redundant BTW.

Does this need to be in a separate manifest.toml file or can it be included in configuration.toml? You will need both a local and docker-ized version of this information anyhow.

internal/system/agent/const.go
```diff
@@ -14,14 +14,25 @@
...
+	DockerLevelOperations bool
+}
+
+type ManifestStruct struct {
+	EdgexServicesNotificationsName         string
+	EdgexServicesNotificationsHost         string
+	EdgexServicesNotificationsPort         int
+	EdgexServicesNotificationsProtocol     string
+	EdgexServicesNotificationsCommandstop  string
+	EdgexServicesNotificationsCommandstart string
```
tsconn23: TOML comments above will affect this. In fact, if you take the recommendations above, You can add a property to the ConfigurationStruct.
`Manifests    map[string] config.ManifestInfo`

which would allow you access to the given manifest by its name or service key (as described above)

internal/system/agent/services.go
```diff
@@ -25,35 +29,349 @@ const (
 	RESTART = "restart"
 )
 
-func invokeAction(action string, services []string, params map[string]string) bool {
+type Wrappee1 ConfigRespMap
+type Wrappee2 MetricsRespMap
```
tsconn23: Wrappee1 and Wrappee2? As a 3rd party looking at this code, I have no idea what the intent of these types are. Need different names. Do they need to be exported? If not, lower case.

internal/system/agent/response_structs.go
```diff
@@ -0,0 +1,36 @@
...
+// For handling the response (containing metrics data) returned by the edgex-support-notifications service.
+type MetricsRespMap struct {
+	Metrics map[string]interface{}
+}
+
+func (b ConfigRespMap) MarshalJSON() ([]byte, error) {
+	return json.Marshal(struct {
+		Service string
+		Wrappee1
+	}{
```
tsconn23: The Wrappee types need different names.

internal/system/agent/router.go
```diff
@@ -19,16 +19,29 @@ import (
...
+
+const (
+	edgexCoreCommand          = "edgex-core-command"
+	edgexCoreData             = "edgex-core-data"
+	edgexCoreMetadata         = "edgex-core-metadata"
+	edgexExportClient         = "edgex-export-client"
+	edgexExportDistro         = "edgex-export-distro"
+	edgexSupportLogging       = "edgex-support-logging"
+	edgexSupportNotifications = "edgex-support-notifications"
+	edgexConfigSeed           = "edgex-config-seed"
```
tsconn23: All of these constants are already available here `internal/constants.go`. Please don't redefine.

internal/system/agent/try.go
```diff
@@ -0,0 +1,30 @@
+package agent
+
+import "errors"
+
+// MaxRetries is the maximum number of retries before bailing.
+var MaxRetries = 10
+
+var errMaxRetriesReached = errors.New("exceeded retry limit")
```
tsconn23: I suggest adding the # of retries to the error message.

internal/system/agent/try.go
```diff
@@ -0,0 +1,30 @@
...
+		cont, err = fn(attempt)
+		if !cont || err == nil {
+			break
+		}
+		attempt++
+		if attempt > MaxRetries {
+			return errMaxRetriesReached
+		}
+	}
+	return err
```
tsconn23: Please note that the above is a tight loop.

internal/system/agent/router.go
```diff
@@ -37,43 +50,53 @@ func LoadRestRoutes() *mux.Router {
...
+	switch action {
 
-	case START:
-		go invokeAction(START, services, params)
-		w.WriteHeader(http.StatusCreated)
-		w.Write([]byte("Response"))
+	// Make asynchronous goroutine call(s) to the appropriate internal function (respectively, to stop, start, or restart the service(s).
+	case STOP:
+		go InvokeOperation(STOP, services, params)
+		w.WriteHeader(http.StatusOK)
```
tsconn23: Without some kind of result from the `go InvokeOperation` how do you know that the operation was successful and we should return a StatusOK? What if the user hands the service an invalid service key?

internal/system/agent/services.go
```diff
@@ -25,35 +29,349 @@ const (
...
+		json.Unmarshal([]byte(responseJSON), &crp.Config)
+	}
+	return crp
+}
+
+func getConfig(services []string) (ConfigRespMap, error) {
+
+	const (
+		configApiPath = "/api/v1/config"
+	)
```
tsconn23: This constant already exists. Please don't redefine. `internal/constants.go`

internal/system/agent/services.go
```diff
@@ -25,35 +29,349 @@ const (
...
+
+	var result = ConfigRespMap{}
 
 	// Loop through requested actions, along with respectively-supplied parameters.
 	for _, service := range services {
-		// TODO: Logging is placeholder for actual functionality to be implemented.
-		LoggingClient.Info(fmt.Sprintf("CONFIG >> Fetching the configuration for the service {%v}!\n", service))
+
+		switch service {
```
tsconn23: These are all done in a sequential loop. You can use goroutines/channels to make this parallel and more efficient.

internal/system/agent/services.go
```diff
@@ -25,35 +29,349 @@ const (
...
+
+			var url string
+
+			if Manifest != nil {
+				url = "http://" + Manifest.EdgexServicesNotificationsHost + ":" + strconv.Itoa(Manifest.EdgexServicesNotificationsPort) + configApiPath
+			}
+			LoggingClient.Debug(fmt.Sprintf("About to fetch the configuration for the service {%v}!", service))
+
+			LoggingClient.Debug("Issuing GET command to: " + url)
+			req, err := http.NewRequest(http.MethodGet, url, nil)
```
tsconn23: This call should be performed through the notifications-client and not via inline http.NewRequest(), as we discussed.

internal/system/agent/services.go
```diff
@@ -25,35 +29,349 @@ const (
...
-	return true
+	return result, nil
 }
 
-func getMetric(services []string, metrics []string) bool {
+func getMetrics(services []string) (MetricsRespMap, error) {
+
+	const (
+		metricsApiPath = "/api/v1/metrics"
+	)
```
tsconn23: Put this elsewhere -- either in internal/constants.go or in /pkg/clients/constants.go

internal/system/agent/services.go
```diff
@@ -25,35 +29,349 @@ const (
...
+
+			var url string
+
+			if Manifest != nil {
+				url = "http://" + Manifest.EdgexServicesNotificationsHost + ":" + strconv.Itoa(Manifest.EdgexServicesNotificationsPort) + metricsApiPath
+			}
+			LoggingClient.Debug(fmt.Sprintf("About to fetch the metrics for the service {%v}!", service))
+
+			LoggingClient.Debug("Issuing GET command to: " + url)
+			req, err := http.NewRequest(http.MethodGet, url, nil)
```
tsconn23: This call should be performed through the notifications-client and not via inline http.NewRequest(), as we discussed.

internal/system/agent/utils.go
```diff
@@ -29,3 +34,65 @@ func pingHandler(w http.ResponseWriter, r *http.Request) {
 		LoggingClient.Error("Error writing pong: " + err.Error())
 	}
 }
+
+const (
+	manifestDirectory = "./res"
+	manifestDefault   = "manifest.toml"
```
tsconn23: As I said above, consider eliminating manifest.toml and folding that information into configuration.toml. If you do that, you don't need this go file with all of its duplicated logic around finding paths, etc.

pkg/clients/management/management.go
```diff
@@ -0,0 +1,143 @@
+package management
```
tsconn23: I specifically clarified a week or so ago that a client for the SMA would not be needed. There is nothing that calls the SMA. Instead, the SMA will be calling clients of the services it manages. This whole file should be removed. As of right now, it's only being used by tests anyway.

pkg/clients/management/management_test.go
```diff
@@ -0,0 +1,49 @@
+package management
```
tsconn23: Remove this file

pkg/models/sma_operation.go
```diff
@@ -0,0 +1,61 @@
...
+ *
+ *
+ * Operation struct
+ */
+type Operation struct {
+	BaseObject        `bson:",inline"`
+	Action   string   `bson:"action" json:"action,omitempty"`
+	Params   []string `bson:"params" json:"params,omitempty"`
+	Services []string `bson:"services,omitempty" json:"services,omitempty"`
+}
```
tsconn23: This file can be removed as well.

internal/system/agent/services.go
```diff
@@ -25,35 +29,349 @@ const (
...
+		switch action {
+
+		case STOP:
+			switch service {
+			case edgexSupportNotifications:
+				if Conf.OsLevelOperations {
+					OsExecuteStop(edgexSupportNotifications, params[0])
+				} else if Conf.DockerLevelOperations {
+					StopDockerContainer(edgexSupportNotifications)
+				}
```
tsconn23: This layer of the code should be using an abstraction rather than if/else with explicit knowledge of OS vs Docker vs something else. That abstraction should be accessed via a global var that is initialized in init.go::Retry()

akramtexas:
Thanks @tsconn23 for the comments as well (i.e. the second round of comments specific to the SMA work), which I'll address immediately after the earlier comments that pertain to the support-notifications portion of this PR.

pkg/clients/management/management.go
akramtexas: Agreed. I evidently misinterpreted the pointer on the creation of "mocks" in the _client_ folder to mean that we did need the functionality after all. Again, agreed, and I will remove the file, and associated test file as well, thanks!

pkg/clients/management/management_test.go
akramtexas: Done.

internal/system/agent/services.go
akramtexas: Done.

internal/system/agent/services.go
akramtexas: Done.

internal/system/agent/services.go
akramtexas: Done.

akramtexas:
- Similar to how we have now delegated (to the respective service) the request (to the SMA) for the respective service's **configuration info**, we will also implement delegation for a given service's _metrics info_.
- Both points above apply, at this time, to the notifications service.

internal/support/notifications/router.go
akramtexas: Done for /config. Will do so shortly for /metrics.

internal/system/agent/services.go
akramtexas: Done for /config. Will do so shortly for /metrics.

akramtexas:
- Created [Issue #648](https://github.com/edgexfoundry/edgex-go/issues/648) to address subsequent work as follows: 

> Having EdgeX Micro Service Management (MSM) API be spread to other services, so as to add the capability of delegating to the respective service the request (to the SMA) for a given service's configuration info and for its metrics info (Currently, this exists only for the support-notifications service.)

akramtexas:
I plan on following up on four Docker-related dependencies in glide.yaml:

- package: github.com/docker/distribution
  version: 16128bbac47f75050e82f7e91b04df33775e0c23
- package: github.com/docker/docker
  version: b1116479b23434de7d70710b6454af7cb2484a58
- package: github.com/docker/go-connections
  version: 97c2040d34dfae1d1b1275fa3a78dbdd2f41cf7e
- package: github.com/docker/go-units
  version: 47565b4f722fb6ceae66b95f853feed578a4a51c

akramtexas:
- Making a note to address any leftover instances of the following:
> Having a package with an identically named file underneath it is referred to in Go as "stutter".

akramtexas:
Recheck

JPWKU:
recheck

PR #629, opened 2018-10-05
Added the option to set the minimum log level severity

### LINKED ISSUE #628: Add log level filter in logging system
Currently all logs are stored/transmitted/printed regardless of their log level. Should be fine to add the option to set the log level in the log client.

If it's fine, I will work on it.

jduranf:
Fix #628 

Added the option to set the minimum log level severity. Now by default is INFO, but in another PR we can add the LogLevel configuration parameter, to set this value. 

### DIFF
--- a/pkg/clients/logging/logger.go
+++ b/pkg/clients/logging/logger.go
@@ -28,15 +28,24 @@ import (
 	"github.com/edgexfoundry/edgex-go/pkg/models"
 )
 
+// These constants identify the log levels in order of increasing severity.
 const (
-	TRACE = "TRACE"
-	DEBUG = "DEBUG"
-	WARN  = "WARN"
-	INFO  = "INFO"
-	ERROR = "ERROR"
+	TraceLog = "TRACE"
+	DebugLog = "DEBUG"
+	InfoLog  = "INFO"
+	WarnLog  = "WARN"
+	ErrorLog = "ERROR"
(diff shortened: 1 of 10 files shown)


tsconn23:
recheck

pkg/clients/logging/logger.go
```diff
@@ -48,6 +57,7 @@ type EdgeXLogger struct {
 	owningServiceName string
 	remoteEnabled     bool
 	logTarget         string
+	logLevel          *string
```
tsconn23: @jduranf Why a pointer?

pkg/clients/logging/logger.go
```diff
@@ -113,39 +148,36 @@ func verifyLogDirectory(path string) {
 	}
 }
 
+// SetLogLevel sets minimum severity log level
+func (lc EdgeXLogger) SetLogLevel(logLevel string) {
+	if IsValidLogLevel(logLevel) == true {
+		*lc.logLevel = logLevel
```
tsconn23: @jduranf Is it because the value set here will be lost when the method exits if logLevel isn't a pointer? I assume you tried it?

jduranf: Yes, if it's not a pointer the value is lost. I tried it and I spent lot of time before to realize I should use a pointer.

PR #636, opened 2018-10-05
Support Scheduler Go initial service creation

xmlviking:
#57 Support Scheduler in Go

Initial Go version port of the java Scheduler Service.
Features:  Scheduler, SchedulerContext, and Router
Includes:  V2 Configuration changes as per the current EdgeX design pattern
                  Bootstrap, Retry design pattern

TODO: 
- Further enhancement of the Support Scheduler to include additional tests
- Examine the use of the core-metadata as an option for our Event.Addressable needs

### DIFF
--- a/cmd/support-scheduler/main.go
+++ b/cmd/support-scheduler/main.go
@@ -1,0 +1,91 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+
+	"github.com/edgexfoundry/edgex-go/internal/pkg/startup"
+	"net/http"
+	"os"
+	"os/signal"
+	"strconv"
+	"syscall"
+	"time"
+
+	"github.com/edgexfoundry/edgex-go"
+	"github.com/edgexfoundry/edgex-go/internal"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/usage"
+	"github.com/edgexfoundry/edgex-go/internal/support/scheduler"
+	"github.com/edgexfoundry/edgex-go/pkg/clients/logging"
+	"github.com/gorilla/context"
+)
+
+func main() {
+
+	start := time.Now()
+	var useConsul bool
+	var useProfile string
+
+	flag.BoolVar(&useConsul, "consul", false, "Indicates the service should use consul.")
+	flag.BoolVar(&useConsul, "c", false, "Indicates the service should use consul.")
+	flag.StringVar(&useProfile, "profile", "", "Specify a profile other than default.")
+	flag.StringVar(&useProfile, "p", "", "Specify a profile other than default.")
+	flag.Usage = usage.HelpCallback
+	flag.Parse()
+
+	params := startup.BootParams{UseConsul: useConsul, UseProfile: useProfile, BootTimeout: internal.BootTimeoutDefault}
+	startup.Bootstrap(params, scheduler.Retry, logBeforeInit)
+
+	ok := scheduler.Init(useConsul)
+	if !ok {
+		logBeforeInit(fmt.Errorf("%s: Service bootstrap failed!", internal.SupportSchedulerServiceKey))
+		return
+	}
+
+	scheduler.LoggingClient.Info("Service dependencies resolved...")
+	scheduler.LoggingClient.Info(fmt.Sprintf("Starting %s %s ", internal.SupportSchedulerServiceKey, edgex.Version))
+
+	// Bootstrap schedulers
+	err := scheduler.AddSchedulers()
+	if err != nil{
+		scheduler.LoggingClient.Error(fmt.Sprintf("Failed to load default schedules and events %s",err.Error()))
+	}
+
+	http.TimeoutHandler(nil, time.Millisecond*time.Duration(scheduler.Configuration.Service.Timeout), "Request timed out")
+	scheduler.LoggingClient.Info(scheduler.Configuration.Service.StartupMsg, "")
+
+	errs := make(chan error, 2)
+	listenForInterrupt(errs)
+	startHttpServer(errs, scheduler.Configuration.Service.Port)
+
+	// Start the ticker
+	scheduler.StartTicker()
+
+	// Time it took to start service
+	scheduler.LoggingClient.Info("Service started in: "+time.Since(start).String(), "")
+	scheduler.LoggingClient.Info("Listening on port: "+strconv.Itoa(scheduler.Configuration.Service.Port), "")
+	c := <-errs
+	scheduler.Destruct()
+	scheduler.LoggingClient.Warn(fmt.Sprintf("terminating: %v", c))
+}
+
+
+func logBeforeInit(err error) {
+	scheduler.LoggingClient = logger.NewClient(internal.CoreCommandServiceKey, false, "")
+	scheduler.LoggingClient.Error(err.Error())
+}
+
+func listenForInterrupt(errChan chan error) {
+	go func() {
+		c := make(chan os.Signal)
+		signal.Notify(c, syscall.SIGINT)
+		errChan <- fmt.Errorf("%s", <-c)
+	}()
+}
+
+func startHttpServer(errChan chan error, port int) {
+	go func() {
+		r := scheduler.LoadRestRoutes()
+		errChan <- http.ListenAndServe(":"+strconv.Itoa(port), context.ClearHandler(r))
+	}()
+}
--- a/internal/support/scheduler/init.go
+++ b/internal/support/scheduler/init.go
@@ -1,0 +1,183 @@
+//
+// Copyright (c) 2018 Tencent
+//
+// Copyright (c) 2018 Dell Inc
+//
+// SPDX-License-Identifier: Apache-2.0
+package scheduler
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/edgexfoundry/edgex-go/internal"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/consul"
+	"github.com/edgexfoundry/edgex-go/pkg/clients"
+	"github.com/edgexfoundry/edgex-go/pkg/clients/logging"
+	"github.com/pkg/errors"
+)
+
+var Configuration *ConfigurationStruct
+var LoggingClient logger.LoggingClient
+
+var chConfig chan interface{} //A channel for use by ConsulDecoder in detecting configuration mods.
+var ticker = time.NewTicker(time.Duration(ScheduleInterval) * time.Millisecond)
+
+
+func Retry(useConsul bool, useProfile string, timeout int, wait *sync.WaitGroup, ch chan error) {
+	now := time.Now()
+	until := now.Add(time.Millisecond * time.Duration(timeout))
+	for time.Now().Before(until) {
+		var err error
+		//When looping, only handle configuration if it hasn't already been set.
+		if Configuration == nil {
+			Configuration, err = initializeConfiguration(useConsul, useProfile)
+			if err != nil {
+				ch <- err
+				if !useConsul {
+					//Error occurred when attempting to read from local filesystem. Fail fast.
+					close(ch)
+					wait.Done()
+					return
+				}
+			} else {
+				//Check against boot timeout default
+				if Configuration.Service.BootTimeout != timeout {
+					until = now.Add(time.Millisecond * time.Duration(Configuration.Service.BootTimeout))
+				}
+				// Setup Logging
+				logTarget := setLoggingTarget()
+				LoggingClient = logger.NewClient(internal.SupportSchedulerServiceKey, Configuration.Logging.EnableRemote, logTarget)
+			}
+		}
+
+		if Configuration != nil {
+			break
+		}
+		time.Sleep(time.Second * time.Duration(1))
+	}
+	close(ch)
+	wait.Done()
+
+	return
+}
+
+func Init(useConsul bool) bool {
+	if Configuration == nil {
+		return false
+	}
+
+	if useConsul {
+		chConfig = make(chan interface{})
+		go listenForConfigChanges()
+	}
+	return true
+}
+
+func Destruct() {
+	if chConfig != nil {
+		close(chConfig)
+	}
+
+	if ticker != nil {
+		StopTicker()
+	}
+}
+
+func initializeConfiguration(useConsul bool, useProfile string) (*ConfigurationStruct, error) {
+	//We currently have to load configuration from filesystem first in order to obtain ConsulHost/Port
+	conf := &ConfigurationStruct{}
+	err := config.LoadFromFileV2(useProfile, conf)
+	if err != nil {
+		return nil, err
+	}
+
+	if useConsul {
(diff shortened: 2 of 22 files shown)


xmlviking:
Updated the docker file.  Missed the new file name and the inclusion of the
docker directory now in the projects folder.
I rebuilt it locally and it worked pushed the change up to the PR.

Thanks!

cmd/support-scheduler/main.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"github.com/edgexfoundry/edgex-go"
+	"github.com/edgexfoundry/edgex-go/internal"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/usage"
+	"github.com/edgexfoundry/edgex-go/internal/support/scheduler"
+	"github.com/edgexfoundry/edgex-go/pkg/clients/logging"
+	"github.com/gorilla/context"
+)
+
+var loggingClient logger.LoggingClient
```
tsconn23: Should not be here. Use scheduler.LoggingClient instead.

internal/support/scheduler/router.go
```diff
@@ -0,0 +1,231 @@
...
+	"io/ioutil"
+	"net/http"
+)
+
+func LoadRestRoutes() http.Handler {
+	r := mux.NewRouter()
+	b := r.PathPrefix("/api/v1").Subrouter()
+
+	//ping
+	b.HandleFunc("/ping", ping).Methods(http.MethodGet)
```
tsconn23: Move this above line 22 and use the internal.ApiPingRoute constant. Also, please provide the config endpoint which the SMA will need. Example:
https://github.com/edgexfoundry/edgex-go/blob/master/internal/support/logging/server.go#L211

internal/support/scheduler/router.go
```diff
@@ -0,0 +1,231 @@
...
+	b.HandleFunc("/callbacks", removeCallbackAlert).Methods(http.MethodDelete)
+
+	return r
+}
+
+const (
+	ContentTypeKey       = "Content-Type"
+	ContentTypeJsonValue = "application/json; charset=utf-8"
+	ContentLengthKey     = "Content-Length"
+)
```
tsconn23: These same strings are used pretty widely as literals. Consider moving these constants to internal/constants.go

internal/support/scheduler/router.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		LoggingClient.Error("read info request error" + err.Error())
+		return
+	}
+
+	str := `{"name"}: {"` + scheduleEvent.Name + `"}"`
+	str1 := `{"frequency"}:{"` + scheduleEvent.Frequency + `"}"`
+
+	io.WriteString(rw, str+str1)
+	// iterate over the current schedulers
```
tsconn23: What is this comment for? Is it a TODO?

cmd/support-scheduler/main.go
```diff
@@ -0,0 +1,96 @@
...
+	if !ok {
+		logBeforeInit(fmt.Errorf("%s: Service bootstrap failed!", internal.SupportSchedulerServiceKey))
+		return
+	}
+
+	scheduler.LoggingClient.Info("Service dependencies resolved...")
+	scheduler.LoggingClient.Info(fmt.Sprintf("Starting %s %s ", internal.SupportSchedulerServiceKey, edgex.Version))
+
+	// Initialize the  Service Routes
+	scheduler.LoadRestRoutes()
```
tsconn23: These routes are being overwritten by the call to startHttpServer on line 67 below. See related comment below.

cmd/support-scheduler/main.go
```diff
@@ -0,0 +1,96 @@
...
+		c := make(chan os.Signal)
+		signal.Notify(c, syscall.SIGINT)
+		errChan <- fmt.Errorf("%s", <-c)
+	}()
+}
+
+func startHttpServer(errChan chan error, port int) {
+	go func() {
+		r := command.LoadRestRoutes()
+		errChan <- http.ListenAndServe(":"+strconv.Itoa(port), context.ClearHandler(r))
```
tsconn23: Please note that the routes being served are coming from the command package, not the scheduler package.

internal/support/scheduler/router.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func info(rw http.ResponseWriter, req *http.Request) {
+	rw.Header().Set(ContentTypeKey, ContentTypeJsonValue)
+	rw.WriteHeader(http.StatusOK)
+
+	vars := mux.Vars(req)
+
+	scheduleEvent, err := schedulerClient.QueryScheduleWithName(vars["name"])
+	if err != nil {
```
tsconn23: After fixing the bug which I called out on the main.go file referring to the wrong routes being loaded, I then used Postman to test this method. I issued the following GET `http://localhost:48085/api/v1/info/test` and received a panic. The panic has to do with the schedulerClient not being initialized anywhere in the service and so every function that uses the schedulerClient will panic. Further, what is the intent here? The scheduler service should not be using its client to call itself. Should it be using `pkg/clients/metadata/schedule.go` instead?

```
INFO: 2018/10/06 18:33:39 This is the Support Scheduler Microservice
INFO: 2018/10/06 18:33:39 Service started in: 15.566394ms
INFO: 2018/10/06 18:33:39 Listening on port: 48085
2018/10/06 18:33:45 http: panic serving [::1]:56923: runtime error: invalid memory address or nil pointer dereference
goroutine 6 [running]:
net/http.(*conn).serve.func1(0xc0001cc0a0)
	/usr/local/go/src/net/http/server.go:1746 +0xd0
panic(0x1339080, 0x1636c70)
	/usr/local/go/src/runtime/panic.go:513 +0x1b9
github.com/edgexfoundry/edgex-go/internal/support/scheduler.info(0x1402c00, 0xc0001d6000, 0xc000168900)
	/Users/tconn/go/src/github.com/edgexfoundry/edgex-go/internal/support/scheduler/router.go:57 +0xe2
net/http.HandlerFunc.ServeHTTP(0x13b8768, 0x1402c00, 0xc0001d6000, 0xc000168900)
	/usr/local/go/src/net/http/server.go:1964 +0x44
github.com/edgexfoundry/edgex-go/vendor/github.com/gorilla/mux.(*Router).ServeHTTP(0xc0000ba1e0, 0x1402c00, 0xc0001d6000, 0xc000168900)
	/Users/tconn/go/src/github.com/edgexfoundry/edgex-go/vendor/github.com/gorilla/mux/mux.go:114 +0xe0
github.com/edgexfoundry/edgex-go/vendor/github.com/gorilla/context.ClearHandler.func1(0x1402c00, 0xc0001d6000, 0xc0001d0100)
	/Users/tconn/go/src/github.com/edgexfoundry/edgex-go/vendor/github.com/gorilla/context/context.go:141 +0x8b
net/http.HandlerFunc.ServeHTTP(0xc0000ae920, 0x1402c00, 0xc0001d6000, 0xc0001d0100)
	/usr/local/go/src/net/http/server.go:1964 +0x44
net/http.serverHandler.ServeHTTP(0xc000099450, 0x1402c00, 0xc0001d6000, 0xc0001d0100)
	/usr/local/go/src/net/http/server.go:2741 +0xab
net/http.(*conn).serve(0xc0001cc0a0, 0x1403000, 0xc000020240)
	/usr/local/go/src/net/http/server.go:1847 +0x646
created by net/http.(*Server).Serve
	/usr/local/go/src/net/http/server.go:2851 +0x2f5
```

internal/support/scheduler/schedulecontext.go
```diff
@@ -0,0 +1,138 @@
+//
+// Copyright (c) 2018 Tencent
+//
+// Copyright (c) 2017 Dell Inc
```
tsconn23: Year in copyright is wrong

internal/support/scheduler/schedule.go
```diff
@@ -0,0 +1,482 @@
...
+	go func() {
+		for range ticker.C {
+			triggerSchedule()
+		}
+	}()
+}
+
+func StopTicker() {
+	ticker.Stop()
+}
```
tsconn23: This function is unused. Should it be called from init.go::Destruct()?

internal/support/scheduler/schedule.go
```diff
@@ -0,0 +1,482 @@
...
+	LoggingClient.Debug("removed the schedule with id : " + scheduleId)
+
+	return nil
+}
+
+//endregion
+
+//region schedule event CRUD
+func queryScheduleEvent(scheduleEventId string) (models.ScheduleEvent, error) {
+	mutex.Lock()
```
tsconn23: GoLand currently identifies this as an unused function. If so, can be removed?

internal/support/scheduler/init.go
```diff
@@ -0,0 +1,265 @@
...
+	go dec.Run()
+
+	select {
+	case <-time.After(2 * time.Second):
+		err = errors.New("timeout loading config from registry")
+	case ex := <-errCh:
+		err = errors.New(ex.Error())
+	case raw := <-updateCh:
+		actual, ok := raw.(*ConfigurationStruct)
+		if !ok {
```
tsconn23: In order for the scheduler to pull its configuration from Consul, you need to make config-seed aware of the service. I don't see that logic in this PR. Do you intend to include it? If so, you will need to add the respective service key constant to the list found here:
https://github.com/edgexfoundry/edgex-go/blob/master/internal/seed/config/populate.go#L188

internal/support/scheduler/init.go
```diff
@@ -0,0 +1,265 @@
...
+				Configuration = actual //Mutex needed?
+			} else {
+				return
+			}
+		}
+	}
+}
+
+func initializeClients(useConsul bool) {
+	// Create metadata clients
```
tsconn23: This is currently an unused function. Whether it remains that way hinges on the question regarding which service client this application should be using.

internal/support/scheduler/init.go
```diff
@@ -0,0 +1,265 @@
...
+
+func setLoggingTarget() string {
+	if Configuration.Logging.EnableRemote {
+		return Configuration.Clients["Logging"].Url() + clients.ApiLoggingRoute
+	}
+	return Configuration.Logging.File
+}
+
+func AddDefaultSchedulers() error {
+
```
tsconn23: Consider moving this function to schedule.go. Add unit tests.

internal/support/scheduler/init.go
```diff
@@ -0,0 +1,265 @@
...
+	LoggingClient.Info(fmt.Sprintf("loading default schedules and schedule events..."))
+
+	defaultSchedules := Configuration.Schedules
+
```
tsconn23: Suggest calling `defaultSchedules` just `schedules`

internal/support/scheduler/init.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+		err := addSchedule(defaultSchedule)
+		if err != nil {
+			return LoggingClient.Error("AddDefaultSchedulers() - failed to load schedule %s", err.Error())
+		} else {
+			LoggingClient.Info(fmt.Sprintf("added default schedule %s", defaultSchedule.Name))
+		}
+	}
+
+	defaultScheduleEvents := Configuration.ScheduleEvents
```
tsconn23: Suggesting calling `defaultScheduleEvents` just `events`

internal/support/scheduler/router.go
```diff
@@ -0,0 +1,231 @@
+//
+// Copyright (c) 2018 Tencent
+//
+// Copyright (c) 2017 Dell Inc.
```
tsconn23: Year is wrong

cmd/support-scheduler/main.go
xmlviking: included

internal/support/scheduler/router.go
xmlviking: Updated router to use the mux style incorporated in your logging/server.go example

internal/support/scheduler/router.go
xmlviking: moved local const to scheduler const.go

internal/support/scheduler/router.go
xmlviking: removed comment

cmd/support-scheduler/main.go
xmlviking: updated the line of code to use the scheduler.LoadRestRoutes()
removed the aforementioned duplication of LoadRestRoutes() as well.

internal/support/scheduler/init.go
xmlviking: I added the service key entry as well as changed the array bounds.

internal/support/scheduler/router.go
xmlviking: Fixed this by removing the client code and adding references to the scheduler functions.
Updated the mux to use the bone style parameters.  
Updated the routes as well to use the style similar to the logging service.

internal/support/scheduler/schedule.go
xmlviking: were now using this.

internal/support/scheduler/schedule.go
xmlviking: reserved for unit tests which will be added. If there is a better paradigm for adding helper functions for unit tests I would love to do it.

internal/support/scheduler/schedulecontext.go
xmlviking: udpated.

cmd/support-scheduler/main.go
xmlviking: I removed the extra line entry for the scheduler.LoadRestRoutes() and it's now being covered in the startHttpServer

internal/support/scheduler/init.go
xmlviking: removed this function as were not currently using this function.

internal/support/scheduler/init.go
xmlviking: Moved this code into the Scheduler.  Were no longer using a bridge from main to inti too scheduler.

internal/support/scheduler/init.go
xmlviking: removed the references to the default in the schedules and scheduleEvent code.
commit to follow shortly.

internal/support/scheduler/init.go
xmlviking: removed the references to the prefix 'default' from scheduleEvents.  events might be too generic in this case. scheduleEvents is contextual to the use of events.

internal/support/scheduler/router.go
xmlviking: updated the router header information to use the 2018 dell inc.

jduranf:
@xmlviking although you added metadata client in configuration, it's not used. support-scheduler should read all associate schedules from metadata at start up, and add into metadata the schedules defined in the configuration which aren't still in metadata. Also, core-metadata should be modified in order to call support-scheduler `callback` when needed.

yanghua:
Congratulations support-scheduler-go was finally merged. Maybe next time, we keep the initial contributor's git information to make it look better. In addition, edgexfoundry-holding/support-scheduler-go can also be deleted or archived.

PR #639, opened 2018-10-08
Update event.go code coverage

brandonforster:
Issue #457


tsconn23:
@brandonforster When you have time, please squash the commits. Thanks.

internal/core/data/event.go
```diff
@@ -236,273 +208,113 @@ func updateEventPushDate(id string) error {
...
-		LoggingClient.Error("Error converting to integer: " + err.Error())
-		return
+		LoggingClient.Error("Unable to send message for event: " + e.String())
 	}
+}
 
-	// Check device
-	if checkDevice(deviceId, w) == false {
-		return
+func testLimit (limit int) error {
```
tsconn23: Since "test" has a specific connotation in Go, I think I'd recommend calling this "checkMaxLimit"

internal/core/data/event.go
```diff
@@ -236,273 +208,113 @@ func updateEventPushDate(id string) error {
...
-// 413 - number exceeds limit
-func readingByDeviceFilteredValueDescriptor(w http.ResponseWriter, r *http.Request) {
-	defer r.Body.Close()
+	return count, nil
+}
 
-	vars := mux.Vars(r)
-	limit := vars["limit"]
+func doScrub()(int, error) {
+	LoggingClient.Info("Scrubbing events.  Deleting all events that have been pushed")
```
tsconn23: Rename to "scrubPushedEvents" to be more explicit about the function's intent.

internal/core/data/event_test.go
```diff
@@ -347,6 +636,60 @@ func newMockDeviceClient() *mocks.DeviceClient {
 	return client
 }
 
+func newMockDb() interfaces.DBClient {
+	db := &dbMock.DBClient{}
```
tsconn23: Question -- Did you notice that in the reset() function in this file, there's a different usage of the dbClient?

`dbClient = &memory.MemDB{}`

I would encourage you to look at that provider and try to use it for your tests. Every test calls reset() to reinitialize state within the MemDB.

brandonforster: Yes. I wanted to be able to get coverage on the error handling parts of the DB interactions, and on some of these memory DB function calls there's so little actually working that it's impossible to generate an error. I probably have too much reliance on mock calls versus the stub memory DB, but it's so much easier to just have the mock throw an error and see how event deals with it than try to organically generate the error that I thought this was worth it.

tsconn23: I'm going to leave this out here overnight and see if either @feclare or @jduranf has any comment on this topic.

tsconn23:
I talked with Brandon today about his use of a mock for the dbClient in the new tests he added. I wanted to clarify my understanding of his intent in using the mock rather than the MemoryDB and I have to say, I agree with his points. His goal is to bring the coverage stats up as much as possible. In order to do that, he's trying to create tests which explicitly exercise the error handling in the business logic. In some execution paths, the only place where an error will be thrown in a production scenario is if the Mongo provider throws an error when attempting a query, such as the following line:
```
err := s.DB(mc.database.Name).C(db.EventsCollection).Find(q).All(&me)
```
The MemoryDB has no way to facilitate this extra coverage because in most paths, it will never throw an error.
```
func (m *MemDB) EventsOlderThanAge(age int64) ([]models.Event, error) {
	currentTime := db.MakeTimestamp()
	events := []models.Event{}
	for _, e := range m.events {
		if currentTime-e.Created >= age {
			events = append(events, e)
		}
	}
	return events, nil
}
```
By using his mock, Brandon is able to take code coverage from this:
screen shot 2018-10-10 at 1 37 19 pm
To this:
screen shot 2018-10-10 at 1 37 53 pm

The difference being in that the second picture, the first error check is now covered by testing.

To achieve even more coverage, a suitable mock for the specific test could be provided that would return the necessary events to allow the bottom loop to execute, whereby an error would be thrown by 
deleteEvent() thus turning that area green as well.

I welcome comments and discussion here but if nothing is forthcoming, I'm going to endorse this approach.

Issue #640, opened 2018-10-08
support-logging can die when started by itself via docker-compose


tsconn23:
Sometimes support-logging will die if it is started by itself via docker compose. Example:

```
~/go/src/github.com/edgexfoundry/edgex-go/docker $ docker-compose run logging
Starting edgex-files ... done
Starting edgex-core-consul ... done
Starting edgex-mongo       ... done
Starting edgex-config-seed ... done
ERROR: 2018/10/08 22:21:34 edgex-support-logging: Service bootstrap failed!
```

When this happens, attempts to connect to the Mongo container and Consul subsequent to the ERROR are fine. Also, the configuration data is populated correctly in Consul.

Please investigate.

PR #641, opened 2018-10-08
Fix improper read from Consul during bootstrap (support-logging)

tsconn23:
#640 

Added content check to verify something received from Consul. If not, force reconnect.
Normalized docker build target name in Makefile

I verified this fix locally as follows:
1.) Stop all Docker containers
2.) Prune all Docker containers `docker container prune`
3.) `make docker_config_seed`
4.) `make docker_support_logging`
5.) cd docker
6.) `docker-compose run logging`

You will see one ERROR line logged followed by successful startup.

### DIFF
--- a/internal/support/logging/init.go
+++ b/internal/support/logging/init.go
@@ -136,6 +136,10 @@ func connectToConsul(conf *ConfigurationStruct) (*ConfigurationStruct, error) {
 			return conf, errors.New("type check failed")
 		}
 		conf = actual
+		//Check that information was successfully read from Consul
+		if len(conf.Persistence) == 0 {
+			return nil, errors.New("error reading from Consul")
+		}
 	}
 	return conf, err
 }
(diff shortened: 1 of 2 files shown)


internal/support/logging/init.go
```diff
@@ -136,6 +136,10 @@ func connectToConsul(conf *ConfigurationStruct) (*ConfigurationStruct, error) {
 			return conf, errors.New("type check failed")
 		}
 		conf = actual
+		//Check that information was successfully read from Consul
+		if len(conf.Persistence) == 0 {
```
jduranf: This looks like a tricky way to know if configuration have been read or not. Is not possible to perform the check in another way?
If we have to perform this check in order to be sure the configuration it's fine, maybe we should implement similar checks in the other microservices.

tsconn23: @jduranf I agree the question now is should we do this in other microservices. Considering that. What's happening is that there's some timing issue between when the support-logging container reads from Consul and when config-seed writes to Consul. In this case, Consul is up but the config-seed hasn't written the necessary key yet, so the service receives a blank structure. Since no error is thrown, I don't know of any other way to test that we got the content we're looking for. In the above scenario, I have to then return a nil up the stack in order to force the initializeConfiguration() function to be called again.

PR #611, opened 2018-10-01
check req error

dvrkps:
Signed-off-by: Davor Kapsa <davor.kapsa@gmail.com>

PR #642, opened 2018-10-09
Refactored several clients code

jduranf:
This is the second part of #609 work. Also should fix #611 issue.
Some considerations:

- The intention of this PR is not fix several hard coded strings, specially in URL
- Also is not the intention of this PR to review the url.QueryEscape() usage, which should be done.
- export-distro client should be refactored because `ApiNotifyRegistrationRoute` should be part of the URL, like it's done with other clients. This will be done in another PR.
- export-client should be refacatored (is not refreshed by Consul, like other clients) and moved in pkg/clients/distro/client. This will be done in another PR.
- pkg/clients/scheduler should be refactored. Anyway, it's not clear for me the purpose of this client and the code is not used. Anyone can explain which is the purpose of this code? Maybe should be better to remove it.


tsconn23:
@jduranf No time to look at this today with all of the other Delhi stuff coming in. With regard to the scheduler client, we do have a PR #636 landing soon that will make that client of use. I hope to be able to look this PR over tomorrow.

jduranf:
@tsconn23 support-scheduler will not use this client, `pkg/clients/scheduler` have the same functionality as `pkg/clients/metadata/schedule.go` and `pkg/clients/metadata/scheduleevent.go`, so will use this second ones.

tsconn23:
@jduranf The Sys Mgmt Agent will need the client for the config/metrics endpoints.

jduranf:
@tsconn23 currently the purpose of this client is to perform those actions in core-metadata:
```
type SchedulerClient interface {
	AddSchedule(schedule models.Schedule) error
	AddScheduleEvent(scheduleEvent models.ScheduleEvent) error
	QuerySchedule(id string) (models.Schedule, error)
	QueryScheduleEvent(id string) (models.ScheduleEvent, error)
	QueryScheduleWithName(scheduleName string) (models.Schedule, error)
	RemoveSchedule(id string) error
	RemoveScheduleEvent(id string) error
	UpdateSchedule(schedule models.Schedule) error
	UpdateScheduleEvent(scheduleEvent models.ScheduleEvent) error
}
```
Those actions are done using the client schedule in metadata:
```
type ScheduleClient interface {
	Add(dev *models.Schedule) (string, error)
	Delete(id string) error
	DeleteByName(name string) error
	Schedule(id string) (models.Schedule, error)
	ScheduleForName(name string) (models.Schedule, error)
	Schedules() ([]models.Schedule, error)
	Update(dev models.Schedule) error
}
```
And scheduleevent in metadata:
```
type ScheduleEventClient interface {
	Add(dev *models.ScheduleEvent) (string, error)
	Delete(id string) error
	DeleteByName(name string) error
	ScheduleEvent(id string) (models.ScheduleEvent, error)
	ScheduleEventForName(name string) (models.ScheduleEvent, error)
	ScheduleEvents() ([]models.ScheduleEvent, error)
	ScheduleEventsForAddressable(name string) ([]models.ScheduleEvent, error)
	ScheduleEventsForAddressableByName(name string) ([]models.ScheduleEvent, error)
	ScheduleEventsForServiceByName(name string) ([]models.ScheduleEvent, error)
	Update(dev models.ScheduleEvent) error
}
```
Also the implementation of this client is completely different of the other ones. So maybe the best option is to remove this client and add again to get config/metrics endpoints using another client as a model.

jduranf:
@tsconn23 my idea is to remove the scheduler client in another PR, also refactoring export clients. There are enough changes in this PR.

PR #647, opened 2018-10-10
Fix snap releases with new docker setup

### LINKED ISSUE #646: snap builds fail to release
The login file location is wrong; the login file is copied into the git root at `$GIT_ROOT/edgex-snap-store-login`, and then in the dockerfile we execute this: https://github.com/edgexfoundry/edgex-go/blob/13a2f8ee5ede957d2082c06417adfd63e4c50c48/snap/Dockerfile.amd64.build#L5

which means that the login file is located in the container at `/build/edgex-snap-store-login`, however the entrypoint.sh script uses `/edgex-snap-store-login`: https://github.com/edgexfoundry/edgex-go/blob/13a2f8ee5ede957d2082c06417adfd63e4c50c48/snap/entrypoint.sh#L12-L20

See for example: https://jenkins.edgexfoundry.org/view/Snap/job/edgex-go-snap-master-stage-snap-arm/111/consoleFull

```
17:15:20 Snapped edgexfoundry_0.7.0-20181009+13a2f8e_arm64.snap
17:15:22 Usage: snapcraft login [OPTIONS]
17:15:22 
17:15:22 Error: Invalid value for "--with": Could not open file: /edgex-snap-store-login: No such file or directory
```

anonymouse64:
Fixes #646

### DIFF
--- a/snap/build.sh
+++ b/snap/build.sh
@@ -46,4 +46,4 @@ docker run --rm -e "IS_RELEASE_JOB=$IS_RELEASE_JOB" -e "RELEASE=$RELEASE" -e "SN
 
 # also delete the image we created so that it doesn't persist on the system as it contains credentials
 docker image rm --force edgex-snap-builder:latest
-docker image prune
+docker image prune --force
--- a/snap/entrypoint.sh
+++ b/snap/entrypoint.sh
@@ -12,7 +12,7 @@ snapcraft
 # only on release jobs release the snap
 if [ "$IS_RELEASE_JOB" = "YES" ]; then
     # login using the provided login
-    snapcraft login --with /edgex-snap-store-login
+    snapcraft login --with /build/edgex-snap-store-login
     # push the snap up to the store and get the revision of the snap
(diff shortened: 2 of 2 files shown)


snap/build.sh
```diff
@@ -46,4 +46,4 @@ docker run --rm -e "IS_RELEASE_JOB=$IS_RELEASE_JOB" -e "RELEASE=$RELEASE" -e "SN
 
 # also delete the image we created so that it doesn't persist on the system as it contains credentials
 docker image rm --force edgex-snap-builder:latest
-docker image prune
+docker image prune --force
```
JPWKU: @anonymouse64 There is a JJB macro you can use in place of this cleanup here https://github.com/edgexfoundry/ci-management/blob/master/jjb/edgex-macros.yaml#L11-L20
You would use it like this https://github.com/edgexfoundry/ci-management/blob/master/jjb/edgex-templates-go.yaml#L202, that is, call it directly after the build script.

anonymouse64: Yeah I like that solution better, I think I will close this for now and do this in the ci-management repo and re-open this with the image handling stuff removed from `snap/build.sh`

Issue #545, opened 2018-09-10
support-logging logfile doesn't follow convention with other services


anonymouse64:
The current support-logging logfile that is specified in the configuration.toml here: https://github.com/edgexfoundry/edgex-go/blob/335486c1d6d1d9e1ca23778dab0a15cd460a6520/cmd/support-logging/res/configuration.toml#L4
doesn't match the rest of the services which prefix the name of the service with `edgex-`.

This isn't a big deal, but changes things slightly for the snap, i.e. see the difference between the sed expression here: https://github.com/edgexfoundry/edgex-go/blob/335486c1d6d1d9e1ca23778dab0a15cd460a6520/snap/snapcraft.yaml#L294  and here: https://github.com/edgexfoundry/edgex-go/blob/335486c1d6d1d9e1ca23778dab0a15cd460a6520/snap/snapcraft.yaml#L299

jpwhitemn:
@xmlviking - Eric, adding to the backlog of work we have for support logging.  Your call if this gets done for Delhi or needs to push to Edinburgh.

anonymouse64:
Fixed in #659

PR #659, opened 2018-10-11
Added prefix edgex- in support-logging and config-seed log name

jduranf:
Two minor issues are fixed:
- Replaced local variable bootTimeout by internal.BootTimeoutDefault in several microservices
- Added prefix edgex- in support-logging and config-seed log name (issue #545)


tsconn23:
@jduranf This looks good, thanks for the good work! I'm going to wait on merging this until #620 is merged because of the overlap with the sys-mgmt-agent. You may have a small conflict to resolve but I'd rather we do it here since there's some pressure to get 620 in. Thanks for your patience.

jduranf:
@tsconn23 don't worry, I will update this PR after #620 have been merged

PR #665, opened 2018-10-11
Use new go support-notifications in the snap

### LINKED ISSUE #658: support-notifications isn't started in the snap
We recently switched the snap to start using the go version of support-notifications in PR #615  

However, we neglected to test it thoroughly enough as the start-edgex script is still attempting to run the java version of support-notifications which isn't in the snap anymore.

Associated with this is the fact that we need to also install the configuration.toml file for support-notifications during the install hook to `$SNAP_DATA/config/support-notifications` like the other services.

anonymouse64:
Fixes #658

### DIFF
--- a/snap/local/bin/start-edgex.sh
+++ b/snap/local/bin/start-edgex.sh
@@ -57,10 +57,8 @@ if [ "$SUPPORT_NOTIFICATIONS" = "y" ] ; then
     sleep 65
     echo "Starting notifications"
 
-    "$JAVA" -jar -Djava.security.egd=file:/dev/urandom -Xmx100M \
-               -Dspring.cloud.consul.enabled=true \
-               -Dlogging.file=$SNAP_COMMON/logs/edgex-notifications.log \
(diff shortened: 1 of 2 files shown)


snap/local/bin/start-edgex.sh
```diff
@@ -57,10 +57,8 @@ if [ "$SUPPORT_NOTIFICATIONS" = "y" ] ; then
     sleep 65
     echo "Starting notifications"
 
-    "$JAVA" -jar -Djava.security.egd=file:/dev/urandom -Xmx100M \
-               -Dspring.cloud.consul.enabled=true \
-               -Dlogging.file=$SNAP_COMMON/logs/edgex-notifications.log \
-               $SNAP/jar/support-notifications/support-notifications.jar &
+    cd $SNAP_DATA/config/support-logging
```
tonyespy: this should be $SNAP_DATA/config/support-notifications

anonymouse64: Yes you're right

Issue #660, opened 2018-10-11
Config V2 -- support-notifications


tsconn23:
Now that the sys-mgmt-agent PR #620 has been merged, the support-notifications service is ready for mods to support config V2. Since this is the final service, note that all duplicated methods providing backward compatibility with as yet unconverted services will need to be removed as part of this issue. Basic other capabilities include:

1.) Reorganize configuration files using hierarchical TOML
2.) Create composable configuration structs that can be used to write/read to and from Consul. This will replace our current flattened structure.
3.) Allow for re-use of individual configuration structs by other services
4. ) When passing configuration information into initialization functions, only pass the relevant struct
5.) When full URLs are present in current config, eliminate them. Instead use host/port to assemble URLs within the given service client.
6.) Ensure the following settings are part of configuration -- period of service bootstrap, service client refresh interval
7.) Enable real-time configuration changes to configuration data via mods to Consul

PR #668, opened 2018-10-11
Migrate support-notifications to Config V2

tsconn23:
#660 

Some additional notes:
- Reworked tests around TOML verification due to the new location and naming of profile configurations.
- This PR does not target the TODO in notifications/normal_distribution.go. That is for another effort. Actually, an audit of this service's functionality should be done since it's newly converted from Java.
- Reworked minor issues around dockerfiles and makefile found during testing
- Remove aliased functions for preserving backward compatibility. Consequently, had to touch everywhere these functions were called.


tsconn23:
recheck

Issue #673, opened 2018-10-13
SMA -- Eliminate manifest.toml


tsconn23:
Per the following comment on issue #620 
https://github.com/edgexfoundry/edgex-go/pull/620/files#r223187910

With further review, I believe the manifest.toml can be completely eliminated. What we have in this file is essentially a Client definition (supported by the existing ClientInfo type) with operations provided by the MgmtClient interface.

There is already a Notifications client in the SMA
https://github.com/edgexfoundry/edgex-go/blob/master/cmd/sys-mgmt-agent/res/configuration.toml#L13

And when the SMA initializes, it should bootstrap an instance of MgmtClient pointed to the Notifications (and other services). This MgmtClient interface will define the operations like Stop,Start so that a manifest describing those is not needed.

PR #685, opened 2018-10-15
snap/entrypoint.sh: include manifest in the snap

### LINKED ISSUE #684: include manifest.yaml in the snap
If one sets the `SNAPCRAFT_BUILD_INFO` environment variable to 1 when building the snap, snapcraft will include a `manifest.yaml` file inside the snap which lists all of the packages, etc. that were used to build the snap. This is turned on automatically when building the snap using build.snapcraft.io or launchpad.net, but since we don't use those it's not turned on.

All we need to do to turn this on is add that environment variable to https://github.com/edgexfoundry/edgex-go/blob/14173639dff77434f23ecc64a352c11e18aa5d37/snap/entrypoint.sh#L1-L20

anonymouse64:
Fixes #684

### DIFF
--- a/snap/entrypoint.sh
+++ b/snap/entrypoint.sh
@@ -4,6 +4,10 @@
 export LC_ALL=C.UTF-8
 export SNAPCRAFT_SETUP_CORE=1
 
+# this tells snapcraft to include a manifest file in the snap
+# detailing which packages were used to build the snap
+export SNAPCRAFT_BUILD_INFO=1
+
 # build the snap
 cd /build
 snapcraft clean


tsconn23:
Would recommend we NOT pull this in. That file is going away. See issue #673

anonymouse64:
@tsconn23 this is a different, snap specific file (also note that it's `manifest.yaml` not `manifest.toml`). The file I'm referring is generated by snapcraft and specifies what debian packages (i.e. from `apt install ...`) the snap was built with, etc.

An example of what this file looks like for the docker snap: https://pastebin.ubuntu.com/p/RnbDR8Ms9G/

This file makes it easy for tracking if the snap has a security vulnerability in it from a debian package, etc.

PR #686, opened 2018-10-15
Issue 675  - Scheduler router info should return 404 with no data

xmlviking:
Addressed issue with empty object being returned when using the info with an invalid schedule name. 
Will not return schedule/event not found as well as the 404 response.


tsconn23:
I've asked @xmlviking to redo this PR since it looks like the effective change is only one line in router.go, not the currently staged 8 files -- most of which have blank lines added in various places.

xmlviking:
closing this PR and were going to simplify this PR with a issue_675b version.

PR #689, opened 2018-10-16
Download jar files for java services in the snap

### LINKED ISSUE #644: Download jar files from nexus for snap java services
As we still have a few java services left in the snap, mainly `support-scheduler`, `support-rulesengine`, and `device-virtual` we build this from scratch using the maven snapcraft plugin. However this takes a long time to build these, and the jars are already built and published to the staging nexus [here](https://nexus.edgexfoundry.org/content/repositories/staging/org/edgexfoundry/support-scheduler/) (for support-scheduler for example).  

We should instead just download these jar files into the snap rather than build them from scratch.

CC @tonyespy

anonymouse64:
This change switches from using the `maven` plugin in snapcraft to build the jars to instead just download them directly from the nexus via cURL.

@JPWKU we spoke about this over rocketchat and it sounds like using the `staging` repo is the right thing for the master builds of the snap.

Resolves #644

### DIFF
--- a/snap/local/bin/nexus-jar-download.sh
+++ b/snap/local/bin/nexus-jar-download.sh
@@ -9,8 +9,10 @@ download_jar_file()
     export JAR_FILE=$SERVICE-$VERSION.jar
 
     # download the jar file and the md5 sum for it
-    curl https://nexus.edgexfoundry.org/content/repositories/$REPO/org/edgexfoundry/$SERVICE/$VERSION/$JAR_FILE -o $JAR_FILE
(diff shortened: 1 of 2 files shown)


anonymouse64:
recheck

anonymouse64:
recheck

JPWKU:
Hi Ian, We need to talk about this a bit more I think, since we'll have to
know the exact staging repo to pull from.  I'm not sure exactly how we
would ensure that yet.  I should be good to circle back in a bit.

anonymouse64:
After discussion with @JPWKU , there's a problem with nexus in that the default `staging` repo at the URL is the oldest one, not the newest one, so we're going to leave this as is and do this after Delhi.

PR #695, opened 2018-10-16
snap/many: use single dockerfile for all arch values with build args

anonymouse64:
This simplification means that the image name is specified by a build argument and we only need a single dockerfile in the source tree.


anonymouse64:
I think it was just that I had a typo in `snap/build.sh`

anonymouse64:
Recheck

feclare:
@anonymouse64 The ubuntu image is multiarch, as you can see in https://hub.docker.com/_/ubuntu/ and it is available in x86 and amd64.
You can use the same dockerfile to build both architectures. The dockerfiles in edgex-go/docker are a good example of that, even when they use alpine instead of ubuntu.

anonymouse64:
@feclare using multiarch images produces errors when using snapcraft to build the snap, so I think it's just as easy to specify the image architecture when building the file.

feclare:
@anonymouse64 I didn't knew about that snapcraft issue. It makes sense now.

anonymouse64:
Actually, that problem I was mentioning doesn't appear to be a problem anymore, let me change it to just use `ubuntu:16.04` and see if it passes on the CI.

anonymouse64:
@feclare looks like the `ubuntu:16.04` image and using snapcraft works properly in that image, the thing that wasn't working was using the `snapcore/snapcraft:stable` docker image which is only published for x86.
```
pi@raspberrypi:~ $ docker run -it snapcore/snapcraft:stable
standard_init_linux.go:190: exec user process caused "exec format error"
pi@raspberrypi:~ $
```

feclare:
@anonymouse64 It looks better now. 
It is sad that creating multiarch dockers is not an easy task.

PR #696, opened 2018-10-16
Add the metrics handler to the router of each EdgeX service.

akramtexas:
- Update (i.e. implement in) the router of each EdgeX service the metrics handler.
- Work per this Issue is taking place on the following branch: [Construction site](https://github.com/akramtexas/edgex-go/tree/update_router_each_in_services).

### DIFF
--- a/internal/support/notifications/router.go
+++ b/internal/support/notifications/router.go
@@ -17,12 +17,10 @@ package notifications
 
 import (
 	"net/http"
-
+	"runtime"
 	"github.com/gorilla/mux"
-	"fmt"
 	"github.com/edgexfoundry/edgex-go/pkg/clients"
 	"github.com/edgexfoundry/edgex-go/internal"
-	"runtime"
 )
 
 func LoadRestRoutes() *mux.Router {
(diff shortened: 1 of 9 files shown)


internal/support/notifications/router.go
```diff
@@ -105,7 +103,7 @@ func metricsHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// The micro-service is to be considered the System Of Record (SOR) in terms of accurate information.
-	// Fetch metrics for the notifications service.
+	// Fetch metrics for the command service.
```
tsconn23: Fix comment -- This is the notifications service

internal/support/notifications/router.go
```diff
@@ -122,9 +120,13 @@ func metricsHandler(w http.ResponseWriter, r *http.Request) {
 	t.LiveObjects = t.Mallocs - t.Frees
 
 	w.Header().Set("Content-Type", "application/json")
-	encode(t, w)
-
-	LoggingClient.Debug(fmt.Sprintf("Fetched these metrics for the notifications service: {%v}:", t))
+	enc := json.NewEncoder(w)
+	err := enc.Encode(t)
+	// Problems encoding
```
tsconn23: Why not use encode(t, w) as before? the encode() function does error handling so you don't need it below.

See internal/support/notifications/utils.go line 28

akramtexas: Will do (I had noticed that some routers were using _encode()_, while others weren't, so I wanted to make that consistent across the routers). But I agree that it's cleaner to use _encode()_ instead: I'll make it so 👍

akramtexas: Never mind, I was looking at the definition of configHandler(), in reference to the note above. Making the use of configHandler() consistent in metricsHandler() as well!

PR #700, opened 2018-10-17
Added tests for event and valuedescriptor

brandonforster:
Related to issue #457


internal/core/data/coredata_test.go
```diff
@@ -0,0 +1,362 @@
...
+	client.On("DeviceForName", mock.MatchedBy(func(name string) bool {
+		return name == "404"
+	})).Return(mockDeviceForNameResultFn, types.ErrNotFound{})
+	client.On("DeviceForName", mock.Anything).Return(mockDeviceForNameResultFn, fmt.Errorf("some error"))
+
+	return client
+}
+
+func newMockDb() interfaces.DBClient {
+	DB := &dbMock.DBClient{}
```
tsconn23: What if all of this mock db stuff was in a internal/core/data/interfaces/mocks/DBClient_test.go?

internal/core/data/coredata_test.go
```diff
@@ -0,0 +1,362 @@
+package data
```
tsconn23: I'm still not thrilled with the name of this file. If we can eliminate some of it, then the remainder could simply be put into the first _test.go file in the package.

internal/core/data/coredata_test.go
brandonforster: I think putting anything in a file we expect to be 100% generated is a bad idea.

brandonforster: To your point about coredata_test, it's not testing the DBClient, it's mocking it. The reason I went with coredata_test is because the name is literally true, even if it doesn't follow the conventions of the language. This is the opposite: it's literally untrue, but follows the language convention. I'd rather be clear than do this.

Issue #697, opened 2018-10-16
Integrate the newly-created GeneralClient into the SMA.


akramtexas:
- Integrate the newly-created GeneralClient into the SMA. The PR (for that GeneralClient) was recently merged into master: [Add management client to fetch microservices configuration and metrics](https://github.com/edgexfoundry/edgex-go/pull/683).
- Updates being made here: [Construction Site](https://github.com/akramtexas/edgex-go/commits/integrate_general_client_sma)

akramtexas:
- This issue is directly related to: [Issue #690](https://github.com/edgexfoundry/edgex-go/issues/690)

akramtexas:
@jpwhitemn @tsconn23 Based on some preliminary testing I did a few minutes ago, my newly-pushed change-set for the related branch is working to get us api/v1/**config** data for all EdgeX services. Now moving on to get us api/v1/**metrics** data as well. To be followed by lots more testing, and, finally, making sure that the (original) CoreCommand-related updates to master branch are reverted out.

PR #701, opened 2018-10-17
Integrate the newly-created GeneralClient into the SMA.

akramtexas:
- This PR contains the work done in accordance with: [Issue Number 697](https://github.com/edgexfoundry/edgex-go/issues/697)
- With a view to providing insight into DEV testing, Postman was used with the **following two HTTP GET requests** repeatedly:

**api/v1/metrics:**
- http://localhost:48090/api/v1/metrics/edgex-support-logging,edgex-core-metadata,edgex-support-notifications,edgex-core-data,edgex-core-command,edgex-export-client,edgex-export-distro,edgex-support-scheduler

**api/v1/config:**
- http://localhost:48090/api/v1/config/edgex-support-logging,edgex-core-metadata,edgex-support-notifications,edgex-core-data,edgex-core-command,edgex-export-client,edgex-export-distro,edgex-support-scheduler

### DIFF
--- a/internal/system/agent/init.go
+++ b/internal/system/agent/init.go
@@ -32,9 +32,15 @@ var Configuration *ConfigurationStruct
 var LoggingClient logger.LoggingClient
 var Manifest *ManifestStruct
 var Conf = &ConfigurationStruct{}
-var nc general.GeneralClient
 var ec interfaces.ExecutorClient
-var mcc general.GeneralClient
+var gccc general.GeneralClient
+var gccd general.GeneralClient
+var gccm general.GeneralClient
+var gcec general.GeneralClient
+var gced general.GeneralClient
+var gcsl general.GeneralClient
+var gcsn general.GeneralClient
+var gcss general.GeneralClient
 
 func Retry(useConsul bool, useProfile string, timeout int, wait *sync.WaitGroup, ch chan error) {
 	until := time.Now().Add(time.Millisecond * time.Duration(timeout))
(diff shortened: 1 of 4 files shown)


akramtexas:
1. Create support-notifications client: ==> (gcsn)
2. Create core-command client: ==> (gccc)
3. Create core-data client: ==> (gccd)
4. Create core-metadata client: ==> (gccm)
5. Create export-client client: ==> (gcec)
6. Create export-distro client: ==> (gced)
7. Create support-logging client: ==> (gcsl)
8. Create support-scheduler client: ==> (gcss)

internal/system/agent/init.go
```diff
@@ -32,9 +32,15 @@ var Configuration *ConfigurationStruct
...
 var ec interfaces.ExecutorClient
-var mcc general.GeneralClient
+var gccc general.GeneralClient
+var gccd general.GeneralClient
+var gccm general.GeneralClient
+var gcec general.GeneralClient
+var gced general.GeneralClient
+var gcsl general.GeneralClient
+var gcsn general.GeneralClient
+var gcss general.GeneralClient
```
tsconn23: For this first pass, the above is fine. Going forward, consider using a map[string]general.GeneralClient. Set the key to one of the service key constants and then in your service.go loops for metrics and config, use the key to fetch the client from the map and make the relevant call. This will be easier to extend if we add more services.

The initializeClients function below could use a similar strategy to populate the map.

jduranf: Completely agree with @tsconn23!

akramtexas:
@tsconn23 My code has, and continues to, benefit from your feedback, thank you!
@jduranf A special thank you for creating the GeneralClient, which is an especially fine piece of code that has vastly simplified the design aspects of this work!
- Given some other pressing items I need to address, before heading overseas to the Open IoT Summit (Edinburgh), I concur, and appreciate your gents' recommendation that:
> For this first pass, the above is fine. Going forward, consider using a map[string]general.GeneralClient. Set the key to one of the service key constants and then in your service.go loops for metrics and config, use the key to fetch the client from the map and make the relevant call. This will be easier to extend if we add more services. (The initializeClients function below could use a similar strategy to populate the map.)

Issue #715, opened 2018-10-22
config-seed panics when support-scheduler config directory is missing


anonymouse64:
Sometime over the past weekish (git bisect hasn't finished running), config-seed stopping working insofar as it panics:

```
INFO: 2018/10/22 13:21:41 Service dependencies resolved...
INFO: 2018/10/22 13:21:41 dir: device-virtual file: application.properties
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x67d378]

goroutine 1 [running]:
github.com/edgexfoundry/edgex-go/vendor/github.com/pelletier/go-toml.(*Tree).ToMap(0x0, 0x49)
	/home/ijohnson/git/edgex-go/parts/edgex-go/build/.gopath/src/github.com/edgexfoundry/edgex-go/vendor/github.com/pelletier/go-toml/tomltree_write.go:318 +0x48
github.com/edgexfoundry/edgex-go/internal/seed/config.ImportConfiguration(0x7ffd541502af, 0x20, 0x0, 0x0, 0x0, 0x0)
	/home/ijohnson/git/edgex-go/parts/edgex-go/build/.gopath/src/github.com/edgexfoundry/edgex-go/internal/seed/config/populate.go:103 +0x81c
main.main()
	/home/ijohnson/git/edgex-go/parts/edgex-go/build/.gopath/src/github.com/edgexfoundry/edgex-go/cmd/config-seed/main.go:53 +0x384
```

config-seed is being invoked as follows:
```
cd "$SNAP_DATA"/config/config-seed

"$SNAP"/bin/config-seed -c "$SNAP_DATA"/config
```

I have confirmed that consul is running before running config-seed here and this is the filesystem layout (`$SNAP_DATA` is `/var/snap/edgexfoundry/x1`):
```
$ tree /var/snap/edgexfoundry/x1/config
/var/snap/edgexfoundry/x1/config
├── config-seed
│   └── res
│       ├── configuration.toml
│       └── properties
│           └── device-virtual
│               └── application.properties
├── core-command
│   └── res
│       └── configuration.toml
├── core-data
│   └── res
│       └── configuration.toml
├── core-metadata
│   └── res
│       └── configuration.toml
├── export-client
│   └── res
│       └── configuration.toml
├── export-distro
│   └── res
│       └── configuration.toml
├── security-api-gateway
│   ├── kong.conf
│   └── res
[...]

PR #716, opened 2018-10-23
Use go implementation of support-scheduler in snap

### LINKED ISSUE #693: Add go version of support-scheduler to snap
We should add the initial go implementation of support-scheduler to the snap for Delhi.

anonymouse64:
* Update start-edgex.sh to use the new go support-scheduler binary
* Update install hook to also install the support-scheduler config files

Fixes #693


anonymouse64:
Closing until tests are setup for Delhi properly...

anonymouse64:
This branch will also somewhat resolve the panic from #715, as we will have the support-scheduler configuration directory available to config-seed so it doesn't panic.

PR #710, opened 2018-10-18
Add support-rulesengine to snap

anonymouse64:
Fixes #552 for delhi/master

PR #717, opened 2018-10-23
Add support-rulesengine to snap

### LINKED ISSUE #552: [snap] Missing support-rulesengine
While running the blackbox tests against the latest California candidate snap (**0.6.1~pre1-20180907+fb544b5 / 31**), I discovered that we'd never added **support-rulesengine** to the snap.

I've started this work [here](https://github.com/tonyespy/edgex-go/tree/cali-add-rulesengine), however there's an issue with the service not finding its template directory which is causing it to crash:

> 
> [2018-09-12 17:58:52.021] boot - 3406  INFO [main] --- RuleEngine: Starting Drools with drl files from:  edgex/rules
> [2018-09-12 17:58:52.022] boot - 3406  INFO [main] --- RuleEngine: Uploading Drool rules...
> [2018-09-12 17:58:52.022] boot - 3406  INFO [main] --- RuleEngine: ...no rules found to upload
> [2018-09-12 17:58:52.102] boot - 3406  WARN [main] --- MavenSettings: Environment variable M2_HOME is not set
> [2018-09-12 17:58:53.784] boot - 3406  WARN [main] --- AbstractKieModule: No files found for KieBase defaultKieBase
> [2018-09-12 17:58:53.864] boot - 3406  INFO [main] --- KieRepositoryImpl: KieModule was added: MemoryKieModule[releaseId=org.default:artifact:1.0.0-SNAPSHOT]
> [2018-09-12 17:58:54.147] boot - 3406 ERROR [main] --- RuleCreator: Problem getting rule template location.edgex/templates does not exist.
> [2018-09-12 17:58:54.163] boot - 3406 ERROR [main] --- SpringApplication: Application startup failed
> org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'ruleCreator': Invocation of init method failed; nested exception is java.io.FileNotFoundException: edgex/templates does not exist.
> 
>

anonymouse64:
Re-opening of #710 
Fixes #552 

* This branch currently suffers from #715 
* We need to install the application.properties file into `$SNAP_DATA/config/config-seed/res/properties/edgex-support-rulesengine` rather than `$SNAP_DATA/config/config-seed/res/properties/support-rulesengine` (note the `edgex-` prefix) because otherwise support-rulesengine fails to read it's configuration from consul properly. This is likely a bug with support-rulesengine and I can file one there if requested.
* The application.properties file has to be processed using sed to use the runtime value of `$SNAP_DATA`, etc.
* Renaming the support-notifications log to be `support-notifications.log` rather than `edgex-support-notifications.log` which is inconsistent with the other services
* Also inserting `$SNAP_COMMON` verbatim in the support-notifications `configuration.toml` when building edgex-go so that this value gets replaced during the install hook properly.


anonymouse64:
Closing until tests are setup for Delhi...

Issue #720, opened 2018-10-24
isIntervalValid() function to check scheduler frequency is useless


jduranf:
This is the current implementation of this function:
```go
func isIntervalValid(frequency string) bool {
	_, err := strconv.Atoi("-42")
	if err != nil {
		return false
	}
	return true
}
```
Which of course always returns `true`. This should be fixed.

PR #721, opened 2018-10-24
Fixed isIntervalValid() used to check scheduler frequency

jduranf:
Fix issue #720 

Also removed some redundant scheduler logs.


jduranf:
recheck

internal/core/metadata/rest_scheduleevent.go
```diff
@@ -28,11 +29,13 @@ import (
 )
 
 func isIntervalValid(frequency string) bool {
-	_, err := strconv.Atoi("-42")
-	if err != nil {
-		return false
+	matched, _ := regexp.MatchString(`^P(\d+Y)?(\d+M)?(\d+D)?(T(\d+H)?(\d+M)?(\d+S)?)?$`, frequency)
```
drasko: Move string constant on the top of the file.

internal/core/metadata/rest_scheduleevent.go
```diff
@@ -611,7 +614,7 @@ func restAddSchedule(w http.ResponseWriter, r *http.Request) {
 	if s.Frequency != "" {
 		if !isIntervalValid(s.Frequency) {
 			err := errors.New("Frequency format incorrect: " + s.Frequency)
-			LoggingClient.Error("Frequency format is incorrect: "+err.Error(), "")
+			LoggingClient.Error(err.Error(), "")
```
tsconn23: Just a comment -- This could be a good candidate for a custom error type.

internal/support/scheduler/schedule.go
```diff
@@ -9,15 +9,16 @@ package scheduler
...
 	"io/ioutil"
 	"net/http"
 	"regexp"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/edgexfoundry/edgex-go/pkg/models"
+	queueV1 "gopkg.in/eapache/queue.v1"
+	"gopkg.in/mgo.v2/bson"
```
tsconn23: Why do we do this anyway -- putting all of the third party stuff at the bottom? By default the IDE (GoLand) alphabetizes imports when they're added. Then we have to manually move this stuff around. Any downside to leaving them alphabetized?

jduranf: I work using atom with [go-plus](https://atom.io/packages/go-plus). Among other plugins, go-plus includes [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), which automatically sorts the imports in the way you have seen every time I save a file, putting all third party stuff at the bottom.

internal/core/metadata/rest_scheduleevent.go
jduranf: You are right, but let's do it in another PR :)

PR #723, opened 2018-10-26
snap/entrypoint.sh: enable silent reports, push metadata

### LINKED ISSUE #713: push snap metadata when publishing the snap
Currently, we only push the snap file itself up to the store, which doesn't always update the metadata about the snap if there's a conflict. We can do so with:

```
snapcraft push-metadata file.snap --force
```

I'm not sure at this point if we should always use --force, I think yes because we want as much of the details/metadata from the snap to come from the code in this repo as it's easier to update.

anonymouse64:
Fixes #713


anonymouse64:
recheck

anonymouse64:
recheck

PR #732, opened 2018-10-31
snapcraft.yaml: skip tests when building java services

anonymouse64:
This speeds up the builds and also avoids a problem with maven not finding org.apache.maven.surefire.booter.ForkedBooter and the tests failing.

This should fix builds such as https://jenkins.edgexfoundry.org/view/Snap/job/edgex-go-snap-delhi-stage-snap/1


anonymouse64:
I messed this PR up, as it thinks it needs checks from both delhi and master... Will re-open

PR #718, opened 2018-10-23
snapcraft.yaml: update device-virtual and rulesengine to delhi branch

anonymouse64:
Fixes #692

PR #733, opened 2018-10-31
snapcraft.yaml: skip tests when building java services

anonymouse64:
This speeds up the builds and also avoids a problem with maven not finding org.apache.maven.surefire.booter.ForkedBooter and the tests failing.

This should fix builds such as https://jenkins.edgexfoundry.org/view/Snap/job/edgex-go-snap-delhi-stage-snap/1/


JPWKU:
Do you not also want to point at delhi branch?

anonymouse64:
That's handled in another PR: https://github.com/edgexfoundry/edgex-go/pull/718

I'd rather merge this in first and then independently handle that fix in the existing PR

PR #736, opened 2018-10-31
internal/seed/config/populate.go: turn info line to debug

anonymouse64:
This line isn't really necessary info and is rather debug info.


tsconn23:
recheck

PR #750, opened 2018-11-06
Support stopping services in snap with SMA

anonymouse64:
This is an initial implementation of stopping services using `snapctl` when running inside the snap and must be specified with a configuration.toml OperationsType value of "snap".

Note that currently starting/restarting services doesn't work with this.

To test this:
1. Build the snap from this branch 
```bash
git clone https://github.com/anonymouse64/edgex-go.git
cd edgex-go
git checkout bugfix/sma-snap
snapcraft
```
2. Install the snap (note that this will automatically start the services from the snap, including sys-mgmt-agent):
```bash
sudo snap install --devmode edgexfoundry*.snap
```
3. Verify that `sys-mgmt-agent` is running:
```bash
snap services | grep sys-mgmt-agent
```

4. Issue a stop command to a service that is running (for example `core-command`):
```bash
curl --header "Content-Type: application/json" --request POST --data '{"action":"stop","services":["edgex-core-command"]}' localhost:48090/api/v1/operation
```

5. Verify that the service stopped is now indeed stopped:

```bash
snap services | grep core-command
```

### DIFF
--- a/internal/system/agent/executor/snap.go
+++ b/internal/system/agent/executor/snap.go
@@ -1,3 +1,17 @@
+/*******************************************************************************
+ * Copyright 2018 Canonical Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
 package executor
 
 import (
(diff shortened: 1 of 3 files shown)


anonymouse64:
Note that this PR depends on snap services #722

akramtexas:
- To test (and see the updates inaction) I'm setting up a Desktop Image of the VM (Ubuntu 16.04.5 LTS) locally.
- Once I get my virtual machine installed, I will run the relevant commands (thanks to @anonymouse64 for them) to build the snap.

internal/system/agent/init.go
```diff
@@ -89,6 +89,8 @@ func newExecutorClient(operationsType string) (interfaces.ExecutorClient, error)
 		return &executor.ExecuteOs{}, nil
 	case "docker":
 		return &executor.ExecuteDocker{}, nil
+	case "snap":
+		return &executor.ExecuteSnap{}, nil
```
akramtexas: 👍

internal/system/agent/executor/snap.go
```diff
@@ -29,7 +29,7 @@ type ExecuteSnap struct {
 }
 
 // StopService of ExecuteSnap will stop a service in the snap using `snapctl`
-func (oe *ExecuteSnap) StopService(service string, params string) error {
+func (oe *ExecuteSnap) StopService(service string) error {
```
akramtexas: 💯

PR #756, opened 2018-11-07
Make export-distro support kafka client

crazyIceChen:
Make export-distro support kafka client


mhall119:
@crazyIceChen to give you some more info on the changes @tsconn23 mentioned above, the Application Services are going to separate export functionality out from all being in the single export-distro service into multiple independent services (similar to how device services are independent per input source). This will allow you to create a Kafka exporter that runs as it's own microservice, and can be maintained independently of the EdgeX Foundry developers.

Issue #752, opened 2018-11-06
Change LogLevel from Error to Warn When Notifying ScheduleEvent Associates


tsconn23:
When adding Schedule Events for the first time, core-metadata runs logic to notify any associated services. An associated service in this case would be a service whose name is the schedule event's target -- in other words, the service providing a callback addressable for the schedule event. In the case where there is no such service (such as defaulted schedule events whose targets are one of the EdgeX core services) this results in an Error being logged and alarm bells going off in the user's / tester's mind. Error is too severe in this case, so we will reduce the log level to Warning.

This issue will be used to track changes against both the Delhi and master branches.

PR #758, opened 2018-11-07
Don't notify associates when adding or removing a schedule

### LINKED ISSUE #757: Don't notify associates when adding or removing a schedule
When adding or removing a schedule there aren't any associated schedule event, so don't try to notify any associated device service because device service are defined in the schedule events.

jduranf:
Fix #757 


tsconn23:
@jduranf Curious if you saw #752. It sounds very similar to this except at the SchedulerEvent level. 

I may not be able to approve this right away. For Delhi, we wanted to ensure backward compatibility with some of the Java device services that might rely on callbacks, thus we didn't remove the logic from schedule events that you've removed here. 

On the other hand, there is a discussion that needs to happen on the Device Service WG calls to spec out this interaction -- so it may be premature to make this change.

Let me copy in @jpwhitemn @tonyespy @xmlviking @SteveOss for visibility.

jduranf:
@tsconn23 currently those lines of code are useless, there is no backwards compatibility issue.

When you add a schedule, there isn't any associated schedule event, first you must create the schedule and then the associated schedule events. The schedule event defines which device service to notify. So, if there isn't any associated schedule event, then there isn't any device service to notify, so it's useless to call `notifyScheduleAssociates()`.

In the same way happens when removing a schedule. You can remove a schedule if there is an associated schedule event (`isScheduleStillInUse()` check this). So when you remove a schedule there isn't any associated schedule event, then there isn't any device service to notify, so it's useless to call `notifyScheduleAssociates()`.

Anyway, if the callback system will be modified in Edinburgh, then you are right this is a premature change.

tsconn23:
@jduranf If you see this today, please rebase and we'll get it pulled in. Thanks.

tonyespy:
This looks OK to me. In both situations the schedule object being acted upon is "service-less", so calling the notifyScheduleAssociates is pointless...

PR #762, opened 2018-11-07
config-seed startup log WARN if expected directory missing

tsconn23:
#715 

Sample output
`WARN: 2018/11/07 15:42:44 open /Users/tconn/go/src/github.com/edgexfoundry/edgex-go/cmd/support-scheduler/res/configuration.toml: no such file or directory`

### DIFF
--- a/internal/seed/config/populate.go
+++ b/internal/seed/config/populate.go
@@ -97,28 +97,32 @@ func ImportConfiguration(root string, profile string) error {
 
 		LoggingClient.Debug("reading toml " + path)
 		// load the ToML file
-		config, _ := toml.LoadFile(path)
-
-		// Fetch the map[string]interface{}
-		m := config.ToMap()
-
-		// traverse the map and put into KV[]
-		kvs, err := traverse("", m)
+		config, err := toml.LoadFile(path)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


internal/seed/config/populate.go
```diff
@@ -97,28 +97,32 @@ func ImportConfiguration(root string, profile string) error {
 
 		LoggingClient.Debug("reading toml " + path)
 		// load the ToML file
-		config, _ := toml.LoadFile(path)
+		config, err := toml.LoadFile(path)
+		if err != nil {
+			LoggingClient.Warn(err.Error())
```
anonymouse64: Is there a reason we don't also propagate this error here with `return err`?

anonymouse64: Oh also, if you do return the error here, you should drop the else block and re-outdent the else block (in accordance with golint)

tsconn23: If I propagate the error then the loop stops. You said in your comment you wanted the rest of the configs to be populated.

anonymouse64: Ah yes you're right sorry I thought ImportConfiguration was run in a loop somewhere, but I see that it runs the loop itself.

PR #774, opened 2018-11-08
Update snap readme, icon, and store description and summary

anonymouse64:
* Adds a nice snap store badge to the README
* cleans up the description and summary of the snap
* adds an icon for the snap

Note a demo version of how this looks in the snap store is here: https://snapcraft.io/edgexfoundry-core-ijohnson and how it looks in gnome software is here:
gnome-software-edgex-snap

and how it looks in the terminal with `snap info`:
terminal-edgex-snap-description


jpwhitemn:
LTGM Ian.  Where do we use Cassandra though?  Is this something special for the Snap?  Also, do you want to include a link to the Snapcraft store where it is located?

anonymouse64:
> LTGM Ian. Where do we use Cassandra though? Is this something special for the Snap? 

Cassandra is the database provider for the security services (specifically Kong needs it's own database, and supports either Postgres or Cassandra - Postgres is used in docker, but we can't use Postgres in the snap so we use Cassandra). I suppose we could just not mention Cassandra as it could be considered part of the security services.

@tonyespy what do you think about dropping mention of Cassandra from the description?

Issue #777, opened 2018-11-08
Simplify docker-related dependencies in glide.yaml


tsconn23:
It is desirable to target an actual release version of the docker libraries utilized by the SMA rather than a SHA. In addition, since we primarily rely on docker/client can the secondary dependencies currently included be removed?

Previously when attempting to target v17.03.2-ce of `github/docker/docker` we received the following build error:
```
# github.com/edgexfoundry/edgex-go/internal/system/agent/executor
internal/system/agent/executor/docker.go:16:14: undefined: client.NewClientWithOpts
internal/system/agent/executor/docker.go:16:39: undefined: client.WithVersion
internal/system/agent/executor/docker.go:45:14: undefined: client.NewClientWithOpts
internal/system/agent/executor/docker.go:45:39: undefined: client.WithVersion
make: *** [cmd/sys-mgmt-agent/sys-mgmt-agent] Error 2
```

Can this be resolved?

tsconn23:
See comment at the end of PR #778.

PR #763, opened 2018-11-07
Remove unwarranted log.Fatal() statements, which can cause the Go runtime to panic if the YAML file does not already exist.

akramtexas:
- This PR is to address the following issue: [Issue #760](https://github.com/edgexfoundry/edgex-go/issues/760)
- The only reason why Go's built-in (overly-rudimentary) logging (viz. log.Fatal()) was used in the first place has to do with how the package structuring that was not allowing the import of the clients/logging package.
- Further testing revealed an issue, leading to removing those unwarranted log.Fatal() statements, primarily in the function FetchDockerComposeYamlAndPath(), which can cause the Go runtime to panic if the YAML file does not already exist.
- Also, the same function, FetchDockerComposeYamlAndPath(), contains a reference to an older YAML, as follows, which is being updated:
`composeUrl := "https://raw.githubusercontent.com/edgexfoundry/developer-scripts/master/compose-files/docker-compose-california-0.6.1.yml"`

PR #778, opened 2018-11-08
Targeting specific docker version for vendoring (SMA)

tsconn23:
#777 

Also changed `OperationsType` value in res/configuration.toml to "os" to accommodate natively running services.


tsconn23:
Closing for now. Applied these changes on top of #763 and they don't work. Specifically, services don't stop even though they're reported as having done so. Will need to re-investigate once the #763 changes are brought into master. Hopefully there's an indication in this PR of how we can approach streamlining the references.

Issue #782, opened 2018-11-09
SMA panics when issued operation without params


anonymouse64:
As per the RAML docs here: 
https://github.com/edgexfoundry/edgex-go/blob/20d25bfd33ac85c17558a6b17d205bc43f42d09e/api/raml/system-agent.raml#L21

the `params` spec is not required and as such, should be able to be left out when issuing a POST to the `/api/v1/operation` endpoint. However leaving it out, or setting it to be an empty list causes `sys-mgmt-agent` to panic. See:

```bash
curl --header "Content-Type: application/json" --request POST --data '{"action":"stop","services":["edgex-core-command"]}' localhost:48090/api/v1/operation
```
and the logs:
```
Nov 09 16:59:51 xenial-classic edgexfoundry.sys-mgmt-agent[5242]: INFO: 2018/11/09 16:59:51 About to {stop} the service {edgex-core-command} with parameters {[]}.
Nov 09 16:59:51 xenial-classic edgexfoundry.sys-mgmt-agent[5242]: 2018/11/09 16:59:51 http: panic serving [::1]:40668: runtime error: index out of range
Nov 09 16:59:51 xenial-classic edgexfoundry.sys-mgmt-agent[5242]: goroutine 31 [running]:
Nov 09 16:59:51 xenial-classic edgexfoundry.sys-mgmt-agent[5242]: net/http.(*conn).serve.func1(0xc4202f5180)
Nov 09 16:59:51 xenial-classic edgexfoundry.sys-mgmt-agent[5242]:         /home/ijohnson/git/edgex-go/parts/go/build/src/net/http/server.go:1726 +0xd0
Nov 09 16:59:51 xenial-classic edgexfoundry.sys-mgmt-agent[5242]: panic(0x73bc00, 0x991a90)
Nov 09 16:59:51 xenial-classic edgexfoundry.sys-mgmt-agent[5242]:         /home/ijohnson/git/edgex-go/parts/go/build/src/runtime/panic.go:502 +0x229
Nov 09 16:59:51 xenial-classic edgexfoundry.sys-mgmt-agent[5242]: github.com/edgexfoundry/edgex-go/internal/system/agent.InvokeOperation(0x7a9f00, 0x4, 0xc4202e50c0, 0x1, 0x4, 0x0, 0x0, 0x0, 0xc4203dc680)
[...]

PR #783, opened 2018-11-10
SMA Addressed panic for missing params property

tsconn23:
#782 

- Added custom unmarshaling from JSON to assign default values for nullable properties
- Removed the params property altogether since it wasn't used anywhere. It was just increased surface area wherein bugs can be found.

### DIFF
--- a/internal/system/agent/router.go
+++ b/internal/system/agent/router.go
@@ -16,12 +16,14 @@
 package agent
 
 import (
-	"net/http"
-	"github.com/gorilla/mux"
-	"fmt"
 	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
 	"strings"
+
 	"github.com/edgexfoundry/edgex-go/pkg/models"
+	"github.com/gorilla/mux"
 )
 
 func LoadRestRoutes() *mux.Router {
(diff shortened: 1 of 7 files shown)


anonymouse64:
I think for Delhi it's fine to just remove `params`, but as we gain are more implementations of how services are managed by the SMA (i.e. snap, native OS, docker) it should allow specifying runtime parameters to how the services are managed. 

Currently building locally...

internal/system/agent/router.go
```diff
@@ -40,56 +42,52 @@ func LoadRestRoutes() *mux.Router {
...
-	vars := mux.Vars(r)
-	action := vars["action"]
-
-	var params []string
-	var o models.Operation
-	dec := json.NewDecoder(r.Body)
-	err := dec.Decode(&o)
+	defer r.Body.Close()
 
+	b, err := ioutil.ReadAll(r.Body)
```
anonymouse64: Personal nitpick, it's more efficient and safer to just do as originally implemented like:

```go
o := models.Operation{}
err := json.NewDecoder(r.Body).Decode(&o)
// other stuff here
```

This has a few advantages:
1. If someone crafts a bad request like `"{} ... 20 GB of worthless random binary data follows"`, with ioutil.ReadAll you will end up attempting to read all of that into memory and have problems.
2. For the same reason, you end up duplicating the memory needed because you'll have one copy of it somewhere in the body stored internally behind `r.Body`, and then you make a copy of it inside `b` here, whereas with the `json.NewDecoder`, it doesn't actually perform any decoding until you call `Decode`, and just puts what it finds exactly into the object referred to by `&o`. 

That said, what you have works fine so I'll approve for Delhi, but just in general using `json.NewDecoder` is safer for decoding HTTP requests.

tsconn23: I see your point about the large request and the memory utilization. However, I do like this approach because it provides a way through the Unmarshaler interface to initialize a type using custom business logic if necessary. It isn't just a deserialization handler but a more full-featured hook. I'll have to look into it further.

anonymouse64: I'm pretty sure that if the obiect implements the Unmarshaler interface that method will still be used, but more efficiently, but I haven't tested this in a while so I may be wrong on that point.

Issue #760, opened 2018-11-07
Fetching docker-compose YAML has unwarranted log.Fatal() statements.


akramtexas:
Fetching docker-compose YAML has unwarranted log.Fatal() statements, which can cause the Go runtime to panic if the YAML file does not already exist.

akramtexas:
- Also, the same function FetchDockerComposeYamlAndPath() contains a reference to an older YAML, which should be updated:
`composeUrl := "https://raw.githubusercontent.com/edgexfoundry/developer-scripts/master/compose-files/docker-compose-california-0.6.1.yml"`

akramtexas:
The error manifests itself like so:

> edgex-sys-mgmt    | INFO: 2018/11/07 19:48:32 About to {start} the service {edgex-config-seed} with parameters {[graceful]}.
edgex-sys-mgmt    | 2018/11/07 19:48:32 http: panic serving 192.168.32.1:33208: runtime error: invalid memory address or nil pointer dereference
edgex-sys-mgmt    | goroutine 146 [running]:
edgex-sys-mgmt    | net/http.(*conn).serve.func1(0xc420308aa0)
edgex-sys-mgmt    | 	/usr/local/go/src/net/http/server.go:1697 +0xd0
edgex-sys-mgmt    | panic(0x74d060, 0x9986f0)
edgex-sys-mgmt    | 	/usr/local/go/src/runtime/panic.go:491 +0x283
edgex-sys-mgmt    | github.com/edgexfoundry/edgex-go/internal/system/agent/executor.FetchDockerComposeYamlAndPath(0x0, 0x0, 0x0, 0x0)
edgex-sys-mgmt    | 	/go/src/github.com/edgexfoundry/edgex-go/internal/system/agent/executor/compose.go:141 +0xc1
edgex-sys-mgmt    | github.com/edgexfoundry/edgex-go/internal/system/agent/executor.RunDockerComposeCommand.func1(0x1, 0x10, 0x75ebc0, 0x1)
edgex-sys-mgmt    | 	/go/src/github.com/edgexfoundry/edgex-go/internal/system/agent/executor/compose.go:108 +0x2f
edgex-sys-mgmt    | github.com/edgexfoundry/edgex-go/internal/system/agent/executor.Do(0xc4202e2ea0, 0xc4202e2ea0, 0x4291f8)
[...]

Issue #780, opened 2018-11-09
Standardize Logging in the SMA


tsconn23:
The System Mgmt Agent should use the standard Logging Client that all EdgeX services use and not utilize log.Println or log.Printf. An example of how to accomplish this is provided here:

https://github.com/tsconn23/edgex-go/tree/issue_763_tsc

akramtexas:
- The _one-and-only_ reason why Go's built-in (overly-rudimentary) logging (viz. log.Fatal()) was used in the first place has to do with how the package structuring that was not allowing the import of the clients/logging package.
- Looking forward to addressing this issue 👍

PR #784, opened 2018-11-12
Remove (on master branch) unwarranted log.Fatal() statements, which can cause the Go runtime to panic if the YAML file does not already exist.

akramtexas:
- This PR addresses (**on the master branch**) the following issue: [Issue #760](https://github.com/edgexfoundry/edgex-go/issues/760), which was (separately) addressed (_on the delhi branch_) via [PR #763](https://github.com/edgexfoundry/edgex-go/pull/763).
- For the testing approach taken here, as well as other details (rather than repeating them here), please refer to [PR #763](https://github.com/edgexfoundry/edgex-go/pull/763) 👍


internal/system/agent/executor/compose.go
```diff
@@ -119,37 +118,27 @@ func RunDockerComposeCommand(service string, dockerComposeService string) {
...
+		// agent.LoggingClient.Error("Call to docker-compose up -d failed with %s\n", err.Error())
 	}
-
-	// composeOutput := string(out)
-	// fmt.Sprintf("For the micro-service {%v}, we got this docker-compose output:\n {%v}", service, composeOutput)
 	println(out)
 
 	if ! WasDockerContainerComposeStarted(service) {
-		log.Fatal(fmt.Sprintf("The container for {%v} was NOT started!" + service))
+		log.Printf(fmt.Sprintf("The container for {%s} was NOT started!" + service))
```
tsconn23: This line should be

`log.Printf(fmt.Sprintf("The container for {%s} was NOT started!" , service))`

internal/system/agent/executor/compose.go
```diff
@@ -119,37 +118,27 @@ func RunDockerComposeCommand(service string, dockerComposeService string) {
...
-	resp, err := http.Get("https://raw.githubusercontent.com/edgexfoundry/developer-scripts/master/compose-files/docker-compose-california-0.6.0.yml")
+	req, _ := http.NewRequest("GET", composeUrl, nil)
+	res, err := http.DefaultClient.Do(req)
 	if err != nil {
-		log.Fatal(fmt.Sprintf("Call to http.Get() failed with %s\n", err))
+		log.Printf(fmt.Sprintf("Call to http.Get() failed with %s\n", err.Error()))
+		// agent.LoggingClient.Error("Call to http.Get() failed with %s\n", err.Error())
 	}
-	defer resp.Body.Close()
+	body, _ := ioutil.ReadAll(res.Body)
```
tsconn23: Response body is never closed.

internal/system/agent/executor/os.go
```diff
@@ -28,7 +28,8 @@ func findAndStopProcess(output string, err error, process string) error {
 
 	var pid int
 	if err != nil {
-		log.Fatalf("Error during findAndStopProcess() as follows: " + err.Error())
+		log.Printf("Error during findAndStopProcess() as follows: %s" + err.Error())
```
tsconn23: This line should be

`log.Printf("Error during findAndStopProcess() as follows: %s" , err.Error())`

internal/system/agent/executor/os.go
```diff
@@ -43,19 +44,19 @@ func findAndStopProcess(output string, err error, process string) error {
 				tokens := strings.Split(line, " ")
 				pid, err = strconv.Atoi(tokens[0])
 				if err != nil {
-					log.Fatalf("Error converting PID to integer: " + err.Error())
+					log.Printf("Error converting PID to integer: %s" + err.Error())
```
tsconn23: This line should be

`log.Printf("Error converting PID to integer: %s" , err.Error())`

internal/system/agent/executor/os.go
```diff
@@ -43,19 +44,19 @@ func findAndStopProcess(output string, err error, process string) error {
...
 		}
 
 		// Now stop the process using the PID found above.
-		// fmt.Sprintf("OS-level >> The micro-service {%v} with the PID of {%v} is now being stopped!", process, pid)
-
 		// Make a system call.
 		proc, err := os.FindProcess(pid)
 		if err != nil {
-			log.Println(err)
+			log.Printf("Error during os.FindProcess(pid) as follows: %s" + err.Error())
```
tsconn23: Same - comma instead of plus sign

internal/system/agent/executor/compose.go
```diff
@@ -158,34 +147,34 @@ func FetchDockerComposeYamlAndPath() (string, error) {
...
 	}
 
 	// [4] Determine whether we have _already_ fetched the contents and written them to the deployed file-system.
 	if os.IsNotExist(err) {
 		println(fileInfo)
-		// fmt.Sprintf("File {%v} does NOT exist. Therefore, create it.", fileInfo)
 		err = ioutil.WriteFile(filename, []byte(body), 0666)
 		if err != nil {
-			log.Fatal(err)
+			log.Printf(err.Error(), "%s We have already fetched the contents and written them to the deployed file-system!", err.Error())
```
tsconn23: `err.Error()` reference twice in one statement. Redundant.

internal/system/agent/executor/compose.go
```diff
@@ -20,7 +20,7 @@ func WasDockerContainerComposeStarted(service string) bool {
 	cmdName := "docker"
 	cmdArgs := []string{"ps"}
 	if cmdOut, err = exec.Command(cmdName, cmdArgs...).Output(); err != nil {
-		fmt.Fprintln(os.Stderr, "There was an error running the docker-compose command: ", err)
+		fmt.Fprintln(os.Stderr, "There was an error running the docker-compose command: ", err.Error())
```
tsconn23: MIssing `%s` in first string. Nothing to replace with err.Error()

internal/system/agent/executor/compose.go
anonymouse64: Shouldn't it just be 
```go
log.Printf("The container for {%s} was NOT started!\n", service)
```
(also noting the trailing `\n`)

internal/system/agent/executor/compose.go
anonymouse64: Also should be consistent between usage of `fmt.Fprintln` and `log.Printf`. Either use one or the other, otherwise logs look really confusing with some lines having the log prefix and some not having the prefix.

akramtexas: > Also should be consistent between usage of `fmt.Fprintln` and `log.Printf`. Either use one or the other, otherwise logs look really confusing with some lines having the log prefix and some not having the prefix.

- We got [Issue #780](https://github.com/edgexfoundry/edgex-go/issues/780) open for that.

akramtexas:
recheck

tsconn23:
recheck

Issue #739, opened 2018-11-02
Rework logging output to key/value format


tsconn23:
Rework logging output using GoKit as shown in the Core WG call 1-Nov-18. The format will support key/value pairs and capture additional information which I will add shortly as examples.

Use console logger provided from private repo as basis for changes required to edgex-go/pkg/clients/logging/logger.go.

The private logger in the support-logging service itself will also probably have to be revamped.

This should be a non-breaking change given that the interface is largely the same. Existing statements throughout the code will continue to work, but the format of their output will change.

Example of new logging output (from sample data client)
```
level=INFO ts=2018-11-02T21:05:48.478044Z source=main.go:28 msg="bootstrap OK"
level=INFO ts=2018-11-02T21:05:48.485321Z source=logging.go:26 duration=7.171019ms function=registration/Register msg="received client id 5bdcbc2cdeafc29792dffb86"
level=INFO ts=2018-11-02T21:05:48.486093Z source=main.go:62 transport=http address=:48050 msg=listening
level=INFO ts=2018-11-02T21:06:00.636649Z source=logging.go:26 duration=20.622007ms function=receiver/AddEvent msg="received event id 1234567895"
```

PR #787, opened 2018-11-12
Update logging to use GoKit key/value logger.

brandonforster:
Issue #739.

Current logging statements are compatible. The logging client implementation requires minor rework.

Support-logging was left as-is since it's a siloed service.

### DIFF
--- a/internal/core/command/init.go
+++ b/internal/core/command/init.go
@@ -59,6 +59,7 @@ func Retry(useConsul bool, useProfile string, timeout int, wait *sync.WaitGroup,
 				// Setup Logging
 				logTarget := setLoggingTarget()
 				LoggingClient = logger.NewClient(internal.CoreCommandServiceKey, Configuration.Logging.EnableRemote, logTarget, Configuration.Logging.Level)
+
 				//Initialize service clients
 				initializeClients(useConsul)
 			}
--- a/internal/export/client/server_test.go
+++ b/internal/export/client/server_test.go
@@ -34,9 +34,7 @@ func (d *distroMockClient) NotifyRegistrations(models.NotifyUpdate) error {
 }
 
 func prepareTest(t *testing.T) *httptest.Server {
-	if LoggingClient == nil {
-		LoggingClient = logger.NewClient(internal.ExportClientServiceKey, false, "", logger.InfoLog)
-	}
+	LoggingClient = logger.NewClient(internal.ExportClientServiceKey, false, "", logger.InfoLog)
 
 	dbClient = &memory.MemDB{}
 	dc = &distroMockClient{}
--- a/internal/export/client/server_test.go
+++ b/internal/export/client/server_test.go
@@ -34,7 +34,7 @@ func (d *distroMockClient) NotifyRegistrations(models.NotifyUpdate) error {
 }
 
 func prepareTest(t *testing.T) *httptest.Server {
-	LoggingClient = logger.NewClient(internal.ExportClientServiceKey, false, "", logger.InfoLog)
+	LoggingClient = logger.NewClient(internal.ExportClientServiceKey, false, "./logs/edgex-export-client-test.log", logger.InfoLog)
 
 	dbClient = &memory.MemDB{}
 	dc = &distroMockClient{}
--- a/pkg/models/log_entry.go
+++ b/pkg/models/log_entry.go
@@ -16,22 +16,22 @@ package models
 import "encoding/json"
 
 type LogEntry struct {
-	Level         string   `json:"logLevel"`
-	Labels        []string `json:"labels"`
-	OriginService string   `json:"originService"`
-	Message       string   `json:"message"`
-	Created       int64    `json:"created"`
+	Level         string        `json:"logLevel"`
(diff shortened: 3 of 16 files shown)


internal/core/command/init.go
```diff
@@ -59,6 +59,10 @@ func Retry(useConsul bool, useProfile string, timeout int, wait *sync.WaitGroup,
 				// Setup Logging
 				logTarget := setLoggingTarget()
 				LoggingClient = logger.NewClient(internal.CoreCommandServiceKey, Configuration.Logging.EnableRemote, logTarget, Configuration.Logging.Level)
+				if err != nil {
+					ch <- err
+				}
```
jduranf: What's the purpose of those three lines? `err` always is `nil`...

brandonforster: During an iteration of this newClient() returned an err and I removed the return but forgot to remove the error handling. Sorry about that!

tsconn23:
In core-data/res/configuration.toml set `EnableRemote = true`. A panic occurs.

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x432c23e]

goroutine 1 [running]:
github.com/go-kit/kit/log.(*context).Log(0xc000093260, 0xc00000c820, 0x2, 0x2, 0xc0000aae08, 0x4498180)
	/Users/tconn/go/src/github.com/go-kit/kit/log/log.go:124 +0x19e
github.com/edgexfoundry/edgex-go/pkg/clients/logging.EdgeXLogger.log(0x442e6dd, 0xf, 0x1, 0xc00015a030, 0x22, 0xc00012f740, 0x4498180, 0xc000122ea0, 0xc000122e70, 0x442b5b2, ...)
	/Users/tconn/go/src/github.com/edgexfoundry/edgex-go/pkg/clients/logging/logger.go:132 +0x231
github.com/edgexfoundry/edgex-go/pkg/clients/logging.EdgeXLogger.Info(0x442e6dd, 0xf, 0x1, 0xc00015a030, 0x22, 0xc00012f740, 0x4498180, 0xc000122ea0, 0xc000122e70, 0x4435748, ...)
	/Users/tconn/go/src/github.com/edgexfoundry/edgex-go/pkg/clients/logging/logger.go:170 +0xca
main.main()
	/Users/tconn/go/src/github.com/edgexfoundry/edgex-go/cmd/core-data/main.go:56 +0x260
```

tsconn23:
Also, services do not appear to be writing to their log file. For example:

```
cd edgex-go
make build
cd cmd/core-data
rm -rf logs
./core-data
```

This will delete and then re-create the logs directory. If you `cd logs` an `edgex-core-data.log` is created but it's 0 bytes.

`-rw-r--r--  1 tconn  staff    0 Nov 13 10:01 edgex-core-data.log`

internal/export/client/server_test.go
```diff
@@ -34,10 +34,15 @@ func (d *distroMockClient) NotifyRegistrations(models.NotifyUpdate) error {
 }
 
 func prepareTest(t *testing.T) *httptest.Server {
+	var err error
 	if LoggingClient == nil {
 		LoggingClient = logger.NewClient(internal.ExportClientServiceKey, false, "", logger.InfoLog)
 	}
 
+	if err != nil {
```
jduranf: This `err` check should be removed

pkg/models/log_entry.go
```diff
@@ -16,22 +16,22 @@ package models
 import "encoding/json"
 
 type LogEntry struct {
-	Level         string   `json:"logLevel"`
-	Labels        []string `json:"labels"`
-	OriginService string   `json:"originService"`
-	Message       string   `json:"message"`
-	Created       int64    `json:"created"`
+	Level         string        `json:"logLevel"`
+	KeyVals       []interface{} `json:"labels"`
```
jduranf: You changed `Labels` by `KeyVals`, but in `internal/support/logging/models/log_entry.go` you changed `Labels` by `Args`. In both cases json label is still `labels`
Is that correct?

pkg/models/log_entry.go
```diff
@@ -16,22 +16,22 @@ package models
...
-		Level         *string  `json:"logLevel"`
-		Labels        []string `json:"labels"`
-		OriginService *string  `json:"originService"`
-		Message       *string  `json:"message"`
-		Created       int64    `json:"created"`
+		Level         *string       `json:"logLevel"`
+		Args          []interface{} `json:"args"`
+		OriginService *string       `json:"originService"`
+		Message       *string       `json:"message"`
+		Created       int64         `json:"created"`
```
jduranf: I know it's out of the scope of this PR, but should be great if you add the `,omitempty` tag in those five parameters. The rest of this PR LGTM (if you want to merge without this change).

akramtexas: @jduranf I agree with the thought of making that additional change (i.e. related to the `omitempty` tag) _separately_, rather than as part of this PR.

akramtexas: Adding that this PR, _as_ it stands, LGTM as well 👍

tsconn23:
recheck

Issue #792, opened 2018-11-14
core-command sometimes fails to get config from consul


anonymouse64:
Sometimes, when restarting a system with the snap installed and `core-command` enabled to get it's configuration from consul, we will see the following log:

```
ijohnson@xenial-classic:~$ journalctl -ef -u snap.edgexfoundry.core-command.service 
-- Logs begin at Wed 2018-11-14 07:26:58 CST. --
Nov 14 07:27:06 xenial-classic systemd[1]: Starting Service for snap application edgexfoundry.core-command...
Nov 14 07:27:06 xenial-classic systemd[1]: Started Service for snap application edgexfoundry.core-command.
Nov 14 07:27:06 xenial-classic edgexfoundry.core-command[1721]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
Nov 14 07:27:06 xenial-classic edgexfoundry.core-command[1721]:                                  Dload  Upload   Total   Spent    Left  Speed
Nov 14 07:27:10 xenial-classic edgexfoundry.core-command[1721]: [474B blob data]
Nov 14 07:27:10 xenial-classic edgexfoundry.core-command[1721]: INFO: 2018/11/14 07:27:10 Service dependencies resolved...
Nov 14 07:27:10 xenial-classic edgexfoundry.core-command[1721]: INFO: 2018/11/14 07:27:10 Starting edgex-core-command 0.7.0
Nov 14 07:27:10 xenial-classic edgexfoundry.core-command[1721]: INFO: 2018/11/14 07:27:10
Nov 14 07:27:10 xenial-classic edgexfoundry.core-command[1721]: INFO: 2018/11/14 07:27:10 Service started in: 17.42021ms
Nov 14 07:27:10 xenial-classic edgexfoundry.core-command[1721]: INFO: 2018/11/14 07:27:10 Listening on port: 0
Nov 14 07:27:10 xenial-classic edgexfoundry.core-command[1721]: ERROR: 2018/11/14 07:27:10 Get http://:0/v1/kv/config/V2/edgex-core-command/?recurse=&wait=1800000ms: dial tcp :0: connect: connection refused
Nov 14 07:27:11 xenial-classic edgexfoundry.core-command[1721]: ERROR: 2018/11/14 07:27:11 Get http://:0/v1/kv/config/V2/edgex-core-command/?recurse=&wait=1800000ms: dial tcp :0: connect: connection refused
[...]

PR #798, opened 2018-11-14
Recheck consul connection in case of bootstrap race (master)

tsconn23:
#792 (for master branch)

Standardizing on service port check == 0 to determine if service is in race condition with config-seed population in Consul.

### DIFF
--- a/internal/support/logging/init.go
+++ b/internal/support/logging/init.go
@@ -133,7 +133,7 @@ func connectToConsul(conf *ConfigurationStruct) (*ConfigurationStruct, error) {
 		}
 		conf = actual
 		//Check that information was successfully read from Consul
-		if len(conf.Persistence) == 0 {
+		if conf.Service.Port == 0 {
 			return nil, errors.New("error reading from Consul")
 		}
 	}
--- a/internal/support/notifications/init.go
+++ b/internal/support/notifications/init.go
@@ -187,7 +187,7 @@ func connectToConsul(conf *ConfigurationStruct) (*ConfigurationStruct, error) {
 		}
 		conf = actual
 		//Check that information was successfully read from Consul
-		if conf.ResendLimit == 0 {
+		if conf.Service.Port == 0 {
 			return nil, errors.New("error reading from Consul")
 		}
 	}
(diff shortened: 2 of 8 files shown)


internal/support/logging/init.go
```diff
@@ -133,7 +133,7 @@ func connectToConsul(conf *ConfigurationStruct) (*ConfigurationStruct, error) {
 		}
 		conf = actual
 		//Check that information was successfully read from Consul
-		if len(conf.Persistence) == 0 {
```
akramtexas: I'm trying to understand how / why the need for this logical check (i.e. `conf.ResendLimit == 0`) was obviated instead of perhaps 'rolling' this logical check into (or alongside) the one for `conf.Service.Port == 0`.

internal/support/notifications/init.go
```diff
@@ -187,7 +187,7 @@ func connectToConsul(conf *ConfigurationStruct) (*ConfigurationStruct, error) {
 		}
 		conf = actual
 		//Check that information was successfully read from Consul
-		if conf.ResendLimit == 0 {
```
akramtexas: Likewise, I'm trying to understand how / why the need for this logical check (i.e. `conf.ResendLimit == 0`) was obviated instead of perhaps 'rolling' this logical check into (or alongside) the one for `conf.Service.Port == 0`.

akramtexas:
Other than the two questions posted above, the PR LGTM 👍

internal/support/logging/init.go
anonymouse64: The linked bug shows that there is a race in using Consul with services where if a service was started *immediately* at the same time as consul, consul would be up and return garbage data to the service, resulting in the service effectively having an uninitialized struct. This check ensures that the struct was in fact initialized with some sensible data. A program is never allowed to explicitly bind to port 0 (well you can make the syscall and this lets the OS choose a different port other than 0 for you) so if the port is 0 then we know that the struct is uninitialized and we need to wait for consul longer.

akramtexas: Make sense, cool 💯

PR #801, opened 2018-11-15
updating docs - getting started guides, some service documents

jpwhitemn:
Signed-off-by: Jim White <james_white2@dell.com>


mhall119:
Is this identical to the changes you made in a previous pull request?

PR #577, opened 2018-09-20
docs/Dockerfile.build: specify version for pip packages

anonymouse64:
This fixes #576 by versioning the pip packages used to build the docs. Apparently Sphinx v1.8 breaks on Linux Foundation CI running Python 2.7.5.

PR #810, opened 2018-11-15
Add Google Analytics to generated documentation

mhall119:
This was a change from the california branch that didn't make it into master


anonymouse64:
When I merged https://github.com/edgexfoundry/edgex-go/pull/577 I also specified `sphinxcontrib-googleanalytics==0.1` I think that should be added as well here.

But since this is last minute I think it's fine to not do it as well.

mhall119:
Yeah, I cherry-picked both those commits

PR #818, opened 2018-11-19
snapcraft.yaml: don't hard-code version in jar file names

anonymouse64:
This will fix build failures such as https://jenkins.edgexfoundry.org/view/Snap/job/edgex-go-snap-delhi-stage-snap/20/console which failed because the device-virtual jar just got upgraded to 0.5.1, which means that the install line failed.
Instead, since there will only ever be a single jar file for that service we can just glob whatever version was installed while building. That will also ensure that we don't suffer from the same problem in the future if the java services have their version revised.


anonymouse64:
Apparently `sh` doesn't like using the `*` in a file name like this, so instead I had to use `find` to get the realt file name of the jar file.

PR #854, opened 2018-11-26
snapcraft.yaml: use go implementation of support-scheduler (master)

anonymouse64:
Duplicate of #716 but for master (i.e. Edinburgh)

* Update start-edgex.sh to use the new go support-scheduler binary
* Update install hook to also install the support-scheduler config files


jpwhitemn:
This PR is still being worked by @tonyespy and @anonymouse64 as of 12/11/18 check in with them.  Leave open for Espy review.

Issue #811, opened 2018-11-16
core-data: Eliminate Mongo Types from Models


tsconn23:
For all models currently used by the core-data service:

- Modify the ID property to be a string instead of a bson.ObjectId
- Remove bson serialization annotations

It will still be necessary to handle BSON IDs natively within the Mongo persistence layer, but these should be transformed to string values and handed out to the rest of the application and external consumers through the contract models.

The solution should be backward compatible with existing data, such as that found in the blackbox tests. GET / PUT / DELETE operations should function when being given either a BSON ID (existing data) or UUID. POST operations that create new records should cause UUID values to be generated and returned.

This issue and others to be created shortly supersede #495 . The work will touch all of EdgeX-Go and PRs should be small enough to be manageable. Future issues for this work should define a scope which a subsequent PR will target. We will most likely have duplicate models within the mongo persistence layer while this work progresses (models in the `mongo` package and `mongo/models` package) to allow backward compatibility, with the idea that all models will be moved into `mongo\models`

PR #856, opened 2018-11-26
Decouple Mongo types from core-data contract models

tsconn23:
#811 
 
- Builds on initial work done by Brandon Forster
- Removed old MongoEvent type
- All core-data blackbox tests passing (companion PR here https://github.com/edgexfoundry/blackbox-testing/pull/109)

Additional notes:

1. All exported MongoClient methods within package db/mongo accept and return "contract" types. This is because they are utilized by the business logic which should not be concerned with Mongo particulars. Within the MongoClient itself, the contract is then mapped into a native model which deals with the Mongo-specific types and queries accordingly based on whether we need to use the BSON ID or UUID.
2. All non-exported MongoClient methods within package db/mongo accept and return db/mongo/models types. That is, they only deal with internal Mongo-specific types.
3. I eliminated the old MongoEvent type for a couple reasons. First, I only wanted to have database access code in the db/mongo package, so I moved the models to a "models" sub-package. Second, the MongoEvent type actually did database access itself when loading readings for an event. This broke separation of concerns (state vs db access). I had to add logic into the db/mongo/data.go file to compensate for removing this.
4. When adding/deleting/updating an item it is necessary to check whether the ID on the incoming contract is a BSON ID or UUID. We then need to run the appropriate query within the given function.
5. Decisions around when to generate new IDs (both BSON and UUID) is now encapsulated within the models and has been removed from db/mongo
6. I have aliased the edgex-go/pkg/models package as "contracts" to make it clearer which layer of the application a given model corresponds to. I may propose that we actually change the name of the above package rather than aliasing it everywhere.


pkg/models/event.go
```diff
@@ -27,37 +25,39 @@ import (
...
-		ID       bson.ObjectId `json:"id"`
-		Pushed   int64         `json:"pushed,omitempty"`
-		Device   *string       `json:"device,omitempty"` // Device identifier (name or id)
-		Created  int64         `json:"created,omitempty"`
-		Modified int64         `json:"modified,omitempty"`
-		Origin   int64         `json:"origin,omitempty"`
-		Schedule *string       `json:"schedule,omitempty"` // Schedule identifier
-		Event    *string       `json:"event,omitempty"`    // Schedule event identifier
-		Readings []Reading     `json:"readings,omitempty"` // List of readings
+		ID       *string   `json:"id"`
```
jduranf: Can you add `omitempty` tag? If an event isn't stored in database, ID field always is empty. After this modifications, maybe you should update some tests.

pkg/models/reading.go
```diff
@@ -28,20 +26,20 @@ import (
...
+	Device   string        `json:"device"`
+	Name     string        `json:"name"`
+	Value    string        `json:"value"` // Device sensor data value
 }
 
 // Custom marshaling to make empty strings null
 func (r Reading) MarshalJSON() ([]byte, error) {
 	test := struct {
-		Id       bson.ObjectId `json:"id"`
+		Id       *string       `json:"id"`
```
jduranf: Same as Event, can you add `omitempty` tag?

pkg/models/value-descriptor.go
```diff
@@ -16,54 +16,54 @@ package models
...
-		Origin       int64         `json:"origin,omitempty"`
-		Name         *string       `json:"name,omitempty"`
-		Min          *interface{}  `json:"min,omitempty"`
-		Max          *interface{}  `json:"max,omitempty"`
-		DefaultValue *interface{}  `json:"defaultValue,omitempty"`
-		Type         *string       `json:"type,omitempty"`
-		UomLabel     *string       `json:"uomLabel,omitempty"`
-		Formatting   *string       `json:"formatting,omitempty"`
-		Labels       []string      `json:"labels,omitempty"`
+		Id           *string      `json:"id"`
```
jduranf: `omitempty` tag? :)

internal/core/data/device_test.go
```diff
@@ -169,19 +166,16 @@ func newMockDb() interfaces.DBClient {
 		return start == 0xBADF00D
 	}), mock.Anything, mock.Anything).Return(nil, fmt.Errorf("some error"))
 
-
```
brandonforster: I added the spacing to try and distinguish between different method mocks since the code is extremely samey. Do we want any sort of separator so it's not just uniform single line gaps or is this fine?

internal/core/data/event_test.go
```diff
@@ -17,6 +17,7 @@ package data
 import (
 	"github.com/edgexfoundry/edgex-go/internal/core/data/errors"
 	"github.com/edgexfoundry/edgex-go/pkg/models"
+	"github.com/satori/go.uuid"
```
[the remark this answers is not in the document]
tsconn23: Will be included in PR update

internal/pkg/db/memory/data.go
```diff
@@ -295,11 +296,11 @@ func (m *MemDB) ReadingsByCreationTime(start, end int64, limit int) ([]models.Re
 	return readings, nil
 }
 
-func (m *MemDB) AddValueDescriptor(value models.ValueDescriptor) (bson.ObjectId, error) {
+func (m *MemDB) AddValueDescriptor(value models.ValueDescriptor) (string, error) {
 	currentTime := db.MakeTimestamp()
 	value.Created = currentTime
 	value.Modified = currentTime
-	value.Id = bson.NewObjectId()
+	value.Id = bson.NewObjectId().Hex()
```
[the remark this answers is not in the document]
tsconn23: Will be included in PR update

internal/pkg/db/mongo/data.go
```diff
@@ -29,76 +31,97 @@ Has functions for interacting with the core data mongo database
...
+	if len(evt.Readings) > 0 {
+		var ui []interface{}
+		for i := range evt.Readings {
+			ui = append(ui, evt.Readings[i])
 		}
-		err := s.DB(mc.database.Name).C(db.ReadingsCollection).Insert(ui...)
+		err = s.DB(mc.database.Name).C(db.ReadingsCollection).Insert(ui...)
 		if err != nil {
-			return e.ID, err
+			return evt.Id.Hex(), err
```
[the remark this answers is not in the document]
tsconn23: Yeah, if it hasn't been added both of those properties will be blank. Or if they aren't blank, then the ID returned is useless and won't be found anywhere in the DB. I did change the line you called out above. Will be in PR update.

pkg/models/reading.go
tsconn23: Will be included in updated PR

pkg/models/value-descriptor.go
tsconn23: Will be included in updated PR

internal/core/data/device_test.go
tsconn23: IMHO a single line of spacing is fine.

tsconn23:
Thanks @jduranf !

Issue #860, opened 2018-11-28
Redis does not assign the device name from an event to each reading in the event when writing to the database


TomFlem:
The Mongodb implementation assigns the device name given in the event to each reading. Redis does not. 

A call to /api/v1/reading thus returns readings where: "device":null

PR #861, opened 2018-11-28
Assign device name from an event to each reading when using Redis

TomFlem:
Issue #860


andresrinivasan:
Looks good @tsconn23 

Thanks for the fix @TomFlem

Issue #855, opened 2018-11-26
Core-metadata applies default values to data transformation fields


iain-anderson:
When creating a device profile in core-metadata, if a deviceResource does not have base, scale, offset etc defined, the following values are autofilled:
```
"mask":"0x00",
"shift":"0",
"scale":"1.0",
"offset":"0.0",
"base":"0"
```
This prevents enforcement of the rule that integer type readings should not be processed with float-type transforms, and complicates the processing of readings where no transforms are needed.
These fields should instead be left null.

iain-anderson:
Note: behavior above is observed when using the /api/v1/deviceprofile/uploadfile endpoint

tsconn23:
Documenting how to duplicate. The screenshot shows how to upload the [sample device profile](https://github.com/edgexfoundry/device-virtual/blob/master/modbus_sample_profiles/GS1-10P5.profile.yaml) through Postman. 
<img width="1085" alt="screen shot 2018-11-28 at 2 02 19 pm" src="https://user-images.githubusercontent.com/1342980/49179224-7525a980-f317-11e8-896e-0c78b086caa7.png">


Once the file has been uploaded, then you can query the mongo database and see the following as part of the device profile JSON
```
{ "description" : "the frequency of collection", 
  "name" : "collectionFrequency", 
  "tag" : "", "properties" : 
    { "value" : 
      { "type" : "Integer", 
        "readWrite" : "W", 
        "minimum" : "0", 
        "maximum" : "600", 
        "defaultValue" : "15", 
        "size" : "4", 
        "word" : "2", 
        "lsb" : "", 
        "mask" : "0x00", 
        "shift" : "0", 
        "scale" : "1.0", 
        "offset" : "0.0", 
        "base" : "0", 
        "assertion" : "", 
        "signed" : true, 
        "precision" : "3" 
      }
``` 
None of the properties Iain called out above are in the device profile document, but they are found in the JSON. This appears to be happening as a result of this function:
https://github.com/edgexfoundry/edgex-go/blob/c1176c36717ef6300818888aac69e5f0ae0abaf2/pkg/models/propertyvalue.go#L146

jpwhitemn:
Is this closed/fixed with #867 ?

PR #867, opened 2018-11-29
Don't apply default values to data transformation fields (core-metadata)

jdharms:
#855 

* Removes default values for data transformation fields when creating a device profile
* Marks json properties as omitempty in the PropertyValue `MarshalJSON()` function.


tsconn23:
LGTM. @iain-anderson I'll give you the final word.

iain-anderson:
LGTM, thanks

Issue #857, opened 2018-11-27
edgex-docker-launch.sh shebang points to (SHELL) but script has BASH syntax


daviur:
Trying to build my own docker containers while following the README.md I got an error in the "sudo make run_docker". It appears to be because the shebang of edgex-docker-launch.sh points to SHELL (#!/bin/sh) but the script has BASH syntax on the if statements with "[[". Simply changing the shebang to "#!/usr/bin/env bash" fixed the issue.

anonymouse64:
On Ubuntu, `/bin/sh` is a symlink to `dash` which is traditional posix `sh`
implementation and not bash.

andresrinivasan:
Ah. So do we prefer a Posix script or can I create a dependency on bash?

anonymouse64:
A dependency on bash is fine, a number of the snap specific scripts all use bash instead of posix sh.

SteveOss:
As a general comment, unless there is a good reason not to, would always write sh scripts, as this maximises portability across Unix variants.

anonymouse64:
Sure, I think this bug is about the fact that we have a `bash` script that declares itself as using `sh`.

SteveOss:
Or, it could be a bug about a sh script that incorrectly uses bash syntax. Depends on how you look at it :)

andresrinivasan:
I'm will update the script to be POSIX compliant for portability per @SteveOss.

andresrinivasan:
Though @daviur I'm still concerned that you're invoking `make run_docker` as root. Is there a requirement on your side for root?

daviur:
I'm running Ubuntu 18.04, and I tried both with root and non-root user with docker group. Same problem. 
Effectively, changing the script, SH did not handle "if [[ ]];" syntax. but it did handle "if [ ];". Also, reading about SH syntax it appears to be expected.

daviur:
To clarify, my first intention was to run "make run_docker" as a non-root user with the docker group appended.

PR #868, opened 2018-11-30
Issue #857 Update edgex-docker-launch.sh for POSIX shell

andresrinivasan:
Re Issue #857 

Removed naughty bash dependencies.


jpwhitemn:
Andre - please submit and issue and let's work through the community.  We haven't seen the problem, but if there is one let's work it and get reaction.  This has been sitting out here for a while now and I don't want it to stay open on an out-of-date branch.

PR #875, opened 2018-12-03
Fixed invalid path in default schedule event

### LINKED ISSUE #873: Missing leading slash in default schedule path
This is a about [this](http://edgex-core-data:48080api/v1/event/scrub) line in the scheduler docker config file.
The support-scheduler generates schedules from its config.
Currently:
```
 Path = 'api/v1/event/scrub'
```
This results in the following addressable path:
```
http://edgex-core-data:48080api/v1/event/scrub
```
Where clearly the slash after the port is missing.

Edit: Consul is disabled

jduranf:
Fix #873

### DIFF
--- a/cmd/support-scheduler/res/docker/configuration.toml
+++ b/cmd/support-scheduler/res/docker/configuration.toml
@@ -45,7 +45,7 @@ File = '/edgex/logs/edgex-support-scheduler.log'
     Protocol = 'http'
     Method = 'DELETE'
     Service = 'core-data'
-    Path = 'api/v1/event/scrub'
+    Path = '/api/v1/event/scrub'
     Schedule = 'midnight'
     Scheduler = 'support-scheduler'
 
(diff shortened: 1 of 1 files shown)


cmd/support-scheduler/res/docker/configuration.toml
```diff
@@ -59,9 +59,3 @@ File = '/edgex/logs/edgex-support-scheduler.log'
     Path = 'api/v1/event/removeold/age/604800000'
```
nfode: Just realized its the same for this line
```suggestion
    Path = '/api/v1/event/removeold/age/604800000'
```

jduranf: You are right, fixed too

tsconn23:
recheck

tsconn23:
@jduranf Do you have time to duplicate this PR targeting the `delhi` branch? Would like to have this in the 0.7.1 release.

cmd/support-scheduler/res/docker/configuration.toml
xmlviking: Is this fixed now?
If so having this in the Delhi would be fantastic

jduranf:
@tsconn23 created #886 for Delhi

Issue #706, opened 2018-10-18
Enhancement: Use a map[string]general.GeneralClient patterns for extensibility benefits.


akramtexas:
- This enhancement is directly related to, and a follow-up, on: [PR #701](https://github.com/edgexfoundry/edgex-go/pull/701)

PR #887, opened 2018-12-04
Use a map of general clients initialized in a loop instead of discrete variables.

brandonforster:
Issue #706

### DIFF
--- a/internal/system/agent/init.go
+++ b/internal/system/agent/init.go
@@ -32,14 +32,7 @@ var Configuration *ConfigurationStruct
 var LoggingClient logger.LoggingClient
 var Conf = &ConfigurationStruct{}
 var ec interfaces.ExecutorClient
-var gccc general.GeneralClient
-var gccd general.GeneralClient
-var gccm general.GeneralClient
-var gcec general.GeneralClient
-var gced general.GeneralClient
-var gcsl general.GeneralClient
-var gcsn general.GeneralClient
-var gcss general.GeneralClient
+var clients map[string]general.GeneralClient
 
 func Retry(useConsul bool, useProfile string, timeout int, wait *sync.WaitGroup, ch chan error) {
 	until := time.Now().Add(time.Millisecond * time.Duration(timeout))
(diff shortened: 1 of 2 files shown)


internal/system/agent/services.go
```diff
@@ -292,100 +93,33 @@ func getMetrics(services []string) (MetricsRespMap, error) {
...
+		} else {
+			m.Metrics[service] = ProcessResponse(responseJSON)
 		}
 	}
 	return m, nil
 }
+
+func isKnownServiceKey(serviceKey string) bool {
+	switch serviceKey {
+	case internal.SupportNotificationsServiceKey:
```
akramtexas: - At first glance, I was looking for the `fallthrough` keyword, which is the directive to fall through to a subsequent case... Maybe it's not required, let's confirm.

brandonforster: Yes, this version of the code doesn't work because of what you described. Please see the latest changes.

akramtexas: OK, gotcha. I will pull the latest changes.

internal/system/agent/init.go
```diff
@@ -43,6 +43,7 @@ var services = map[string]string{
 	internal.ExportDistroServiceKey: "Distro",
 	internal.SupportLoggingServiceKey: "Logging",
 	internal.SupportSchedulerServiceKey: "Scheduler",
+	internal.ConfigSeedServiceKey: "Config",
```
akramtexas: - The `config-seed` service had intentionally been left out (as another service with which the SMA could also interact), the idea being that `config-seed` service is ephemeral, being launched at the outset to seed `consul` and shortly thereafter tearing itself down, thus thereafter remaining unavailable.
- While I'm not opposed to adding the `config-seed` service back, as I noticed you have per this PR, let's discuss as I want to hear your thoughts on the merits of doing so.

akramtexas:
- Let's chat about the process whereby the returned bool variable `exists` is set and returned according to the following function in the SMA's `services.go` source file:
```
func isKnownServiceKey(serviceKey string) bool {
	 _, **exists** := services[serviceKey]

	 return **exists**
}

```

akramtexas:
> I had remotely added your branch to my git edgex-go repo locally, and tested the updates per this PR using Postman (with the same `SMA Testing` config that I had passed on to you yesterday). Not seeing any issues during testing the `api/v1/config/` and api/v1/metrics/ endpoint-based operations.

internal/system/agent/init.go
brandonforster: I added it because the Docker files were looking for it and it didn't seem to do any harm in init. I'll remove it tomorrow.

brandonforster:
> * Let's chat about the process whereby the returned bool variable `exists` is set and returned according to the following function in the SMA's `services.go` source file:
> 
> ```
> func isKnownServiceKey(serviceKey string) bool {
> 	 _, exists := services[serviceKey]
> 
> 	 return exists
> }
> ```

From [the Golang blog](https://blog.golang.org/go-maps-in-action)

>A two-value assignment tests for the existence of a key:

>`i, ok := m["route"]`
>In this statement, the first value (i) is assigned the value stored under the key "route". If that key >doesn't exist, i is the value type's zero value (0). The second value (ok) is a bool that is true if the key >exists in the map, and false if not.

>To test for a key without retrieving the value, use an underscore in place of the first value.

akramtexas:
> > * Let's chat about the process whereby the returned bool variable `exists` is set and returned according to the following function in the SMA's `services.go` source file:
> > 
> > ```
> > func isKnownServiceKey(serviceKey string) bool {
> > 	 _, exists := services[serviceKey]
> > 
> > 	 return exists
> > }
> > ```
> 
> From [the Golang blog](https://blog.golang.org/go-maps-in-action)
> 
> > A two-value assignment tests for the existence of a key:
> 
> > `i, ok := m["route"]`
> > In this statement, the first value (i) is assigned the value stored under the key "route". If that key doesn't exist, i is the value type's zero value (0). The second value (ok) is a bool that is true if the key exists in the map, and false if not.
> 
> > To test for a key without retrieving the value, use an underscore in place of the first value.

Yep, that works, makes sense 👍

internal/system/agent/init.go
akramtexas: - The more thought that I gave to `config-seed` service having been intentionally been left out (as another service with which the SMA could also interact), the more I'm convinced that is should _remain_ left out (as another service with which the SMA could also interact). 
- Note that `config-seed` is not a _regular_ service and could be done by some other system/bootstrap capability in the future. 
- Yes, please remove it 👍

akramtexas:
@brandonforster Make sure to squash your commits in the end, _before_ we merge this PR into `master`.

akramtexas:
> @brandonforster Make sure to squash your commits in the end, _before_ we merge this PR into `master`.
@brandonforster I notice that you've squashed the commits. I'm assuming that the intended updates, for example the removal of the following line, etc., are _still_ to be made:
`internal.ConfigSeedServiceKey: "Config",`
- Please ping me via a note here once the PR is ready for review again, thanks.

brandonforster:
It's ready.

Issue #322, opened 2018-06-23
Add a test for addressable client under core/clients/metadata


yanghua:
cc @tsconn23

jpwhitemn:
Not the only place where more testing is needed, but point taken.  @tsconn23 and team are working on more unit and black box testing for core and supporting services.  Probably not going to happen until Edinburgh or later unless volunteer resources come forth to help.

PR #890, opened 2018-12-05
#322 Small refactoring to core-metadata; unit tests

jdharms:
#322 

Minor refactoring to a few functions in core-metadata's routing
layer to introduce a layer to contain business logic that will
be more easily testable.  Also add some unit tests covering
some of these functions.


internal/core/metadata/rest_addressable.go
```diff
@@ -51,31 +53,29 @@ func restAddAddressable(w http.ResponseWriter, r *http.Request) {
...
 	}
-	var id bson.ObjectId
-	id, err = dbClient.AddAddressable(&a)
+
+	id, err := addAddressable(a)
 	if err != nil {
-		if err == db.ErrNotUnique {
+		switch err.(type) {
+		case types.ErrDuplicateAddressableName:
 			http.Error(w, "Duplicate name for addressable", http.StatusConflict)
```
jduranf: The string can be replaced by `err.Error()`. This should be done also in the other case.

jdharms: Thank you!

Issue #892, opened 2018-12-05
Remove unused functions/behavior in support-notifications


jdharms:

jdharms:
Resolved with #893

PR #893, opened 2018-12-05
Small refactoring to support-notifications

jdharms:
#892


xmlviking:
The removal of those constants was that in regards to the removal of the unused functions?  As well as the notifications pkg reference?

jdharms:
@xmlviking No, the constants just aren't used (at the moment).  If they are important to keep around let me know and I'll revert those changes.

PR #894, opened 2018-12-05
Add new device services

tonyespy:
This change adds the following new device services to the snap:

 - device-modbus-go
 - device-mqtt-go
 - device-random

They are all disabled by default, and therefore must be enabled using:

'snap set edgexfoundry `service name`=on'

Enabling device-random will trigger a trio of random signed integer readings to be pushed to Core Data every 5 seconds. 

Enabling device-modbus or device-mqtt will start the respective device service, however no readings will be generated automatically as there are no device profiles configured by default. In order to produce readings from either service, device profiles must be copied to service's configuration directory ($SNAP_DATA/config/service name/res) before starting.

This PR has been tested on Ubuntu 18.04 desktop and on a Dell Gateway 3000 running Ubuntu Core 16.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -150,7 +150,30 @@ apps:
       after:
         - core-config-seed
         - mongo-worker
-
+  device-modbus:
+    command: bin/go-services-wrapper.sh device-modbus
+    daemon: simple
+    plugs: [network, network-bind]
+    passthrough:
+      after:
+        - core-config-seed
+        - mongo-worker
+  device-mqtt:
+    command: bin/go-services-wrapper.sh device-mqtt
+    daemon: simple
+    plugs: [network, network-bind]
+    passthrough:
+      after:
+        - core-config-seed
+        - mongo-worker
+  device-random:
+    command: bin/go-services-wrapper.sh device-random
+    daemon: simple
+    plugs: [network, network-bind]
+    passthrough:
+      after:
+        - core-config-seed
+        - mongo-worker
 
   # helper commands the snap exposes
   mongo:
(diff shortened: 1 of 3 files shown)


snap/snapcraft.yaml
```diff
@@ -150,7 +150,30 @@ apps:
       after:
         - core-config-seed
         - mongo-worker
-
+  device-modbus:
+    command: bin/go-services-wrapper.sh device-modbus
+    daemon: simple
+    plugs: [network, network-bind]
```
anonymouse64: I assume at some point this will change, but for now is it expected that modbus can only talk to modbus devices over TCP? I think so but just want affirmation...

snap/snapcraft.yaml
```diff
@@ -835,3 +858,100 @@ parts:
         -e "s:host = \"edgex-support-rulesengine\":host = \"localhost\":" \
         -e "s:host = \"edgex-device-virtual\":host = \"localhost\":" \
         $SNAPCRAFT_PART_INSTALL/config/security-api-gateway/res/configuration.toml
+  device-modbus:
+    source: https://github.com/edgexfoundry/device-modbus-go.git
+    source-depth: 1
+    #source-branch: dehli
```
anonymouse64: I see that this, `device-mqtt-go` and `device-random` all only have a single master branch and no tags. As such I would rather see all these pin to specific to the SHA's  that those repos' master is currently on, otherwise as daily builds continue after the stable release more recent code will automatically for these services will be pulled into `delhi/edge` releases, and if those repos ever make breaking changes `delhi/edge` will become automatically broken at some point...

snap/snapcraft.yaml
```diff
@@ -835,3 +858,100 @@ parts:
...
+        "$SNAPCRAFT_PART_INSTALL/config/device-random/res/configuration.toml"
+
+      install -T "./cmd/res/device.random.yaml" \
+        "$SNAPCRAFT_PART_INSTALL/config/device-random/res/device.random.yaml"
+
+      # FIXME: no Attribution.txt for 3rd party licenses; probably needs to be
+      # an aggregation of SDK's and this device service...
+
+      install -DT "./LICENSE-2.0.txt" \
+         "$SNAPCRAFT_PART_INSTALL/usr/share/doc/device-random/LICENSE-2.0.txt"
```
anonymouse64: So close to 1000 line snapcraft.yaml!!! We should just add another 43 lines of comments or whitespace :tada:

snap/snapcraft.yaml
anonymouse64: Actually nevermind, we talked about this and we don't want to tag just yet

Issue #848, opened 2018-11-21
When sending post command with a body which has an invalid formatting string to /valuedescriptor of Core Data API, 409 should be returned


FelixTing:
OS version: ubuntu 16.04
EdgeX version: 1.0.0
Environment details: Docker container
Steps to reproduce:

1. start up EdgeX
2. send post command with a body which has an invalid formatting string /valuedescriptor of Core Data API
for example, with the following request URL and body:
```
Method: POST
URL: http://localhost:48080/api/v1/valuedescriptor
Body:
  {
  "name":"co2",
  "min":"12",
  "max":"15",
  "type":"F",
  "uomLabel":"degreecel",
  "defaultValue":"0",
  "formatting":"%",
  "labels":["NHCO2","hvac"]
  }
```

Actual results: Got returned http status code 400 BadRequest
Expected results: Got returned http status code 409

RAML: https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-data.raml#L657

PR #897, opened 2018-12-06
#848 POST to /valuedescriptor with duplicate name or invalid fmt stri…

jdharms:
…ng should 409

#848

Return a 409 status code and informative error when either an in-use name
or an improperly formatted formatting string is passed to /valuedescriptor's
POST handler.

### DIFF
--- a/internal/core/data/valuedescriptor.go
+++ b/internal/core/data/valuedescriptor.go
@@ -15,6 +15,7 @@ package data
 
 import (
 	"encoding/json"
+	"fmt"
 	"io"
 	"regexp"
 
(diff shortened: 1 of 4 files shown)


internal/core/data/valuedescriptor.go
```diff
@@ -200,7 +201,8 @@ func decodeValueDescriptor(reader io.ReadCloser) (vd contract.ValueDescriptor, e
 		return contract.ValueDescriptor{}, err
 	}
 	if !match {
```
tsconn23: It seems like this check and the handling for it should be in the validateFormatString() function itself. In which case, the function would just return an error of type `ErrValueDescriptorInvalid`. That would get logged and returned.

jdharms: Thanks, that makes a lot of sense.  I think I've addressed this now.

tsconn23:
@jdharms Can you update the branch via rebase and squash commits plz?

internal/core/data/valuedescriptor.go
```diff
@@ -194,14 +207,9 @@ func decodeValueDescriptor(reader io.ReadCloser) (vd contract.ValueDescriptor, e
...
-	match, err := validateFormatString(v)
+	err = validateFormatString(v)
 	if err != nil {
-		LoggingClient.Error("Error checking for format string for value descriptor " + v.Name)
-		return contract.ValueDescriptor{}, err
-	}
-	if !match {
-		LoggingClient.Error("Error posting value descriptor. Format is not a valid printf format.")
-		return contract.ValueDescriptor{}, errors.NewErrValueDescriptorInvalid(v.Name, err)
+		return contract.ValueDescriptor{}, nil
```
tsconn23: There's a bug here. Notice that the error is being dropped because instead of returning `err` we're returning `nil`.

Issue #901, opened 2018-12-07
Add Consul/config-seed flag to not repopulate


michaelestrin:
Ensure second and subsequent execution of config-seed does not overwrite preexisting /config/v2/* values.

michaelestrin:
#827

PR #902, opened 2018-12-07
Ensure second and subsequent execution of config-seed does not overwr…

michaelestrin:
…ite preexisting /config/v2/* values.

### DIFF
--- a/internal/seed/config/populate.go
+++ b/internal/seed/config/populate.go
@@ -67,7 +67,7 @@ func ImportProperties(root string) error {
 
 // Import configuration files using the specified path to the cmd directory where service configuration files reside.
 // Also, profile indicates the preferred deployment target (such as "docker")
-func ImportConfiguration(root string, profile string) error {
+func ImportConfiguration(root string, profile string, overwrite bool) error {
 	dirs := listDirectories()
 	absRoot, err := determineAbsRoot(root)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


michaelestrin:
#901

michaelestrin:
@anonymouse64 - updated to add command line switch to support optional overwrite of existing configuration values in Consul.

internal/seed/config/populate.go
```diff
@@ -120,8 +120,10 @@ func ImportConfiguration(root string, profile string) error {
 			// Put config properties to Consul K/V store.
 			for _, v := range kvs {
 				p := &consulapi.KVPair{Key: prefix + v.Key, Value: []byte(v.Value)}
-				if _, err := (*consulapi.KV).Put(Registry.KV(), p, nil); err != nil {
-					return err
+				if kvPair, _, _ := (*consulapi.KV).Get(Registry.KV(), p.Key, nil); overwrite || kvPair == nil {
+					if _, err := (*consulapi.KV).Put(Registry.KV(), p, nil); err != nil {
+						return err
+					}
```
tsconn23: Tested locally and this certainly works. However for every key in a given service's config it does a GET to check if the key exists in Consul and then a PUT if it doesn't or `overwrite` is true. I had originally thought we would check Consul at the level of the service key itself. If the service key is present, then we skip the whole thing. Like so:

```
// Put config properties to Consul K/V store.
prefix := internal.ConfigV2Stem + internal.ServiceKeyPrefix + d

options := &consulapi.QueryOptions{}

if kvPairs, _, _ := Registry.KV().List(prefix, options); overwrite || kvPairs == nil {
    // Put config properties to Consul K/V store.
    for _, v := range kvs {
        p := &consulapi.KVPair{Key: prefix + "/" + v.Key, Value: []byte(v.Value)}
        if _, err := (*consulapi.KV).Put(Registry.KV(), p, nil); err != nil {
            return err
        }
    }
}
```
Note that I also moved the slash concatenation if you decide to adopt this approach.

michaelestrin: @tsconn23 

Code as implemented allows for subsequent creation of additional properties (added in future releases) without modifying the preexisting properties.  Performance of doing multiple GET calls wasn't considered since config-seed is typically run once to initialize the environment.

Do you still want me to make your suggested changes in light of the explanation above?

tsconn23: @michaelestrin OK we can leave it as is for now. I still don't like all of the GETs but the additive modification of the configuration keys is what we talked about during last week's Core WG call (for point release versioning). If we check at the service key level, there's no way to do this. 

LGTM then.

tsconn23:
@anonymouse64 Yes, that's correct. That's what I used to test it.

Issue #845, opened 2018-11-20
When sending get/put command with an inexistent id to /device/{id}/command/{commandId} of Core Command API, 404 should be returned


FelixTing:
OS version: ubuntu 16.04
EdgeX version: 1.0.0
Environment details: Docker container
Steps to reproduce:

1. start up EdgeX
2. send get/put command with an inexistent device id to /device/{id}/command/{commandId} of Core Command API
for example:
```
Method: GET
URL: http://localhost:48082/api/v1/device/57f7937ae4AAAAe2d67c20b5/command/57f5e11fe4b060e2071a5c00

Method: PUT 
URL: http://localhost:48082/api/v1/device/57f7937ae4AAAAe2d67c20b5/command/57f5e11fe4b060e2071a5c00
Body: {"temp":74}
```

Actual results: Got returned http status code 500 InternalServerError
Expected results: Got returned http status code 404 NotFound

RAML: 

- https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-command.raml#L75
- https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-command.raml#L59

brandonforster:
Fixed by PR#919

PR #917, opened 2018-12-10
#845 Return a status 404 for core-command calls with bad device ID

jdharms:
#845 

Possibly a regression from commit: https://github.com/edgexfoundry/edgex-go/commit/40337b3b7d1dcbb47afc1e140e71d89914688f67#diff-2311174e4ebc3d4be67d0cf62dc5e1d2R72

It appears that other code (possibly just test code) depends on this case being handled by an ErrNotFound instead of an ErrServiceClient, so I've just updated the caller to handle this error appropriately.


jdharms:
Closing to rework.

Issue #844, opened 2018-11-20
When using Core Command API to update the device opState/adminState with an inexistent id(or name) , 404 should be returned


FelixTing:
OS version: ubuntu 16.04
EdgeX version: 1.0.0
Environment details: Docker container
Steps to reproduce:

1. start up EdgeX
2. send put command with an inexistent id(or name) to /device/{id}/adminstate/{adminState} of Core Command API
for example, with the following request URL:
`Method: PUT`
`URL: http://localhost:48082/api/v1/device/57f7937ae4AAAAe2d67c20b5/adminstate/LOCKED`

Actual results: Got returned http status code 500 InternalServerError
Expected results: Got returned http status code 404 NotFound

RAML: 

- https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-command.raml#L100
- https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-command.raml#L123
- https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-command.raml#L168
- https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-command.raml#L191

brandonforster:
Fixed by PR#919

PR #919, opened 2018-12-10
#845, #844 Fix error handling in core-command service

jdharms:
#844, #845

I believe this PR fixes some regressions caused by (or related to) commit: https://github.com/edgexfoundry/edgex-go/commit/40337b3b7d1dcbb47afc1e140e71d89914688f67#diff-2311174e4ebc3d4be67d0cf62dc5e1d2R72 around error handling in the core-command service.

Error codes in this service were being passed up to the rest layer alongside golang errors, and then discarded and 500's written out anyway.  This change causes the rest layer to respect the status codes returned by these functions even when an err value is returned.

### DIFF
--- a/internal/core/command/restDevice.go
+++ b/internal/core/command/restDevice.go
@@ -56,7 +56,6 @@ func restPutDeviceAdminState(w http.ResponseWriter, r *http.Request) {
 	status, err := putDeviceAdminState(did, as)
 	if err != nil {
 		LoggingClient.Error(err.Error(), "")
-		w.WriteHeader(http.StatusInternalServerError)
 	}
 	w.WriteHeader(status)
 
@@ -69,7 +68,6 @@ func restPutDeviceAdminStateByDeviceName(w http.ResponseWriter, r *http.Request)
 	status, err := putDeviceAdminStateByName(dn, as)
 	if err != nil {
 		LoggingClient.Error(err.Error(), "")
-		w.WriteHeader(http.StatusInternalServerError)
 	}
 	w.WriteHeader(status)
 }
@@ -81,7 +79,6 @@ func restPutDeviceOpState(w http.ResponseWriter, r *http.Request) {
 	status, err := putDeviceOpState(did, os)
 	if err != nil {
 		LoggingClient.Error(err.Error(), "")
-		w.WriteHeader(http.StatusInternalServerError)
 	}
 	w.WriteHeader(status)
 }
@@ -93,7 +90,6 @@ func restPutDeviceOpStateByDeviceName(w http.ResponseWriter, r *http.Request) {
 	status, err := putDeviceOpStateByName(dn, os)
 	if err != nil {
 		LoggingClient.Error(err.Error(), "")
-		w.WriteHeader(http.StatusInternalServerError)
 	}
 	w.WriteHeader(status)
 }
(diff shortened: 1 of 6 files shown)


pkg/clients/request.go
```diff
@@ -69,9 +69,7 @@ func GetRequest(url string) ([]byte, error) {
 		return nil, err
 	}
 
-	if resp.StatusCode == http.StatusNotFound {
-		return nil, types.ErrNotFound{}
-	} else if (resp.StatusCode != http.StatusOK) && (resp.StatusCode != http.StatusAccepted) {
+	if (resp.StatusCode != http.StatusOK) && (resp.StatusCode != http.StatusAccepted) {
 		return nil, types.NewErrServiceClient(resp.StatusCode, bodyBytes)
 	}
```
tsconn23: The refactored request.go looks correct. The intent of the ErrServiceClient is to allow a service client to return an error when calling its target service. The error type allows the http status code and any message returned from the failed call to be used by upstream logging (by conforming to the Error interface). Therefore no special dispensation as applies to http.StatusNotFound should be necessary.

jduranf: @jdharms your proposed fix can have side effects, because in some cases it's expected a `types.ErrNotFound`. Like here:
https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/data/router.go#L339
or here:
https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/data/valuedescriptor.go#L161

jdharms: @jduranf Thanks, I will look into addressing these side effects and will update when finished.

jdharms: @tsconn23 @jduranf I've updated my changes to (hopefully) fix all other places where this error value is relied upon.  I've tested with the blackbox tests and unit tests.

internal/core/command/restDevice.go
```diff
@@ -132,7 +128,7 @@ func restGetAllCommands(w http.ResponseWriter, _ *http.Request) {
 	status, devices, err := getCommands()
 	if err != nil {
 		LoggingClient.Error(err.Error(), "")
-		w.WriteHeader(http.StatusInternalServerError)
+		w.WriteHeader(status)
 	} else if status != http.StatusOK {
```
jduranf: This else can be removed, and just do something like this:
```go
if err != nil {
}
w.WriteHeader(status)
```

internal/core/data/device_test.go
```diff
@@ -108,7 +108,7 @@ func newMockDeviceClient() *mocks.DeviceClient {
 	})).Return(mockDeviceResultFn, nil)
 	client.On("Device", mock.MatchedBy(func(id string) bool {
 		return id == "404"
-	})).Return(mockDeviceResultFn, types.ErrNotFound{})
+	})).Return(mockDeviceResultFn, types.NewErrServiceClient(404, []byte{}))
```
jduranf: Please, replace all 404 by `http.StatusNotFound`

internal/core/command/restDevice.go
jdharms: I address this in PR #933 

We actually don't want to fall through in the case where `status != http.StatusOK` because we don't want to set a content type.

internal/core/data/device_test.go
jdharms: Thank you

brandonforster:
Rebase and I'll merge.

jdharms:
recheck

Issue #904, opened 2018-12-07
Remove Support for Labels from Logging


tsconn23:
As discussed on the Core Working Group call (6-Dec-18) we will be removing functionality supporting "labels" from log entries. This is because there is nowhere in the current Go-based services that this field is used. Therefore the logic and APIs to support its use are just cruft.

It appears from preliminary review that changes will be contained to the support-logging service.

- Remove APIs which take labels as parameters
- Remove any persistence logic for labels
- Remove any parsing for queries or marshaling/unmarshaling dealing with labels.

Related blackbox test issue https://github.com/edgexfoundry/blackbox-testing/issues/112

jduranf:
If it's fine, I will work on this issue

PR #925, opened 2018-12-11
Update SMA logging to use structured logging library and style.

brandonforster:
Issue #780


internal/system/agent/executor/compose.go
```diff
@@ -20,19 +31,19 @@ func WasDockerContainerComposeStarted(service string) bool {
 	cmdName := "docker"
 	cmdArgs := []string{"ps"}
 	if cmdOut, err = exec.Command(cmdName, cmdArgs...).Output(); err != nil {
-		fmt.Fprintln(os.Stderr, "There was an error running the docker-compose command: ", err.Error())
+		logs.LoggingClient.Error("error running the docker-compose command", "error message", err.Error())
```
jduranf: Labels will be removed from the logging system (issue #904), so please don't use it.

akramtexas: This looks right to me in that you are aligned with the rework of logging output using GoKit as shown in the Core WG call 1-Nov-18 👍

Issue #834, opened 2018-11-20
SMA start/stop/restart refactor (Edinburgh Roadmap Item)


jpwhitemn:
The SMA start/stop/restart capabilities will be refactored to better address Docker environments as well as generic executable calls.

akramtexas:
As part of **POC C**, the following **Dockerfile** has been created:
`FROM docker:dind

RUN apk add --no-cache py-pip openssl bash
RUN apk --no-cache add curl
RUN pip install docker-compose
RUN curl -o docker-compose.yml https://raw.githubusercontent.com/edgexfoundry/developer-scripts/master/compose-files/docker-compose-delhi-0.7.0.yml

FROM docker:dind AS builder

RUN make cmd/sys-mgmt-agent/sys-mgmt-agent

FROM docker:dind 

COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/sys-mgmt-agent/
COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/res/docker/configuration.toml /res/docker/configuration.toml

ENTRYPOINT ["/bin/bash", "-c"]
`

akramtexas:
Unexpectedly, though, running into the following issue at the moment:
> docker build .
> Sending build context to Docker daemon  16.38kB
> Error response from daemon: Dockerfile parse error line 14: COPY requires at least two arguments, but only one was provided. Destination could not be determined.

akramtexas:
- `Dockerfile` for **POC A (Completed)** is as follows:
```
RUN apk add --no-cache py-pip openssl bash
RUN pip install docker-compose
RUN apk update \
RUN apk install curl \
    curl "https://github.com/docker/compose/releases/download/1.23.1/docker-compose-$(uname -s)-$(uname -m)" /usr/bin/docker-compose \
    chmod +x /usr/bin/docker-compose
ENTRYPOINT ["/bin/bash", "-c"]
```
- And the `Dockerfile` for **POC B (Completed)** is as follows:
```
FROM docker:dind
RUN apk add --no-cache py-pip openssl bash
RUN apk --no-cache add curl
RUN pip install docker-compose
RUN curl -o docker-compose.yml https://raw.githubusercontent.com/edgexfoundry/developer-scripts/master/compose-files/docker-compose-delhi-0.7.0.yml
ENTRYPOINT ["/bin/bash", "-c"]
```

[...]

PR #926, opened 2018-12-11
Add Dockerfile.sys-mgmt-agent (and related updates) as part of POC C …

JPWKU:
- The work done in this PR is per [Issue #834](https://github.com/edgexfoundry/edgex-go/issues/834), which is for **the SMA start/stop/restart refactor** (Edinburgh Roadmap Item).
- Assistance from @jpwhitemn and @JPWKU has been indispensable, and is greatly appreciated 👍 
- The SMA (via the `DinD` container hosting it) can now start, stop, restart EdgeX services, responding to REST calls.
- We don’t want the SMA to be responsible for the *creation* of containers; only for *starting, stopping, and restarting* services in containers. Some research confirmed that running `docker` (instead of `docker-compose` fits the bill perfectly.)
- In other words, the SMA needs to be able to start and stop the EdgeX containers (more precisely the _services_, because we have the _inside-DinD_ and _outside-of-docker_ scenarios) 
- The _preceding_ is the latest update incorporated in the work done in this PR.
- _Other_ salient aspects are covered by notes (aka comments) below.
- Postman was used for testing.
- Here is a `sample payload` corresponding to the *start* operation during use of `http://localhost:48090/api/v1/operation` in Postman:
```
{
   "action":"start",
   "services":[
   	  "edgex-support-logging",
   	  "edgex-core-metadata",
      "edgex-support-notifications",
      "edgex-core-data",
      "edgex-core-command",
      "edgex-export-client",
      "edgex-export-distro"
   ],
   "params":[
   	"graceful"
   	]
}
```
- During development and especially during testing, these commands helped a lot, and for which I'm grateful to both @jpwhitemn and @JPWKU:

1. `docker run -d -p 127.0.0.1:48090:48090 -v /var/run/docker.sock:/var/run/docker.sock IMAGEID`
2. `docker exec -it CONTAINERID /bin/sh`

Signed-off-by: Akram Ahmad <sftwr2020@gmail.com>


docker/Dockerfile.sys-mgmt-agent
```diff
@@ -0,0 +1,30 @@
...
+# So we can try these.
+RUN echo http://nl.alpinelinux.org/alpine/v3.7/main > /etc/apk/repositories; \
+    echo http://nl.alpinelinux.org/alpine/v3.7/community >> /etc/apk/repositories
+
+RUN apk update && apk add make
+COPY . .
+RUN make cmd/sys-mgmt-agent/sys-mgmt-agent
+
+#Next image - Copy built Go binary into new workspace
+FROM alpine:3.7
```
JPWKU: FROM docker:dind

docker/Dockerfile.sys-mgmt-agent
```diff
@@ -0,0 +1,30 @@
...
+
+LABEL license='SPDX-License-Identifier: Apache-2.0' \
+      copyright='Copyright (c) 2018: Dell, Cavium'
+
+COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/sys-mgmt-agent /
+COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/res/configuration.toml /res/configuration-docker.toml
+
```
JPWKU: Add in bits to install docker-compose and download docker-compose file.

docker/Dockerfile.sys-mgmt-agent
```diff
@@ -0,0 +1,30 @@
...
+FROM golang:1.11.2-alpine3.7 AS builder
+WORKDIR /go/src/github.com/edgexfoundry/edgex-go
+
+# The main mirrors are giving us timeout issues on builds periodically.
+# So we can try these.
+RUN echo http://nl.alpinelinux.org/alpine/v3.7/main > /etc/apk/repositories; \
+    echo http://nl.alpinelinux.org/alpine/v3.7/community >> /etc/apk/repositories
+
+RUN apk update && apk add make
```
JPWKU: also add `apk add glide` and `apk add git`

docker/Dockerfile.sys-mgmt-agent
```diff
@@ -0,0 +1,30 @@
...
+COPY . .
```
JPWKU: After this line add `RUN make prepare`

akramtexas:
@jpwhitemn Getting feedback from the local environment and / or `Docker` itself has been trickier than anticipated. Basically, verifying whether this worked or not: `COPY . .` Tried several variants, including some online wisdom on [How can I inspect the file system of a failed docker build?](https://stackoverflow.com/questions/26220957/how-can-i-inspect-the-file-system-of-a-failed-docker-build)

docker/Dockerfile.sys-mgmt-agent
akramtexas: Done.

docker/Dockerfile.sys-mgmt-agent
akramtexas: Done.

docker/Dockerfile.sys-mgmt-agent
akramtexas: Done.

docker/Dockerfile.sys-mgmt-agent
akramtexas: Done.

docker/Dockerfile.sys-mgmt-agent
```diff
@@ -0,0 +1,41 @@
...
+RUN echo http://nl.alpinelinux.org/alpine/v3.7/main > /etc/apk/repositories; \
+    echo http://nl.alpinelinux.org/alpine/v3.7/community >> /etc/apk/repositories
+
+RUN apk update && apk add make glide git
+COPY . .
+
+RUN make prepare
+RUN make cmd/sys-mgmt-agent/sys-mgmt-agent
+
+#Next image - Copy built Go binary into new workspace
```
JPWKU: You can delete lines 23-29, they are not needed.

docker/Dockerfile.sys-mgmt-agent
```diff
@@ -0,0 +1,41 @@
...
+
+# Layer-in the Docker-in-Docker image now...
+FROM docker:dind
+
+COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/sys-mgmt-agent /
+COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/res/configuration.toml /res/configuration-docker.toml
+
+RUN apk --no-cache add curl
+# Do we (still) need the following directive (for adding 'make')?
+# RUN apk --no-cache add make
```
JPWKU: Uncomment and change this to `RUN apk --no-cache add py-pip`.  You will not need make anymore here I think.

JPWKU:
> @jpwhitemn Getting feedback from the local environment and / or `Docker` itself has been trickier than anticipated. Basically, verifying whether this worked or not: `COPY . .` Tried several variants, including some online wisdom on [How can I inspect the file system of a failed docker build?](https://stackoverflow.com/questions/26220957/how-can-i-inspect-the-file-system-of-a-failed-docker-build)

@akramtexas If you look at the console logs of edgex-go-master-verify and/or edgex-go-master-verify-arm you will find the specific build error there.

akramtexas:
The Jenkins build eventually runs aground, but has interesting feedback: [Jenkins build](https://jenkins.edgexfoundry.org/job/edgex-go-master-verify-go/1468/console) and which I'm parsing for clues 🔍

docker/Dockerfile.sys-mgmt-agent
akramtexas: Excellent point! **Done**

docker/Dockerfile.sys-mgmt-agent
akramtexas: Done.

akramtexas:
- Things looking up: *11:19:25 Finished: SUCCESS*
- That's per the Jenkins' build: [Jenkins Console](https://jenkins.edgexfoundry.org/job/edgex-go-master-verify-go/1474/console)
- Meanwhile, I’m testing it locally, by running `make docker` locally, and then try to run the image (locally) to verify that all works as expected per *POC C*. However, getting this error, locally, on running `make docker`:

```
[WARN]  Unable to checkout github.com/mitchellh/consulstructure
[ERROR] Update failed for github.com/mitchellh/consulstructure: Unable to get repository: Cloning into '/root/.glide/cache/src/https-github.com-mitchellh-consulstructure'...
fatal: unable to access 'https://github.com/mitchellh/consulstructure/': Empty reply from server
: exit status 128
[WARN]  Unable to checkout github.com/mitchellh/copystructure
[ERROR] Update failed for github.com/mitchellh/copystructure: Unable to get repository: Cloning into '/root/.glide/cache/src/https-github.com-mitchellh-copystructure'...
fatal: unable to access 'https://github.com/mitchellh/copystructure/': Empty reply from server
: exit status 128

```

akramtexas:
Using the following, two-step process to shell-in to the `Docker` image...
> docker run -itd -v /var/run/docker.sock:/var/run/docker.sock IMAGEID
> docker exec -it CONTAINERID /bin/bash
...I get the following feedback at this time, using *locally-generated* `Docker` images:

akramtexas:
Looking into how we might adjust the last entry point in the `Dockerfile` (Thanks for the tip, @JPWKU) since the container does start running, but only momentarily, stopping unbidden:

> docker ps -a
> CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS                      PORTS               NAMES
> 3bf37022c05c        6a691ee9f9dd        “/sys-mgmt-agent --p…”   3 minutes ago       Exited (0) 3 minutes ago                        eager_cocks
> e2223a063ec8        6a691ee9f9dd        “/sys-mgmt-agent --p…”   4 minutes ago       Exited (0) 4 minutes ago                        upbeat_hofstadter

Right now we got in in `Dockerfile.sys-mgmt-agent`: `ENTRYPOINT ["/sys-mgmt-agent","--profile=docker","--confdir=/res"]`

akramtexas:
@jpwhitemn Thanks to tips from @JPWKU (i.e. in `Dockerfile`, replacing ENTRYPOINT ["/sys-mgmt-agent","--profile=docker","--confdir=/res"] by `ENTRYPOINT ["/bin/bash", "-c"]`, etc.) we aimed to `exec` in to the generate imaged, start the SMA, and see if there is anything wrong. 
- Got the following, and investigating:
> bash-4.4# ./sys-mgmt-agent --profile=docker --confdir=/res
> level=ERROR ts=2018-12-13T00:02:01.494654173Z source=main.go:91 msg="edgex-sys-mgmt-agent: Service bootstrap failed!"

akramtexas:
Check

akramtexas:
Preserving some `Docker`-related info (often-used, diagnostic commands) that will be helpful down-the-road, for myself and others:
As an intermediate (testing) step, to to start all EdgeX services from _outside_ the container (effectively bypassing the SMA), run: `docker exec -it CONTAINERID docker-compose -d up`

akramtexas:
- While testing through the SMA's REST interface, the response as in the snippet below is expected when (other-than-SMA) EdgeX services are _down_. I had issued this REST call via the PostMan client, after having gotten port-forwarding going via `docker run -it -p 127.0.0.1:48090:48090 8767fe8f90bf`:
`http://localhost:48090/api/v1/config/edgex-support-logging,edgex-core-metadata,edgex-support-notifications,edgex-core-data,edgex-core-command,edgex-export-client,edgex-export-distro`
- And got a response, of which a partial snippet is like this:

> level=INFO ts=2018-12-14T16:18:02.7891318Z source=main.go:72 msg="Service dependencies resolved..."
> level=INFO ts=2018-12-14T16:18:02.7892279Z source=main.go:73 msg="Starting edgex-sys-mgmt-agent 1.0.0 "
> level=INFO ts=2018-12-14T16:18:02.7893398Z source=main.go:84 msg="Listening on port: 48090"
> level=ERROR ts=2018-12-14T16:18:22.3871968Z source=services.go:78 msg="edgex-support-logging get config error: Get http://edgex-support-logging:48061/api/v1/config: dial tcp: lookup edgex-support-logging on 192.168.65.1:53: no such host"
> level=ERROR ts=2018-12-14T16:18:22.4518932Z source=services.go:78 msg="edgex-core-metadata get config error: Get http://edgex-core-metadata:48081/api/v1/config: dial tcp: lookup edgex-core-metadata on 192.168.65.1:53: no such host

akramtexas:
Thanks to a crucial insight from @jpwhitemn, the SMA (running inside the DinD `Docker container`) can now start and stop EdgeX services when it (i.e. the DinD `Docker container` hosting the SMA) is started with bindings as illustrated in the command below:

> docker run -d -p 48090:48090 --name akram-sma -v /var/run/docker.sock:/var/run/docker.sock IMAGE-ID-of-DinD

akramtexas:
> * Using the following, two-step process to shell-in to the `Docker` image...
> 
> 1. docker run -itd -v /var/run/docker.sock:/var/run/docker.sock IMAGEID
> 2. docker exec -it CONTAINERID /bin/bash
> ...
> docker run -itd -v /var/run/docker.sock:/var/run/docker.sock **6a691ee9f9dd**
> 7a6bd045d8c58e5d5a7394e933f2326587415be7fc7e602dc7db5353dd34ca9c
> ...

- Starting the custom `DinD Docker container` (hosting the SMA) with (_similar_) **proper bindings** was attempted (as in the quoted comment that appears earlier in this PR) at one time, during the work on _predecessors_ of POC C, this information had now been applied (thanks to Jim, as mentioned in the preceding note).

akramtexas:
- Separate from the preceding two comments, in a parallel effort, I created the SMA executable with the `make` command, then did a docker build with a `Dockerfile` that simply copies the SMA executable (rather than building it in the Docker image).
- Here is the outcome of that parallel effort:

> bash-4.4# ./sys-mgmt-agent 
> bash: ./sys-mgmt-agent: cannot execute binary file: Exec format error
> bash-4.4# whoami
> root

akramtexas:
@jpwhitemn While the SMA now can, via the `DinD Docker image with SMA`, start all EdgeX services, it is, oddly enough, at the moment, unable to _stop_ them 🛑 I am investigating 🔍

akramtexas:
> @jpwhitemn While the SMA now can, via the `DinD Docker image with SMA`, start all EdgeX services, it is, oddly enough, at the moment, unable to _stop_ them 🛑 I am investigating 🔍

@jpwhitemn I `exec'd` in to running container and found telltale signs in the logging by SMA, and am investigating:

> 2018/12/19 15:21:15 http: panic serving 172.17.0.1:54228: runtime error: invalid memory address or nil pointer dereference
> goroutine 23 [running]:
> net/http.(*conn).serve.func1(0xc0002bb2c0)
>         /usr/local/go/src/net/http/server.go:1746 +0xd0
> panic(0x7555e0, 0xa70a40)
>         /usr/local/go/src/runtime/panic.go:513 +0x1b9
> github.com/edgexfoundry/edgex-go/internal/system/agent.InvokeOperation(0x7bf65e, 0x4, 0xc0003c0090, 0x7, 0x9, 0x0)
>         /go/src/github.com/edgexfoundry/edgex-go/internal/system/agent/services.go:48 +0x5a7
> github.com/edgexfoundry/edgex-go/internal/system/agent.operationHandler(0x822260, 0xc0004ea000, 0xc00038e500)
>         /go/src/github.com/edgexfoundry/edgex-go/internal/system/agent/router.go:69 +0x5d2
> net/http.HandlerFunc.ServeHTTP(0x7d74e0, 0x822260, 0xc0004ea000, 0xc00038e500)
>         /usr/local/go/src/net/http/server.go:1964 +0x44
> github.com/edgexfoundry/edgex-go/vendor/github.com/gorilla/mux.(*Router).ServeHTTP(0xc0002c8280, 0x822260, 0xc0004ea000, 0xc00038e500)
>         /go/src/github.com/edgexfoundry/edgex-go/vendor/github.com/gorilla/mux/mux.go:114 +0xe
> 

  553  docker run -d -p 127.0.0.1:48090:48090 -v /var/run/docker.sock:/var/run/docker.sock 5b844e9acd6b
  554  docker exec -it a827a5252080 /bin/sh

- As noted earlier, the SMA now can, via the `DinD Docker image with SMA`, _start_ all EdgeX services, for example as in this logging:

> level=INFO ts=2018-12-19T15:20:23.1481388Z source=services.go:36 actiontype=start servicename=edgex-export-distro msg="invoking operation on service"
> [370/1536]0xc0004ae600
> level=INFO ts=2018-12-19T15:21:15.4507536Z source=services.go:36 actiontype=stop servicename=edgex-support-logging msg="invoking operation on service"

akramtexas:
> @jpwhitemn I `exec'd` in to running container and found telltale signs in the logging by SMA, and am investigating:
> 
> > 2018/12/19 15:21:15 http: panic serving 172.17.0.1:54228: runtime error: invalid memory address or nil pointer dereference

Identified the problem, or at least a major part that is problematic, which is that this call (made during execution of the SMA function `func (de *ExecuteDocker) StopService(service string) error {...}`) is returning `nil`:

> containers, err := cli.ContainerList(ctx, types.ContainerListOptions{})

- No containers are found by the SMA, and so it can stop none...
- Investigating the cause of the disconnect (i.e. unreachability of the containers that host the other EdgeX services)

akramtexas:
- Captured this (annotated) runtime information through debugger, related to the `containers` list being empty:

akramtexas:
`Through a combination of `Docker` exec'ing and stepping through code, figured out and fixed the cause of `stop` operation not working through `DinD` (Tested and pushed the change-set to remote repo).`

docker/Dockerfile.sys-mgmt-agent
```diff
@@ -0,0 +1,37 @@
...
+FROM docker:latest
+
+COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/sys-mgmt-agent /
+COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/res/docker/configuration.toml /res/docker/configuration.toml
+COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/res/docker/configuration.toml /res/configuration.toml
+
+RUN apk --no-cache add py-pip
+RUN pip install docker-compose
+RUN apk --no-cache add curl
+RUN curl -o docker-compose.yml https://raw.githubusercontent.com/edgexfoundry/developer-scripts/master/compose-files/docker-compose-delhi-0.7.0.yml
```
tsconn23: Just calling out you might want to reference the delhi-0.7.1 file here

internal/pkg/config/loader.go
```diff
@@ -41,12 +43,14 @@ func LoadFromFile(profile string, configuration interface{}) error {
 	}
 	contents, err := ioutil.ReadFile(fileName)
 	if err != nil {
+		LoggingClient.Error("could not load configuration file (%s): %v", fileName, err.Error())
 		return fmt.Errorf("could not load configuration file (%s): %v", fileName, err.Error())
```
tsconn23: Can just build the string once and use it for logging as well as the returned error. Also err.Error() returns a string so you can use `%s` instead of `%v`

```
msg := fmt.Sprintf("could not load configuration file (%s): %s", fileName, err.Error())
LoggingClient.Error(msg)
return errors.New(msg)
```

internal/system/agent/executor/docker_exec.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil
+	} else {
+		newError := fmt.Errorf("unknown service: %v", service)
+		logs.LoggingClient.Error(newError.Error())
+
+		return newError
+	}
+}
+
+func RunDockerCommands(service string, dockerService string, operation string) {
```
tsconn23: Function isn't called outside of this package, so doesn't need to be exported (capitalized)

internal/system/agent/executor/docker_exec.go
```diff
@@ -0,0 +1,134 @@
...
+			logs.LoggingClient.Warn("docker start operation failed", "service name", service)
+		}
+	} else if operation == "stop" {
+		if ! FindDockerContainerStatus(service, "Exited") {
+			logs.LoggingClient.Warn("docker stop operation failed", "service name", service)
+		}
+	}
+}
+
+func FindPathToRunDocker() (string, error) {
```
tsconn23: Function isn't called outside of this package, so doesn't need to be exported (capitalized)

internal/system/agent/services.go
```diff
@@ -41,17 +41,17 @@ func InvokeOperation(action string, services []string) bool {
 
 		switch action {
 		case START:
-			executor.StartDockerContainerCompose(service, Configuration.ComposeUrl)
+			executor.ExecuteDockerCommands(service, "start")
 			break
 
 		case STOP:
-			ec.StopService(service)
+			executor.ExecuteDockerCommands(service, "stop")
```
tsconn23: Note that the interface method `StopService` is no longer being used here, meaning it currently isn't used _anywhere_ in the application. I'm assuming this will be rectified or else the `ExecutorClient` interface will be removed.

internal/system/agent/services.go
```diff
@@ -110,19 +110,18 @@ func getMetrics(services []string) (MetricsRespMap, error) {
...
-		internal.ConfigSeedServiceKey: {},
+	var services = map[string]string{
+		internal.SupportNotificationsServiceKey: "edgex-support-notifications",
+		internal.CoreDataServiceKey:             "edgex-core-data",
+		internal.CoreMetaDataServiceKey:         "edgex-core-metadata",
+		internal.CoreCommandServiceKey:          "edgex-core-command",
+		internal.ExportClientServiceKey:         "edgex-export-client",
+		internal.ExportDistroServiceKey:         "edgex-export-distro",
+		internal.SupportLoggingServiceKey:       "edgex-support-logging",
+		internal.ConfigSeedServiceKey:           "edgex-config-seed",
```
tsconn23: This list is already built once in `init.go` via `func initializeClients()`. This implementation is different than that one. Recommend these be common and that the list be stored globally somewhere rather than populated each time we need to do a lookup.

jpwhitemn: @akramtexas - given holiday season and your about to be out in a few short minutes, I am going to merge - but please put this as first task to address on your return and try to hit as many of Trevor's suggestions as possible.

internal/system/agent/services.go
anonymouse64: This also breaks the snap, as this interface is how the SMA is interacting with the snap.

internal/system/agent/services.go
akramtexas: - @jpwhitemn This will be my number one priority on returning after the holiday season 👍 
- @tsconn23 Making a note of your comments, thanks. In particular, I will definitely _refactor_ this code to align with your proposal; actually, this was a quick fix done in the past 24 hours to undo some functionality-breaking effects of a refactoring that was done earlier, in a PR elsewhere (I had actually reviewed that PR for the submitter a couple of weeks ago). Hence the reason for re-introducing the global definition in the spirit of this being POC work. But for sure, I will address and clean this up in particular 👍

jpwhitemn:
@akramtexas - based on Ian's breakage note, let's hold off on this.  Let's get it cleaned up in the new year.  Thanks Ian, Thanks Trevor and thanks much Akram.  We'll get it right.

akramtexas:
> @akramtexas - based on Ian's breakage note, let's hold off on this. Let's get it cleaned up in the new year. Thanks Ian, Thanks Trevor and thanks much Akram. We'll get it right.

@jpwhitemn Checking the status of PR quick while on holiday and notice Ian rightly pointing out the breakage. Yes, will get it cleaned up and make it right early in the new year 👍

internal/system/agent/services.go
akramtexas: _Prior_ to the refactoring work done earlier, separately, per [PR-706](https://github.com/edgexfoundry/edgex-go/issues/706), a _single_ list had been created and stored globally, rather than populated each time we need to do a lookup. But given the distinction between the contents of the strings appearing in the two lists, for example "Notifications", "Command", "CoreData", etc. **versus** "edgex-support-notifications", "edgex-core-data", "edgex-core-metadata", etc., **respectively**, a logical check had been put in place to handle the 'mapping'. As best as I can tell, that was undone during the refactoring work. A similar handling can be reintroduced...

akramtexas:
Recheck

akramtexas:
Jenkins build is still failing with this logging:

> 07:42:48 [0;31m[ERROR]	[mError scanning golang.org/x/net/internal/socks: cannot find package "." in:
07:42:48 	/home/jenkins/.glide/cache/src/https-golang.org-x-net/internal/socks
07:42:48 [0;32m[INFO]	[m--> Fetching github.com/hashicorp/go-immutable-radix
07:42:49 [0;32m[INFO]	[m--> Fetching github.com/hashicorp/golang-lru
07:42:49 [0;31m[ERROR]	[mFailed to retrieve a list of dependencies: Error resolving imports
07:42:49 make: *** [prepare] Error 1
07:42:49 Build step 'Execute shell' marked build as failure

akramtexas:
- Updates to Dockerfile to bring in the bits for the executor are currently a WIP:
> Step 8/20 : RUN go build main.go
 ---> Running in 9e76041d57da
stat main.go: no such file or directory
The command '/bin/sh -c go build main.go' returned a non-zero code: 1
make: *** [docker_sys_mgmt_agent] Error 1

docker/Dockerfile.sys-mgmt-agent
```diff
@@ -7,24 +7,42 @@
...
 RUN apk update && apk add make glide git
 COPY . .
 
-RUN make prepare
-RUN make cmd/sys-mgmt-agent/sys-mgmt-agent
+# RUN make prepare
+# RUN make cmd/sys-mgmt-agent/sys-mgmt-agent
+
+# Build the golang "executor" bits (an SMA-related reference implementation) by fetching the source from Github.
+RUN git clone https://github.com/akramtexas/docker-compose-executor
```
JPWKU: Clone this to /go/src/github.com/docker-compose-executor since that what you have WORKDIR set as.  You may need to reset WORKDIR for building each of docker-compose-executor and SMA.

akramtexas:
I’m close to a fully-functioning `Dockerfile`. This glitch, however, is currently holding me back. Both *SMA* and *executor* build fine, but the *executor* is not getting `COPY` over as it’s not found by `Docker` runtime. Reached out to @JPWKU 👍 

> Status: Downloaded newer image for docker:latest
 ---> 57d6f6487e82
Step 12/21 : COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/sys-mgmt-agent /
 ---> da8441825aef
Step 13/21 : COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/res/docker/configuration.toml /res/docker/configuration.toml
 ---> 331fd288e07b
Step 14/21 : COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/res/docker/configuration.toml /res/configuration.toml
 ---> fbac498ab7f4
Step 15/21 : COPY --from=builder /go/src/github.com/docker-compose/main /
COPY failed: stat /var/lib/docker/overlay2/c332cccd64f8f2e6f16fda79aa646eb62e9648a94b4060c7b9c080caea73b88e/merged/go/src/github.com/docker-compose/main: no such file or directory
make: *** [docker_sys_mgmt_agent] Error 1

docker/Dockerfile.sys-mgmt-agent
```diff
@@ -0,0 +1,49 @@
...
+FROM docker:latest
+
+# Copy over the SMA bits.
+COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/sys-mgmt-agent /
+COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/res/docker/configuration.toml /res/docker/configuration.toml
+COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/res/docker/configuration.toml /res/configuration.toml
+
+# Copy over the golang "executor" bits.
+# COPY /go/src/github.com/docker-compose/main /
+COPY --from=builder /go/src/github.com/docker-compose/main /
```
JPWKU: This is built in the previous stage in `WORKDIR /go/src/github.com/docker-compose` so I think maybe the COPY should be from `/go/src/github.com/docker-compose/docker-compose/main`

JPWKU: either that or make the WORKDIR `/go/src/github.com`

akramtexas:
> This PR will break the SMA working with the snap

Following several discussions with Ian and Tony, and indeed in the Sys Mgmt WG meetings, we got three new abstractions now (the following) that address Ian’s comments:

`type ServiceStarter interface {
	Start(service string) error
}

type ServiceStopper interface {
	Stop(service string, ) error
}

type ServiceRestarter interface {
	Restart(service string) error
}
`

internal/pkg/config/loader.go
```diff
@@ -41,13 +44,17 @@ func LoadFromFile(profile string, configuration interface{}) error {
 	}
 	contents, err := ioutil.ReadFile(fileName)
 	if err != nil {
-		return fmt.Errorf("could not load configuration file (%s): %v", fileName, err.Error())
+		msg := fmt.Sprintf("could not load configuration file (%s): %s", fileName, err.Error())
```
anonymouse64: You don't need to explictly call `err.Error()` here, go will just call the Error method for you. It's easier to read if you leave out the `Error()` method.

cmd/sys-mgmt-agent/res/docker/configuration.toml
```diff
@@ -0,0 +1,64 @@
...
+ServiceTimeout = 5000
+CheckInterval = '10s'
+
+[Logging]
+EnableRemote = false
+File = '/edgex/logs/edgex-sys-mgmt-agent.log'
+
+# OperationsType = 'os'
+OperationsType = 'docker'
+ComposeUrl = 'https://raw.githubusercontent.com/edgexfoundry/developer-scripts/master/compose-files/docker-compose-delhi-0.7.0.yml'
```
anonymouse64: Shouldn't this be deleted too since now the ComposeUrl will be living inside the executable that we run instead of in the SMA?

cmd/sys-mgmt-agent/res/docker/configuration.toml
```diff
@@ -0,0 +1,64 @@
...
+
```
anonymouse64: Shouldn't this config file also contain the new `AppPath` (or as I recommended changing it `ExecutorPath`)?

glide.yaml
```diff
@@ -42,3 +34,9 @@ import:
   version: =1.2.1
 - package: gopkg.in/yaml.v2
   version: eb3733d160e74a9c7e442f435eb3bea458e1d19f
+- package: golang.org/x/net
```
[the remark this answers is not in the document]
akramtexas: So this is purely a _temporary_ (experimental) fix. Details here: [Issue #1070](https://github.com/edgexfoundry/edgex-go/issues/1070)

akramtexas:
Please note the new coordinates for the cleaned-up repo, which houses a reference implementation for an SMA executor: [Temporary home](https://github.com/akramtexas/docker-compose).

anonymouse64:
Is the reference implementation for the docker-compose executor not going to be pulled into edgex-go?

akramtexas:
> Is the reference implementation for the docker-compose executor not going to be pulled into edgex-go?

The thinking there is to have a separate repo (_outside_ of edgex-go) where this reference implementation, which of course is for the docker-compose executor (_and_ other reference implementations) to find a (unifying) home as well.

Makefile
```diff
@@ -65,7 +65,7 @@ test:
 	go vet ./...
 
 prepare:
-	glide install
+	glide update && glide install
```
JPWKU: Hmm I am seeing the same failure as before.  Can you remove the `glide install` here and just have `glide update` as a check?

anonymouse64:
> The thinking there is to have a separate repo (_outside_ of edgex-go) where this reference implementation, which of course is for the docker-compose executor (_and_ other reference implementations) to find a (unifying) home as well.

I guess I still don't understand... Is the code you wrote for the docker-compose-executor going to be committed to edgex-go (either as a part of this PR or a future one)? I think it should, otherwise you have 2 scenarios:
1. The default distribution of EdgeX doesn't include this and users have to someone inject this new binary into their running docker images, which is a very bad user experience OR
2. We have to modify this repository's docker build scripts so that it builds a different repository and then includes it into the SMA image. I think this is wrong and will prove to be difficult to maintain.

akramtexas:
> > The thinking there is to have a separate repo (_outside_ of edgex-go) where this reference implementation, which of course is for the docker-compose executor (_and_ other reference implementations) to find a (unifying) home as well.
> 
> I guess I still don't understand... Is the code you wrote for the docker-compose-executor going to be committed to edgex-go (either as a part of this PR or a future one)? I think it should, otherwise you have 2 scenarios:
> 
> 1. The default distribution of EdgeX doesn't include this and users have to someone inject this new binary into their running docker images, which is a very bad user experience OR
> 2. We have to modify this repository's docker build scripts so that it builds a different repository and then includes it into the SMA image. I think this is wrong and will prove to be difficult to maintain.

- Good question @anonymouse64 - To simplify things, we're going to track that related work (i.e. the code I wrote for the docker-compose-executor) here: [Issue #1091](https://github.com/edgexfoundry/edgex-go/issues/1091)

akramtexas:
Recheck

akramtexas:
Recheck

cmd/sys-mgmt-agent/main.go
```diff
@@ -56,7 +56,7 @@ func main() {
 	flag.Usage = usage.HelpCallback
 	flag.Parse()
 	// [1] Removed two lines of code above (having to do with accepting the "--consul" aka "--c" flag) because the SMA is
-	// designed to operate independently, in a stand-alone fashion. In particular, the SMA should not rely on Consul,
+	// designed to operate independently,d in a stand-alone fashion. In particular, the SMA should not rely on Consul,
```
anonymouse64: I just noticed the typo here, but what is the purpose of this comment in the code? It doesn't appear to be describing anything in the code right here so I don't think it's necessary. If you choose to merge it in with this code here, please file an issue for a followup PR to remove such comments

cmd/sys-mgmt-agent/res/configuration.toml
```diff
@@ -16,9 +16,10 @@ ServiceAddress = 'localhost'
 CheckInterval = '10s'
 EnableRemoteLogging = false
 LoggingFile = './logs/edgex-sys-mgmt-agent.log'
-# OperationsType = 'os'
-OperationsType = 'docker'
-ComposeUrl = 'https://raw.githubusercontent.com/edgexfoundry/developer-scripts/master/compose-files/docker-compose-delhi-0.7.0.yml'
+
+ExecutorPath = '/path/to/the/file'
+# The following (ExecutorPath) is DEBUG_ONLY:
+# ExecutorPath = '/Users/akramahmad/go/src/github.com/edgexfoundry/docker-compose-executor/main'
```
anonymouse64: I still would prefer this to not be here, but it's at least commented out so :man_shrugging:

docker/Dockerfile.sys-mgmt-agent
```diff
@@ -0,0 +1,48 @@
...
+
+RUN apk update && apk add make glide git
+COPY . .
+
+RUN make prepare
+RUN make cmd/sys-mgmt-agent/sys-mgmt-agent
+
+# Build the golang "executor" bits (an SMA-related reference implementation) by fetching the source from Github.
+WORKDIR /go/src/github.com/docker-compose
+RUN git clone https://github.com/akramtexas/docker-compose
```
anonymouse64: I'm more just commenting here because I want to raise attention to this point on why I think doing it this way is a bad idea. Here, in order to include the docker-compose executable from this external repo, we have to clone another external git repo inside this docker build file, which obviously works right now but has the following issues:

* you will now have duplicated build code for both this repo here and the docker-compose repo as well (I assume that repo will be adopted into `edgexfoundry` and as such will need it's own build code in that repository separate from this repo. That code is going to be out-of-sync with this code because likely no one is going to independently test/run that repo, they will only ever use the version in edgex-go and that's the sort of thing that leads to bit-rotting and old confusing code that doesn't run anymore. Having all the build code for a single thing in one place is much better. All that being said, you could also just put the build code inside a Makefile in the other repo and call `make build` here and that would be better
* because you are cloning this repo here, and because I assume that this other repo will need to be released with the rest of EdgeX come release day, the other repo will have it's branch change and someone will need to remember to come in to the release branch on this repo and change this git clone to checkout the correct branch. Building the other code inside this repo removes that problem entirely.
* testing changes to the other, external repo will become burdensome and difficult for reviewers because they will need to change this line in the docker compose file to checkout your branch (or otherwise build up all of edgex-go in their own docker containers) in order to build a working container which can run the things needed. This may not seem like a big deal now, but we want to make it easier to test changes to components of EdgeX. For example, it means that writing automated CI solutions that can run integration tests that would run against a specific PR that is opened would need to perform this kind of modification step in order to build the container in this repo and test it which is highly unfortunate and difficult for the devops team to setup. 
* I also highly discourage pulling in code that we build like this from a github repo under your personal account. I think before this PR can be merged that code has to be pulled into edgexfoundry where it can be reviewed and will have a higher visibility than under your personal account. Additionally this bit brings up the point about who will review (or has reviewed) this external go code? If this PR was merged right now and my assumption that no one has reviewed your code there yet then we are essentially bringing in completely un-reviewed code and building it.

internal/system/agent/execution_engine.go
```diff
@@ -0,0 +1,60 @@
...
+	// Preparing the call to the executor app.
+	cmd := exec.Command(Configuration.ExecutorPath, service, operation)
+
+	cmd.Dir = Configuration.ExecutorPath
+
+	_, err := cmd.CombinedOutput()
+	if err != nil {
+		logs.LoggingClient.Error(fmt.Sprintf("the error occurred in the invocation of executor: %v ", err.Error()))
+	} else {
+		logs.LoggingClient.Info("invocation of executor succeeded")
```
anonymouse64: It would be nice if this message could contain what service the executor handled and what operation it did with it. But this can be filed as an issue and handled in a followup PR.

internal/system/agent/init.go
```diff
@@ -73,19 +75,9 @@ func Retry(useConsul bool, useProfile string, timeout int, wait *sync.WaitGroup,
...
-	case "os":
-		return &executor.ExecuteOs{}, nil
-	case "docker":
-		return &executor.ExecuteDocker{}, nil
-	case "snap":
-		return &executor.ExecuteSnap{}, nil
-	default:
-		return nil, nil
-	}
+func newExecutorClient() (interface{}, error) {
```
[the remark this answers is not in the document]
anonymouse64: I still would like to see this function removed as it now does nothing

PR #927, opened 2018-12-12
Removed support for labels from logging

### LINKED ISSUE #904: Remove Support for Labels from Logging

jduranf:
Fix #904.

Also `internal/support/logging/models/log_entry.go` have been removed, in order to use `pkg/models/log_entry.go`, which have the same functionality.

Finally an issue in support-logging have been fixed. When using file persistence in support-logging, the limit parameter wasn't used.


internal/support/logging/persistence_test.go
```diff
@@ -62,13 +50,9 @@ func testPersistenceFind(t *testing.T, persistence persistence) {
 	persistence.add(le)
 	le.Message = message2
 	le.OriginService = sampleService2
-	le.Message = message2
 	persistence.add(le)
 	le.Message = message1
 	persistence.add(le)
-	le.Message = message2
-	le.Args = args2
```
brandonforster: Do we not want to persist key/value pairs?

pkg/clients/logging/logger.go
```diff
@@ -49,11 +48,11 @@ var LogLevels = []string{
 
 type LoggingClient interface {
 	SetLogLevel(logLevel string) error
-	Debug(msg string, args ...interface{}) error
-	Error(msg string, args ...interface{}) error
-	Info(msg string, args ...interface{}) error
-	Trace(msg string, args ...interface{}) error
-	Warn(msg string, args ...interface{}) error
+	Debug(msg string) error
```
brandonforster: The entire point of #739 was adding this functionality.

internal/support/logging/types.go
```diff
@@ -47,7 +47,7 @@ func newPrivateLogger() privLogger {
 	return p
 }
 
-func (l privLogger) log(logLevel string, msg string, args ...interface{}) {
+func (l privLogger) log(logLevel string, msg string) {
```
tsconn23: The changes in this file are too aggressive. Since PR #787 was merged, the args parameter should be used to specify key/value pairs for inclusion in our log messages. It no longer corresponds to labels and should be preserved. I don't think any changes in this file are required.

pkg/clients/logging/logger.go
tsconn23: Agreed, see my additional comment on logging/types.go

tsconn23:
I'm fine with the removal of the routes, RAML adjustments and any relevant testing mods. Please see the comment above on logging/types.go

Also, the majority of this PR DOES address a valid problem though (albeit tangential) wherein we have statements throughout the application which specify an empty string for the `args` parameter. This prevents the statement from being logged at all, so I definitely want those changes.

From an email to @brandonforster from 2-Dec-18
```
One thing of which you may already be aware – During testing I found that with the new logger implementation, entries like this will log:
data.LoggingClient.Info("Service dependencies resolved...")

Entries like this will not:
data.LoggingClient.Info(data.Configuration.Service.StartupMsg, "")
```

jduranf:
ok, second attempt, without touching the key/value system

Anyway, this is how a log entry is stored in the persistence system:
`{"logLevel":"INFO","args":["msg","Starting edgex-core-data 1.0.0 "],"originService":"edgex-core-data","message":"Starting edgex-core-data 1.0.0 ","created":1544704140918}`

The message is stored twice times. Is this the desired behaviour? If this isn't the desired behaviour, I can fix this...

tsconn23:
@jduranf Yes, please do address the duplication of the msg field. We shouldn't put that into args but any free-form key/vals which may have been provided at the point LoggingClient was called should be put into args.

From an email to @brandonforster (11-Nov-18)
```
One final thing to look at -- Using PR 787, entries that are written to Mongo look like this:
{ "_id" : ObjectId("5bf1ec4e3b91dc8388fe5973"), "logLevel" : "INFO", "labels" : [ "msg", "Deleting events by age: 1000000000000" ], "originService" : "edgex-core-data", "message" : "Deleting events by age: 1000000000000", "created" : NumberLong("1542581326675") }

As you can see, what's in "labels" is the same content as what's in "message". Think about how can we instead put the optional fields that will come to be found in our log messages into this field instead (such as "source" right now)
```

jduranf:
third attempt :)
Fixed some issues in the logging system:
1. Fixed msg field duplication
2. Now support-logging uses gokit log system
3. Now support-logging generates args parameter properly

Issue #842, opened 2018-11-20
When sending post command with an empty device profile to /deviceprofile/uploadfile of Core Metadata API, 400 should be returned


FelixTing:
OS version: ubuntu 16.04
EdgeX version: 1.0.0
Environment details: Docker container
Steps to reproduce:

1. start up EdgeX
2. send post command with an empty device profile to /deviceprofile/uploadfile of Core Metadata API

Actual results: Got returned http status code 200 OK
Expected results: Got returned http status code 400 BadRequest
RAML: https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-metadata.raml#L779

PR #935, opened 2018-12-13
#842 Check for empty yaml file at /deviceprofile/uploadfile

jdharms:
#842

Check for empty yaml file at /deviceprofile/uploadfile of
core metadata api

### DIFF
--- a/internal/core/metadata/rest_deviceprofile.go
+++ b/internal/core/metadata/rest_deviceprofile.go
@@ -387,6 +387,12 @@ func restAddProfileByYaml(w http.ResponseWriter, r *http.Request) {
 		LoggingClient.Error(err.Error())
 		return
 	}
+	if len(data) == 0 {
+		err := errors.New("YAML file is empty")
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		LoggingClient.Error(err.Error())
+		return
+	}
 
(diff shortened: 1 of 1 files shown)


internal/core/metadata/rest_deviceprofile.go
```diff
@@ -385,6 +385,12 @@ func restAddProfileByYaml(w http.ResponseWriter, r *http.Request) {
 		LoggingClient.Error(err.Error(), "")
 		return
 	}
+	if len(data) == 0 {
+		err := errors.New("YAML file is empty")
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		LoggingClient.Error(err.Error(), "")
```
brandonforster: I understand you're following along with the other statements in the file, but this is not how we want to do logging going forward.

The syntax for a logging statement is `(message string, args string...)` where message is a required message to the user and args is a collection of key/value pairs corresponding to some additional info you want to pass to the user. We're moving away from using `fmt.sprintf` and transitioning statements over to a string literal as the message with any variables as a value in a k/v.

I'd recommend updating this statement to
`LoggingClient.Error("failed to add profile by YAML", "error message", "err.Error()")`

If you're feeling saucy, it would be greatly appreciated if you did the same to more of the statements in the file. At the very least however we don't want to introduce more of the old style into the codebase.

Issue #841, opened 2018-11-20
Device Profile can be created with a null name


FelixTing:
OS version: ubuntu 16.04
EdgeX version: 1.0.0
Environment details: Docker container
Steps to reproduce:

1. start up EdgeX
2. send post command to /deviceprofile of Core Metadata API
for example, with the following request URL and body:
```
Method: POST
URL: http://localhost:48081/api/v1/deviceprofile
Body:
{
  "origin":1471806386918,
  "description":"BACnet Honeywell thermostats",
  "manufacturer":"Honeywell",
  "model":"ABC123",
  "labels":["bacnet","thermostat","Honeywell"],
  "commands":[
    {
      "origin":1471806386919,
      "name":"cooling point",
      "get":
        {
          "path":"/cooling",
          "response":{
            "code":"200",
            "errorDescription":"not a float",
            "expectedValues":"temperature"
          }
        },
      "put":{
        "path":"/cooling",
        "response":{
          "code":"200",
          "errorDescription":"not a float",
          "expectedValues":"temperature"
        },
        "parameters":[  
               {  
                  "name":"coolingpoint",
                  "value":"float",
                  "uom":"string"
               }
        ]
      }
    }
  ],
  "objects":{
    "resources":{
      "CurrentHumidity":{
        "get":[
          {
            "parameter":"CurrentHumidity",
            "property":"presentValue",
            "operation":"get",
            "object":"AnalogValue_22"
          }
        ]
      }
    }
  }
}
```

Actual results: Got returned http status code 200 OK
Expected results: Got returned http status code 400 BadRequest
RAML: https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-metadata.raml#L921

FelixTing:
Once there is a profile data with a null name exist in the database,
send a put command with an inexistent id and a null name to /deviceprofile of Core Metadata API, we will get the returned http status code 200(OK). However, it should be 404(NotFound)

PR #936, opened 2018-12-13
#841 Check for empty device profile name and return status 400

jdharms:
#841 

Check for empty device profile name in mongo client and
handle the error with a 400 status code.

### DIFF
--- a/internal/core/metadata/rest_deviceprofile.go
+++ b/internal/core/metadata/rest_deviceprofile.go
@@ -73,6 +73,8 @@ func restAddDeviceProfile(w http.ResponseWriter, r *http.Request) {
 	if err := dbClient.AddDeviceProfile(&dp); err != nil {
 		if err == db.ErrNotUnique {
 			http.Error(w, "Duplicate name for device profile", http.StatusConflict)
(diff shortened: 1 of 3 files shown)


jduranf:
Although the code looks good, in my opinion the check should be performed in rest_deviceprofile.go. Doing that, the check will be performed if another database it's used.

brandonforster:
@jduranf I'm not sure I agree. That would add business logic to the REST layer and force us to throw an error on an empty DB all the time.

Off the top of my head, I can see that negatively impacting tests using the MemDB where a developer has an empty DB but doesn't need or want anything in it.

I think leaving it the way Daniel has it and requiring the DB driver to determine when there is an error  and letting the REST layer react to "something terrible has happened" is the best solution.

tsconn23:
@jduranf 's point is valid, however what we should really have is a separation like we have in core-data/event whereby the router is decoupled from biz logic, which is further decoupled from persistence. If we had, for example, a deviceprofile.go which handled biz logic (like we have event.go in core-data) then _THAT_ is where this check should be. I am hoping subsequent refactoring will move this function and others like it in the desired direction. Will go ahead and merge.

Issue #932, opened 2018-12-13
Core-Metadata should return NotFoundException (HTTP 404) if no addressable with the provided id is found


weichou1229:
**OS version:** macos 10.14  
**EdgeX version:** 1.0.0
**Environment details:** Docker container  
**Steps to reproduce:**

1.Start up EdgeX

2.Update addressable

```
Method: PUT
URL: http://localhost:48081/api/v1/addressable
Body:
{
    "Uuid": "0874cf01-bbd4-4121-9020-d3e667665fd0",
    "origin": 1471806386924,
    "name": "gasmeter",
    "protocol": "HTTP",
    "address": "172.17.0.6",
    "port": 48093,
    "path": "/gasmeter",
    "publisher": "DELLPUB",
    "user": "test",
    "password": "test123",
    "topic": "kWhData"
}
```

Actual result: 500Internal Server Error

This API shoud return 404 according to RAML - [core-metadata.raml](https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-metadata.raml#L2216)

PR #938, opened 2018-12-13
Decoupling mongo types from metadata/Addressable endpoint

tsconn23:
#881 
    
- GET functions should return values, not populate pointers to types passed as arguments. Side-effect hell.
- Implemented Mongo provider logic, also MemoryDB changes
- Adding fix for issue #932

### DIFF
--- a/internal/core/metadata/rest_scheduleevent.go
+++ b/internal/core/metadata/rest_scheduleevent.go
@@ -107,43 +107,16 @@ func restAddScheduleEvent(w http.ResponseWriter, r *http.Request) {
 	// Check for the addressable
 	// Try by ID
 	var a models.Addressable
-	if err := dbClient.GetAddressableById(&a, se.Addressable.Id.Hex()); err != nil {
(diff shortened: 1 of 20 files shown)


internal/core/metadata/rest_scheduleevent.go
```diff
@@ -233,9 +206,11 @@ func updateScheduleEventFields(from models.ScheduleEvent, to *models.ScheduleEve
 	if (from.Addressable.String() != models.Addressable{}.String()) {
 		// Check if the new addressable exists
 		// Try by ID
-		if err := dbClient.GetAddressableById(&from.Addressable, from.Addressable.Id.Hex()); err != nil {
+		var addr models.Addressable
```
jduranf: This line can be removed

tsconn23: Removed, thx!

michaelestrin:
Addressable black box test fails:

image

PR #939, opened 2018-12-13
Turn all snap services into daemons (for master)

### LINKED ISSUE #480: Add support for systemd-notify to snap services
Using systemd-notify for snap services would allow us to properly support EdgeX micro-services as actual services in the snap. The current problem with doing this is that after a given service is started there's some amount of "initialization time" before the service is "ready" for it's dependent services to run. 
If we were able to use systemd-notify inside the micro-services, then we could specify in the snapcraft.yaml for the services to be considered "ready" by systemd only after receiving the notify signal from the services. This would also give us easy watchdog support when starting the snaps. 
See more discussion at issue #441.

### LINKED ISSUE #485: Convert services to daemons in snap
We want to convert all of the services to daemons so that the snap is able to start up on boot automatically and systemd can manage services going down and coming back up, etc. which currently the snap does not do at all.

### LINKED ISSUE #519: [snap] Config hook support for enabling/disabling individual services
The California snap start & stop scripts read a file called **$SNAP_DATA/edgex-service-env** to determine which services to start/stop.  In Dehli, the snap instead should provide configure hook settings for each service, with a single setting controlling whether or not the security services should be started.

**Note** - this should probably be done at the same time we get rid of the start/stop scripts and and make all of the services daemons, or after that work is finished.

anonymouse64:
Duplicate of #722, but for master (Edinburgh). Note that since the original implementation had issues with using daemon-notify, I have also pulled in the fix changing all the services to forking from #885.

In order to pull in the fix from #885, I had to pull in a few commits from #744, notably https://github.com/edgexfoundry/edgex-go/pull/744/commits/28b170bcee0929764261dc7b83fd1610cc4a0f1c and https://github.com/edgexfoundry/edgex-go/pull/744/commits/218f602232f3b043bb0f06eb6686d7f6968f71c8 which are unrelated to adding sys-mgmt-agent to the snap. I think it's simpler this way and ensures that for master we do a clean switch from no services to all services. I will open a followup PR after this one which adds sys-mgmt-agent to the snap properly.

Closes #485, closes #480, and closes #519 for Edinburgh


snap/README.md
```diff
@@ -161,7 +186,24 @@ $ docker run -it -v"$PWD":/build snapcore/snapcraft:stable bash -c "apt update &
...
+```
+
+After connecting these restart the services in the snap with:
+
+```bash
+$ sudo snap restart edgexfoundry
 ```
 
-After installing the snap, you will need to connect interfaces and restart the snap as explained in the previous section on [Configuration](https://github.com/edgexfoundry/edgex-go/blob/master/snap/README.md#configuration).
+Note that you cannot install a local snap without using `--devmode` because some of the services require using `systemd-notify` to startup, which in turn requires the `daemon-notify` interface. If this interface is not connected (or the snap is not in devmode), then the services will fail to start and the snap's installation will be aborted. This is not a problem for the snap as downloaded/installed from the store because the assertions to auto-connect the `daemon-notify` interface are included when you download the snap from the store, but no such assertions exist for the locally developed snap because it hasn't been published to the snap store yet.
```
tonyespy: Your PR description says that you've switched all the daemons from type=notify to type=forking, so section of text shouldn't be necessary anymore?

snap/hooks/configure
```diff
@@ -0,0 +1,54 @@
...
+ALL_SERVICES="$ALL_SERVICES core-data"
+ALL_SERVICES="$ALL_SERVICES core-metadata"
+ALL_SERVICES="$ALL_SERVICES core-config-seed"
+ALL_SERVICES="$ALL_SERVICES core-command"
+
+# support services
+ALL_SERVICES="$ALL_SERVICES support-notifications"
+ALL_SERVICES="$ALL_SERVICES support-logging"
+ALL_SERVICES="$ALL_SERVICES support-scheduler"
+ALL_SERVICES="$ALL_SERVICES support-rulesengine"
```
tonyespy: We just removed support-rulesengine in the previous snap merged, so this should be dropped here as well.

snap/hooks/install
```diff
@@ -76,3 +71,9 @@ if [ ! -f ${SNAP_DATA}/config/security-api-gateway/kong.conf ]; then
     # finally set the make database to be cassandra
     sed -i "s@#database = postgres@database = cassandra@" ${SNAP_DATA}/config/security-api-gateway/kong.conf
 fi
+
+# finally, disable services in the snap that aren't meant to come up by default
+# by default, we want the export-*, support-*, and device-virtual services off
+for svc in export-distro export-client support-notifications support-scheduler support-rulesengine support-logging device-virtual; do
```
tonyespy: Drop support-rulesengine

snap/local/runtime-helpers/bin/mongo-worker-wrapper.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -e
+
+
+while true; do
```
tonyespy: Hmmmm, is this identical to what's in delhi? Shouldn't there be a retry limit so that this doesn't run forever on a mis-configured system? I'm fine if you want to file an issue for this.

snap/local/runtime-helpers/bin/mongod-wrapper.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
+
+# check the mongo database path
+MONGO_DATA_DIR="$SNAP_DATA"/mongo/db
+if [ -e "$MONGO_DATA_DIR" ] ; then
+    rm -rf "${MONGO_DATA_DIR:?}"/*
```
tonyespy: We also fixed this in delhi. Definitely don't want to blow away the db on every restart!

snap/local/runtime-helpers/bin/support-rulesengine-wrapper.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -e
```
tonyespy: This file can be dropped...

snap/local/runtime-helpers/bin/go-services-wrapper.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -e
+
+# wait for consul to come up
+$SNAP/bin/wait-for-consul.sh $1
+
+# if we get here, just assume that consul is up, otherwise this 
+# service will fail after 10 seconds
```
tonyespy: Is 10s the default timeout for all the core services?

snap/snapcraft.yaml
```diff
@@ -26,48 +26,130 @@ confinement: strict
...
+        - mongo-worker
+  core-command:
+    command: bin/fork-wrapper.sh $SNAP/bin/go-services-wrapper.sh core-command --consul
+    daemon: forking
+    plugs: [network, network-bind, ]
+    passthrough:
+      after:
+        - core-config-seed
+        - mongo-worker
+  support-rulesengine:
```
tonyespy: Remove...

snap/README.md
anonymouse64: Yes this is not necessary anymore I will remove it.

snap/hooks/configure
anonymouse64: Good catch

snap/local/runtime-helpers/bin/mongo-worker-wrapper.sh
anonymouse64: Hmm it is indeed the same... I will file an issue for this.

snap/local/runtime-helpers/bin/mongod-wrapper.sh
anonymouse64: Do you want me to incorporate your fix for this into this PR? I was trying to keep the forward-porting PR's more or less in line with how we did them for Delhi, but since I already pull in a couple of commits from other PR's to fix bugs it's probably okay to also pull that one in too

snap/local/runtime-helpers/bin/go-services-wrapper.sh
anonymouse64: yes it's hard-coded in the go code for the edgex-go services

snap/snapcraft.yaml
anonymouse64: yes my cherry-pick must have picked this back up as well

snap/local/runtime-helpers/bin/mongo-worker-wrapper.sh
anonymouse64: Filed as #944

snap/local/runtime-helpers/bin/mongod-wrapper.sh
tonyespy: I'm fine either way, just didn't want it to get lost in the shuffle...

anonymouse64: I've include it here for posterity.

snap/local/runtime-helpers/bin/support-rulesengine-wrapper.sh
anonymouse64: Done

snap/hooks/install
anonymouse64: Done

snap/snapcraft.yaml
```diff
@@ -80,22 +154,40 @@ apps:
   curl:
     command: usr/bin/curl
     plugs: [network]
+  kong:
+    command: bin/kong-wrapper.sh
```
tonyespy: Is there a reason kong is listed under apps when it's part of security-services?

anonymouse64: This is so folks could run kong commands manually, i.e. to get status checks of the gateway. It's only for debugging purposes and the wrapper script is somewhat complicated so providing an app here makes it easier to use.

PR #940, opened 2018-12-14
Refactored logging code to remove error returns from basic log functions.

brandonforster:


tsconn23:
Once @brandonforster returns, I'm going to need to get some context on this PR. Things have moved on quite a bit in this area in the last week. Need to know if there's a particular issue requiring these changes, or if the changes are still relevant in light of where we are now. Putting on "Hold"

internal/support/scheduler/loader.go
```diff
@@ -25,19 +25,22 @@ func AddSchedulers() error {
 	// load data from core-metadata
 	err := loadCoreMetadataInformation()
 	if err != nil {
-		return LoggingClient.Error("failed to load information from core-metadata", err.Error())
+		LoggingClient.Error("failed to load information from core-metadata", err.Error())
```
tsconn23: Should this (and other statements like it) be

`LoggingClient.Error(fmt.Sprintf("failed to load information from core-metadata %s", err.Error()))`

As this is written in the PR branch I believe it won't actually log anything due to mismatched keys to values.

jdharms:
@tsconn23 I think the context is just that it doesn't make sense for the logging functions to return error values that are never checked--they're best effort.

brandonforster:
@tsconn23 Daniel is correct. There was no issue or work here beyond "the logging statements return a value that is used in one place in the code." Further, I think it's self-evident that the best course of action is not to update every instance of a call to logging to do a check and make sure it worked, but rather have logging give it a best effort and do what it can if it fails.

PR #950, opened 2018-12-18
Upgrade snapcraft version and refactor build scripts

### LINKED ISSUE #888: Define a base and use snapcraft 3.X for the snap
Snapcraft 3.0 (which was just released to stable via the snap) will no longer support building natively when specifying what base to use for a snap and instead will only perform builds inside multipass VM's. This has a number of implications for us that we need to consider:

* If we upgrade to specifying a base for the snap, we also need to install multipass when building the snap, and can no longer build the snap inside a docker container
* Even running multipass may be complicated on the VM's setup by the LF - they currently are already running as VM's, so using multipass would mean running a VM in a VM...
* Specifying `base: core18` would give us access to newer packages in the ubuntu archive, which might be useful for getting newer java packages, a newer libzmq3-dev package (4.2 is in bionic, while 4.1 is in xenial), luarocks (which we currently build manually because the version in the xenial archive is too old, but the one in bionic archive is basically same as the version we are building), etc.
* Remote parts no longer work which means we would need to stop using the `go` part and instead use `build-snaps` or manually manage how we get a version of the go compiler to use with the build
* The `core18` base is significantly trimmed down, so we may lose some utilities that we are using, i.e. the security services used to need base64 from the core (aka core16) snap, so that might not be available to us anymore

There may be other implications here not sure yet. For the time being, I recommend just staying with not defining a base for the snap which keeps the legacy behavior however it's not clear how long that will be maintained.

### LINKED ISSUE #1079: Make jenkins jobs for releasing the snap
Currently, we need someone from the LF to run a snapcraft command with the EdgeX login credentials to release a snap revision that was uploaded from the automatic daily jobs to a different channel (all the daily jobs release to edge channels).

We should create a manually triggered jenkins job that just does this with the credentials so that all we have to do when we want to release a new version of the snap to a different channel is trigger that job with the snap revision and the channel to release to as parameters to the job.

I attempted to get the scripts in edgex-go ready to do this in PR https://github.com/edgexfoundry/edgex-go/pull/950 and it seems to work well enough, so the only other part to this would be to create the jenkins jobs in the ci-management repo.

anonymouse64:
* Adjusts the build scripts + Dockerfile to install snapcraft from the stable snap channel rather than through debian packages. As per the snapcraft maintainers, snapcraft 3.0 will not be backported/SRU'd to xenial, and so to get bugfixes and other features with snapcraft 3.0 we need to start installing from the snap. The trick here used to install it is exactly the same on that upstream snapcraft does, but adapted slightly to work for non-amd64 architectures. 
* Adjusts build scripts to allow for a new "release" job in jenkins (specified by the JOB_NAME jenkins environment variable). This job will specify an env var `SNAP_REVISION` and `SNAP_CHANNEL`, and release that snap revision to that snap channel. This release job is meant to only be manually triggered by users in jenkins. 
* Explicitly specify the supported architectures in the snapcraft.yaml so that it fails quicker if someone inadvertently tries to build it on other architectures.

For the corresponding jenkins job see https://github.com/edgexfoundry/ci-management/pull/281

Closes #888 
Closes #1079 (from the edgex-go side)

### DIFF
--- a/snap/Dockerfile.build
+++ b/snap/Dockerfile.build
@@ -1,15 +1,41 @@
 FROM ubuntu:16.04
 
-# this is basically the same as the amd64 upstream:
-# https://github.com/snapcore/snapcraft/blob/master/Dockerfile
+# allow specifying the architecture from the build arg command line
+ARG ARCH
+
+# this is essentially the same as the upstream dockerfile
+# here: https://github.com/snapcore/snapcraft/blob/master/docker/stable.Dockerfile
+# except we also specify the architecture to download so that this works
+# on other architectures
+# basically, we send a command to the snap store for the info on the core +
+# snapcraft snaps, extract the download link from the result and 
+# download and extract the snaps into the docker container
+# we do this because we can't easily run snapd (and thus snaps) inside a docker
+# container without disabling important security protections enabled for 
+# docker containers
 RUN apt-get update && \
   apt-get dist-upgrade --yes && \
   apt-get install --yes \
-  git \
-  snapcraft \
-  && \
+  curl sudo jq squashfs-tools && \
+  curl -s -L $(curl -s -H 'X-Ubuntu-Series: 16' -H "X-Ubuntu-Architecture: $ARCH" 'https://api.snapcraft.io/api/v1/snaps/details/core' | jq '.download_url' -r) --output core.snap && \
+  mkdir -p /snap/core && unsquashfs -n -d /snap/core/current core.snap && rm core.snap && \
+  curl -s -L $(curl -s -H 'X-Ubuntu-Series: 16' -H "X-Ubuntu-Architecture: $ARCH" 'https://api.snapcraft.io/api/v1/snaps/details/snapcraft' | jq '.download_url' -r) --output snapcraft.snap && \
+  mkdir -p /snap/snapcraft && unsquashfs -n -d /snap/snapcraft/current snapcraft.snap && rm snapcraft.snap && \
+  apt remove --yes --purge curl jq squashfs-tools && \
   apt-get autoclean --yes && \
   apt-get clean --yes
+
+# the upstream dockerfile just uses this file locally from the repo, but 
+# rather than copy that file here, we can just download it here
+# while unlikely it is possible that the file location could move in the git repo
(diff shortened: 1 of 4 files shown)


snap/build.sh
```diff
@@ -16,22 +16,33 @@ if [ ! -z "$JENKINS_URL" ]; then
...
-docker build -t edgex-snap-builder:latest -f ${SCRIPT_DIR}/Dockerfile.build $GIT_ROOT
+# build the container image - providing the relevant architecture we're on
+# to determine which snap arch to download in the docker container
+case $(arch) in 
+    x86_64)
+        arch="amd64";;
+    aarch64)
+        arch="arm64";;
+    arm*)
+        arch="armhf";;
```
tonyespy: If **armhf** isn't yet supported, why include it here?

snap/Dockerfile.build
```diff
@@ -1,16 +1,36 @@
...
+# here: https://github.com/snapcore/snapcraft/blob/master/docker/stable.Dockerfile
+# except we also specify the architecture to download so that this works
+# on other architectures
 RUN apt-get update && \
   apt-get dist-upgrade --yes && \
   apt-get install --yes \
-  git \
-  snapcraft \
-  && \
+  curl sudo jq squashfs-tools && \
```
tonyespy: I realize this logic is cut & pasted from the upstream dockerfile, but it might be helpful to add a short bullet list that describes each step below.

It looks to me like you're querying the snap details for both core and snapcraft, downloading both, and unsquashing both under /snap/<name>/current, which is a hacky way to install snapcraft w/out requiring that snapd run in the snap...

snap/build.sh
anonymouse64: Because I think eventually we will build the snap for armhf and it's easier to handle these changes now rather than go back and add them later

snap/Dockerfile.build
anonymouse64: Yes it is a very hacky way to do this, but it's the official upstream way the docker image works, so I think it's okay as long as we document it and explain what and why we're doing it. I will add more comments explaining what we're doing and why though

PR #931, opened 2018-12-13
Issue 882 - core-metadata Command: Eliminate Mongo Types from Models

michaelestrin:
#882 

Signed-off-by: Michael Estrin <m.estrin@dell.com>

PR #955, opened 2018-12-19
Issue 954 - core-metadata Command: Eliminate Mongo Types from Models

michaelestrin:
PR #931 for issue #882 contained incomplete implementation.

* Command get/put fields not populating correctly.
* GetCommands not properly mapping database retrieval result to contract
* GetCommandById not properly handling uuid-based ids
* AddCommand not properly populating Id field with bson.NewObjectId
* etc.

Requires https://github.com/edgexfoundry/blackbox-testing/issues/125 be addressed for core-metadata's command_error_4xx blackbox tests to succeed.


internal/pkg/db/mongo/device_profile.go
```diff
@@ -101,16 +102,22 @@ func (mdp *mongoDeviceProfile) SetBSON(raw bson.Raw) error {
...
-	var commands []models.Command
+	var commands []contract.Command
 
 	// Get all of the commands from the references
 	for _, cRef := range decoded.Commands {
 		var c models.Command
-		err := comCol.FindId(cRef.Id).One(&c)
+
+		query := comCol.FindId(cRef.Id)
+		if n, _ := query.Count(); n > 0 {
```
tsconn23: Bear in mind we want to eventually move this database access code out of this model (see issue #878). It's fine to make these short term edits in order to satisfy blackbox tests and ensure we have clean functionality.

internal/pkg/db/mongo/metadata.go
```diff
@@ -1037,7 +1036,14 @@ func (m *MongoClient) UpdateCommand(c *contract.Command, r *contract.Command) er
...
 	}
 
-	return col.UpdateId(id, r)
+	var err error
+	if model.Id.Valid() {
+		err = s.DB(m.database.Name).C(db.Command).UpdateId(model.Id, model)
+	} else {
+		query := bson.M{"uuid": model.Uuid}
+		err = s.DB(m.database.Name).C(db.Command).Update(query, model)
+	}
```
tsconn23: Help me understand -- We instantiate a models.Command above (line 1018) and populate it from the contract. We use that instance in the UpdateId/Update calls, but we never set any new values on it. Since we map it from `c`, inferring only from the logic above, I guess the `model` instance contains all the new values we need, so we just update. If that's so, why preserve the lines setting `r` properties from `c`? It looks to me like `r` is useless here.

internal/pkg/db/mongo/models/action.go
```diff
@@ -14,42 +14,8 @@
 
 package models
 
-import (
-	contract "github.com/edgexfoundry/edgex-go/pkg/models"
-	"github.com/pkg/errors"
-)
-
 type Action struct {
```
tsconn23: I would probably do away with this type since everything about it is managed in the Get/Put types. It's literally got zero responsibilities and exists only for this reason:
```
type Get struct {
	Action `bson:",inline"`
}
```
which is an artifact of the way the original contract was written -- kind of like with `BaseObject`

internal/pkg/db/mongo/models/command.go
```diff
@@ -88,10 +88,12 @@ func (c *Command) FromContract(from contract.Command) error {
 	}
 
 	c.Name = from.Name
+	c.Get = &Get{}
 	err := c.Get.FromContract(*from.Get)
```
tsconn23: `*from.Get`

We're eventually going to get rid of these pointers, right? This is just a holdover from the existing code and is pending @jdharms work on removing the MemoryDB, correct?

michaelestrin: Correct.

internal/pkg/db/mongo/models/action.go
michaelestrin: Yep.  Removed.

internal/pkg/db/mongo/device_profile.go
michaelestrin: Moved database code out of mongoDeviceProfile model.

internal/pkg/db/mongo/metadata.go
michaelestrin: Inherited code but will take responsibility for it.  Refactored to remove second parameter from dbClient.UpdateCommand().

PR #791, opened 2018-11-13
Update security services for Delhi to use go binaries

anonymouse64:
Fixes #623 for Delhi.

This PR does a few things:
* It switches to using the new `vault-worker` and `pkisetup` binaries from security-secret-store for generating certs and loading them into vault, then also uploading them into kong.
* It now correctly checks if the certs have already been generated when re-running and doesn't regenerate them if they already exist
* We also now clear the glide cache before all go build parts that use glide because there's conflicts in the version of packages that `edgex-go`, `security-security-secret-store`, and `security-api-gateway`. This slows down builds somewhat as the packages will have to be fetched clean for each part, but ensures the builds always pass
* The kong routes now correctly present the certificate generated for Kong. 

The main things to test with this PR is that:
* The `security-services` service in the snap starts up correctly on install and on reboot. (you can inspect logs more easily with `journalctl -ef -u snap.edgexfoundry.security-services.service` than `sudo snap logs edgexfoundry.security-services`, `snap logs` seems to miss some things)
* The certificate presented by kong when accessing the proxy are signed by the `EdgeXFoundryCA.pem` file we generate using pkisetup. This can be tested by using curl with the `--cacert` option pointed at the generated root CA file like so (using the route configured for `core-command` as an example):
```bash
sudo cp /var/snap/edgexfoundry/current/vault/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem .
curl --cacert EdgeXFoundryCA.pem https://localhost:8443/command
```
* The certificate files are not regenerated after restarting the snap. To test this, run the previous test to make sure the certificates are correct, then do:
```
sudo snap restart edgexfoundry
```
And after all the services are back up again, do the same test with the EdgeXFoundryCA.pem file you originally copied, curl should still be able to connect to the server over HTTPS successfully.

PR #960, opened 2018-12-20
Update security services from Delhi, split into separate daemons

### LINKED ISSUE #623: Update snap security services to go implementation
Currently, the master branch of the snap is still using the california version of the `security-secret-store` parts, using the shell scripts originally wrote in bash/shell. We need to update this to use the go versions recently implemented.

### LINKED ISSUE #775: Split security services up in snap
Once proper support for `before`/`after` has landed in snapd stable, we will want to go back and split up the security services into independent daemons.
However, because `after` does implement direct dependencies, it won't block service startup if a service is disabled. So what we can do is we can order all of the other services to run after the security services and if the security services are disabled by the user, then everything will still be able to startup successfully. 

Strictly speaking right now there's not a direct dependency between any of the normal services and the actual security services, but this dependency is being discussed and may eventually present itself.

### LINKED ISSUE #916: snap logs shouldn't contain curl output
Currently all of the go services use a small wrapper script which uses curl to ensure that consul has at least started up, but this produces unnecessary and potentially confusing output in the logs:
```
Dec 10 15:04:57 hostname systemd[1]: Starting Service for snap application edgexfoundry.core-data...
Dec 10 15:04:57 hostname systemd[1]: Started Service for snap application edgexfoundry.core-data.
Dec 10 15:04:57 hostname edgexfoundry.core-data[5635]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
Dec 10 15:04:57 hostname edgexfoundry.core-data[5635]:                                  Dload  Upload   Total   Spent    Left  Speed
Dec 10 15:04:57 hostname edgexfoundry.core-data[5635]: [158B blob data]
Dec 10 15:04:58 hostname edgexfoundry.core-data[5635]: ERROR: 2018/12/10 15:04:58 error reading from Consul
Dec 10 15:04:59 hostname edgexfoundry.core-data[5635]: ERROR: 2018/12/10 15:04:59 error reading from Consul
Dec 10 15:05:00 hostname edgexfoundry.core-data[5635]: ERROR: 2018/12/10 15:05:00 error reading from Consul
Dec 10 15:05:01 hostname edgexfoundry.core-data[5635]: ERROR: 2018/12/10 15:05:01 error reading from Consul
Dec 10 15:05:02 hostname edgexfoundry.core-data[5635]: INFO: 2018/12/10 15:05:02 Service dependencies resolved...
Dec 10 15:05:02 hostname edgexfoundry.core-data[5635]: INFO: 2018/12/10 15:05:02 Starting edgex-core-data 0.7.1
Dec 10 15:05:02 hostname edgexfoundry.core-data[5635]: INFO: 2018/12/10 15:05:02 This is the Core Data Microservice
Dec 10 15:05:02 hostname edgexfoundry.core-data[5635]: INFO: 2018/12/10 15:05:02 Service started in: 4.101312641s
Dec 10 15:05:02 hostname edgexfoundry.core-data[5635]: INFO: 2018/12/10 15:05:02 Listening on port: 48080
```
We should ensure that the ordering specs we have with `after` work correctly in conjunction with the services waiting for their dependencies so that we don't need this curl command anymore. However, if that doesn't work, we should just silence the curl output in the script

### LINKED ISSUE #921: Add jq to snap for debugging
We already add `curl` to the snap and expose it for simple debugging, but we should also add `jq` for easy consumption of the JSON returned from `curl`.

anonymouse64:
This PR updates the security services in the snap to use delhi code, and also refactors how they are used within the snap so that they are now separate daemons rather than one single monolithic "security-services" daemon. As such, it partly builds on the work done in #791, but goes further and splits up the security services, which we couldn't do then as the various fixes in snapd needed hadn't landed yet at that point.  ~~Note that we are still building security-api-gateway from the `delhi` branch rather than master as all the delhi changes have not yet been forward-ported to master. See open PR's: https://github.com/edgexfoundry/security-api-gateway/pulls~~. **Update**: the upstream delhi forwardport PR has been merged and we can build from master now for security-api-gateway. 

Additionally note that while now the individual components for the security services are split up, they are meant to be configured as a whole, either using the keys `security-api-gateway` or `security-secret-store` with snap set to enable/disable them. This means that when a user disables one of those, we will disable all the individual daemons for that "service". Also note that since `security-api-gateway` depends on `security-secret-store` for accessing it's certificates, if one disables security-secret-store, `security-api-gateway` will also be disabled.

Also some miscellaneous things like adding jq as an app for the snap, silencing curl output when checking for consul status, and adding the removable-media interface to the mongo app.

# Testing
* All enabled services startup properly (except the oneshot daemons, which should still run, but showup in `snap services` as `inactive`. Test this after installation and also after a reboot. You should only see `core-config-seed`, `pkisetup`, `edgexproxy`, `vault-worker`, and `mongo-worker` when running 
```
snap services edgexfoundry | grep inactive | grep enabled
``` 
* Test that the security services as a whole can be turned on/off with the `security-api-gateway` and `security-secret-store` config keys. for example after installing the snap and verifying everything starts up, run:
`sudo snap set edgexfoundry security-api-gateway=off` 

which should result in `kong-daemon`, `cassandra`, and `edgexproxy` all being disabled. Similarly for `security-secret-store` except `vault` and `vault-worker` should also be disabled.

* The certificate presented by kong when accessing the proxy are signed by the EdgeXFoundryCA.pem file we generate using pkisetup. This can be tested by using curl with the --cacert option pointed at the generated root CA file like so (using the route configured for core-command as an example):
```
sudo cp /var/snap/edgexfoundry/current/vault/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem .
curl --cacert EdgeXFoundryCA.pem https://localhost:8443/command
```
* The certificate files are not regenerated after restarting the snap. To test this, run the previous test to make sure the certificates are correct, then do:
```
sudo snap restart edgexfoundry
```
And after all the services are back up again, do the same test with the EdgeXFoundryCA.pem file you originally copied, curl should still be able to connect to the server over HTTPS successfully.

* Test OAuth is working with Kong. To run this test, you can just refer to [this comment](https://github.com/edgexfoundry/edgex-go/pull/791#issuecomment-438534454) for how to test that.

Closes #921
Closes #775 
Closes #623 
Closes #916


anonymouse64:
> I've re-tested this a number of times and so far I've only been able to reproduce this scenario once. Is this a race condition that I'm seeing (i.e. will the core services show as inactive before they've actually started)?

This could be a race condition, I do believe it is the case that services which have not yet been started because their dependent services have not finished starting show up in snapd as `inactive`. If you ever see this again, you should see what `systemctl status` says about the service.

> 
> **Note** - I'm also seeing weird behavior on 16.04 too. After installing the snap and rebooting, I'm prevented from logging in. I enter my password, it's accepted, but then the screen resets and I'm back at the login prompt. I'll re-install a fresh 16.04 tomorrow. It could be that there's something wrong with this install (again a Parallels guest).

This is very odd, I've never seen this with the snap, but I have encountered a similar error with Ubuntu 18.04 natively installed because I had a gnome extension installed which conflicted with the default one installed and I had to remove the default one shipped with Ubuntu so that the one I installed was the only one used. I have no idea how the snap could block login though. Are you able to reach the machine over SSH when this happens to see what the status of the snap services are in this scenario?

anonymouse64:
I just tested and immediately after being able to login over SSH I see the following services listed as `inactive`:

```
$ snap services edgexfoundry | grep enabled | grep inactive
edgexfoundry.core-command           enabled   inactive  -
edgexfoundry.core-data              enabled   inactive  -
edgexfoundry.core-metadata          enabled   inactive  -
edgexfoundry.edgexproxy             enabled   inactive  -
edgexfoundry.kong-daemon            enabled   inactive  -
edgexfoundry.mongo-worker           enabled   inactive  -
edgexfoundry.pkisetup               enabled   inactive  -
edgexfoundry.sys-mgmt-agent         enabled   inactive  -
edgexfoundry.vault-worker           enabled   inactive  -
```

but after waiting about a minute I see these:

```
$ snap services edgexfoundry | grep enabled | grep inactive
edgexfoundry.core-config-seed       enabled   inactive  -
edgexfoundry.edgexproxy             enabled   inactive  -
edgexfoundry.mongo-worker           enabled   inactive  -
edgexfoundry.pkisetup               enabled   inactive  -
edgexfoundry.vault-worker           enabled   inactive  -
```

which is what we should see so I think this is fine.

Issue #956, opened 2018-12-19
core-metadata Command: update query methods


michaelestrin:
Update core-metadata command's query methods from ```accepting function parameter pointer to result``` to ```return function-allocated structure```.  These changes will align core-metadata's command implementation with the implementation approach elsewhere.

michaelestrin:
db/mongo/metadata.go - AddCommand().  Refactor to remove redundant assignments (already made in call to FromContract).

michaelestrin:
#972

PR #972, opened 2019-01-02
Issue 956 - core-metadata Command: update query methods

michaelestrin:
* Removed redundant assignments.
* Converted GetCommandById().
* Converted GetCommandByName().
* Converted GetAllCommands().

Issue #956

### DIFF
--- a/internal/pkg/db/mongo/metadata.go
+++ b/internal/pkg/db/mongo/metadata.go
@@ -532,8 +532,10 @@ func (m *MongoClient) AddDeviceProfile(dp *contract.DeviceProfile) error {
 		return db.ErrNotUnique
 	}
 	for i := 0; i < len(dp.Commands); i++ {
-		if err := m.AddCommand(&dp.Commands[i]); err != nil {
+		if newId, err := m.AddCommand(dp.Commands[i]); err != nil {
 			return err
+		} else {
+			dp.Commands[i].Id = newId
 		}
 	}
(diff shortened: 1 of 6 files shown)


internal/pkg/db/mongo/metadata.go
```diff
@@ -1034,13 +1031,12 @@ func (m *MongoClient) AddCommand(c *contract.Command) error {
 	s := m.session.Copy()
 	defer s.Close()
 
+	c.Id = ""
+	c.Created = 0
```
tsconn23: Is there a specific use case you saw that requires these reset lines? Also, c parameter should not be a pointer. For example:

https://github.com/edgexfoundry/edgex-go/blob/29a4bebcaf459d7aad88a7f62a5784fcebb874d5/internal/pkg/db/mongo/metadata.go#L643

michaelestrin: Thinking was that contract.Command passed into AddCommand shouldn't be able to dictate its Id, Uuid, or Created fields.

Removed assignments.  Also corrected the AddCommand function signature.

tsconn23:
As part of this PR @michaelestrin has changed the receiver of some methods to not use pointers. This is because no modification of the receiver's state is taking place. A follow up issue has been created to make this uniform throughout the mongo/metadata.go file.

https://github.com/edgexfoundry/edgex-go/issues/974

PR #975, opened 2019-01-02
Issue 974 - Update metadata function signatures

michaelestrin:
* Updated metadata function signatures so they reference receiver as pointer only if the corresponding function modifies the receiver (in this case, none currently do).


tsconn23:
This is the first part of an ongoing effort to remove the need for pointer receivers on the DBClient implementation. No state is ever modified internally by these functions, so no need for a pointer. Yes, it will be inconsistent while this is happening, and we will need to rework client initialization so that connection is a part of it.

Issue #977, opened 2019-01-03
Mongo test fails


jduranf:
When trying to perform Mongo tests, this failure happens:
```
~/go/src/github.com/edgexfoundry/edgex-go/internal/pkg/db/mongo$ go test -tags mongoRunning
--- FAIL: TestMongoDB (0.07s)
	client_test.go:24: This test needs to have a running mongo on localhost
	db_data.go:251: Error updating reading ObjectIDs must be exactly 12 bytes long (got 0)
FAIL
exit status 1
FAIL	github.com/edgexfoundry/edgex-go/internal/pkg/db/mongo	0.068s
```

If `TestDataDB()` is commented, then another error happens while performing `TestMetadataDB()`:

```
~/go/src/github.com/edgexfoundry/edgex-go/internal/pkg/db/mongo$ go test -tags mongoRunning
--- FAIL: TestMongoDB (0.04s)
	client_test.go:24: This test needs to have a running mongo on localhost
panic: invalid input to ObjectIdHex: "5a846fa2-09a6-4627-bce8-052bc1200525" [recovered]
	panic: invalid input to ObjectIdHex: "5a846fa2-09a6-4627-bce8-052bc1200525"

goroutine 6 [running]:
testing.tRunner.func1(0xc42011c0f0)
	/usr/local/go/src/testing/testing.go:742 +0x29d
panic(0x71e140, 0xc420217ab0)
	/usr/local/go/src/runtime/panic.go:502 +0x229
github.com/edgexfoundry/edgex-go/vendor/github.com/globalsign/mgo/bson.ObjectIdHex(0xc420214600, 0x24, 0x0, 0x0)
	/home/jduran/go/src/github.com/edgexfoundry/edgex-go/vendor/github.com/globalsign/mgo/bson/bson.go:212 +0x163
github.com/edgexfoundry/edgex-go/internal/pkg/db/test.populateCommand(0x7efc20, 0xc4201220e0, 0x64, 0x0, 0x0, 0xc42027bc88, 0x5)
	/home/jduran/go/src/github.com/edgexfoundry/edgex-go/internal/pkg/db/test/db_metadata.go:131 +0x53
github.com/edgexfoundry/edgex-go/internal/pkg/db/test.testDBCommand(0xc42011c0f0, 0x7efc20, 0xc4201220e0)
	/home/jduran/go/src/github.com/edgexfoundry/edgex-go/internal/pkg/db/test/db_metadata.go:573 +0x265
[...]

Issue #976, opened 2019-01-02
Eliminate MongoClient Receiver Pointers


tsconn23:
In discussion with @michaelestrin today, we thought it would be a good idea to revise the MongoClient implementation of the DBClient to not use receiver pointers in any of the implementation methods. The primary reason for this is that none of the DBClient methods modify anything about the state of the MongoClient. Since the MongoClient struct is tiny, we think the cost to copy it will be insignificant.

The methods found in mongo/metadata.go have already been modified per #974 without any negative impacts w/r/t testing. The same should be done for data, export and notifications in the internal/db/mongo package.

Also
- Refactor NewClient to perform connection during instantiation. This will allow us to discard the current `config` member on the MongoClient struct.
- Eliminate `getCurrentMongoClient()` through ongoing work to decouple Mongo types from models (since Mongo models currently violate separation of concerns -- [example](https://github.com/edgexfoundry/edgex-go/blob/17d2ea12037649a4bd34c24067a6c47d717bdfab/internal/pkg/db/mongo/device_profile.go#L95))
- Retry logic in services will now need to utilize NewClient() instead of Connect()
- The means of calling `CloseSession()` will need to be refactored since its pointer will no longer reference a global variable. This could necessitate its removal from the DBClient interface.

PR #979, opened 2019-01-03
Removed receiver pointers from MongoClient

tsconn23:
#976 

Planning to address issues raised in #977 in separate PR.

### DIFF
--- a/internal/export/db.go
+++ b/internal/export/db.go
@@ -20,8 +20,6 @@ import (
 type DBClient interface {
 	CloseSession()
 
-	Connect() error
-
 	// ********************** REGISTRATION FUNCTIONS *****************************
 	// Return all the registrations
 	// UnexpectedError - failed to retrieve registrations from the database
--- a/internal/pkg/db/mongo/client.go
+++ b/internal/pkg/db/mongo/client.go
@@ -14,54 +14,48 @@
 package mongo
 
 import (
-	"errors"
 	"strconv"
 	"time"
 
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
-	"github.com/google/uuid"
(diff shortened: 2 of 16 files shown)


internal/pkg/db/mongo/client.go
```diff
@@ -14,71 +14,61 @@
...
 	}
 }
 
 // Get the current Mongo Client
-func getCurrentMongoClient() (*MongoClient, error) {
-	if currentMongoClient == nil {
-		return nil, errors.New("No current mongo client, please create a new client before requesting it")
-	}
-
+func getCurrentMongoClient() (MongoClient, error) {
```
michaelestrin: Why not remove `error` from the return list?

internal/export/db.go
```diff
@@ -20,8 +20,6 @@ import (
 type DBClient interface {
 	CloseSession()
```
michaelestrin: If creation is a detail outside the scope of the contract, it feels like destruction should be as well.

tsconn23: Destruction is dependent on the instance though. Creation happens before the instance exists. Very metaphysical stuff ;-)

tsconn23:
@michaelestrin Regarding retry on connection failure, it's in there. Look at Retry() in init.go which calls `connectToDatabase()`.

michaelestrin:
@tsconn23 Retry exists if the initial connection attempt fails exists.  I was referring to retry on loss of connection.  Once the initial connection is made, Retry() exits.

internal/export/db.go
michaelestrin: Just as creation is.  I guess what I'm saying is we should be instantiating and destructing at the same level.  Since we create using a switch, perhaps we should destruct at that same level using a switch.

Issue #880, opened 2018-12-03
core-metadata Device: Eliminate Mongo Types from Models 


tsconn23:
For all contract models (edgex-go/pkg/models) currently used by the core-metadata service when serving the following route and all sub-routes

- [/api/v1/device](https://github.com/edgexfoundry/edgex-go/blob/31e4491729ff378ebd20d54d11bc576314cbf1f0/internal/core/metadata/router.go#L51)

1.) Modify the ID property to be a string instead of a bson.ObjectId
2.) Remove bson serialization annotations
3.) It will still be necessary to handle BSON IDs natively within the Mongo persistence layer, but these should be transformed to string values and handed out to the rest of the application and external consumers through the contract models.

The solution should be backward compatible with existing data, such as that found in the blackbox tests. GET / PUT / DELETE operations should function when being given either a BSON ID (existing data) or UUID. POST operations that create new records should cause UUID values to be generated and returned.

The core-data service has already been refactored in this manner. Please refer to the additional information on the relevant PR #856 about implementation specifics. In particular, the internal/pkg/db/mongo/device.go model [utilizes DBRefs](https://github.com/edgexfoundry/edgex-go/blob/31e4491729ff378ebd20d54d11bc576314cbf1f0/internal/pkg/db/mongo/device.go#L37) and violates separation of concerns by [accessing the database](https://github.com/edgexfoundry/edgex-go/blob/31e4491729ff378ebd20d54d11bc576314cbf1f0/internal/pkg/db/mongo/device.go#L98). The same strategy used in core-data for events that referred to readings should be undertaken here.

The above device model also has child properties referring to DeviceService and DeviceProfile. There are issues created for the same refactoring to be done for those metadata application paths. It might be a good idea to sequence this issue after those (see #877, #878)

greenrobot:
> POST operations that create new records should cause UUID values to be generated and returned.

This part caught my attention. Forcing IDs to be UUID would have performance and space impact: Integer based IDs have significant better efficiency in databases like SQLite and ObjectBox (and probably any SQL based solution where you could opt to an 64 bit int).

It would be great to have clarify the intention here.

michaelestrin:
Dependency on #878

PR #985, opened 2019-01-05
Issue 877 - core-metadata DeviceService: Eliminate Mongo Types from Models

michaelestrin:
Issue 877 - core-metadata DeviceService: Eliminate Mongo Types from Models

* Updated contract to return uuid *string for Id and removed BSON annotations.
* Updated query methods to eliminated references and return result values instead.
* Updated AddDeviceService() to return Id.
* Collapsed common code that translates Id into bson.ObjectId or UUID.

#877

### DIFF
--- a/internal/core/metadata/rest_device.go
+++ b/internal/core/metadata/rest_device.go
@@ -81,10 +81,10 @@ func restAddNewDevice(w http.ResponseWriter, r *http.Request) {
 
 	// Service Check
 	// Try by name
-	err = dbClient.GetDeviceServiceByName(&d.Service, d.Service.Service.Name)
+	d.Service, err = dbClient.GetDeviceServiceByName(d.Service.Service.Name)
 	if err != nil {
 		// Try by ID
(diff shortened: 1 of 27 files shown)


michaelestrin:
Requires corresponding blackbox test change (https://github.com/edgexfoundry/blackbox-testing/pull/129) in order for all metadata devices service tests to pass.  

Note the core-metadata-importer blackbox collection will fail to import device report data.  This should be fixed when #880 is worked.

michaelestrin:
Rebased

internal/core/metadata/rest_device.go
```diff
@@ -81,10 +81,10 @@ func restAddNewDevice(w http.ResponseWriter, r *http.Request) {
 
 	// Service Check
 	// Try by name
-	err = dbClient.GetDeviceServiceByName(&d.Service, d.Service.Service.Name)
+	d.Service, err = dbClient.GetDeviceServiceByName(d.Service.Service.Name)
```
tsconn23: `d.Service.Service.Name`

That's lovely. Clearly something needs refactoring here...

Probably out of scope for this PR

PR #989, opened 2019-01-07
Dockerfile restructuring

rsdmike:
This is what it would look like if we moved Dockerfile to their appropriate directories and called the file `Dockerfile`


rsdmike:
Closing since commits weren't signed

mhall119:
@rsdmike you can redo the commit with --signoff and then git push with --force to the same remote branch and it will update the pull request

Issue #981, opened 2019-01-03
Suggestion: Rename files in /docker to be [servicename].Dockerfile instead of Dockerfile.[servicename]


rsdmike:
In working in an IDE like VSCode, the files in the docker directory aren't recognized as actual Dockerfiles by the editor given their file extension isn't .Dockerfile. Furthermore, they are not recognized as dockerfiles by the Docker extension thus preventing them from showing up in menu options to build with the IDE. It's a small thing for developer experience, but I'd be more then willing to help out in updating the names and makefiles. However, I'm not sure the impact it would have on CI/CD if any. 

![image](https://user-images.githubusercontent.com/22204562/50666560-9743c480-0f72-11e9-9c26-f353f3472076.png)

As a second thought, It might be a better idea yet to move the Dockerfiles into their respective cmd/ folders and simply call them "Dockerfile" which could potentially simplify any tooling such that the the filename would not need to be specified. 

Thoughts?

tsconn23:
@rsdmike The top of line proposal is easy enough and will be self-contained to the edgex-go repo. The CI/CD pipeline simply calls `make docker` which is driven from the makefile. So if the filenames change and the make targets are updated, all should continue to work. We'll know quickly from the verify jobs if that's not the case.

I'm not sure of the benefit of the second thought. FWIW I use GoLand. Waiting to see if any other VSCode users out there chime in on this.

andresrinivasan:
I solved the first problem by adding another file association to my VSC `settings.json` as this is an issue with other projects I am working with:

```json
"files.associations": {
        "Dockerfile.*": "dockerfile"
    }
```

I'm not a huge fan of creating a directory for a single file.

jpwhitemn:
Agree with Andre on this.  That is what file association preferences in any IDE are about.  As soon as we change it to meet the default for one IDE, it will run counter to the default of another IDE.  Not in favor of making changes unless there appears to be a bigger benefit.  

Mike, thanks for your suggestion and if you still feel this is needed, I'd encourage you to make your case on a broader context and let us get input from DevOps working group chair on impact.


    On Thursday, January 3, 2019, 5:54:59 PM CST, André Srinivasan <notifications@github.com> wrote:  
 
 
I solved the first problem by adding another file association to my VSC settings.json as this is an issue with other projects I am working with:
"files.associations": {
        "Dockerfile.*": "dockerfile"
    }
I'm not a huge fan of creating a directory for a single file.

[...]

PR #990, opened 2019-01-07
Restructuring dockerfiles and Updating makefile

rsdmike:
This is what it would look like if we moved Dockerfile to their appropriate directories and called the file `Dockerfile`. #981


tsconn23:
With no activity on this and little momentum for the change from the community, I'm inclined to close.

rsdmike:
I feel ya...though I want to stress that there was two primary points to this pull request. One was naming to follow Docker standards, the second was location and to have them next to their main.go file and based of the discussion it definitely seemed like there was confusion on the thread as to what that looks like. So at the very least I'd like to get @jamesrgregg as the DevOps guy to see if he has any input for the pipeline and experience he has.  So give it another week, please :) Then we can close?

Issue #969, opened 2019-01-01
Config-seed: Refactor stem assignment for configuration path


tsconn23:
Currently all Go service configurations in the registry (Consul) live at the following stem "config/V2/". This is governed by the ConfigV2Stem constant. We need to eliminate this constant (since V2 has no meaning for a 1.0 release) and create a new one (perhaps `ConfigRegistryStem`) that points to "edgex/core/1.0". The Config-seed should utilize this new constant to write service configurations for the edgex-go repo beneath this new stem.

The edgex-go services which utilize the old ConfigV2Stem constant will need to be repointed to the new constant. This should affect the `connectToConsul()` and `listenForConfigChanges()` functions in each service's init.go

PR #991, opened 2019-01-07
Updated config stem.

brandonforster:
Issue #969


internal/constants.go
```diff
@@ -19,7 +19,7 @@ const (
 	BootTimeoutDefault   = 30000
 	ClientMonitorDefault = 15000
 	ConfigFileName       = "configuration.toml"
-	ConfigV2Stem         = "config/V2/"
+	ConfigRegistryStem   = "edgex/core/1.0"
```
tsconn23: Missing slash at the end. Result is as follows in Consul.

screen shot 2019-01-08 at 12 15 47 pm

PR #996, opened 2019-01-09
Issue 877 - core-metadata DeviceService: Eliminate Mongo Types from Models (Refactored)

michaelestrin:
Issue #877 - core-metadata DeviceService: Eliminate Mongo Types from Models (Refactored)

* Eliminated unnecessary mongoDeviceService struct and moved SetBSON/GetBSON into DeviceService model.

* Separated responsibilities -- Mongo model updated to represent data as it exists in Mongo (including DBRefs) and conversion (ToContract/FromContract) now responsible for transform to/from DBRefs.

* Added addressableTransform contract to facilitate conversion of addressable to/from DBRef within individual model's ToContract/FromContract.


internal/pkg/db/mongo/models/addressable.go
```diff
@@ -37,7 +43,7 @@ type Addressable struct {
 	Origin     int64         `bson:"origin"`
 }
 
-func (a Addressable) ToContract() contract.Addressable {
+func (a *Addressable) ToContract() contract.Addressable {
```
tsconn23: I don't think the receiver needs to be a pointer here. No internal state is mutated. Small potatoes.

internal/pkg/db/mongo/models/event.go
```diff
@@ -34,7 +34,7 @@ type Event struct {
 	dbRefs   []mgo.DBRef
 }
 
-func (e Event) ToContract() contract.Event {
+func (e *Event) ToContract() contract.Event {
```
tsconn23: Again, I don't understand why some of these receivers were made into pointers.

tsconn23:
Overall LGTM. Will leave as is for others who may be interested to review. Would like to know about the rationale for the new pointer receivers. It's a small point, but would like to understand.

jdharms:
> Overall LGTM. Will leave as is for others who may be interested to review. Would like to know about the rationale for the new pointer receivers. It's a small point, but would like to understand.

From The Go Programming Language (book):

> In a realistic program, convention dictates that if any method of Point has a pointer receiver, then all methods of Point should have a pointer receiver, even ones that don’t strictly need it.

Where "Point" is a type they're writing for illustrative purposes in the book.

I think this is @michaelestrin 's reasoning here as well.

michaelestrin:
@jdharms beat me to it.  My response was in flight when his showed up.  That's exactly it.

tsconn23:
@jdharms @michaelestrin You missed a couple spots then.

```
func (e Event) GetBSON() (interface{}, error) 
func (e Event) GetDBRefs() []mgo.DBRef 
```

michaelestrin:
> @jdharms @michaelestrin You missed a couple spots then.
> 
> ```
> func (e Event) GetBSON() (interface{}, error) 
> func (e Event) GetDBRefs() []mgo.DBRef 
> ```

Fixed.

Issue #998, opened 2019-01-09
Refactor core-data events


michaelestrin:
Separate responsibilities -- update Mongo model to represent data as it exists in Mongo (including DBRefs) and conversion (ToContract/FromContract) to be responsible for transform to/from DBRefs.

PR #1001, opened 2019-01-10
Issue 998 - Refactor core-data events

michaelestrin:
Separated responsibilities -- updated Mongo model to represent data as it exists in Mongo (including DBRefs) and conversion (ToContract/FromContract) to be responsible for transform to/from DBRefs.

#998


brandonforster:
recheck

Issue #970, opened 2019-01-01
Config-seed: Writable Section for all Edgex-go Services


tsconn23:
Per the [Edinburgh configuration proposal](https://wiki.edgexfoundry.org/download/attachments/7602423/EdgeX-Config-Proposal-v3.pdf?version=1&modificationDate=1546379949609&api=v2) each service should support a Writable section in their respective configuration. Any setting located in the Writable section can be changed in real time with the change taking effect immediately without a service restart. The two key aspects of this include

- Changes to the `configuration.toml` files wherein settings not assigned to a section will be grouped under a new Writable section. _Be sure to change the `res/docker/configuration.toml` as well!_
- Changes to each service's Configuration types.
- Changes to the path of each service's Consul watcher

For the first point, taking `core-data` as an example, the `res/configuration.toml` file contains the following free-form settings
```
DeviceUpdateLastConnected = false
MetaDataCheck = false
PersistData = true
ServiceUpdateLastConnected = false
ValidateCheck = false
```
These should all be grouped like so:
```
[Writable]
DeviceUpdateLastConnected = false
MetaDataCheck = false
PersistData = true
ServiceUpdateLastConnected = false
ValidateCheck = false
```

For the second point, in order to support the new structure in the `configuration.toml`, each service that requires a Writable section will need to have its own WritableInfo type. The above settings are currently loaded into core-data's `ConfigurationStruct` via [these properties](https://github.com/edgexfoundry/edgex-go/blob/29a4bebcaf459d7aad88a7f62a5784fcebb874d5/internal/core/data/config.go#L19). Instead, these properties will need to be moved to a WritableInfo type specific to core-data (and by extension each service since each one could be unique) and the new type should be included in the ConfigurationStruct.
```
type ConfigurationStruct struct {
        Writable                   WritableInfo
	MessageQueue               config.MessageQueueInfo
	Clients                    map[string]config.ClientInfo
	Databases                  map[string]config.DatabaseInfo
	Logging                    config.LoggingInfo
	Registry                   config.RegistryInfo
	Service                    config.ServiceInfo
}
```

References in the code which used to point to the `ConfigurationStruct` properties will then need to be corrected to point to the `ConfigurationStruct.Writable` properties.

[...]

PR #1002, opened 2019-01-10
Updated core data to use writable config.

brandonforster:
Issue #970

### DIFF
--- a/cmd/core-data/res/configuration.toml
+++ b/cmd/core-data/res/configuration.toml
@@ -1,3 +1,4 @@
+[Writable]
 DeviceUpdateLastConnected = false
 MetaDataCheck = false
 PersistData = true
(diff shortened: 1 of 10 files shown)


tsconn23:
@brandonforster When you have a second, see if you can reproduce / troubleshoot the following:

1.) Build branch
2.) Start local consul via `consul agent -dev`
3.) Run config seed to populate
4.) Start core-data service with `--consul` flag
5.) Get the following endpoint `http://localhost:48080/api/v1/config`

Here's what I see:
screen shot 2019-01-10 at 1 15 29 pm

brandonforster:
It turns out that modifying the prefix makes all sorts of things mad. The prefix HAS to be that otherwise nothing with Consul works at all.

I was able to achieve the functionality we want by simply only looking at the `Writable` object in the `ListenForChanges` function and ignoring the rest of the `ConfigurationStruct`

Issue #1012, opened 2019-01-14
export-client: Eliminate Mongo types from Models


jdharms:
For all contract models currently used by the export-client service:

1.) Modify the ID property to be a string instead of a bson.ObjectId
2.) Remove bson serialization annotations
3.) It will still be necessary to handle BSON IDs natively within the Mongo persistence layer, but these should be transformed to string values and handed out to the rest of the application and external consumers through the contract models.

The solution should be backward compatible with existing data, such as that found in the blackbox tests. GET / PUT / DELETE operations should function when being given either a BSON ID (existing data) or UUID. POST operations that create new records should cause UUID values to be generated and returned.

This will also involve some work to refactor the export-client model usage to more closely conform to the other services' model.

PR #1013, opened 2019-01-14
Refactor export client service models

jdharms:
#1012 Refactors the export client service's models to conform
to the other services and abstract away Mongo specific fields.


internal/pkg/db/mongo/models/registration.go
```diff
@@ -0,0 +1,91 @@
...
+	r.Filter.FromContract(c.Filter)
+
+	r.Encryption = EncryptionDetails{}
+	r.Encryption.FromContract(c.Encryption)
+
+	r.Compression = c.Compression
+	r.Enable = c.Enable
+	r.Destination = c.Destination
+
+	//return toContractId(r.ID, r.Uuid), nil
```
michaelestrin: Dependency on code in #878

internal/pkg/db/mongo/export.go
```diff
@@ -14,51 +14,73 @@
...
 	s := mc.getSessionCopy()
 	defer s.Close()
 
-	reg.Created = db.MakeTimestamp()
-	reg.ID = bson.NewObjectId()
+	registration := &models.Registration{}
+	id, err := registration.FromContract(r)
+
+	if err != nil {
+		return id, err
```
michaelestrin: Returning an "empty" value for id instead is consistent with how error conditions are handled elsewhere.

internal/pkg/db/mongo/export.go
```diff
@@ -14,51 +14,73 @@
...
+	col := s.DB(mc.database.Name).C(db.ExportCollection)
+
+	registration.Created = db.MakeTimestamp()
 
 	// Add the registration
-	err := s.DB(mc.database.Name).C(db.ExportCollection).Insert(reg)
+	err = col.Insert(registration)
 	if err != nil {
-		return reg.ID, err
+		return id, err
```
michaelestrin: Returning an "empty" value for id instead is consistent with how error conditions are handled elsewhere.  If not, this `if` statement is redundant.

internal/pkg/db/mongo/export.go
```diff
@@ -14,51 +14,73 @@
...
+	}
+
+	mapped.Modified = db.MakeTimestamp()
 
-	err := s.DB(mc.database.Name).C(db.ExportCollection).UpdateId(reg.ID, reg)
-	if err == mgo.ErrNotFound {
-		return db.ErrNotFound
+	c := s.DB(mc.database.Name).C(db.ExportCollection)
+
+	var err error
```
michaelestrin: Here to end can be simplified to:

```
	if mapped.ID.Valid() {
		return c.UpdateId(mapped.ID, mapped)
	} 
	return c.Update(bson.M{"uuid": mapped.Uuid}, mapped)
```

internal/pkg/db/mongo/export.go
```diff
@@ -67,28 +89,59 @@ func (mc MongoClient) UpdateRegistration(reg export.Registration) error {
 // Get a registration by ID
 // UnexpectedError - problem getting in database
 // NotFound - no registration with the ID was found
-func (mc MongoClient) RegistrationById(id string) (export.Registration, error) {
-	if !bson.IsObjectIdHex(id) {
-		return export.Registration{}, db.ErrInvalidObjectId
+func (mc MongoClient) RegistrationById(id string) (contract.Registration, error) {
+	reg, err := mc.registrationById(id)
```
michaelestrin: I would inline `registrationById` here and refactor `RegistrationById` as this is its only reference.

internal/pkg/db/mongo/export.go
```diff
@@ -122,11 +175,11 @@ func (mc MongoClient) getRegistrations(q bson.M) ([]export.Registration, error)
 }
 
 // Get a single registration for the passed query
-func (mc MongoClient) getRegistration(q bson.M) (export.Registration, error) {
+func (mc MongoClient) getRegistration(q bson.M) (models.Registration, error) {
```
michaelestrin: If you modify this function to return `models.Registration{}` explicitly when err != nil, you can simplify the calling code (`registrationById`, `registrationByName`).

internal/pkg/db/mongo/export.go
```diff
@@ -67,28 +89,59 @@ func (mc MongoClient) UpdateRegistration(reg export.Registration) error {
...
+	return reg, nil
 }
 
 // Get a registration by name
 // UnexpectedError - problem getting in database
 // NotFound - no registration with the name was found
-func (mc MongoClient) RegistrationByName(name string) (export.Registration, error) {
-	return mc.getRegistration(bson.M{"name": name})
+func (mc MongoClient) RegistrationByName(name string) (contract.Registration, error) {
+	reg, err := mc.registrationByName(name)
```
michaelestrin: I would inline `registrationByName` here and refactor `RegistrationByName` as this is its only reference.

internal/pkg/db/mongo/export.go
```diff
@@ -14,51 +14,73 @@
...
 
 // ****************************** REGISTRATIONS ********************************
 
 // Return all the registrations
 // UnexpectedError - failed to retrieve registrations from the database
-func (mc MongoClient) Registrations() ([]export.Registration, error) {
-	return mc.getRegistrations(bson.M{})
+func (mc MongoClient) Registrations() ([]contract.Registration, error) {
+	var mrs []models.Registration
+	mrs, err := mc.getRegistrations(bson.M{})
```
michaelestrin: I would inline `getRegistrations` and `mapRegistrations` and refactor `Registrations` as these are their own references.

internal/pkg/db/mongo/export.go
jdharms: I'm sort of fond of having the db querying and the conversion to the contract model separated out like this.

internal/pkg/db/mongo/export.go
brandonforster: > If you modify this function to return models.Registration{} explicitly when err != nil, 

Do you mean when `err == nil`?

jdharms:
Most feedback addressed.  I left the functions unrolled for matters of taste, but can inline them if that's the consensus.

internal/pkg/db/mongo/export.go
jdharms: No, I think he means return the empty struct when there's an error (err != nil).

PR #1022, opened 2019-01-16
Issue #880 - core-metadata Device: Eliminate Mongo Types from Models

michaelestrin:
* Updated query methods to remove reference parameter.
* Refactored device contracts to remove Mongo dependency.
* Eliminated unnecessary MongoDevice struct.
* Removed pass-by-ref parameter from UpdateCommand().
* Extracted common code into new idToQuery() and idToQueryParameters() methods.


internal/core/metadata/rest_addressable.go
```diff
@@ -333,7 +333,7 @@ func restGetAddressableByAddress(w http.ResponseWriter, r *http.Request) {
 func notifyAddressableAssociates(a models.Addressable, action string) error {
 	// Get the devices
 	var d []models.Device
-	if err := dbClient.GetDevicesByAddressableId(&d, a.Id); err != nil {
```
jdharms: Need to keep `d` here for use on line 345

internal/core/metadata/rest_device.go
```diff
@@ -523,10 +522,11 @@ func restCheckForDevice(w http.ResponseWriter, r *http.Request) {
 			LoggingClient.Debug(fmt.Sprintf("device %s %v", token, err))
 		}
 	}
+
 	//If lookup by name failed, see if we were passed the ID
 	if len(dev.Name) == 0 {
 		if bson.IsObjectIdHex(token) {
```
jdharms: Don't we need to handle the case where it's a UUID as well as a bson.ObjectId here?

internal/pkg/db/mongo/models/device.go
```diff
@@ -0,0 +1,131 @@
+/*******************************************************************************
+ * Copyright 2017 Dell Inc.
```
jdharms: 2019

internal/core/metadata/rest_device.go
```diff
@@ -287,8 +285,8 @@ func restGetDevicesWithLabel(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	res := make([]models.Device, 0)
-	err = dbClient.GetDevicesWithLabel(&res, label)
+	var res []models.Device
+	res, err = dbClient.GetDevicesWithLabel(label)
```
tsconn23: I don't think you need the `var` line here if you change the assignment on 289 to `:=`

Just breezing over this, I think there are other `var` declarations that are no longer required (like line 255) since we aren't passing pointers into these methods as parameters anymore.

internal/core/metadata/rest_device.go
```diff
@@ -315,8 +313,8 @@ func restGetDeviceByProfileId(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	res := make([]models.Device, 0)
-	err = dbClient.GetDevicesByProfileId(&res, pid)
+	var res []models.Device
+	res, err = dbClient.GetDevicesByProfileId(pid)
```
tsconn23: Same note on the `var`

internal/core/metadata/rest_device.go
tsconn23: Actually that check should be inside of dbClient.GetDeviceById. The controller shouldn't do this at all. The validation responsibility is farther down, so this bson check should be removed in my view (provided it's in the dbClient implementation)

internal/core/metadata/rest_device.go
```diff
@@ -558,8 +558,7 @@ func restSetDeviceOpStateById(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Check if the device exists
-	var d models.Device
-	err := dbClient.GetDeviceById(&d, did)
+	d, err := dbClient.GetDeviceById(did)
```
tsconn23: See, like here the `var` was eliminated. I think we should do this wherever possible.

internal/core/metadata/rest_deviceprofile.go
```diff
@@ -360,7 +366,7 @@ func deleteDeviceProfile(dp models.DeviceProfile, w http.ResponseWriter) error {
 		return err
 	}
 	if len(pw) > 0 {
-		err := errors.New("Cant delete device profile, the profile is still in use by a provision watcher")
+		err = errors.New("Cant delete device profile, the profile is still in use by a provision watcher")
```
tsconn23: Is it worth creating custom error types and using those here? See internal/core/metadata/errors/types.go

We'll probably do that anyway at some point when we decouple biz logic from these router files.

internal/core/metadata/rest_deviceprofile.go
```diff
@@ -464,14 +470,15 @@ func addDeviceProfileYaml(data []byte, w http.ResponseWriter) {
 
 func restGetProfileByModel(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
+	var err error
```
tsconn23: Why do you need this?

internal/pkg/db/mongo/metadata.go
```diff
@@ -294,106 +297,103 @@ func (m MongoClient) UpdateDeviceReport(dr contract.DeviceReport) error {
...
-	ts := db.MakeTimestamp()
-	d.Created = ts
-	d.Modified = ts
-	d.Id = bson.NewObjectId()
 
-	addr, err := m.getAddressableByName(d.Addressable.Name)
-	if err != nil {
-		return err
+	var device models.Device
+	if err := device.FromContract(d, m, m, m, m); err != nil {
```
tsconn23: The use of `m` here is kind of unfortunate. In all of these transform operations, you're passing in the MongoClient instance anyway. Could you remove the transform params and have the following method:

`func (d *Device) FromContract(from contract.Device, mc MongoClient)`

and then cast the `mc` to the appropriate transform where necessary within the function? For example:
```
aTransform := mc.(addressableTransform)
if d.Addressable, err = aTransform.AddressableToDBRef(aModel); err != nil {
    return
}
```

internal/pkg/db/mongo/metadata.go
```diff
@@ -294,106 +297,103 @@ func (m MongoClient) UpdateDeviceReport(dr contract.DeviceReport) error {
...
 }
 
-func (m MongoClient) GetDevicesByProfileId(d *[]contract.Device, pid string) error {
+func (m MongoClient) GetDevicesByProfileId(pid string) ([]contract.Device, error) {
 	if bson.IsObjectIdHex(pid) {
-		return m.GetDevices(d, bson.M{"profile.$id": bson.ObjectIdHex(pid)})
-	} else {
-		err := errors.New("mgoGetDevicesByProfileId Invalid Object ID " + pid)
-		return err
+		return m.getDevices(bson.M{"profile.$id": bson.ObjectIdHex(pid)})
```
tsconn23: Couldn't pid also be a UUID?

internal/core/metadata/rest_device.go
jdharms: Ah, right.  I thought this was `pkg/db/mongo/metadata.go` for some reason.  That makes more sense.

internal/pkg/db/mongo/metadata.go
```diff
@@ -294,106 +297,103 @@ func (m MongoClient) UpdateDeviceReport(dr contract.DeviceReport) error {
...
 }
 
-func (m MongoClient) GetDevicesByServiceId(d *[]contract.Device, sid string) error {
+func (m MongoClient) GetDevicesByServiceId(sid string) ([]contract.Device, error) {
 	if bson.IsObjectIdHex(sid) {
-		return m.GetDevices(d, bson.M{"service.$id": bson.ObjectIdHex(sid)})
-	} else {
-		err := errors.New("mgoGetDevicesByServiceId Invalid Object ID " + sid)
-		return err
+		return m.getDevices(bson.M{"service.$id": bson.ObjectIdHex(sid)})
```
tsconn23: Could sid also be a UUID?

internal/core/metadata/rest_addressable.go
michaelestrin: Good catch.  I think I initially moved the variable to the left side of the declaration and the IDE told me it was an unused variable.  Assignment to a previously declared variable within an if initialization statement requires the variable be used within the body of the if.  Compiler throws an error if it isn't.  I'll refactor.

internal/core/metadata/rest_device.go
michaelestrin: I concur with Trevor.  It's actually already being handled inside the Mongo implementation of GetDeviceById.  The check for `bson.IsObjectIdHex(token)` has been removed.

internal/pkg/db/mongo/models/device.go
michaelestrin: Updated.

internal/core/metadata/rest_device.go
michaelestrin: Updated file to remove unnecessary var statements.

internal/pkg/db/mongo/metadata.go
```diff
@@ -294,106 +297,103 @@ func (m MongoClient) UpdateDeviceReport(dr contract.DeviceReport) error {
 	s := m.session.Copy()
 	defer s.Close()
 
-	model := &models.DeviceReport{}
-	if err := model.FromContract(dr); err != nil {
+	var err error
```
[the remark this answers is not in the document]
michaelestrin: Refactored function to remove unnecessary `var` declarations.

internal/pkg/db/mongo/metadata.go
```diff
@@ -294,106 +297,103 @@ func (m MongoClient) UpdateDeviceReport(dr contract.DeviceReport) error {
...
+	if err = mapped.FromContract(dr); err != nil {
 		return errors.New("FromContract failed")
 	}
 
-	var err error
-	if model.Id.Valid() {
-		err = s.DB(m.database.Name).C(db.DeviceReport).UpdateId(dr.Id, dr)
-	} else {
-		err = s.DB(m.database.Name).C(db.DeviceReport).Update(bson.M{"uuid": model.Uuid}, model)
+	var query bson.M
```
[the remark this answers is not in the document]
michaelestrin: Using `:=` on line 307 will scope query to if statement and it's used after the if statement body.  Refactored to remove the `var` declaration.

internal/pkg/db/mongo/metadata.go
michaelestrin: Why is the existing implementation lacking? 

Design decision to separate transform contracts by type was based on interface segregation principle.  MongoClient happens to implement them all.  It doesn't have to in the future (if, for instance, we extract type-specific repositories from the current mono-repository that is MonoClient).
 
Passing MongoClient to disambiguate to individual interfaces in individual FromContract methods violates encapsulation -- it equates to passing around a global struct and letting callers pick from it what they want/need.

internal/pkg/db/mongo/metadata.go
michaelestrin: Refactored.

internal/pkg/db/mongo/metadata.go
michaelestrin: Refactored.

internal/core/metadata/rest_device.go
michaelestrin: Updated.

internal/pkg/db/mongo/metadata.go
tsconn23: Valid point.

`It doesn't have to in the future (if, for instance, we extract type-specific repositories from the current mono-repository that is MonoClient).`

This is something that occurred to me last night. Maybe it's not optimal to have the MongoClient implement those interfaces. Perhaps they should be their own types. On the other hand, Device is a special case thus far in the exercise and the only place where we've encountered this.

I won't block the PR as is, but I still think it's a code smell.

Issue #1017, opened 2019-01-15
Services Should Return Code When Exiting


tsconn23:
Per discussion on issue #638, services in edgex-go should call `os.Exit()` when shutting down instead of just `return`. They should return different codes in the case of shutdown as normal (0) or from failure (1). This change should be made to all services.

tsconn23:
@SteveOss @tingyuz Heads up on this issue in case you need to make the same changes in Dev Services / Security for consistency.

PR #1024, opened 2019-01-16
Updated return codes in all mains to call exit instead of return.

brandonforster:
Issue #1017

### DIFF
--- a/cmd/config-seed/main.go
+++ b/cmd/config-seed/main.go
@@ -16,6 +16,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"os"
 	"sync"
 
 	"github.com/edgexfoundry/edgex-go/internal"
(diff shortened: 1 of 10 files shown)


tsconn23:
recheck

cmd/config-seed/main.go
```diff
@@ -71,7 +74,7 @@ func bootstrap(profile string) {
 				if ok {
 					config.LoggingClient.Error(e.Error())
 				} else {
-					return
```
lenny-intel: This is a return from the Go func, not the service. I don'r think we want to exit the service here.

jdharms:
`os.Exit()` causes an immediate exit--and will skip any deferred functions.  We need to either 1) Never defer a function in our main functions or 2) wrap our current main functions something that calls `os.Exit()` once and convert our current main functions to return `int`s.

Something like this:

```
func inner_main() int {
    defer f.close()

    return 0
}

func main() {
    os.Exit(inner_main())
}
```

Something to keep in mind moving forward.

Issue #1025, opened 2019-01-16
Logging client race condition


jdharms:
There is a race condition in the logging client related to access to to the `levelLoggers` map in the `EdgeXLogger` struct.

Clip from the core-metadata `go run -race main.go`:

```==================
WARNING: DATA RACE
Read at 0x00c000175b00 by goroutine 83:
level=WARN ts=2019-01-16T21:50:54.049635Z app=edgex-core-metadata source=rest_scheduleevent.go:134 msg="Problem notifying associated device services for schedule event: Item not found"
  github.com/edgexfoundry/edgex-go/vendor/github.com/go-kit/kit/log.(*context).Log()
      /Users/jdharms/workspace/go/src/github.com/edgexfoundry/edgex-go/vendor/github.com/go-kit/kit/log/log.go:124 +0x21b
  github.com/edgexfoundry/edgex-go/pkg/clients/logging.EdgeXLogger.log()
      /Users/jdharms/workspace/go/src/github.com/edgexfoundry/edgex-go/pkg/clients/logging/logger.go:179 +0x5d6
  github.com/edgexfoundry/edgex-go/pkg/clients/logging.EdgeXLogger.Warn()
      /Users/jdharms/workspace/go/src/github.com/edgexfoundry/edgex-go/pkg/clients/logging/logger.go:227 +0xe4
  github.com/edgexfoundry/edgex-go/pkg/clients/logging.(*EdgeXLogger).Warn()
      <autogenerated>:1 +0x101
  github.com/edgexfoundry/edgex-go/internal/core/metadata.restAddScheduleEvent()
      /Users/jdharms/workspace/go/src/github.com/edgexfoundry/edgex-go/internal/core/metadata/rest_scheduleevent.go:134 +0xd0c
  net/http.HandlerFunc.ServeHTTP()
      /usr/local/Cellar/go/1.11.4/libexec/src/net/http/server.go:1964 +0x51
  github.com/edgexfoundry/edgex-go/vendor/github.com/gorilla/mux.(*Router).ServeHTTP()
      /Users/jdharms/workspace/go/src/github.com/edgexfoundry/edgex-go/vendor/github.com/gorilla/mux/mux.go:162 +0x17e
  github.com/edgexfoundry/edgex-go/vendor/github.com/gorilla/context.ClearHandler.func1()
[...]

PR #1026, opened 2019-01-16
#1025 Synchronize access to EdgeXLogger.levelLoggers map

jdharms:
#1025 Fixes a race condition in pkg/clients/logging/logger.go

### DIFF
--- a/pkg/clients/logging/logger.go
+++ b/pkg/clients/logging/logger.go
@@ -81,9 +81,6 @@ func NewClient(owningServiceName string, isRemote bool, logTarget string, logLev
 		logLevel:          &logLevel,
 	}
 
-	// Set up the loggers
-	lc.levelLoggers = map[string]log.Logger{}
-
 	if !lc.remoteEnabled && logTarget != "" { // file based logging
 		verifyLogDirectory(lc.logTarget)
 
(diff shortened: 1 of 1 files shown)


pkg/clients/logging/logger.go
```diff
@@ -160,10 +163,19 @@ func (lc EdgeXLogger) log(logLevel string, msg string, args ...interface{}) {
 		args = append(args, "msg", msg)
 	}
 
+	levelLock.RLock()
 	if lc.levelLoggers[logLevel] == nil {
+		levelLock.RUnlock()
+		// Upgrade lock to write
+		levelLock.Lock()
 		lc.levelLoggers[logLevel] = log.WithPrefix(lc.rootLogger, "level", logLevel)
+		levelLock.Unlock()
```
tsconn23: If we were to pull this in, I think the only locking you'd need is this writable lock since the levelLoggers map is being modified. All of the other read lock mgmt shouldn't be necessary.

But in order to prove that out I need to be able to reliably reproduce and as I said in my comment on issue #1025 I cannot.

jdharms: On my machine at least, I need both locks to avoid the race condition warnings.

jdharms:
Updated PR to implement lock-free solution @tsconn23 suggested today.

pkg/clients/logging/logger.go
```diff
@@ -160,10 +167,8 @@ func (lc EdgeXLogger) log(logLevel string, msg string, args ...interface{}) {
 		args = append(args, "msg", msg)
 	}
 
-	if lc.levelLoggers[logLevel] == nil {
-		lc.levelLoggers[logLevel] = log.WithPrefix(lc.rootLogger, "level", logLevel)
-	}
-
+	levelLock.RLock()
+	defer levelLock.RUnlock()
```
tsconn23: Are these statements still needed?

jdharms: Yeah, they're not needed.

PR #1032, opened 2019-01-18
Generate newDBClient for user-specified DB provider

vaind:
This PR is a part of decoupling the MongoDB and providing a way to use other persistence-layer implementations. This is based on the talks in slack corewg-persist channel as well as the corewg meeting held on January 17, 2019.

The basic idea is:
* each service has a single file that initializes the DBClient -  `initdb.go`
* this file is generated before build, i. e. during the `make prepare` step. 
* by default, the internal mongoDb implementation is used at the moment (as it was before - so by default this PR doesn't change EdgeX behaviour)
* If the developer building the platform would want to use another persistence provider, she would do so by overriding the default value of DBPROVIDER, e. g. like this: `make prepare DBPROVIDER=github.com/objectbox/edgex-db` 
* as the DB "type" is now set during build time and can't be changed in runtime, I've removed the Database.Type configuration - it was only used during `newDBClient()`

A few additional notes
* this has no effect on CI/CD pipelines - doesn't require any changes
* this PR is independent of the decisions about modules, plugins, etc.
* originally, I've wanted to extract-out the mongoDb implementation (internal/pkg/db/mongo). This, however, isn't currently possible because it imports "internal/pkg/db" which, of course, is an internal package and thus can't be imported from the other go projects. 
* if the developer specifies a different provider, he is responsible for making sure that the EdgeX can be built with it, i. e. that it implements the necessary interfaces & that it's available in go path (or added using glide/go dep/go module) - I've opted for this as opposed to actually going via some automatic installation because of two reasons: 1. it's an ongoing discussion and not an easy one; 2. in reality, it doesn't matter how the dependencies are going to be managed, there still would need to be a way to switch the provider and this PR implements the pretty obvious one to do so during build time

-----

As the initdb.go files are not part of the repository (they're generated and ignored by git), here's an example from the core-data service (the comment is in a "standardized" format recognized by go tools & IDEs):
```go
// Code generated by EdgeX; DO NOT EDIT.

package data

import (
	"github.com/edgexfoundry/edgex-go/internal/core/data/interfaces"
	impl "github.com/edgexfoundry/edgex-go/internal/pkg/db/mongo"
)

// Return the dbClient interface
func newDBClient(config db.Configuration) (interfaces.DBClient, error) {
	return impl.NewClient(config)
}
```


JPWKU:
recheck

vaind:
Any feedback on this?

Makefile
```diff
@@ -11,6 +11,9 @@
 GO=CGO_ENABLED=0 go
 GOCGO=CGO_ENABLED=1 go
 
+DBPROVIDER ?= github.com/edgexfoundry/edgex-go/internal/pkg/db/mongo
```
tsconn23: How would you support more than one DB provider at a time? For example in the case where a customer wanted Redis for core-data and Mongo for everything else? Additional vars in the makefile?

cmd/core-data/res/configuration.toml
```diff
@@ -45,7 +45,6 @@ File = './logs/edgex-core-data.log'
   Port = 27017
   Username = ''
   Timeout = 5000
-  Type = 'mongodb'
```
tsconn23: Removal kind of concerns me because if I don't have access to the build time logs, I don't have a clear indication of which DB is being used. The way it is now, I can always just go to configuration. And since the Type property has been removed from the DatabaseInfo struct, I can't obtain this information via the sys-mgmt-agent either.

Makefile
vaind: Yes, that is one option, another is reading a configuration & setting stuff up during build time. That would be necessary in case the db-provider needs some additional changes/resources during the build process, such as a docker-compose fragment if the DB is not embedded but needs a server running. 

I've not included any solution to this in the PR exactly because it would be too opinionated and would side-track from the goal of the PR, i.e. import & initialize the chosen database.

cmd/core-data/res/configuration.toml
vaind: I understand your concerns. The removal was IMHO necessary because this option is not read so changing it in the config file would not do anything. 

What we could do is provide a runtime log during DB initialization to see which DB is being launched for each service. The logger call could be added to the newDBClient().

Not sure about how the sys-mgmt-agent works yet but I guess there would be a way to provide this info there as well.

Issue #1030, opened 2019-01-17
Error when importing edgex-go into a module -- ";" in path 


tsconn23:
When attempting to setup device-sdk-go as a module and referencing edgex-go as a dependency, the following error is encountered due to the legacy properties directories in the config-seed/cmd/res directory:

```
~/Documents/code/git/edgexfoundry/device-sdk-go $ go build
go: finding github.com/edgexfoundry/edgex-go 0.7.1
go: downloading github.com/edgexfoundry/edgex-go v0.0.0-20181210184951-f7a0b44a0188
go: finding github.com/hashicorp/serf/coordinate latest
go: finding github.com/hashicorp/go-rootcerts latest
-> unzip /Users/tconn/go/pkg/mod/cache/download/github.com/edgexfoundry/edgex-go/@v/v0.0.0-20181210184951-f7a0b44a0188.zip: malformed file path "cmd/config-seed/res/properties/device-bacnet;docker/application.properties": invalid char ';'
internal/clients/init.go:11:2: unknown import path "github.com/edgexfoundry/edgex-go/pkg/clients": unzip /Users/tconn/go/pkg/mod/cache/download/github.com/edgexfoundry/edgex-go/@v/v0.0.0-20181210184951-f7a0b44a0188.zip: malformed file path "cmd/config-seed/res/properties/device-bacnet;docker/application.properties": invalid char ';'
```

In the above example, I'm attempting to import Delhi via go.mod file into device-sdk-go. There are a couple issues about this in the Go repo
https://github.com/golang/go/issues/28001
https://github.com/golang/go/issues/26672

tsconn23:
SDK-Go contract mods were merged early this morning

https://github.com/edgexfoundry/device-sdk-go/pull/157

I've gotten latest and gone through the approach outlined above. It appears to work.
```
~/Documents/code/git/edgexfoundry/device-sdk-go $ make build
CGO_ENABLED=0 go build -ldflags "-X github.com/edgexfoundry/device-sdk-go.Version=1.0.0" -o example/cmd/device-simple/device-simple ./example/cmd/device-simple
go build ./...
```

I need to check one more thing to ensure adding the go.mod to the config-seed properties folder won't break the CI-CD process. Assuming no problems, I'll create an issue in edgex-go to drop the file in place.

anonymouse64:
As @Burning1020 points out, it would be nice if this were backported to the Delhi branch.

PR #1035, opened 2019-01-21
Added go.mod to config-seed properties folder

tsconn23:
#1030 

This is a preparatory effort and does not have a material effect on the build since GO111MODULE is not explicitly set. Per the issue above, we need to add this file to get the device-sdk-go module to ignore the config-seed/res/properties directories containing semi-colons. Device-sdk-go is our target first module project.


Burning1020:
Could u plz merge this commit into delphi branch?

Issue #1021, opened 2019-01-16
Eliminate Mongo Types from Models - Clean up


michaelestrin:
Normalize/clean up code that eliminated Mongo types from models across several issues/PRs (in core-data, core-metadata, support-notifications, and export-client).

michaelestrin:
Dependency on #883.

PR #1036, opened 2019-01-21
Issue #1021 - Eliminate Mongo Types from Models - Clean up

michaelestrin:
* Updated to consistently return id from FromContract for models with ObjectId and Uuid identity.
* Ensured pw.Profile is updated in restAddProvisionWatcher().
* Ensured err results from Mongo calls are consistently passed through errorMap.
* Cleaned up var usage in repositories.
* Standardized update functions
* Standardized MongoClient receiver name.
* Addressed GoLand IDE errors/warnings in repositories.
* Removed unnecessary SetBSON(), GetBSON() functions
* Standardized how get...() repository functions disambiguate contract id to a bson.ObjectId or uuid.
* Inlined second class models.


pkg/models/interval.go
```diff
@@ -20,39 +20,32 @@ import (
...
-	Created   int64  `json:"created"`
-	Modified  int64  `json:"modified"`
-	Origin    int64  `json:"origin"`    // not currently used reserved for interval origin source
-	Name      string `json:"name"`      // non-database identifier for a shcedule (*must be quitue)
-	Start     string `json:"start"`     // Start time i ISO 8601 format YYYYMMDD'T'HHmmss 	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyymmdd'T'HHmmss")
-	End       string `json:"end"`       // Start time i ISO 8601 format YYYYMMDD'T'HHmmss 	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyymmdd'T'HHmmss")
-	Frequency string `json:"frequency"` // how frequently should the event occur according ISO 8601
-	Cron      string `json:"cron"`      // cron styled regular expression indicating how often the action under interval should occur.  Use either runOnce, frequency or cron and not all.
-	RunOnce   bool   `json:"runOnce"`   // boolean indicating that this interval runs one time - at the time indicated by the start
+	BaseObject
```
tsconn23: Not sure we should be adding BaseObject back in here. We want to get rid of it, as discussed, so I believe @xmlviking created this type with inline properties for that reason.

pkg/models/interval_action.go
```diff
@@ -21,10 +21,8 @@ import (
 )
 
 type IntervalAction struct {
+	BaseObject
```
tsconn23: Not sure we should be adding BaseObject back in here. We want to get rid of it, as discussed, so I believe @xmlviking created this type with inline properties for that reason.

pkg/models/reading.go
```diff
@@ -26,34 +26,28 @@ import (
...
 type Reading struct {
-	Id       string `json:"id"`
-	Pushed   int64  `json:"pushed"`  // When the data was pushed out of EdgeX (0 - not pushed yet)
-	Created  int64  `json:"created"` // When the reading was created
-	Origin   int64  `json:"origin"`
-	Modified int64  `json:"modified"`
-	Device   string `json:"device"`
-	Name     string `json:"name"`
-	Value    string `json:"value"` // Device sensor data value
+	BaseObject
```
tsconn23: Not sure we should be adding BaseObject back in here. We want to get rid of it, as discussed, and I think a previous PR refactored this type to remove it.

pkg/models/registration.go
```diff
@@ -59,10 +59,8 @@ const (
 // Registration - Defines the registration details
 // on the part of north side export clients
 type Registration struct {
+	BaseObject
```
tsconn23: Not sure we should be adding BaseObject back in here. We want to get rid of it, as discussed, and I think a previous PR refactored this type to remove it.

pkg/models/event.go
```diff
@@ -25,35 +25,29 @@ import (
  * Event struct to hold event data
  */
 type Event struct {
+	BaseObject
```
tsconn23: Not sure we should be adding BaseObject back in here. We want to get rid of it, as discussed, and I think a previous PR refactored this type to remove it.

internal/support/scheduler/loader.go
```diff
@@ -162,16 +162,16 @@ func loadConfigIntervals() error {
...
 			Name:      intervals[i].Name,
 			Start:     intervals[i].Start,
 			End:       intervals[i].End,
 			Frequency: intervals[i].Frequency,
 			Cron:      intervals[i].Cron,
 			RunOnce:   intervals[i].RunOnce,
 		}
+		interval.Created = 0
+		interval.Modified = 0
+		interval.Origin = 0
```
tsconn23: Probably don't even need these lines since an Interval type is being instantiated and the default value of int64 is zero.

pkg/models/interval.go
michaelestrin: Corrected.

pkg/models/interval_action.go
michaelestrin: Corrected.

pkg/models/reading.go
michaelestrin: Corrected.

pkg/models/registration.go
michaelestrin: Corrected.

pkg/models/event.go
michaelestrin: Corrected.

internal/support/scheduler/loader.go
michaelestrin: Reverted.

internal/pkg/db/mongo/export.go
```diff
@@ -176,33 +151,28 @@ func (mc MongoClient) getRegistration(q bson.M) (models.Registration, error) {
...
+	return errorMap(s.DB(mc.database.Name).C(db.ExportCollection).Remove(q))
 }
 
 func mapRegistrations(registrations []models.Registration, err error) ([]contract.Registration, error) {
 	if err != nil {
 		return []contract.Registration{}, err
 	}
 
-	var mapped []contract.Registration
+	mapped := make([]contract.Registration, 0)
```
jdharms: Is there some reason why this is the preferred format for this?

michaelestrin: In this particular case, when `registrations` is empty `make` results in an empty slice and `var` results in nil.  This specific change was made because convention is to return an empty slice instead of nil.

If the question is why use `make([]contract.Registration, 0)` in place of `[]contract.Registration{}`, it's to address the `Empty slice declaration via literal` warning in GoLand.

jdharms: Thanks

jdharms: @tsconn23 His point was that if append isn't called at all then `nil` will be returned, which is not what most of the methods have been doing in this case (returning empty slices).  I agree that `nil` should be treated as an empty slice by the caller, but to Michael's point the consistency might be worth it here.

michaelestrin: @tsconn23 I think I'm missing something here.  If `append()` is not called, using `var` leaves `mapped` nil.

tsconn23: ah, got it. Yes, I was just focused on the append operation. Makes sense.

PR #1050, opened 2019-01-23
sort lists with limits to return newest items

### LINKED ISSUE #966: Query API with limit $n should return the latest $n records instead of the first $n records
EdgeX version: master

Steps to reproduce:

1. Start up EdgeX with device-random
2. Invoke the core-data API to get the events limited to 10 records
http://localhost:48080/api/v1/event/device/Random-Integer-Generator01/10

Actual results:
The first 10 events are returned, so the result is always the same even if there are more new events received

Expected results:
The last 10 events are returned
The order is not defined in RAML, but users expect to get the latest ones.
https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-data.raml#L139

This issue could apply to all other query with limit argument.

Quyrean:
Fix #966 Query API with limit $n should return the latest $n records instead of the first $n records

queries affected:
 /event/device/{deviceId}/{limit}: 
 /event/{start}/{end}/{limit}: 
 /event/device/{deviceId}/valuedescriptor/{valuedescriptor}/{limit}: 
 /reading/device/{deviceId}/{limit}: 
 /reading/name/{name}/device/{device}/{limit}: 
 /reading/name/{name}/{limit}: 
 /reading/uomlabel/{uomLabel}/{limit}: 
 /reading/label/{label}/{limit}: 
 /reading/type/{type}/{limit}: 
 /reading/{start}/{end}/{limit}: 

The results for these queries are sorted before imposing the limit.  I had to hard code the name of the field "created" as I could not find a function or const that had the value.

The existing sort function in the code uses mongo orderby, so I just called that.  The caveat is that any sorts can only hold up to 32 meg.  I have no idea what the scale requirements are.  I did not use natural order because it is not guaranteed.  The good news is that if the column is not defined, there is no error.

I ran "make test", but do not know of any other test programs.  I tested using postman and curl.  I do not have any means of doing performance and/or scale testing.

This is a fail-fast approach where making all the calls consistant will exercise the code more frequently quickly uncovering any problems.


Quyrean:
> 
> 
> @Quyrean Sign-off is missing. You'll need to fix that. Please see the [Contributor Process](https://wiki.edgexfoundry.org/display/FA/Contributor+Process) document. Note the yellow box which describes the keys you will need.
> 
> Also, since there's a conflict, you'll need to rebase your feature branch from master.

Ok, taken care of, sorry about that.

cloudxxx8:
@Quyrean please squash these three commits and add some meaningful commit message for each one.
@tsconn23 @andresrinivasan this fix only modify the mongo driver.  do we need to also modify the redis driver?

tsconn23:
@cloudxxx8 We don't yet have the Redis driver in the master branch, only in Delhi. @andresrinivasan will need to incorporate this into his upcoming work.

michaelestrin:
Changes establish the convention that events are always returned in reverse `created` order.  Is this sufficient?  

What about updated events?  Their `created` date doesn't change but their data implicitly does.  If I wanted to get the last `n` changes, those changes might include updates to previously created events.

tsconn23:
To be fair, the original issue doesn't specify `created` versus `modified`, however I agree with @michaelestrin that `modified` is probably more useful in this case. Querying by origin would be a separate issue if we were to receive a requirement asking for it.

Quyrean:
@cloudxxx8   I think I got my fork fixed and squashed the old commits correctly with a useful commit message.  Thanks.

I updated to using the `modified ` field which is present in the examples, and has the same value as as the created field.

cloudxxx8:
Thanks, @Quyrean  and @tsconn23 , the change looks good to me

tsconn23:
@Quyrean You may want to rebase. Something has happened with your commit history. Notice the # of lines that are now added/removed is much larger. I also see commits from other people in your commit history now, plus a file conflict. Maybe this is due to a pull vs rebase prior to your last force push.

tsconn23:
Hi @Quyrean -- Unfortunately as you can see the DCO Sign-off check is now hung. We've seen this happen a couple times recently. It's some issue with the GitHub job integration. The only way to fix it (since you've got the sign-off on your commits) is to close this PR and open a new one. Once the sign-off check is in this state, it will never work.

I looked at the latest version of the changes yesterday and they appear fine to me. There is one thing I'm looking at a bit more closely that may be attributable to a gap in our blackbox tests, but hopefully can resolve this sometime tomorrow.

Quyrean:
Thank you for looking into this.  I should be able to open a new request tomorrow morning.

tsconn23:
@Quyrean When you create the new PR be sure to rebase your feature branch as I've merged a few things since your last commit.

Issue #1014, opened 2019-01-14
Add correlation ID tracking to all edgex-go services


tsconn23:
The purpose of a correlation ID in a microservice architecture is to allow for visible tracing of service interactions initiated by a caller at the edge of the eco-system. For example, when we ingest a new event, that POST can cause a series of interactions between the EdgeX services. These interactions can change based on configuration settings.

We would like to be able to map these interactions as well as obtain latency metrics at each hop.

I have a proof of concept available in a [feature branch](https://github.com/tsconn23/edgex-go/tree/correlation_WIP) that shows how we can trace the interactions between core-data and core-metadata when adding an event. The MetaDataCheck flag in core-data's configuration.toml is set to `true` which tells core-data to call core-metadata to verify that the event's source device is registered.

Example output
```
level=INFO ts=2019-01-14T19:19:59.203195Z app=edgex-core-data source=middleware.go:44 correlation-id=75d4313e-2428-450b-872e-ba9f1ab0f09e path=/api/v1/event msg="Begin request"
level=INFO ts=2019-01-14T19:19:59.209241Z app=edgex-core-data source=router.go:216 msg="Posting Event: {\"device\":\"Nebulizer123\",\"origin\":1471806386919,\"readings\":[{\"origin\":1471806386919,\"name\":\"luminousIntensity1\",\"value\":\"40\"}]}"
level=INFO ts=2019-01-14T19:19:59.214442Z app=edgex-core-metadata source=middleware.go:44 correlation-id=75d4313e-2428-450b-872e-ba9f1ab0f09e path=/api/v1/device/check/Nebulizer123 msg="Begin request"
level=INFO ts=2019-01-14T19:19:59.21531Z app=edgex-core-metadata source=middleware.go:37 correlation-id=75d4313e-2428-450b-872e-ba9f1ab0f09e duration=869.648µs
level=ERROR ts=2019-01-14T19:19:59.215805Z app=edgex-core-data source=router.go:228 msg="Item not found"
level=INFO ts=2019-01-14T19:19:59.216133Z app=edgex-core-data source=middleware.go:37 correlation-id=75d4313e-2428-450b-872e-ba9f1ab0f09e duration=12.934262ms
```

Here you can see that for each service, we denote the beginning of the request and its completion. Each completion has a duration. Core-data's duration signifies the total duration of the operation since it is at the bottom of the stack. Every call has the same correlation ID.

PR #1052, opened 2019-01-23
Add basic tracing and metrics via a simple correlation ID.

brandonforster:
Issue #1014. This work is a production ready wrap up of @tsconn23's PoC located at branch tsconn23/correlation_WIP.


cmd/core-data/res/configuration.toml
```diff
@@ -1,6 +1,6 @@
 [Writable]
 DeviceUpdateLastConnected = false
-MetaDataCheck = false
+MetaDataCheck = true
```
jduranf: I think this change is not related with this PR

tsconn23: Correct. This is part of what appears to be a cherry-picked commit from my feature branch above. It is intended to demonstrate how the correlation ID is passed from core-data to core-metadata, but should not necessarily be part of the default implementation.

Issue #987, opened 2019-01-07
Core-metadata should return 409 when updating the command which the name is not unique for the Device Profile


weichou1229:
**OS version:** macos 10.14  
**EdgeX version:** 1.0.0
**Environment details:** Docker container  
**Steps to reproduce:**

1.Start up EdgeX

2-1.Post a new device profile

```
Method: POST
URL: http://localhost:48081/api/v1/deviceprofile
Body:
{
  "name": "testProfileName",
  "commands":[ 
    { "name":"testCommanndName" },
    { "name":"boilingpoint" }
  ]
}
```

2-1.Find command by name

```
Method: GET
URL: http://localhost:48081/api/v1/command/name/testCommanndName

Response:
[
  {
    id:"bf14c6f2-bbcb-4398-a366-8bf5f1748c13",
    name:"testCommanndName"
  }
]
```

2-2.Update command

```
Method: PUT
URL: http://localhost:48081/api/v1/command
Body:
 {
    "id": "bf14c6f2-bbcb-4398-a366-8bf5f1748c13",
    "name": "boilingpoint",
    ...
 }

Response:
true
Response status code: 200
```

According to [RAML](https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-metadata.raml#L2333), API should return 409 when updating the command which the name is not unique for the Device Profile. Actually, it returns 200.

Issue #986, opened 2019-01-07
Core-metadata should return 409 when deleting the command which is still associated with a Device Profile


weichou1229:
**OS version:** macos 10.14  
**EdgeX version:** 1.0.0
**Environment details:** Docker container  
**Steps to reproduce:**

1. Start up EdgeX

2. Add a new device profile

```
Method: POST
URL: http://localhost:48081/api/v1/deviceprofile
Body:
{
  "name": "testProfileName",
  "commands":[ { "name":"testCommanndName" } ]
}
```

3. Find command by name

```
Method: GET
URL: http://localhost:48081/api/v1/command/name/testCommanndName

Response:
[
  {
    id:"bf14c6f2-bbcb-4398-a366-8bf5f1748c13",
    name:"testCommanndName"
  }
]
```

4. Delete command by ID

```
Method: DELETE
URL: http://localhost:48081/api/v1/command/id/bf14c6f2-bbcb-4398-a366-8bf5f1748c13

Response:
true
Response status code: 200
```

According to [RAML](https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-metadata.raml#L2246), API should return 409 when deleting the command which is still associated with a Device Profile. Actually, it returns 200.

PR #1058, opened 2019-01-25
Fix issues with commands in device profile service

brandonforster:
This PR addresses two issues and two issues #986 and #987:

First, the logic for determining if a command was associated with a device profile was broken and would always return that there were no commands associated with a device profile.

Second, the logic for determining whether to update a command was out of sync with the RAML and included an unpublished requirement that the ID of the updated command be different. This has been removed.

### DIFF
--- a/internal/pkg/db/mongo/metadata.go
+++ b/internal/pkg/db/mongo/metadata.go
@@ -557,12 +557,12 @@ func (mc MongoClient) UpdateDeviceProfile(dp contract.DeviceProfile) error {
 
 // Get the device profiles that are currently using the command
 func (mc MongoClient) GetDeviceProfilesUsingCommand(c contract.Command) ([]contract.DeviceProfile, error) {
-	query, err := idToBsonM(c.Id)
+	command, err := mc.getCommandById(c.Id)
 	if err != nil {
 		return []contract.DeviceProfile{}, err
 	}
 
-	dps, err := mc.getDeviceProfiles(bson.M{"commands": bson.M{"$elemMatch": query}})
+	dps, err := mc.getDeviceProfiles(bson.M{"commands.$id": command.Id})
 	if err != nil {
 		return []contract.DeviceProfile{}, err
 	}
--- a/internal/core/metadata/rest_command.go
+++ b/internal/core/metadata/rest_command.go
@@ -95,7 +95,7 @@ func restUpdateCommand(w http.ResponseWriter, r *http.Request) {
 		// Loop through matched device profiles to ensure the name isn't duplicate
 		for _, profile := range dps {
 			for _, command := range profile.Commands {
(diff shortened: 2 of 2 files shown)


internal/pkg/db/mongo/metadata.go
```diff
@@ -557,12 +557,12 @@ func (mc MongoClient) UpdateDeviceProfile(dp contract.DeviceProfile) error {
...
 // Get the device profiles that are currently using the command
 func (mc MongoClient) GetDeviceProfilesUsingCommand(c contract.Command) ([]contract.DeviceProfile, error) {
-	query, err := idToBsonM(c.Id)
+	command, err := mc.getCommandById(c.Id)
 	if err != nil {
 		return []contract.DeviceProfile{}, err
 	}
 
-	dps, err := mc.getDeviceProfiles(bson.M{"commands": bson.M{"$elemMatch": query}})
+	dps, err := mc.getDeviceProfiles(bson.M{"commands.$id": command.Id})
```
tsconn23: Before approving this PR, I want to document reasoning for the approval w/r/t the above line change.

The older version of the line above was introduced by @michaelestrin who was intending to use a Mongo projection to find all device profiles containing a command associated with the given ID. See [the docs](https://docs.mongodb.com/manual/reference/operator/query/elemMatch/) on that capability.

However that doesn't work if the incoming `command.Id` is a UUID because the array of commands on the device profile is an array of DBRefs, which means BSON IDs. Yet one more artifact of using Mongo as a faux relational database when we should just be storing documents.

Given that, it is necessary to obtain the command Mongo model so that we have the BSON ID and can then query for the relevant device profiles via the DBRef BSON ID.

Sample document for illustration:
```
{ 
"_id" : ObjectId("5c4f6e62deafc2723239b3a4"), 
"created" : NumberLong("1548709474557"), 
"modified" : NumberLong("1548709474557"), 
"origin" : NumberLong("1471806386919"), 
"description" : "BACnet Honeywell thermostats",
 "uuid" : "91b5e0c0-2752-43e0-9cd0-d9b23fd03eb2", 
"name" : "boiling point profile", 
"manufacturer" : "Honeywell", 
"model" : "ABC123", 
"labels" : [ "bacnet", "thermostat", "Honeywell" ], 
"objects" : 
    { "resources" : 
        { "CurrentHumidity" : 
            { "get" : [ 
                { "parameter" : "CurrentHumidity", "property" : "presentValue", "operation" : "get", "object" : "AnalogValue_22" }
                 ] 
            }
         }
    }, 
    "deviceResources" : [ ],
    "resources" : [ ], 
    "commands" : [
        DBRef("command", ObjectId("5c4f6e62deafc2723239b394")), 
        DBRef("command", ObjectId("5c4f6e62deafc2723239b39e")) 
    ] 
}
```

PR #1059, opened 2019-01-26
SenML format supported in export microservice

juanmagal:
This PR is introducing the possibility of reporting with SenML format to an external entity via export micro-services.

Please check code since I cannot be considered a go lang expert.

I have tested with a mosquitto broker and with mainflux server and it works.


tsconn23:
@juanmagal At a minimum, you need to fix the build/test issues before anyone looks at this.

```
10:10:14 --- FAIL: TestReading_String (0.00s)
10:10:14     --- FAIL: TestReading_String/reading_to_string (0.00s)
10:10:14     	reading_test.go:74: Reading.String() = {"pushed":123,"created":123,"origin":123,"modified":123,"device":"test device name","name":"Temperature","value":"45","unit":null,"type":0}, want {"pushed":123,"created":123,"origin":123,"modified":123,"device":"test device name","name":"Temperature","value":"45"}
```

juanmagal:
My apologies since I forgot to run the unit tests (and create new ones). Please, check if now it is ok.

internal/export/distro/format.go
```diff
@@ -275,3 +280,63 @@ func (af biotFormatter) Format(event *models.Event) []byte {
...
+		case v.Type == models.Float64:
+			b, err := base64.StdEncoding.DecodeString(v.Value)
+			if err != nil {
+				LoggingClient.Error(fmt.Sprintf("Error parsing SenML Float value. Error: %s", err.Error()))
+			} else {
+				value := Float64frombytes(b)
+				reading.Value = &value
+			}
+			// SenML Data not supported yet
+		}
```
tsconn23: Do you want a default case in here if the `v.Type` isn't one of the above specific types? Should that default populate v.DataValue?

Also @SteveOss @tonyespy notice the assumption that `float` is encoded to `base64` above. This PR is now pending a decision on that representation.

pkg/models/reading.go
```diff
@@ -21,32 +21,82 @@ import (
 /*
  * This file is for the Reading model in EdgeX
  * Holds data that was gathered from a device
- *
- *
+ */
+
+// ValueType indicates the type of value being passed back
+// from a ProtocolDriver instance.
+type ValueType int
```
tsconn23: The indication of type is already present in a value descriptor. 
https://github.com/edgexfoundry/edgex-go/blob/4c54a00c141161af257cef1d7c161527c4d51ce7/pkg/models/value-descriptor.go#L34

Each reading is based on a value descriptor and we validate the reading accordingly upon ingestion into core-data. However it looks like for the SenML export, you need type information at export time in addition to the reading value and (understandably) don't want to go lookup the value descriptor for every reading to be exported. The SenML solution has a minimal typing in its reading that you're attempting to map to rather than using a default string property value.

Because of the change to the contract, I can't let this PR go through without discussion with the Device Services working group.

tsconn23:
Related https://github.com/edgexfoundry/device-sdk-go/pull/164

pkg/models/service.go
```diff
@@ -16,6 +16,7 @@ package models
 
 import (
 	"encoding/json"
+
```
drasko: Why this space?

pkg/models/registration.go
```diff
@@ -97,8 +98,10 @@ func (reg Registration) Validate() (bool, error) {
 		reg.Format != FormatAWSJSON &&
 		reg.Format != FormatCSV &&
 		reg.Format != FormatThingsBoardJSON &&
-		reg.Format != FormatNOOP {
-		return false, fmt.Errorf("Format invalid: %s", reg.Format)
+		reg.Format != FormatNOOP &&
+		reg.Format != FormatSenMLJSON {
+
```
drasko: No need for this extra space.

pkg/models/reading.go
juanmagal: Not sure I got your point... The problem I saw with export (or at least on how Device Services export data) is everything seems to be encoded as a string. However, Sen ML RFC explicitly states whether a value must be encoded as a boolean, as a string, as a float or as binary data in the json file. For instance, for the examples I am using (humidity, pressure and temperature), all of them are explicitly defined in the RFC as "float". 
So, if float is encoded using base64 as it is defined in pkg/models/commandvalue.go in device service (device sdk) and sent as a string, I am not able to know the type. I could not find another way to get it, but explicitly encoding, probably there are other ways to do it, but unfortunately I am a newbie when it comes to go lang and definitively don't know all the details of the Edge X Foundry architecture.
Of course I fully agree this is something to be discussed with Device Services.

pkg/models/registration.go
juanmagal: Ok, I will remove it once it is clear how the communication between device services and export service is done (in relation to the type)

JPWKU:
recheck

juanmagal:
Thanks @JPWKU 
I'm not sure I understand your comment, may I ask you to be a bit more specific on what I should recheck?

juanmagal:
@tsconn23, do you have some more input on the discussion with Device Services?

pkg/models/service.go
juanmagal: Space removed

pkg/models/registration.go
juanmagal: Space removed

tsconn23:
@juanmagal The next Device Service call isn't until Monday. I do have some feedback from other folks involved in project oversight and they're recommending we wait on this PR for the reason I called out in the review. That is, the need to have an indication of type on the reading itself. It can be done, but this may not be the right place and there a groups that feel this would be better served through the revised export services that we are about to start working on.

pkg/models/reading.go
tonyespy: @tsconn23 @juanmagal 

The current approach where clients are required to use value descriptors in order introspect string reading values is a core concept of EdgeX. While the original Java code that was released to the community used a slightly different schema for value properties and the related value descriptors, the idea was still the same, the low-level device service protocol layer was responsible for translating native data types to a string value contained in a reading object. If a client needed detailed type information (i.e. like in your SenML export logic), the naive implementation would be to query the associated value descriptors on-demand. As the set of value descriptors in an EdgeX implementation don't change all that often, a client could easily cache all of the value descriptors locally and only query if/when a reading was encountered with an unknown descriptor.

With all that said, as the new type model is much simpler, we could in fact consider adding the type to the reading object, as it is only a single `int` field. The original Java code used a type scheme with multiple attributes, which I imagine was one of the factors in not replicating them all in each reading, but instead storing them in the value descriptor objects shared across a single EdgeX instance.

juanmagal: Thanks! 
Besides the measure type, I am also including in the reading object the "unit" of the measure read which is a string, would you also consider including the unit as part of the reading?
@tsconn23 @tonyespy , should I assume the introduction of the type (and the unit) in the reading object will be accepted? Should I change something on my code?

tsconn23: @tonyespy To your second point on adding the type to the reading object, I don't think @jpwhitemn is going to be onboard with that for Edinburgh. In discussions I've had with him, this falls under the auspices of the new application services which could provide the representation @juanmagal is looking for before exporting via SenML. I think Jim is going to resist changing the core model as is done in this PR.

juanmagal: Thanks @tsconn23 
I am of course open to other alternatives if you think this change on the data model will not be accepted. IMHO I don't know if the best option is waiting till this new application service is ready, do you have any expected lead time? 
I could try to check the ValueDescriptor option even from performance point of view it might not be the best option. However, I am not sure how I can get a ValueDescriptor from the Reading object.

drasko: @tsconn23 @tonyespy without knowing in depth new changes - how will an app in the cloud know the type of `Value` if all it recieves is a string? Under which type will it store it in the DB?

Regarding the timestamp - it can be always the current time in the moment of sending the message. It can be also set to `0`, as a relative offset of server time (time when message is recieved). Finally - `Time` field in SenML is optional - if it does not exist then server will probably attach timestamp of the receiving of the message.

tsconn23:
Closing this PR per the discussion above. This will not be an Edinburgh item.

PR #1086, opened 2019-02-07
Add new device services for master

### LINKED ISSUE #958: Add new device services to snap
We added the new device services, `device-modbus-go`, `device-mqtt-go`, and `device-random` to the snap for Delhi with #894 

This should also be forward-ported to master for Edinburgh.

anonymouse64:
Duplicate of #894 but for master (Edinburgh)

Closes #958

### DIFF
--- a/snap/hooks/install
+++ b/snap/hooks/install
@@ -24,7 +24,7 @@ export LD_LIBRARY_PATH=$SNAP_LIBRARY_PATH:$LD_LIBRARY_PATH
 # install all the config files from $SNAP/config/SERVICE/res/configuration.toml 
 # into $SNAP_DATA/config
 mkdir -p ${SNAP_DATA}/config
-for service in security-api-gateway security-secret-store core-command config-seed core-data core-metadata export-client export-distro support-logging support-notifications support-scheduler sys-mgmt-agent; do
+for service in security-api-gateway security-secret-store core-command config-seed core-data core-metadata export-client export-distro support-logging support-notifications support-scheduler sys-mgmt-agent device-modbus device-mqtt device-random; do
     if [ ! -f "${SNAP_DATA}/config/${service}/res/configuration.toml" ]; then
         mkdir -p "${SNAP_DATA}/config/${service}/res"
         cp ${SNAP}/config/${service}/res/configuration.toml "${SNAP_DATA}/config/${service}/res/configuration.toml"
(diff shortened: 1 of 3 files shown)


snap/hooks/install
```diff
@@ -33,10 +33,30 @@ for service in security-api-gateway security-secret-store core-command config-se
...
-    mkdir -p "${SNAP_DATA}/config/config-seed/res/properties/device-virtual"
-    cp "${SNAP}/config/config-seed/res/properties/device-virtual/application.properties" "${SNAP_DATA}/config/config-seed/res/properties/device-virtual/application.properties"
+# handle device-mqtt driver configuration
+cp ${SNAP}/config/device-mqtt/res/configuration-driver.toml "${SNAP_DATA}/config/device-mqtt/res/configuration-driver.toml"
+
+# handle device-random device profile
+cp ${SNAP}/config/device-random/res/device.random.yaml "${SNAP_DATA}/config/device-random/res/device.random.yaml"
+
+# also handle java services' application.properties
+for jsvc in device-virtual; do
```
tonyespy: Since we're down to a single service, doesn't it make sense to get rid of this **for** loop, and just keep the existing logic in the master version?

snap/hooks/install
```diff
@@ -33,10 +33,30 @@ for service in security-api-gateway security-secret-store core-command config-se
...
+    if [ ! -f "${SNAP_DATA}/config/config-seed/res/properties/$jsvc/application.properties" ]; then
+        mkdir -p "${SNAP_DATA}/config/config-seed/res/properties/$jsvc"
+        cp "${SNAP}/config/config-seed/res/properties/$jsvc/application.properties" "${SNAP_DATA}/config/config-seed/res/properties/$jsvc/application.properties"
+        # also replace SNAP_DATA and SNAP_COMMON in the application files
+        sed -i -e "s@\$SNAP_COMMON@$SNAP_COMMON@g" -e "s@\$SNAP_DATA@$SNAP_DATA@g" "${SNAP_DATA}/config/config-seed/res/properties/$jsvc/application.properties"
+    fi
+done
+
+# create support-rulesengine directories for templates/rules
```
tonyespy: Isn't rules-engine till Java? If so, I guess we do need to keep the afore-mentioned (pun intended) **for** loop, and add rules-engine to the list of svcs?

tonyespy: Actually, after review, we haven't yet added rules-engine to master (it's yet another contribution to delhi that we didn't forward port at the time), so we should just remove all refs to rules-engine in this file and keep the **for** loop after all.

snap/hooks/install
anonymouse64: On the off chance that we need to add more java services, I'd rather leave it here.

snap/hooks/install
anonymouse64: This was leftover from my merge resolution with cherry-picking, I removed it.

Issue #1089, opened 2019-02-07
Pass CorrelationID through EventPublisher


tsconn23:
We need to extend handling of the correlation-id value through the EventPublisher that allows core-data to send events to export-distro. Soon, core-data will also be sent events to other endpoints via pub/sub such as the app-functions-sdk. The event being published needs to retain the correlation-id so that it can be utilized on the subscription side in order to add the value to an HTTP request. For example, in both export-distro and app-functions-sdk, the event can be forwarded onto the next level of processing via HTTP. We need to track this egress from edge-x and log it along with the duration of time taken to obtain a response.

rsdmike:
Hey @tsconn23 just want to clarify, this sounds like it is focused on the existing export distro. Is this intended to include the "Envelope" idea that we discussed at the Boston F2F? Or did you want to create a separate issue for that?

PR #1090, opened 2019-02-08
Propogating correlation-id through EventPublisher

tsconn23:
#1089 

Note the following:
* Aliased all imports from `github.com/edgexfoundry/pkg/models` as `contract`
* Introduction of internal/pkg/correlation/model/event.go which wraps the contract event adding a CorrelationID property.
* The above model is sent through the EventPublisher to export-distro
* Export-distro continues to export the regular contract model, as it should, however if exporting via HTTP the correlation-id header is passed to the next endpoint.
* In the case of HTTP export, a log entry is written showing the correlation-id and duration of the call

I have also tested this using a small export client that I've written and I confirmed receipt of the same correlation-id that was generated when the event was ingested by core-data.

### DIFF
--- a/internal/core/data/device_test.go
+++ b/internal/core/data/device_test.go
@@ -12,19 +12,20 @@ import (
 	"time"
 
 	"github.com/edgexfoundry/edgex-go/internal/core/data/messaging"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/correlation/models"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 
 	"github.com/edgexfoundry/edgex-go/pkg/clients/logging"
 	"github.com/edgexfoundry/edgex-go/pkg/clients/metadata/mocks"
 	"github.com/edgexfoundry/edgex-go/pkg/clients/types"
-	"github.com/edgexfoundry/edgex-go/pkg/models"
+	contract "github.com/edgexfoundry/edgex-go/pkg/models"
 
 	"github.com/globalsign/mgo/bson"
 	"github.com/gorilla/mux"
 	"github.com/stretchr/testify/mock"
 )
 
-var testEvent models.Event
+var testEvent contract.Event
 var testRoutes *mux.Router
 
 const (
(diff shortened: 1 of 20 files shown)


tsconn23:
@rsdmike You're welcome to comment/review as well. I couldn't add you as a reviewer.

internal/pkg/correlation/models/event.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	for _, r := range e.Readings {
+		event.Readings = append(event.Readings, r)
+	}
+	return &event
+}
+
+func (e Event) MarshalJSON() ([]byte, error) {
+	test := struct {
```
lenny-intel: Why using the name "test" here? i.e. had to look back to see it this was a unit test.

internal/pkg/correlation/models/event.go
```diff
@@ -0,0 +1,88 @@
...
+
+package models
+
+import (
+	"encoding/json"
+
+	contract "github.com/edgexfoundry/edgex-go/pkg/models"
+)
+
+type Event struct {
```
lenny-intel: It is confusing that this has the same type name as edgex-go/pkg/models/Event. How about EventEnvelope or something?

rsdmike:
I took a look with @lenny-intel, he's gonna add a couple comments. I get why its put in `internal/` to minimize impact but that means I cant import it in the SDK right?

internal/core/data/device_test.go
```diff
@@ -12,19 +12,20 @@ import (
...
 
 	"github.com/edgexfoundry/edgex-go/internal/core/data/messaging"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/correlation/models"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 
 	"github.com/edgexfoundry/edgex-go/pkg/clients/logging"
 	"github.com/edgexfoundry/edgex-go/pkg/clients/metadata/mocks"
 	"github.com/edgexfoundry/edgex-go/pkg/clients/types"
-	"github.com/edgexfoundry/edgex-go/pkg/models"
+	contract "github.com/edgexfoundry/edgex-go/pkg/models"
```
lenny-intel: Will this be "contract" once moved to the core-contracts module resolving the namespace clash?

internal/pkg/correlation/models/event.go
tsconn23: Well it's technically in another namespace so there isn't a name collision and it's got a different scope.

lenny-intel: Yes technically different types. We are coming from a readability point of view and ease to understand which type we are dealing with at a point in the code. Make sense?

PR #1097, opened 2019-02-11
Fixed issues in core-metadata DB test

jduranf:
This PR fixes several issues in data base test (related with #977).
Anyway, MongoDB test still fails in core-metadata:
```
$ go test -tags mongoRunning
--- FAIL: TestMongoDB (0.20s)
    client_test.go:24: This test needs to have a running mongo on localhost
    db_metadata.go:684: There should be 1 deviceServices instead of 0
FAIL
exit status 1
FAIL	github.com/edgexfoundry/edgex-go/internal/pkg/db/mongo	0.198s
```
This seems to be a bad implementation in `GetDeviceServicesByAddressableId()` and several similar functions (`GetScheduleEventsByAddressableId()` or `GetDevicesByProfileId()`) 

### DIFF
--- a/internal/core/metadata/interfaces/db.go
+++ b/internal/core/metadata/interfaces/db.go
@@ -72,7 +72,7 @@ type DBClient interface {
 	GetDeviceProfilesByManufacturerModel(man string, mod string) ([]contract.DeviceProfile, error)
 	GetDeviceProfilesByManufacturer(man string) ([]contract.DeviceProfile, error)
 	GetDeviceProfileByName(n string) (contract.DeviceProfile, error)
-	GetDeviceProfilesUsingCommand(c contract.Command) ([]contract.DeviceProfile, error)
+	GetDeviceProfilesByCommandId(id string) ([]contract.DeviceProfile, error)
 
 	// Addressable
 	UpdateAddressable(a contract.Addressable) error
--- a/internal/core/metadata/interfaces/mocks/DBClient.go
+++ b/internal/core/metadata/interfaces/mocks/DBClient.go
@@ -836,13 +836,13 @@ func (_m *DBClient) GetDeviceProfilesByModel(m string) ([]models.DeviceProfile, 
 	return r0, r1
 }
 
-// GetDeviceProfilesUsingCommand provides a mock function with given fields: c
-func (_m *DBClient) GetDeviceProfilesUsingCommand(c models.Command) ([]models.DeviceProfile, error) {
-	ret := _m.Called(c)
+// GetDeviceProfilesByCommandId provides a mock function with given fields: c
+func (_m *DBClient) GetDeviceProfilesByCommandId(id string) ([]models.DeviceProfile, error) {
+	ret := _m.Called(id)
 
 	var r0 []models.DeviceProfile
-	if rf, ok := ret.Get(0).(func(models.Command) []models.DeviceProfile); ok {
-		r0 = rf(c)
+	if rf, ok := ret.Get(0).(func(string) []models.DeviceProfile); ok {
+		r0 = rf(id)
 	} else {
 		if ret.Get(0) != nil {
 			r0 = ret.Get(0).([]models.DeviceProfile)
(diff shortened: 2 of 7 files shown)


tsconn23:
The error from db_metadata.go above can be fixed with the following mod:

```
func (mc MongoClient) GetDeviceServicesByAddressableId(id string) ([]contract.DeviceService, error) {
	addr, err := mc.getAddressableById(id)
	if err != nil {
		return []contract.DeviceService{}, err
	}
	return mc.getDeviceServices(bson.M{"addressable.$id": addr.Id})
}
```

However, the next test in the suite is this:
```
deviceServices, err = db.GetDeviceServicesByAddressableId(uuid.New().String())
if err != nil {
	t.Fatalf("Error getting deviceServices by addressable id")
}
```
I don't understand why that should be handled as a fatal since if the test asks for an addressable for a brand new ID that it generates on the fly, the appropriate response is db.NotFoundErr. The test should instead check for `if err == nil` in my opinion.

jduranf:
@tsconn23 agreed with booth points. I fixed the related issues in MongoDB and in DB test.
There are still two fails:
```
$ go test -tags mongoRunning
--- FAIL: TestMongoDB (0.39s)
    client_test.go:24: This test needs to have a running mongo on localhost
    db_metadata.go:982: Error getting ScheduleEvents mgoGetScheduleEventsByAddressableId Invalid Object ID
FAIL
exit status 1
FAIL	github.com/edgexfoundry/edgex-go/internal/pkg/db/mongo	0.395s
```
and:
```
$ go test -tags mongoRunning
--- FAIL: TestMongoDB (0.29s)
    client_test.go:24: This test needs to have a running mongo on localhost
    db_metadata.go:1164: Error getting deviceProfiles Item not found
FAIL
exit status 1
FAIL	github.com/edgexfoundry/edgex-go/internal/pkg/db/mongo	0.289s
```

Sorry for not trying to fix those issues, but we don't use MongoDB. We use DB test because is a great tool to check if our DB implementations works fine or not. Anyway, if you know how to fix it, I will modify again the PR.

tsconn23:
@jduranf The `mgoGetScheduleEventsByAddressableId` should be fixable in the same manner as my edits to the `GetDeviceServicesByAddressableId`. The issue is that the incoming ID could be either a UUID or a BSON so we have to go and load the Addressable itself. We do this because the lookup on the DBRef requires a BSON ID.

On the `Error getting deviceProfiles` it looks like the test is wrong. It's another case where a new ID is generated on the fly and so we should expect an `errNotFound` instead of failing if we receive one.
```
c.Id = bson.NewObjectId().Hex()
deviceProfiles, err = db.GetDeviceProfilesUsingCommand(c)
if err != nil {
	t.Fatalf("Error getting deviceProfiles %v", err)
}
```

jduranf:
@tsconn23 completely agree with the second point, in fact (if it's fine) I changed `GetDeviceProfilesUsingCommand(c contract.Command)` by `GetDeviceProfilesByCommandId(id string)`, which is more clear and similar to other calls.

Regarding the first issue, if you try to get a schedule event using Postman, you will receive something like this:
```
{
    "created": 1550007465939,
    "modified": 1550007465939,
    "origin": 0,
    "id": "5c633ca964562a6780e00596",
    "name": "name0",
    "schedule": "name0",
    "addressable": {
        "created": 0,
        "modified": 0,
        "origin": 0,
        "name": "se_name0",
        "protocol": "se_name0",
        "address": "se_name0",
        "path": "se_name0",
        "publisher": "se_name0",
        "user": "se_name0",
        "password": "se_name0",
        "topic": "se_name0",
        "baseURL": "se_name0://se_name0:0",
        "url": "se_name0://se_name0:0se_name0"
    },
    "parameters": "name0",
    "service": "name0"
}
```
In this case, the addressable ID it's missing. If you get the addressable, the ID isn't missing, but it's an UUID (I don't know if this is the cause of the issue or not).

As the schedule events will be removed, I don't know if it's worth to spend more time trying to fix this issue or should be better to just remove schedule checks from db test.

tsconn23:
@jduranf I would recommend removing "schedule event" tests. I see that [this file](https://github.com/edgexfoundry/edgex-go/blob/master/internal/pkg/db/mongo/schedule_event.go) is still hanging around and wasn't moved into the models folder. And the reason for that is because we plan to remove schedule events. However we can't actually do that until after the next Device Services working group meeting on Feb-25. So yes, I would remove the relevant tests from test/db_metadata.go.

Also, good time to rebase this branch now that the module work has gone in.

jduranf:
@tsconn23 removed all DB schedule tests from core-metadata, and added support-schedule DB tests. Now all tests are passed :smiley:

Issue #1098, opened 2019-02-11
Add Provision Watcher client


jduranf:
Provision Watcher client is not implemented and should be implemented in order to enable provision watchers feature in the device service

jduranf:
Will be done in [go-mod-core-contracts](https://github.com/edgexfoundry/go-mod-core-contracts/issues/5)

PR #1099, opened 2019-02-11
Added provision watcher client

jduranf:
Implements #1098 

### DIFF
--- a/pkg/clients/metadata/provision_watcher.go
+++ b/pkg/clients/metadata/provision_watcher.go
@@ -1,0 +1,145 @@
+/*******************************************************************************
+ * Copyright 2018 Dell Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+package metadata
+
+import (
+	"context"
+	"encoding/json"
+	"net/url"
+
+	"github.com/edgexfoundry/edgex-go/pkg/clients"
+	"github.com/edgexfoundry/edgex-go/pkg/clients/types"
+	"github.com/edgexfoundry/edgex-go/pkg/models"
+)
+
+/*
+Provision Watcher client for interacting with the provision watcher section of metadata
+*/
+type ProvisionWatcherClient interface {
+	Add(dev *models.ProvisionWatcher, ctx context.Context) (string, error)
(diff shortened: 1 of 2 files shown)


pkg/clients/metadata/provision_watcher.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/edgexfoundry/edgex-go/pkg/models"
+)
+
+/*
+Provision Watcher client for interacting with the provision watcher section of metadata
+*/
+type ProvisionWatcherClient interface {
+	Add(dev *models.ProvisionWatcher, ctx context.Context) (string, error)
+	Delete(id string, ctx context.Context) error
+	ProvisionWatcher(id string, ctx context.Context) (models.ProvisionWatcher, error)
```
tsconn23: Rename to GetById?

pkg/clients/metadata/provision_watcher.go
```diff
@@ -0,0 +1,145 @@
...
+	ProvisionWatcherForName(name string, ctx context.Context) (models.ProvisionWatcher, error)
```
tsconn23: Rename to GetByName?

pkg/clients/metadata/provision_watcher.go
```diff
@@ -0,0 +1,145 @@
...
+	ProvisionWatchers(ctx context.Context) ([]models.ProvisionWatcher, error)
```
tsconn23: Rename to List()?

tonyespy:
Note `ProvisionWatchers` are part of dynamic device discovery, which is currently a stretch goal for the device SDKs in Edinburgh.

jduranf:
@tsconn23 I agree with your comments, but all other clients in [core-data](https://github.com/edgexfoundry/edgex-go/blob/master/pkg/clients/coredata/value_descriptor.go), [core-metadata](https://github.com/edgexfoundry/edgex-go/blob/master/pkg/clients/metadata/device.go) or [scheduler](https://github.com/edgexfoundry/edgex-go/blob/master/pkg/clients/scheduler/interval.go) have the same structure, which I tried to follow. Maybe should be fine to rename all clients to remove the redundant prefix in another PR, but for coherency, in this PR I prefer to keep those names.

tsconn23:
@tonyespy Yes, I had the same thought and considered possibly putting a hold on this PR until the Fuji dev cycle. @jduranf Are you in practical need of this change or are you contributing simply because it's a gap in our functionality?

jduranf:
@tsconn23 I'm in practical need. If you want to add this feature in Fuji, should be great for me to, at least, add the provision watcher client in edgex-go (and maybe the callback system in another PR, but I have to check that).

tsconn23:
@jduranf Depending on the timing of edgex-go's conversion to a module and subsequent import of https://github.com/edgexfoundry/go-mod-core-contracts you might think about making this change in the go-mod-core-contracts repo.

Once go-mod-core-contracts is imported to edgex-go, the /pkg folder in edgex-go will be deleted.

We'll be discussing timing on the Core WG call tomorrow.

jduranf:
This will be done in [go-mod-core-contracts](https://github.com/edgexfoundry/go-mod-core-contracts/pull/6)

PR #1105, opened 2019-02-14
snap/many: remove deprecated device-virtual service

tonyespy:
This PR removes the last Java device service, device-virtual, as it's scheduled to be replaced with a new Go-based version as part of the Edinburgh release.

This PR can be tested by building the snap, installing it, and verifying that device-virtual no longer appears as a service using the following command:

$ snap services | grep device-virtual


snap/hooks/install
```diff
@@ -123,7 +113,7 @@ if [ ! -f ${SNAP_DATA}/config/security-api-gateway/kong.conf ]; then
 fi
 
 # finally, disable services in the snap that aren't meant to come up by default
-# by default, we want the export-*, support-*, and device-virtual services off
-for svc in export-distro export-client support-notifications support-scheduler support-logging device-virtual device-modbus device-mqtt device-random; do
+# by default, we want the export-* and support-* off
```
anonymouse64: small nitpick, can we make this 
```
# by default, we want the export-*, support-*, and device-* services off
```

(since the device-mqtt, etc. are also disabled by default)

PR #1106, opened 2019-02-14
snapcraft.yaml: fix architectures key

anonymouse64:
The current architectures key worked with the old snapcraft, but this
is now interpreted as the snap being a multiarch snap.

Currently, if the snap is built it with a recent version of snapcraft it shows up as "edgexfoundry_..._multi.snap", which is wrong.


anonymouse64:
recheck

Issue #1107, opened 2019-02-14
Convert edgex-go to use Go Modules


tsconn23:
Convert edgex-go to use go module support, and remove glide.

PR #1108, opened 2019-02-14
Eliminate glide. Use go modules.

tsconn23:
#1107 

- In `master` we have refs to docker dependencies in glide.yaml. In order to eliminate these from the `go.mod` I had to remove the code from the SMA that depended on it. These changes should align with PR #926.
- Tagging @tonyespy and @anonymouse64 as reviewers in case this affects the Snap. If we need to consider timing of the merge, please let me know in the comments below.
- The Makefile `prepare` section will be removed altogether once I can coordinate with the DevOps team to remove the `make prepare` statement from their JJB scripts. For now, it's been blanked out and shouldn't do anything.


akramtexas:
Recheck

go.mod
```diff
@@ -0,0 +1,28 @@
+module github.com/edgexfoundry/edgex-go
+
+require (
```
akramtexas: Confirming that this is exactly the set of `require`s that I, too, got generated separately, when I created a `go module` out of the edgex-go repo 👍

internal/system/agent/init.go
```diff
@@ -80,8 +80,6 @@ func newExecutorClient(operationsType string) (interfaces.ExecutorClient, error)
 	switch operationsType {
 	case "os":
 		return &executor.ExecuteOs{}, nil
-	case "docker":
```
akramtexas: This is correct for now, but will change (later) _after_ another feature branch, per [PR#926](https://github.com/edgexfoundry/edgex-go/pull/926), has been merged into `master`.

akramtexas:
In addition to `go.mod` (which is already checked-in), I recommend checking-in `go.sum` as well: I hear support (in online forums) for checking-in `go.sum`. The idea is: _So checking in go.sum file can help others to verify that they are using the same modules as you._

akramtexas:
- I'm able to check-out branch `issue_1107` and successfully build edgex-go repo as a `go module` (locally). Only tweak I had to make was to tweak the default `go.mod` that was generated on two lines:
1. Top-line tweaked to: `module github.com/edgexfoundry/edgex-go`
2. This line, also in `go.mod` removed: `github.com/edgexfoundry/edgex-go v0.0.0-20190214175459-0eb9eaa966a0`
- LGTM 👍

akramtexas:
- The only other comment I've got is the observation of the following error [in the `Jenkins` build](https://jenkins.edgexfoundry.org/job/edgex-go-master-verify-go/1911/console):
```
14:00:25 internal/core/data/interfaces/mocks/DBClient.go:5:8: cannot find package "github.com/stretchr/testify/mock" in any of:
14:00:25 	/usr/local/go/src/github.com/stretchr/testify/mock (from $GOROOT)
14:00:25 	/home/jenkins/1911/gopath/src/github.com/stretchr/testify/mock (from $GOPATH)
14:00:25 make: *** [test] Error 1
```

Makefile
```diff
@@ -65,7 +65,6 @@ test:
 	go vet ./...
 
 prepare:
```
jduranf: Why no t remove `prepare` completely?

jduranf:
I agree with @lenny-intel. This is the way it's done in [device-sdk-go](https://github.com/edgexfoundry/device-sdk-go/blob/master/Makefile#L3)

Makefile
tsconn23: See third bullet in PR comments at the top.

tsconn23:
@lenny-intel @anonymouse64 I'll add the GO111MODULE env var to the Makefile now. I was hoping to avoid that since I've been working locally outside of my GOPATH and if it's set, it could cause issues for local devs. But in the interest of getting this through CI, will make the change.

lenny-intel:
@tsconn23 , I think it is actually needed for local devs that still have their fork cloned into their GOPATH, just as the CI is doing.

tsconn23:
@lenny-intel Set environment variable. Build is still failing to import deps
```
09:49:33 + make prepare
09:49:33 make: Nothing to be done for `prepare'.
09:49:33 + make test
09:49:33 go test -cover ./...
09:49:35 internal/pkg/config/loader.go:25:2: cannot find package "github.com/BurntSushi/toml" in any of:
09:49:35 	/usr/local/go/src/github.com/BurntSushi/toml (from $GOROOT)
09:49:35 	/home/jenkins/1913/gopath/src/github.com/BurntSushi/toml (from $GOPATH)
```

Makefile
lenny-intel: Checking with DevOps, but I think the empty make prepare will need to stay since the CI template also builds delhi and california branches which require CI to call make prepare.

lenny-intel:
@tsconn23 , I'll work with DevOps to figure out why.

Makefile
lenny-intel: @tsconn23 , @jduranf , DevOps confirmed that the empty make prepare needs to stay in order to support the CI template build of delhi and california branches.

lenny-intel: @tsconn23 , all references to "go" need to be "$(GO)" so GO111MODULE=on is used. 

These are the targets that need to be updated:
build: $(MICROSERVICES)
	go build ./...

test:
	go test -cover ./...
	go vet ./...

tsconn23:
recheck

tsconn23:
recheck

tsconn23:
@anonymouse64 Can you verify that I've addressed your review feedback please?

Makefile
```diff
@@ -61,11 +60,10 @@ clean:
 	rm -f $(MICROSERVICES)
 
 test:
-	go test -cover ./...
-	go vet ./...
+	GO111MODULE=on go test -cover ./...
```
anonymouse64: Why is this not just `$(GO)`?

tsconn23: I get zmq test error when I do that. Probably something to do with the CGO enabled=off. Note we build the services relying on zeromq with CGO enabled.

anonymouse64: Understood, that makes sense

Issue #1124, opened 2019-02-21
Update Dockerfile base image to Golang 1.11-alpine


tsconn23:
Update the service Dockerfiles to use `golang:1.11-alpine` as the base image.

tsconn23:
I'm going to add some information here based on my experience when creating the associated PR. It appears to be the case that the checksum calculation of go module dependencies can change between patch releases. This means that if your local machine (or a build agent) is running Go 1.11.2, for example, and your images are based on Go 1.11.5 you might see an error like this:

```
go: verifying github.com/hashicorp/go-rootcerts@v1.0.0: checksum mismatch
	downloaded: h1:Rqb66Oo1X/eSV1x66xbDccZjhJigjg0+e82kpwzSwCI=
	go.sum:     h1:ueI78wUjYExhCvMLow4icJnayNNFRgy0d9EGs/a1T44=
make: *** [Makefile:42: cmd/export-client/export-client] Error 1
```

In order to get the PR through, I have commented out the `COPY go.sum .` directive from our Dockerfiles. This is because the go.sum in the short term will be created on a Jenkins build agent which has Go 1.11.2 and will cause the above error when we go to build inside the container using Go 1.11.5. Once the build agent is modified to also utilize the `golang:1.11-alpine` as a base, we can uncomment this line from the Dockerfiles as the go.sum generated by the build agent should be identical to what would be generated inside the container.

Pending resolution of this issue
https://github.com/edgexfoundry/ci-management/issues/310

PR #1126, opened 2019-02-21
Use golang:1.11-alpine as the base image. Ingore go.sum for the moment.

ernestojeda:
Changes made:
  - Ignore go.sum. Currently there is an issue with the way checksums are computed across patch versions of Go. We have observed that if a go.sum is generated from Go 1.11.2 it may not work with Go 1.11.5. Typically this is seen as a checksum mismatch error on specific modules (github.com/hashicorp/go-rootcerts for example). Until we can get some resolution on this, go.sum will be ignored.
  - Changed the base image to be golang:1.11-alpine as was discussed in the DevOps WG call.
  - Commented out APK repo pinning.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,6 @@
 docs
+
+# Ignoring the go.sum file. There are checksum issues with
+# different versions of Golang. Remove once all Go versions
+# are in sync
+go.sum
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ glide.lock
 docker/docker-compose.yml
 .project
 .vscode/
+go.sum
 
 # binary files
 cmd/config-seed/config-seed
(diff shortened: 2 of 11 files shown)


ernestojeda:
@tsconn23 @jamesrgregg @MightyNerdEric

tsconn23:
So this is a duplicate of #1127 which I started working on after this morning's DevOps call in which I created issue #1124 and assigned it to myself. Let's coordinate better in the future.

The one thing this doesn't have that #1127 does is the following line to set the source for the update repositories in several of the Dockerfiles
`RUN sed -e 's/dl-cdn[.]alpinelinux.org/nl.alpinelinux.org/g' -i~ /etc/apk/repositories`

ernestojeda:
I am closing this one.

PR #1131, opened 2019-02-25
Fix update command and other issues

### LINKED ISSUE #1130: Update command doesn't work
When trying to update a command, this error is received:
`level=ERROR ts=2019-02-25T13:51:05.282653404Z app=edgex-core-metadata source=rest_command.go:100 msg="Error updating command: duplicate command name in device profile"
`

jduranf:
This PR fixes #1130 and also:
* Makes update addressable database independent
* Fixed panic error when trying to get commands and max limit is reached
* When trying to add an existing value descriptor, now the error is "duplicate name for value descriptor" instead of "value descriptor still referenced by readings"

And ok, maybe I should create different PR for each of those issues, but at least there is one commit for each one :smiley:

### DIFF
--- a/internal/core/data/errors/types.go
+++ b/internal/core/data/errors/types.go
@@ -15,6 +15,7 @@ package errors
 
 import (
 	"fmt"
+
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 )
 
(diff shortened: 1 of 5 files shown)


tsconn23:
@jduranf I'm curious, are we missing a test case in our blackbox tests that might have caught this? I'm not aware of an error currently being thrown in this execution path.

tsconn23:
FYI that blackbox test core-data/valuedescriptor_error_4xx/104 is failing. It's expecting that a descriptor of name "co2" not be present but it is, so the test fails. Doesn't look to me like the logic is wrong, so maybe the test is incorrect based on your above changes. Please take a look.

```
POST http://localhost:48080/api/v1/valuedescriptor

BODY
{
  "name":"co2",
  "min":"12",
  "max":"15",
  "type":"F",
  "uomLabel":"degreecel",
  "defaultValue":"0",
  "formatting":"%s",
  "labels":["NHCO2","hvac"]
}
```
I run the core-data/valuedescriptor tests first, and then the error_4xx tests

jduranf:
@tsconn23 I found this issue while adapting our database system to the database changes done in Edinburgh (we use BoltDB instead of MongoDB).

Regarding Blackbox test failing, the same issue also happens using the master branch, so doesn't seems related with this PR.

Finally, rebase have been done :)

Issue #1101, opened 2019-02-11
Change Device Addressable for More Broad Protocol Support


tsconn23:
Resulting from the Device Service Working Group call 11-Feb-19

- The [Device](https://github.com/edgexfoundry/edgex-go/blob/master/pkg/models/device.go) struct should contain a map of the properties necessary for connecting to a device. This will be a more flexible structure to connect via any number of protocols rather than overloading a type ([Addressable](https://github.com/edgexfoundry/edgex-go/blob/master/pkg/models/addressable.go)) that is specific to HTTP
- Since the Addressable type will remain for other purposes, the above supporting property should be called `ProtocolProperties` of type `map[string]string`
- The `ProtocolProperties` will be stored as part of the Device document in the database. No sharing is required as in the case of the existing Addressable, thus no requirements for uniqueness.
- Remember to modify the DB storage logic to remove the Addressable mgmt for Devices.

tsconn23:
@SteveOss @tonyespy for visibility

SteveOss:
@tsconn23 Need to remember that a device can support multiple protocol types (e.g. modbus-ip/modbus-rtu) so in terms of modelling a device I think we have (using Thrift as a definition):

struct ProtocolProperty
{
  1: string name;
  2: string value;
}
typedef list <ProtocolProperty> ProtocolProperties;

struct Protocol
{
  1: string type;
  2: ProtocolProperties properties;
}
typedef sequence <Protocol> Protocols;

struct Device
{
  1: Protocols protocols;
}

tsconn23:
This issue is going to get moved to the new https://github.com/edgexfoundry/go-mod-core-contracts repo

tsconn23:
Closing this issue. Moved to core-contracts repo.

tsconn23:
Re-opening issue to track work done within the edgex-go repo itself to support this.

Issue #946, opened 2018-12-17
Add AutoEvents (Subscriptions) to device model


iain-anderson:
The purpose of this change is to provide a new mechanism for device services to autonomously generate events, and to be able to take advantage of devices which can push new readings rather than relying on polling.

We propose that a "subscriptions" field is added to "device". This will contain 0..n subscriptions which are defined as { resource, interval, onChange }.
resource : the name of the resource in the device profile which describes the event to generate.
interval : for devices which must be polled, the time between queries. For devices which can push, the minimum interval between generated events (ie, a rate limit).
onChange : boolean. If true, only generate events when readings change.

tonyespy:
A few comments on this approach:

 * This provides less functionality than the existing `Schedule/ScheduleEvent` model:
   *  A `ScheduleEvent` specifies an endpoint `Addressable.Path` that can be called vs. a resource name (by which I think you mean a device resource?). A `ScheduleEvent` can call a device resource or resource (command).
   * A `ScheduleEvent` also has an optional `Addressable.Method` field and an optional `Parameters` field allowing PUT commands to be scheduled. We could just say that scheduling PUT commands must be done via the Scheduler service, and punt on supporting this via the internal SDK scheduler, but we're still removing functionality which exists today.
   * I prefer **frequency** vs. **interval**, as this keeps the terminology similar to a `Schedule` object. I'm also not sure we should overload this field in order to rate-limit a device which can push. I think it probably deserves its own field called **limit** or **rate-limit**.
   * **onChange** can only be supported if we implement a readings cache (not sure if we agreed to implement one for Edinburgh? If so, we should make sure there's an issue filed).

My original idea for this was to leverage device profiles similar to what @cloudxxx8 did in his original device-virtual device service. The idea would be to add a new `ScheduleProperty` to each device resource, so a device resource could have a **value**, **unit**, and **schedule** properties. There are problems with this approach too:

 * it requires the device service to have its own persistent store
 * like your proposal, it doesn't allow scheduling of resource commands, only device resources
 * it means all devices inherit the same frequency, which could lead to a Core Data meltdown unless some we built in some kind random +/- delta added

So, I think your approach, with some tweaks to address some of the points of raised is probably the best approach, however we should probably discuss more in the new year...

iain-anderson:
With regard to the above,

I'm not sure what you mean in the discussion of endpoints. The ScheduleEvent
targets a command by specifying its path, and the proposed mechanism targets
[...]

PR #1136, opened 2019-02-27
Removed Device-Addressable relationship

tsconn23:
#1101 
#946 

- Addressables are no longer associated with Devices, no need for uniqueness constraints
- Removal of endpoints related to obtaining devices by their addressable
- Includes RAML updates
- Note associated blackbox test issue https://github.com/edgexfoundry/blackbox-testing/issues/169 It might make sense to leave this PR in place for testing with that issue. If we merge this PR before the tests are ready, the daily run will have errors in it.


internal/core/metadata/rest_device.go
```diff
@@ -63,20 +63,13 @@ func restAddNewDevice(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		// Try by ID
-		addressable, err = dbClient.GetAddressableById(d.Addressable.Id)
-		if err != nil {
-			LoggingClient.Error(err.Error())
-			http.Error(w, err.Error()+": A device must be associated to an Addressable", http.StatusBadRequest)
-			return
-		}
+	// Protocol check
+	if len(d.Protocols) == 0 {
```
jduranf: In case of device-random or device-simple, Protocols length can be 0. So maybe should be better to remove this check.

internal/core/metadata/rest_addressable.go
```diff
@@ -331,24 +322,12 @@ func restGetAddressableByAddress(w http.ResponseWriter, r *http.Request) {
 // Notify the associated device services for changes in the device addressable
 func notifyAddressableAssociates(a models.Addressable, action string) error {
```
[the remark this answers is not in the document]
tsconn23: @tonyespy @cloudxxx8 for visibility and in case you have any input.

@jduranf I think I prefer to leave this for now because if, as you say, it doesn't matter to the Device Service when an addressable is modified, then the logic farther down the stack wherein an Addressable's callback URL is called will never fire. This is because, presumably, no callback URL will have been specified. Unless of course, Tony or Cloud +1 your comment for confirmation. Then I'll remove it.

internal/core/metadata/rest_device.go
tsconn23: I'll take this under advisement. In the case of device-random, it currently creates an addressable from the configuration.toml
```
[[DeviceList]]
  Name = "Random-Integer-Generator01"
  Profile = "Random-Integer-Generator"
  Description = "Example of Device Random"
  Labels = [ "device-random-example" ]
  [DeviceList.Addressable]
    Address = "device-random-01"
    Protocol = "OTHER"
```
Presumably the Device Service team is going to create a corollary for protocol entries from the configuration, and I expect they'll convert this one to populate some kind of dummy protocol entry.

@tonyespy @cloudxxx8 for visibility

akramtexas:
Recheck

internal/pkg/db/mongo/models/device.go
```diff
@@ -64,12 +67,19 @@ func (d *Device) ToContract(dsTransform deviceServiceTransform, dpTransform devi
 	result.AdminState = d.AdminState
 	result.OperatingState = d.OperatingState
 
-	aModel, err := aTransform.DBRefToAddressable(d.Addressable)
+	p := make(map[string]map[string]string)
```
lenny-intel: I notice a lot of single character variable names. Why not just use protocols in this instance? Makes it clear what it is going to be used for.

tsconn23: In Go, similar to C, it's an idiom to be concise and so, similar to C, if the scope of a variable is short enough from declaration to usage then the name can be less descriptive. In this case, the lifespan of variable `p` is 6 lines. Also, `p` doesn't really have much conceptual weight. It's allocated as a bucket for unmarshalling a byte array. If we were to instead do something of importance with `p`, then yes I'd agree it should be called `protocols`.

internal/core/metadata/rest_device.go
jduranf: Currently device-random creates an addressable because otherwise it's not possible to create the device. As you said, with this check you are forcing to use a dummy protocol entry in some cases, which can be avoided.

internal/pkg/db/mongo/models/device.go
SteveOss: Looking at the diff above, using something more descriptive than "p" makes it more readable to me. In general I agree with being concise, but I don't interpret this as meaning use single letter variables all over the place.

internal/core/metadata/rest_addressable.go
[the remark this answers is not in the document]
cloudxxx8: I can confirm the Device addressable callback handler is going to be removed.  Thus, we could remove it safely.  The only addressable in Device Service is Device Service itself, and we do not need to notify Device Service when its addressable is changed.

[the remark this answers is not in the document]
tsconn23: Per @cloudxxx8 's confirmation, this has been removed.

internal/pkg/db/mongo/models/device.go
lenny-intel: @tsconn23 , that is the part of C that Go devs have inherited that I dislike... ;-) Moving on... :-)

Issue #1078, opened 2019-02-05
Add CPU telemetry retrieved from each service to the metrics handler


brandonforster:
* Add metrics (telemetry) retrieved from each service to the metricsHandler in https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/data/router.go

Subtask of #832. Related to #690. Will require #1077 to be complete.

brandonforster:
Complete.

tsconn23:
Which PR is this a part of?

Issue #1077, opened 2019-02-05
Update metrics handler in SMA to return RAM and/or CPU usage.


brandonforster:
Update SMA to return CPU in addition to memory metric on getMetrics function in https://github.com/edgexfoundry/edgex-go/blob/master/internal/system/agent/services.go.

Subtask of #832. This service should be written so that the metrics service can serve requests for:

* RAM only
* CPU only
* RAM and CPU

Additionally, this should be implemented with the abstraction layer defined in #1076 and should be implemented with an eye toward extensibility for the future.

brandonforster:
Closing. It was determined that CPU and RAM do not need be separated. If that decision changes, this issue can be reopened.

Issue #1076, opened 2019-02-05
Create a CPU metrics abstraction layer for the SMA to use


brandonforster:
provide an abstraction that allows different implementations if necessary

Subtask of #832. Must complete #1074 first.

brandonforster:
Complete. The `Telemetry` interface will abstract OS dependent implementations.

tsconn23:
Which PR is this a part of?

Issue #1074, opened 2019-02-05
PoC: Capture CPU metrics data for Linux


brandonforster:
find appropriate means to capture CPU (seems to be OS dependent see:  https://stackoverflow.com/questions/11356330/getting-cpu-usage-with-golang)

Subtask of #832. This item will be complete when there is a library/function in EdgeX that can monitor CPU usage on the machine it is running on.

brandonforster:
Complete.

tsconn23:
Which PR is this a part of?

sjt-xfwh:
i advice to use this package "github.com/shirou/gopsutil", this package support both windows and linux platform, and we can get more information, like cpu disk mem host net and so on.

PR #1137, opened 2019-02-27
Added a CPU usage metrics endpoint for Linux.

brandonforster:
Issues #832, #1137, #1074, #1076, #1077, #1078


go.sum
```diff
@@ -0,0 +1,326 @@
+bitbucket.org/bertimus9/systemstat v0.0.0-20180207000608-0eeff89b0690/go.mod h1:Ulb78X89vxKYgdL24HMTiXYHlyHEvruOj1ZPlqeNEZM=
```
jduranf: This file should not be added

internal/core/command/init.go
```diff
@@ -18,22 +18,26 @@ import (
...
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logging"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
 	"github.com/pkg/errors"
+
+	"github.com/edgexfoundry/edgex-go/internal"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/consul"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/startup"
+	"github.com/edgexfoundry/edgex-go/internal/system/agent/telemetry"
```
jduranf: In my opinion, telemetry package should be located in `github.com/edgexfoundry/edgex-go/pkg/telemetry`

internal/system/agent/telemetry/telemetry.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type CpuUsage struct {
+	Busy, // time used by all processes. this ideally does not include system processes.
+	Idle, // time used by the idle process
+	Total uint64 // reported sum total of all usage
+}
+
+func GetCpuUsageAverage(usageAvg *float64, lastSample *CpuUsage) {
+	for {
```
jduranf: Should be great to have a [Do Once](https://golang.org/pkg/sync/#Once.Do) control

brandonforster: That's a great idea. I had no idea this was a construct. Thank you.

internal/core/command/init.go
tsconn23: I agree with this too. Telemetry is a general use package, so having it internal to system/agent isn't its proper context.

internal/system/agent/telemetry/telemetry.go
```diff
@@ -0,0 +1,52 @@
...
+ *******************************************************************************/
+
+package telemetry
+
+import "time"
+
+type Telemetry interface {
+	PollCpu() (cpuSnapshot CpuUsage)
+	AvgCpuUsage(init, final CpuUsage) (avg float64)
+}
```
tsconn23: This interface isn't used anywhere. Please remove.

internal/support/scheduler/router.go
```diff
@@ -91,14 +91,16 @@ func metricsHandler(w http.ResponseWriter, _ *http.Request) {
...
+	t.Memory.TotalAlloc = rtm.TotalAlloc
+	t.Memory.Sys = rtm.Sys
+	t.Memory.Mallocs = rtm.Mallocs
+	t.Memory.Frees = rtm.Frees
 
 	// Live objects = Mallocs - Frees
-	t.LiveObjects = t.Mallocs - t.Frees
+	t.Memory.LiveObjects = t.Memory.Mallocs - t.Memory.Frees
+
+	t.CpuBusyAvg = usageAvg
```
tsconn23: Suggest encapsulating all of the above logic into your `telemetry.SystemUsage` type.

```
t := telemetry.NewSystemUsage(usageAvg)
encode(t, w)
```
where NewSystemUsage() contains all of the above gathering/assigning of stats. This logic is the same in all routers, so would be good to consolidate/refactor.

internal/system/agent/logger/logs.go
```diff
@@ -25,4 +25,8 @@ var LoggingClient logger.LoggingClient
 
 func BuildLoggingClient(config *interfaces.ConfigurationStruct, logTarget string) {
 	LoggingClient = logger.NewClient(internal.SystemManagementAgentServiceKey, config.EnableRemoteLogging, logTarget, config.LoggingLevel)
+}
+
+func BuildDefaultClient() {
```
tsconn23: This function can be eliminated if telemetry is moved into its own package and out of system/agent

go.sum
tsconn23: Yes, go.sum should be in the .gitignore

internal/pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,78 @@
...
+	// Live objects = Mallocs - Frees
+	s.Memory.LiveObjects = s.Memory.Mallocs - s.Memory.Frees
+
+	s.CpuBusyAvg = usageAvg
+	
+	return s
+}
+
+func GetCpuUsageAverage(usageAvg *float64, lastSample *CpuUsage) {
+	var once sync.Once
```
jduranf: I think should better to have usageAvg as a variable in this package, not declared in all init functions. Also the once variable should declared out of this function, otherwise I think it's useless. So this function can be something like:
```go
var once sync.Once
var usageAvg float64

func StartCpuUsageAverage() {
	lastSample := telemetry.PollCpu()

	once.Do(func() {
	...
```

internal/pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,78 @@
...
+	LiveObjects uint64
+}
+
+type CpuUsage struct {
+	Busy, // time used by all processes. this ideally does not include system processes.
+	Idle, // time used by the idle process
+	Total uint64 // reported sum total of all usage
+}
+
+func NewSystemUsage(usageAvg float64) (s SystemUsage) {
```
jduranf: Taking in account my other comment, this can function can by changed by:
```go
func GetSystemUsage() (s SystemUsage) {
	...
}
```

brandonforster:
Changes made.

Issue #833, opened 2018-11-19
SMA will provide health/status API (Edinburgh Roadmap Item)


jpwhitemn:
The SMA will add an API to provide the health/status check of a service, but this will be a proxy to the configuration/registry service.

PR #1139, opened 2019-02-28
Added a health handler to the SMA.

brandonforster:
Issue #833 

This health handler checks whether microservices are alive via the Consul URL.


internal/system/agent/services.go
```diff
@@ -131,6 +136,49 @@ func getMetrics(services []string, ctx context.Context) (MetricsRespMap, error)
 	return m, nil
 }
 
+func getHealth(services []string, ctx context.Context) HealthRespMap {
```
akramtexas: Instead of returning a `HealthRespMap` struct alone from this func, return _along_ with it a var of type `error`.

internal/system/agent/router.go
```diff
@@ -127,6 +131,22 @@ func metricsHandler(w http.ResponseWriter, r *http.Request) {
...
+func healthHandler(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	logs.LoggingClient.Debug("health status data requested", "service names", vars)
+
+	list := vars["services"]
+	var services []string
+	services = strings.Split(list, ",")
+
+	ctx := r.Context()
+	send := getHealth(services, ctx)
```
akramtexas: - I like how you've followed (for `getHealth()`) the pattern each that was established for `getMetrics()` and `getConfig()` - This makes for code consistency and ease of understanding 👍 
- Even though we _currently_ end up discarding the `error` returned from the respective call each to `getMetrics()` and `getConfig()`, I advise making the flow consistent with the latter two as we may, for example, uncover (or find) a use case where we need to make use of `error`. (See my other note regarding returning a tuple from the func `getHealth()`.

internal/system/agent/services.go
```diff
@@ -131,6 +136,49 @@ func getMetrics(services []string, ctx context.Context) (MetricsRespMap, error)
...
+
+	for _, service := range services {
+
+		h.Health[service] = ""
+
+		if !isKnownServiceKey(service) {
+			logs.LoggingClient.Warn("unknown service found getting health", "service name", service)
+		}
+
+		err := client.IsServiceAvailable(service)
```
akramtexas: - With the addition of `go-mod-registry` (to the go.mod file) in place and `go-mod-registry` downloaded into the module cache: I would be interested in hearing about any testing steps you may have carried out (using Postman, etc.) to trace the flow that involves the `Registry` service.

internal/system/agent/services.go
```diff
@@ -131,6 +136,49 @@ func getMetrics(services []string, ctx context.Context) (MetricsRespMap, error)
 	return m, nil
 }
 
+func getHealth(services []string, ctx context.Context) HealthRespMap {
+
+	h := HealthRespMap{}
+	h.Health = map[string]interface{}{}
```
tsconn23: Per my comment above, this could be
`health := map[string]interface{}{}`

`health` can then be returned along with the error, if any, as Akram points out below. When this is then serialized into JSON as the HTTP response, the consumer won't have the extra 'Health' node in the JSON to parse (this applies to the existing Config and Metrics endpoints as well). In my opinion, including `Health` or `Metrics` or `Config` in the response is redundant because that's already indicated by the path of the REST request. It will also require case based parsing on the consumer side.

Having said all of that, if you still want to have a type that wraps the response data (which is a [legit pattern](https://martinfowler.com/eaaCatalog/dataTransferObject.html)), you can do this with just a single type instead of three.

internal/system/agent/services.go
```diff
@@ -131,6 +136,49 @@ func getMetrics(services []string, ctx context.Context) (MetricsRespMap, error)
...
+		if !isKnownServiceKey(service) {
+			logs.LoggingClient.Warn("unknown service found getting health", "service name", service)
+		}
+
+		err := client.IsServiceAvailable(service)
+		// the registry service returns nil for a healthy service
+		if err != nil {
+			h.Health[service] = err.Error()
+		} else {
+			h.Health[service] = "alive and healthy"
```
tsconn23: I would just make this value "true" for easy conversion to a boolean

internal/system/agent/services.go
brandonforster: I started Consul, populated it with configuration from `config-seed`, then started `core-data` with Consul configuration, then started the SMA and tested health. I confirmed that health is green when `core-data` was running, and red when I killed `core-data`

internal/system/agent/router.go
```diff
@@ -103,8 +107,7 @@ func configHandler(w http.ResponseWriter, r *http.Request) {
 	services = strings.Split(list, ",")
 
 	ctx := r.Context()
-	var send = ConfigRespMap{}
-	send, _ = getConfig(services, ctx)
+	send, _ := getConfig(services, ctx)
```
anonymouse64: this error should be checked here

internal/system/agent/router.go
```diff
@@ -127,6 +130,21 @@ func metricsHandler(w http.ResponseWriter, r *http.Request) {
...
 
+func healthHandler(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	logs.LoggingClient.Debug("health status data requested", "service names", vars)
+
+	list := vars["services"]
+	var services []string
+	services = strings.Split(list, ",")
+
+	send, _ := getHealth(services)
```
anonymouse64: the error here should be checked

internal/system/agent/services.go
```diff
@@ -81,54 +86,94 @@ func InvokeOperation(action string, services []string) bool {
...
+	}
+	client, err := factory.NewRegistryClient(config)
+	if err != nil {
+		logs.LoggingClient.Error("could not create registry client", "errMsg", err.Error())
+		return health, err
+	}
+
+	for _, service := range services {
+
+		health[service] = ""
```
anonymouse64: why is this initialized here? It's always set to something else below in all control flows

internal/system/agent/services.go
```diff
@@ -81,54 +86,94 @@ func InvokeOperation(action string, services []string) bool {
...
 
-	m := MetricsRespMap{}
-	m.Metrics = map[string]interface{}{}
+	metrics := make(map[string]interface{})
 
 	// Loop through requested actions, along with respectively-supplied parameters.
 	for _, service := range services {
 
-		m.Metrics[service] = ""
+		metrics[service] = ""
```
anonymouse64: why does this get set here? It is reset to something else on all control flows below

internal/system/agent/router.go
```diff
@@ -120,7 +129,34 @@ func metricsHandler(w http.ResponseWriter, r *http.Request) {
...
+func healthHandler(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	logs.LoggingClient.Debug("health status data requested", "service names", vars)
+
+	list := vars["services"]
+	var services []string
+	services = strings.Split(list, ",")
+
+	send, err := getHealth(services)
+
```
anonymouse64: please remove this newline

internal/system/agent/router.go
```diff
@@ -103,8 +107,13 @@ func configHandler(w http.ResponseWriter, r *http.Request) {
 	services = strings.Split(list, ",")
 
 	ctx := r.Context()
-	var send = ConfigRespMap{}
-	send, _ = getConfig(services, ctx)
+	send, err := getConfig(services, ctx)
+
```
anonymouse64: please remove this newline

Issue #1075, opened 2019-02-05
PoC: Capture CPU metrics data for Windows


brandonforster:
find appropriate means to capture CPU (seems to be OS dependent see: https://stackoverflow.com/questions/11356330/getting-cpu-usage-with-golang)

Subtask of #832. This item will be complete when there is a library/function in EdgeX that can monitor CPU usage on the machine it is running on.

This subtask is a lower priority than #1074.

PR #1143, opened 2019-03-05
Adding support for Windows CPU Telemetry

rsdmike:
Took a stab at this. 

I can't add reviewers :(  @brandonforster can you take a look? #1075 

Signed-off-by: Mike Johanson <michael.johanson@intel.com>


internal/pkg/telemetry/windows_cpu.go
```diff
@@ -0,0 +1,84 @@
+// +build windows
```
brandonforster: You're going to need to exclude Windows from `unimplented_cpu`.

internal/pkg/telemetry/windows_cpu.go
```diff
@@ -0,0 +1,84 @@
...
+		Idle:  init.Idle,
+		Total: init.Total,
+	}
+
+	windowsFinal := systemstat.CPUSample{
+		Idle:  final.Idle,
+		Total: final.Total,
+	}
+
+	avg = systemstat.GetSimpleCPUAverage(windowsInit, windowsFinal).BusyPct
```
brandonforster: I used SystemStat's average calculation because I was already importing the library for polling CPU usage. Importing it here just to use their simple mean calculation seems unnecessary.

internal/pkg/telemetry/windows_cpu.go
```diff
@@ -0,0 +1,84 @@
...
+		uintptr(unsafe.Pointer(lpIdleTime)),
+		uintptr(unsafe.Pointer(lpKernelTime)),
+		uintptr(unsafe.Pointer(lpUserTime)))
+
+	return ret != 0
+}
+
+// FILETIME Struct from: http://msdn.microsoft.com/en-us/library/windows/desktop/ms724284.aspx
+type FILETIME struct {
+	DwLowDateTime  uint32
```
brandonforster: Using Hungarian notation doesn't seem like the answer in Go.

internal/pkg/telemetry/windows_cpu.go
```diff
@@ -0,0 +1,84 @@
...
+	avg = systemstat.GetSimpleCPUAverage(windowsInit, windowsFinal).BusyPct
+
+	if avg < .000001 || math.IsNaN(avg) {
+		return 0.0
+	}
+
+	return avg
+}
+
+func GetSystemTimes(lpIdleTime, lpKernelTime, lpUserTime *FILETIME) bool {
```
brandonforster: This should probably be a private function.

internal/pkg/telemetry/windows_cpu.go
```diff
@@ -0,0 +1,84 @@
...
+	ret, _, _ := procGetSystemTimes.Call(
+		uintptr(unsafe.Pointer(lpIdleTime)),
+		uintptr(unsafe.Pointer(lpKernelTime)),
+		uintptr(unsafe.Pointer(lpUserTime)))
+
+	return ret != 0
+}
+
+// FILETIME Struct from: http://msdn.microsoft.com/en-us/library/windows/desktop/ms724284.aspx
```
brandonforster: This seems like an awkward way to give attribution.

internal/pkg/telemetry/windows_cpu.go
rsdmike: Yup makes sense, will do.

internal/pkg/telemetry/windows_cpu.go
rsdmike: I totally agree with this, but since I didn't know exactly what was in that library and since is was already pulled in for linux i decided for the sake of staying consistent that I'd use it. Happy to remove it.

internal/pkg/telemetry/windows_cpu.go
rsdmike: No requirement that I'm aware of, just did it to match the Windows API so it was easy to follow what was what.

internal/pkg/telemetry/windows_cpu.go
rsdmike: If you've got a better way to do it, I'm all for it :)  I just put the link there so it was easy to see where/why/what that struct is doing. Wasn't so much for attribution as it was FYI.

tsconn23:
recheck

tsconn23:
@rsdmike LGTM. Needs rebase.

tsconn23:
@rsdmike Why are there so many other people besides you showing up on the commit stream? Are you rebasing or cherry picking?

rsdmike:
I have no clue, i ran `git pull --rebase upstream master` ? Cherry-basing maybe?

rsdmike:
Look like should be good now after squashing?

brandonforster:
recheck

PR #1151, opened 2019-03-12
build cassandra from source

### LINKED ISSUE #1103: Implement more resilient method of downloading cassandra for the snap

anonymouse64:
This will resolve #1103 for edinburgh/master.

Tested by building the snap, installing it and making sure that the security services come up correctly.


anonymouse64:
recheck

anonymouse64:
recheck

PR #1152, opened 2019-03-12
build cassandra from source in snap for delhi

### LINKED ISSUE #1103: Implement more resilient method of downloading cassandra for the snap

anonymouse64:
Duplicate of #1151 but for Delhi.

Closes #1103 for Delhi

Note this is to fix failing edge builds such as https://jenkins.edgexfoundry.org/view/Snap/job/edgex-go-snap-delhi-stage-snap/134/ in the unlikely event we need to do another Delhi release


anonymouse64:
recheck

Issue #1144, opened 2019-03-05
Remove metadata/scheduler endpoints


tsconn23:
All routes under the following parents need to be removed from core-metadata

/api/v1/schedule
/api/v1/scheduleevent

tsconn23:
Creation of PR for this issue pending merge of contract changes
https://github.com/edgexfoundry/go-mod-core-contracts/pull/28

jduranf:
Done via #1155

PR #1155, opened 2019-03-12
Removal of Schedule and ScheduleEvent

tsconn23:
#1144 


tsconn23:
@jduranf Thanks for comments. I think I addressed the items you pointed out. As for documentation, yes that will be handled via separate PRs. @mhall119 Has created an issue https://github.com/edgexfoundry/edgex-go/issues/1158

PR #1160, opened 2019-03-12
Fix SMA's ExecutorPath; implement snap-specific executor

### LINKED ISSUE #1132: Add snap SMA executor
With the new implementation of the SMA executor from #926 and #1122 we will need to add an executor script for the snap to use `snapctl` to control services from the SMA.

### LINKED ISSUE #1159: SMA is unable to use the configured Executor
After configuring the SMA's executor to be 

```toml
ExecutorPath = '/snap/edgexfoundry/x1/bin/sys-mgmt-agent-snap-executor.sh'
```

and running the SMA from the snap, it is unable to use the configured executor because the `exectorClient` is never initialized in internal/system/agent/init.go. This is testable by issuing a stop command to the SMA with:

```
curl --header "Content-Type: application/json" --request POST --data '{"action":"stop","services":["edgex-core-command"]}' localhost:48090/api/v1/operation
```

and seeing the log files show a failure like this:

```
Mar 12 11:15:24 edgexfoundry.sys-mgmt-agent[8102]: level=WARN ts=2019-03-12T16:15:24.107334021Z app=edgex-sys-mgmt-agent source=services.go:70 msg="stopping not supported with specified executor"
```

anonymouse64:
* Fixes SMA support for using the interfaces defined for each operation 
* Has SMA use consul now that it has support for it
* Add snap specific SMA executor that uses `snapctl` to start/stop/restart services

To test this, build the snap and after installing, run:

```
$ curl --header "Content-Type: application/json" --request POST --data '{"action":"stop","services":["edgex-core-command"]}' localhost:48090/api/v1/operation
Done. Started the requested services.
```

you should then see `core-command` go down is inactive:

```
$ snap services edgexfoundry.core-command
Service                    Startup  Current   Notes
edgexfoundry.core-command  enabled  inactive  -
```

Then start the service with:

```
$ curl --header "Content-Type: application/json" --request POST --data '{"action":"start","services":["edgex-core-command"]}' localhost:48090/api/v1/operation; echo
Done. Started the requested services.
```

and core-command should be alive again:

```
$ snap services edgexfoundry.core-command
Service                    Startup  Current  Notes
edgexfoundry.core-command  enabled  active   -
```

Finally, test `restart` with:
```
$ curl --header "Content-Type: application/json" --request POST --data '{"action":"restart","services":["edgex-core-command"]}' localhost:48090/api/v1/operation; echo
Done. Restarted the requested services.
```
and see that core-command is running again:

```
$ snap services edgexfoundry.core-command
Service                    Startup  Current  Notes
edgexfoundry.core-command  enabled  active   -
```

Closes #1132 
Closes #1159

### DIFF
--- a/internal/system/agent/init.go
+++ b/internal/system/agent/init.go
@@ -100,6 +100,7 @@ func Init(useRegistry bool) bool {
 	if Configuration == nil {
 		return false
 	}
+	executorClient = &ExecuteApp{}
 
 	if useRegistry && registryClient != nil {
 		chErrors = make(chan error)
--- a/cmd/sys-mgmt-agent/res/configuration.toml
+++ b/cmd/sys-mgmt-agent/res/configuration.toml
@@ -5,6 +5,8 @@
 # In the first release of the SMA, the manifest will be static.
 # In the future, the manifest may be more dynamic or even provided by some 3rd party orchestrator.
 
+ExecutorPath = '/path/to/the/file'
+
 [Writable]
 ResendLimit = 2
 LogLevel = 'INFO'
@@ -20,7 +22,6 @@ ReadMaxLimit = 100
 StartupMsg = 'This is the System Management Agent Service'
 Timeout = 5000
 FormatSpecifier = '%(\\d+\\$)?([-#+ 0(\\<]*)?(\\d+)?(\\.\\d+)?([tT])?([a-zA-Z%])'
-ExecutorPath = '/path/to/the/file'
 
 [Registry]
 Host = 'localhost'
--- a/internal/system/agent/execution_engine.go
+++ b/internal/system/agent/execution_engine.go
@@ -13,8 +13,6 @@ func runExec(service string, operation string) error {
 	// Preparing the call to the executor app.
 	cmd := exec.Command(Configuration.ExecutorPath, service, operation)
 
-	cmd.Dir = Configuration.ExecutorPath
-
 	_, err := cmd.CombinedOutput()
 	if err != nil {
 		LoggingClient.Error(fmt.Sprintf("an error occurred in calling executor on service %s where requested operation was %s: %v ", service, operation, err.Error()))
--- a/snap/hooks/install
+++ b/snap/hooks/install
@@ -28,8 +28,11 @@ for service in security-api-gateway security-secret-store core-command config-se
     if [ ! -f "${SNAP_DATA}/config/${service}/res/configuration.toml" ]; then
         mkdir -p "${SNAP_DATA}/config/${service}/res"
         cp ${SNAP}/config/${service}/res/configuration.toml "${SNAP_DATA}/config/${service}/res/configuration.toml"
-        # do replacement of the SNAP_DATA and SNAP_COMMON environment variables in the config files
-        sed -i -e "s@\$SNAP_COMMON@$SNAP_COMMON@g" -e "s@\$SNAP_DATA@$SNAP_DATA@g" "${SNAP_DATA}/config/${service}/res/configuration.toml"
+        # do replacement of the $SNAP, $SNAP_DATA, $SNAP_COMMON environment variables in the config files
+        sed -i -e "s@\$SNAP_COMMON@$SNAP_COMMON@g" \
+            -e "s@\$SNAP_DATA@$SNAP_DATA@g" \
+            -e "s@\$SNAP@$SNAP@g" \
+            "${SNAP_DATA}/config/${service}/res/configuration.toml"
     fi
 done
 
--- a/snap/local/runtime-helpers/bin/sys-mgmt-agent-snap-executor.sh
+++ b/snap/local/runtime-helpers/bin/sys-mgmt-agent-snap-executor.sh
@@ -1,0 +1,37 @@
+#!/bin/bash -e
+
+# the first argument is the service
+# the second argument is the operation, one of "stop", "start", "restart"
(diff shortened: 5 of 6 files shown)


anonymouse64:
recheck

akramtexas:
Recheck

anonymouse64:
recheck

anonymouse64:
recheck

anonymouse64:
recheck

anonymouse64:
recheck

akramtexas:
Recheck

anonymouse64:
recheck

internal/system/agent/init.go
```diff
@@ -100,6 +100,7 @@ func Init(useRegistry bool) bool {
 	if Configuration == nil {
 		return false
 	}
+	executorClient = &ExecuteApp{}
```
akramtexas: - Please confirm your use of `executorClient` (the empty interface) since it, in conjunction with `type ExecuteApp struct`, no longer performs any (useful) functionality...

akramtexas: - Other than this (and other references to `executorClient`), the PR LGTM.

anonymouse64: Note that the `executorClient` is _of type_ `interface {}`, which means that it could hold any type (or also be `nil`). It importantly does perform useful functionality because this `executorClient` variable is used to actually run commands here: https://github.com/edgexfoundry/edgex-go/blob/4d274133c690b91bd282f994267e04970c2eb073/internal/system/agent/services.go#L49 we attempt to type assert this variable into the various operation interfaces. If we never initialize this variable like being done here, then when we get to that snippet of code above, `executorClient` is nil, and as such can't match the interface of say `ServiceStarter`, and as such the operation fails to be executed.

It's not clear to me the code could ever have worked without this initialization of the `executorClient` variable. 

Also note that while I did originally ask that you remove the function initializing this variable, I only meant that the _function_ was useless because it was just assigning a variable, I did not mean the variable was useless. Apologies if that comment on your earlier PR was unclear.

akramtexas: The detailed response is very helpful @anonymouse64 👍 - I'll catch up with you offline, so we can get this squared away, thanks.

akramtexas:
Recheck

internal/system/agent/init.go
akramtexas: - Was trying to catch up and chat... In summary, please proceed with the initialization of the `executorClient` variable (the way you have it currently should work).

tsconn23:
@anonymouse64 @tonyespy Since this has been approved does it just need a rebase before merge?

anonymouse64:
@tsconn23 I don't want to merge this until Tony has reviewed it

anonymouse64:
recheck

anonymouse64:
We should also take the time to make sure that the SMA is registered with consul using the `--registry` flag in this PR since that's now a thing the SMA can do.

akramtexas:
Recheck

akramtexas:
> We should also take the time to make sure that the SMA is registered with consul using the `--registry` flag in this PR since that's now a thing the SMA can do.

- Yes, we’re telling the SMA to start with the `--registry` flag, and effectively using `consul` primarily (and exclusively) as a passthrough. 
- That's right, this is a relatively _recent_ update to `system mgmt`,  given how we’re making use of the `Registry` interface to `consul`, with the `Registry` design and implementation having been contributed to EdgeX by Lenny Goodell (Intel).
- This PR, once merged with `master` branch, will pick up those updated, having to do with how  the SMA is getting registered with consul, via `config-seed`, like so, in `../edgex-go/internal/seed/config/populate.go`:
```
func listDirectories() [9]string {
	var names = [9]string{internal.CoreMetaDataServiceKey, internal.CoreCommandServiceKey, internal.CoreDataServiceKey,
		internal.ExportDistroServiceKey, internal.ExportClientServiceKey, internal.SupportLoggingServiceKey,
		internal.SupportSchedulerServiceKey, internal.SupportNotificationsServiceKey, internal.SystemManagementAgentServiceKey}
...
...
```
- So yes, we need to run `config-seed` at the outset to ensure the SMA, along with other `core services`, is getting registered with consul 👍

anonymouse64:
@tonyespy @akramtexas I figured out what the issue was with not importing configuration here, it was becuase the `ExecutorPath` configuration item in the configuration.toml file was in the wrong location. It should have been before any of the "sub" struct items in the configuration file, so that it would properly get imported as a root level key, instead it was under "Service" and so the toml importer just silently ignored it when trying to store the map of keys into the configuration struct as the config.ServiceInfo struct doesn't have an ExecutorPath key.

This is good to be reviewed again

anonymouse64:
Note the snap builds are failing due to https://github.com/edgexfoundry/edgex-go/issues/1225

anonymouse64:
Note that before you can actually test the snap, we will need to pull in https://github.com/edgexfoundry/edgex-go/pull/1232

anonymouse64:
@tonyespy see my previous comment :-)

akramtexas:
> Looks good to me. The only thing that was surprising is that the requests seem to be synchronous (i.e. the response isn't sent till the SMA operation finishes).

@tonyespy If you're referring to the results from the `/api/v1/operation` endpoint (aka *operation* endpoint) of the `SMA`, that's intentional as we want to include in the response (from the `SMA`) the result (i.e. `success` or `failure`) of the *operation*.

Issue #1167, opened 2019-03-14
Got "500 Internal Server Error" when posting /api/v1/valuedescriptor with duplicate name


cherrycl:
OS version: ubuntu 16.04
EdgeX version: 1.0.0
Environment details: Docker container
Steps to reproduce:

1. start up EdgeX
2. Post /api/v1/valuedescriptor with the following body which is existed on DB.

> Method: POST
URL: http://localhost:48080/api/v1/valuedescriptor
Body:
  {
  "name":"co2",
  "min":"12",
  "max":"15",
  "type":"F",
  "uomLabel":"degreecel",
  "defaultValue":"0",
  "formatting":"%s",
  "labels":["NHCO2","hvac"]
  }

Actual results: Got return code "500" and  response body "duplicate name for value descriptor: co2".
Expected results: Return code 409 with response body "duplicate name for value descriptor: co2"

brandonforster:
fixed by #1178

PR #1178, opened 2019-03-19
Added missing error case to ValueDescriptor router

brandonforster:
Issue #1167.


brandonforster:
@cherrycl, tagging you for visibility. This should fix the issue; good catch.

Issue #769, opened 2018-11-08
SMA doesn't handle errors when using operation endpoint


anonymouse64:
For example, when using SMA with docker, attempting to stop a non-existent service returns HTTP 200 and this message:

```
Done. Stopped the requested services.
```

To reproduce do:
```
curl --header "Content-Type: application/json" --request POST --data '{"action":"stop","services":["edgex-not-a-thing"]}' localhost:48090/api/v1/operation; echo
```

tsconn23:
@akramtexas Confirmed this is still an issue. Is there a plan for resolution?

akramtexas:
> @akramtexas Confirmed this is still an issue. Is there a plan for resolution?

@tsconn23 Yep, this is right next up on my `TO-DO` list, and I plan on resolving this as soon as I'm done with the work underway for `integrating sys-mgmt with device-sdk-go` 👍

akramtexas:
This work is next up as I wrap up the `sys-mgmt` integration work with `device-sdk-go`.

akramtexas:
Thanks for picking up the bug-fix @brandonforster 👍

brandonforster:
Closed by #1184

PR #1184, opened 2019-03-19
Addressed SMA operation error handling issues

brandonforster:
Issue #769 

### DIFF
--- a/internal/system/agent/router.go
+++ b/internal/system/agent/router.go
@@ -17,10 +17,12 @@ package agent
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"strings"
 
+	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 	"github.com/edgexfoundry/go-mod-core-contracts/models"
 	"github.com/gorilla/mux"
 
@@ -63,7 +65,7 @@ func operationHandler(w http.ResponseWriter, r *http.Request) {
 	err = o.UnmarshalJSON(b)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
-		LoggingClient.Error("error during decoding: %v", err.Error())
+		LoggingClient.Error("error during decoding")
 		return
 	} else if o.Action == "" {
 		http.Error(w, err.Error(), http.StatusBadRequest)
@@ -72,24 +74,38 @@ func operationHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	switch o.Action {
-
-	// Call the appropriate internal function (respectively, to stop, start, or restart the service(s)).
 	case STOP:
-		InvokeOperation(STOP, o.Services)
-		w.WriteHeader(http.StatusOK)
-		w.Write([]byte("Done. Stopped the requested services."))
+		err := InvokeOperation(STOP, o.Services)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			w.Write([]byte(fmt.Sprintf("Error: %s", err.Error())))
+		} else {
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte("Done. Stopped the requested services."))
+		}
+
 		break
 
 	case START:
-		InvokeOperation(START, o.Services)
-		w.WriteHeader(http.StatusOK)
-		w.Write([]byte("Done. Started the requested services."))
+		err := InvokeOperation(START, o.Services)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			w.Write([]byte(fmt.Sprintf("Error: %s", err.Error())))
+		} else {
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte("Done. Started the requested services."))
+		}
 		break
 
 	case RESTART:
-		InvokeOperation(RESTART, o.Services)
-		w.WriteHeader(http.StatusOK)
-		w.Write([]byte("Done. Restarted the requested services."))
+		err:= InvokeOperation(RESTART, o.Services)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			w.Write([]byte(fmt.Sprintf("Error: %s", err.Error())))
+		} else {
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte("Done. Restarted the requested services."))
+		}
 		break
 
 	default:
@@ -99,7 +115,7 @@ func operationHandler(w http.ResponseWriter, r *http.Request) {
 
 func configHandler(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
-	LoggingClient.Debug("service names: ", vars)
+	LoggingClient.Debug("retrieved service names")
 
 	list := vars["services"]
 	var services []string
@@ -113,14 +129,14 @@ func configHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	w.Header().Add("Content-Type", "application/json")
+	w.Header().Add(clients.ContentType, clients.ContentTypeJSON)
 	encode(send, w)
 	return
 }
 
 func metricsHandler(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
-	LoggingClient.Debug("service names: ", vars)
+	LoggingClient.Debug("retrieved service names")
 
 	list := vars["services"]
 	var services []string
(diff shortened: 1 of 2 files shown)


internal/system/agent/router.go
```diff
@@ -63,7 +65,7 @@ func operationHandler(w http.ResponseWriter, r *http.Request) {
 	err = o.UnmarshalJSON(b)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
-		LoggingClient.Error("error during decoding: %v", err.Error())
+		LoggingClient.Error("error during decoding", "errMsg", err.Error())
```
akramtexas: Avoid multiple keys (for example, here we have _two_: `error during decoding` and `errMsg`). Collapse them like so:
```
LoggingClient.Error("error during decoding: %v", err.Error())
```

Or simply:
```
```

internal/system/agent/router.go
```diff
@@ -72,24 +74,38 @@ func operationHandler(w http.ResponseWriter, r *http.Request) {
...
-
-	// Call the appropriate internal function (respectively, to stop, start, or restart the service(s)).
 	case STOP:
-		InvokeOperation(STOP, o.Services)
-		w.WriteHeader(http.StatusOK)
-		w.Write([]byte("Done. Stopped the requested services."))
+		err := InvokeOperation(STOP, o.Services)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			w.Write([]byte(fmt.Sprintf("Error: %s", err.Error())))
```
akramtexas: Please confirm the syntax, but I would simplify this like so:
```
w.Write([]byte(fmt.Sprintf(err.Error())))
```

internal/system/agent/router.go
```diff
@@ -72,24 +74,38 @@ func operationHandler(w http.ResponseWriter, r *http.Request) {
...
 		break
 
 	case START:
-		InvokeOperation(START, o.Services)
-		w.WriteHeader(http.StatusOK)
-		w.Write([]byte("Done. Started the requested services."))
+		err := InvokeOperation(START, o.Services)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			w.Write([]byte(fmt.Sprintf("Error: %s", err.Error())))
```
akramtexas: Please confirm the syntax, but I would simplify this like so:
```
w.Write([]byte(fmt.Sprintf(err.Error())))
```

internal/system/agent/router.go
```diff
@@ -72,24 +74,38 @@ func operationHandler(w http.ResponseWriter, r *http.Request) {
...
 		break
 
 	case RESTART:
-		InvokeOperation(RESTART, o.Services)
-		w.WriteHeader(http.StatusOK)
-		w.Write([]byte("Done. Restarted the requested services."))
+		err:= InvokeOperation(RESTART, o.Services)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			w.Write([]byte(fmt.Sprintf("Error: %s", err.Error())))
```
akramtexas: Please confirm the syntax, but I would simplify this like so:
```
w.Write([]byte(fmt.Sprintf(err.Error())))
```

internal/system/agent/router.go
```diff
@@ -99,7 +115,7 @@ func operationHandler(w http.ResponseWriter, r *http.Request) {
 
 func configHandler(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
-	LoggingClient.Debug("service names: ", vars)
+	LoggingClient.Debug("retrieved service names","names", vars)
```
akramtexas: Here also, and elsewhere, please avoid multiple keys. Collapse them as indicated in the earlier comment 👍

internal/system/agent/router.go
```diff
@@ -113,14 +129,14 @@ func configHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	w.Header().Add("Content-Type", "application/json")
+	w.Header().Add(clients.ContentType, clients.ContentTypeJSON)
```
akramtexas: - This update, involving `clients.ContentType` etc. here (and similar updates elsewhere in this PR) is fine by me. I have recognized, though, that the _other_ style, involving strings instead, such as `"Content-Type"` etc., is heavily prevalent in the code base.
- While this may sounds like a nit, having a consistent style across the code base promotes easier comprehension by making searches (`grepping`) that much easier 👍

internal/system/agent/services.go
```diff
@@ -33,14 +33,14 @@ const (
 	RESTART = "restart"
 )
 
-func InvokeOperation(action string, services []string) bool {
+func InvokeOperation(action string, services []string) error {
 
 	// Loop through requested operation, along with respectively-supplied parameters.
 	for _, service := range services {
-		LoggingClient.Info("invoking operation %s on service %s", action, service)
+		LoggingClient.Info("invoking operation", "operation", action, "service", service)
```
akramtexas: Similar comments (as earlier) apply here, etc., too, regarding multiple keys in logging.

internal/system/agent/router.go
brandonforster: I thought having a "presentation" to the user was valuable since this is going to be returned in the API. A success returns "Done", "Error:" for an error seemed appropriate. I can easily change it, however.

You don't need the `Sprintf` if there's nothing to format.

internal/system/agent/router.go
```diff
@@ -72,24 +74,38 @@ func operationHandler(w http.ResponseWriter, r *http.Request) {
...
 	switch o.Action {
-
-	// Call the appropriate internal function (respectively, to stop, start, or restart the service(s)).
 	case STOP:
-		InvokeOperation(STOP, o.Services)
-		w.WriteHeader(http.StatusOK)
-		w.Write([]byte("Done. Stopped the requested services."))
+		err := InvokeOperation(STOP, o.Services)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
```
akramtexas: 👍 Reporting back with an `http.StatusInternalServerError` makes the most sense.

internal/system/agent/router.go
tsconn23: We should utilize the constants here IMHO.

internal/system/agent/router.go
```diff
@@ -150,24 +166,24 @@ func healthHandler(w http.ResponseWriter, r *http.Request) {
...
 func encode(i interface{}, w http.ResponseWriter) {
-	w.Header().Add("Content-Type", "application/json")
+	w.Header().Add(clients.ContentType, clients.ContentTypeJSON)
 
 	enc := json.NewEncoder(w)
 	err := enc.Encode(i)
 
 	if err != nil {
-		LoggingClient.Error("error during encoding: %v", err.Error())
+		LoggingClient.Error("error during encoding", "errMsg", err.Error())
```
tsconn23: `errMsg` key is redundant.

internal/system/agent/router.go
```diff
@@ -150,24 +166,24 @@ func healthHandler(w http.ResponseWriter, r *http.Request) {
 	send, err := getHealth(services)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
-		LoggingClient.Error("could not retrieve health status: %v: ", err.Error())
+		LoggingClient.Error("could not retrieve health status", "errMsg", err.Error())
```
tsconn23: `errMsg` key is redundant

internal/system/agent/router.go
```diff
@@ -113,14 +129,14 @@ func configHandler(w http.ResponseWriter, r *http.Request) {
...
-	w.Header().Add("Content-Type", "application/json")
+	w.Header().Add(clients.ContentType, clients.ContentTypeJSON)
 	encode(send, w)
 	return
 }
 
 func metricsHandler(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
-	LoggingClient.Debug("service names: ", vars)
+	LoggingClient.Debug("retrieved service names","names", vars)
```
tsconn23: No `names` key please

internal/system/agent/services.go
```diff
@@ -33,14 +33,14 @@ const (
...
 
 		if !isKnownServiceKey(service) {
-			LoggingClient.Warn("unknown service %s during invocation", service)
+			LoggingClient.Warn("unknown service", "service", service)
```
tsconn23: No `service` key please. Same applies below.

internal/system/agent/services.go
```diff
@@ -49,34 +49,43 @@ func InvokeOperation(action string, services []string) bool {
...
 				err := starter.Start(service)
 				if err != nil {
-					LoggingClient.Error(fmt.Sprintf("error starting service %s: %v", service, err.Error()))
+					LoggingClient.Error("error starting service", "service", service)
+					return err
 				}
 			} else {
-				LoggingClient.Warn(fmt.Sprintf("starting not supported with specified executor"))
+				err := fmt.Errorf("operation not supported with specified executor")
+				LoggingClient.Error(err.Error(), "operation", action, "service", service)
```
tsconn23: No `operation` key please.

internal/system/agent/router.go
akramtexas: Agreed. Having an additional, leading `Error: ` is fine (The way you currently have it, works) 👍

akramtexas:
Recheck

brandonforster:
recheck

PR #1189, opened 2019-03-21
Create "floatEncoding" field

### LINKED ISSUE #1188: Create "floatEncoding" field to "PropertyValue" and "ValueDescriptor" struct in Mongo db models
According to https://github.com/edgexfoundry/device-sdk-go/issues/107 and https://github.com/edgexfoundry/go-mod-core-contracts/issues/22, we need additional field `floatEncoding` in `PropertyValue` and `ValueDescriptor`

> That said, we re-hashed this once again during today's Core WG meeting. The decision was that we'll add a new attribute (strawman name "floatEncoding" to the ValueProperty & ValueDescriptor, which if specified as "printfFloat" (other suggestions welcome, but this is the most accurate I could come up with), then floats will be encoded using the C printf style encoding (e.g. "3.14159265e+00"). If this new attribute isn't specified, then the default will be base64 as currently implemented.

cloudxxx8:
Create "floatEncoding" field to "PropertyValue" and "ValueDescriptor"
struct in Mongo db models
fix #1188

Signed-off-by: Cloud Tsai <cloudxxx8@gmail.com>

### DIFF
--- a/internal/pkg/db/mongo/models/value_descriptor.go
+++ b/internal/pkg/db/mongo/models/value_descriptor.go
@@ -20,20 +20,22 @@ import (
 )
 
 type ValueDescriptor struct {
-	Id           bson.ObjectId `bson:"_id,omitempty"`
-	Uuid         string        `bson:"uuid"`
-	Created      int64         `bson:"created"`
-	Description  string        `bson:"description,omitempty"`
-	Modified     int64         `bson:"modified"`
-	Origin       int64         `bson:"origin"`
-	Name         string        `bson:"name"`
-	Min          interface{}   `bson:"min,omitempty"`
-	Max          interface{}   `bson:"max,omitempty"`
-	DefaultValue interface{}   `bson:"defaultValue,omitempty"`
-	Type         string        `bson:"type,omitempty"`
-	UomLabel     string        `bson:"uomLabel,omitempty"`
-	Formatting   string        `bson:"formatting,omitempty"`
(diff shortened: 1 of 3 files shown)


cloudxxx8:
add @iain-anderson for awareness

cloudxxx8:
also added MediaType into ValueDescriptor

internal/pkg/db/mongo/models/value_descriptor.go
```diff
@@ -20,20 +20,22 @@ import (
...
+	Name          string        `bson:"name"`
+	Min           interface{}   `bson:"min,omitempty"`
+	Max           interface{}   `bson:"max,omitempty"`
+	DefaultValue  interface{}   `bson:"defaultValue,omitempty"`
+	Type          string        `bson:"type,omitempty"`
+	UomLabel      string        `bson:"uomLabel,omitempty"`
+	Formatting    string        `bson:"formatting,omitempty"`
+	Labels        []string      `bson:"labels,omitempty"`
+	MediaType     string        `bson:"mediaType"`
+	FloatEncoding string        `bson:"floatEncoding"`
```
tsconn23: @cloudxxx8 It looks like these need to have the `omitempty` attribute to conform with the other string types above.

cloudxxx8: @tsconn23 sorry, I missed them. Just added.  Thanks for the review

Issue #1194, opened 2019-03-22
Possible Duplicate Paths for core-data/readings by device and value descriptor


tsconn23:
While working on the GoDoc content for go-mod-core-contracts, I noticed that we have in the core-data ReadingClient what appears to be two functions that do the same thing. These are:

```
ReadingsForNameAndDevice(name string, deviceId string, limit int, ctx context.Context) ([]models.Reading, error)
```
and
```
func (r *readingRestClient) ReadingsForDeviceAndValueDescriptor(deviceId string, vd string, limit int, ctx context.Context) ([]models.Reading, error)
```
I include the receiver on the second function because it's not actually included in the ReadingClient interface right now. This would seem to indicate a preference in usage for the former. When I look at the implementation, while both achieve the same thing, the former's operation is more efficient.

`ReadingsForNameAndDevice` [uses](https://github.com/edgexfoundry/edgex-go/blob/3c1d3d5f30f66af437d5d945f68bacfa3c825c6e/internal/core/data/router.go#L72) `readingByValueDescriptorAndDeviceHandler` which ultimately [runs a query](https://github.com/edgexfoundry/edgex-go/blob/3c1d3d5f30f66af437d5d945f68bacfa3c825c6e/internal/core/data/reading.go#L262) at the database level to obtain its result.

There are two problems with `ReadingsForDeviceAndValueDescriptor`. The first is that its handler is actually defined on the `event` [segment](https://github.com/edgexfoundry/edgex-go/blob/3c1d3d5f30f66af437d5d945f68bacfa3c825c6e/internal/core/data/router.go#L58) of the core-data API. This means if someone were able to call the above method on the ReadingClient it wouldn't work anyway. Second, when reviewing the [handler](https://github.com/edgexfoundry/edgex-go/blob/3c1d3d5f30f66af437d5d945f68bacfa3c825c6e/internal/core/data/router.go#L612), it first gets all events for a given device and then performs [two loops](https://github.com/edgexfoundry/edgex-go/blob/3c1d3d5f30f66af437d5d945f68bacfa3c825c6e/internal/core/data/reading.go#L123). For each event, it loops through the readings. For each reading it loops through the value descriptors looking for a match on the name. If it finds one, it adds the reading to the list.

[...]

PR #1209, opened 2019-03-27
Removed duplicate path for readings by device and value descriptor

brandonforster:
Issue #1194


tsconn23:
recheck

Issue #1206, opened 2019-03-27
Update to latest go-mod-registry and fix imports


lenny-intel:
Once PR https://github.com/edgexfoundry/go-mod-registry/pull/6 has been merged,
`go-mod-registry` will have been refactored to conform with using `/pkg` and `/internal` appropriately. This will require some imports and namespace changes.

First update go.mod to pull in latest go-mode-registry
Then update any failing imports and namespaces.

brandonforster:
closed via #1211

PR #1211, opened 2019-03-28
Updates for latest go-mod-registry

lenny-intel:
The PR addresses issue #1206 
Update for import and namespace changes.

### DIFF
--- a/internal/seed/config/populate.go
+++ b/internal/seed/config/populate.go
@@ -24,8 +24,8 @@ import (
 	"strings"
 
 	"github.com/edgexfoundry/edgex-go/internal"
-	"github.com/edgexfoundry/go-mod-registry"
-	"github.com/edgexfoundry/go-mod-registry/pkg/factory"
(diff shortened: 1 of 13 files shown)


internal/seed/config/populate.go
```diff
@@ -18,14 +18,14 @@ import (
 	"bufio"
 	"errors"
 	"fmt"
+	"github.com/edgexfoundry/go-mod-registry/registry"
```
brandonforster: Can you move this import to be with the others?

lenny-intel: Yep. @tsconn23 do you know any setting in Goland for it not insert imports in the std lib section. Thx

tsconn23: I haven't played around with it, but there's a setting under `Tools --> Go Tools --> Goimports file` that could help. Also see https://godoc.org/golang.org/x/tools/cmd/goimports

tsconn23:
recheck

Issue #1214, opened 2019-03-29
Error while removing an Interval or an IntervalAction


jduranf:
I have those Interval Actions:
```
[
    {
        "id": "09efe7f9-03ee-48e3-bbc7-9201386cfaf4",
        "created": 1553874878745,
        "modified": 1553874878745,
        "name": "scrub-aged-events",
        "interval": "midnight",
        "target": "core-data",
        "protocol": "http",
        "httpMethod": "DELETE",
        "address": "localhost",
        "port": 48080,
        "path": "/api/v1/event/removeold/age/604800000"
    },
    {
        "id": "3f12b944-7cc9-476b-b684-b0aaac2732c5",
        "created": 1553874935579,
        "modified": 1553874935579,
        "name": "scrub-pushed-events",
        "interval": "midnight",
        "target": "core-data",
        "protocol": "http",
        "httpMethod": "DELETE",
        "address": "localhost",
        "port": 48080,
        "path": "/api/v1/event/scrub"
    }
]
```
And this interval:
```
[
    {
        "id": "21da99b3-49ee-4d3d-a67c-628428339142",
        "created": 1553874878734,
        "modified": 1553874878734,
        "name": "midnight",
        "start": "20180101T000000",
        "frequency": "P1D"
    }
]
```
I try to remove the interval by sending a DELETE in `http://localhost:48085/api/v1/intervalaction/09efe7f9-03ee-48e3-bbc7-9201386cfaf4`
I receive a `400 Bad Request` with `interval still in use by intervalAction(s) name:  midnight`
I try to remove again the interval, but now I receive `500 Internal Server Error` with `no interval found for id: 21da99b3-49ee-4d3d-a67c-628428339142`
At this point, if I try to remove an interval action, I have the same issue and is not possible to remove the interval action.

After restart the support-scheduler, the service works fine again and is possible to remove first the interval action and then the interval.

PR #1215, opened 2019-04-01
Fixed issue while trying to remove an Interval still in use

jduranf:
Fix issue #1214 


tsconn23:
recheck

PR #1218, opened 2019-04-02
PR for Issue 788 - Redis updated for UUID, contracts, and propagated across services

andresrinivasan:
This brings Redis to Core Data, Core MetaData, Export Client, and Scheduler. The notification service will be merged separately. There is no plan to support the Logger which defaults to using a file.

Please note changes to `metadata/rest_command.go` and `metadata/rest_deviceprofile.go`; there are inconsistencies with how `err == ErrNotFound` and `err == nil` are treated and this is my proposed workaround. At some point we should make a decision on what the right way is and propagate those changes.

There are a few Black Box tests that are failing from what I believe are minor bugs in the tests. I will need to revisit that.

### DIFF
--- a/bin/edgex-docker-launch.sh
+++ b/bin/edgex-docker-launch.sh
@@ -31,11 +31,8 @@ fi
 
 EDGEX_CORE_DB=${EDGEX_CORE_DB:-"mongo"}
 
-echo "Starting Mongo"
-docker-compose -f $COMPOSE_FILE up -d mongo
-
 if [ ${EDGEX_CORE_DB} != mongo ]; then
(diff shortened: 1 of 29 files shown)


tsconn23:
@andresrinivasan Thanks for the PR. When you have a chance, can you squash the commits please?

tsconn23:
@andresrinivasan I think this looks good enough to pull in so we can make additional tweaks. Testing via blackbox looks pretty good. Can you rebase and squash your commits please?

andresrinivasan:
Hmmm. That didn't work out the way I expected. Backing out.

andresrinivasan:
Much better.

tsconn23:
recheck

Issue #1223, opened 2019-04-03
Add core-command endpoint to call commands by name


mhall119:
The current core-command endpoint for calling a device currently requires knowing the generated IDs of both the target device and the command to be called. Since these are generated when the device is added, there's no way to know them ahead of time making it impossible to provide pre-configured solutions.

For example, in [this example RaspberryPi service](https://github.com/mhall119/edgex-device-rpi/tree/master/examples/MotionDetector#configure-node-red) additional steps are required to replace the Command URL in Node-Red, even though the device and command are known.

I propose adding a new endpoint that will perform a lookup by device and command name, rather than a generated ID, so the above use case could be:

http://edgex-core-command:48082/api/v1/device/name/RPiMotionDetector /command/Set_Green_Led

PR #1233, opened 2019-04-07
Issue #1223 Supports both GET/PUT, restructured to used same executio…

scsosna99:
Added matching RESTful call /device/api/v1/device/name/{device-name}/command/{command-name}, finding the corresponding device and command ids and passing them to the methods that operate off of IDs used by /device/api/v1/device/{device-id}/command/{command-id}.

To convert command name into appropriate ID for the device, used the device's profile and iterate through names looking for a match.  Unless the number of commands is substantial - and not sure how large that would be - seemed that overhead of creating the map would be more than just walking through the array.  Not opposed to making a map but didn't seem necessary (but remember, I'm new here).

Tested by following camera control example in docs https://docs.edgexfoundry.org/Ch-WalkthroughData.html and trying various combinations of commands or invalid commands.  Also updated .raml with new calls (basically cut-and-paste of existing ID calls with changes to reflect the use of names).

### DIFF
--- a/api/raml/core-command.raml
+++ b/api/raml/core-command.raml
@@ -169,7 +169,50 @@ schemas:
                 description: if no device exists by the name provided.
             "500": 
                 description: for unanticipated or unknown issues encountered.
-/device/name/{name}/opstate/{opState}: 
+/device/name/{name}/command/{command}:
+    displayName: Issue command
+    description: Example - http://localhost:48082/api/v1/device/name/RPIMotionDetector/command/Set_Green_Led
+    uriParameters:
+        name:
+            displayName: name
+            type: string
+            required: false
+            repeat: false
+        command:
+            displayName: commandname
+            type: string
+            required: false
+            repeat: false
+    get:
+        description: Issue the get command referenced by the command name to the device/sensor (also referenced by name) it is associated to via the device service. ServiceException (HTTP 500) for unanticipated or unknown issues encountered. Throws NotFoundException (HTTP 404) if no device exists by the name provided. Throws LockedException (HTTP 423) if the device is locked (admin state).
+        responses:
+            "200":
+                description: String as returned by the device/sensor via the device service.
+            "400":
+                description: if the request is malformed or unparsable
+            "404":
+                description: if no device exists by the name provided
+            "423":
+                description: if the device is locked (admin state).
+            "500":
+                description: for unanticipated or unknown issues encountered.
+    put:
+        description: Issue the get command referenced by the command name to the device/sensor (also referenced by name) it is associated to via the device service. ServiceException (HTTP 500) for unanticipated or unknown issues encountered. Throws NotFoundException (HTTP 404) if no device exists by the name provided. Throws LockedException (HTTP 423) if the device is locked (admin state).
+        body:
+            application/json:
+                example: '{"temp":72}'
+        responses:
+            "200":
+                description: String as returned by the device/sensor via the device service.
+            "400":
+                description: if the request is malformed or unparsable
+            "404":
+                description: if no device exists by the name provided
+            "423":
+                description: if the device is locked (admin state)
+            "500":
+                description: for unanticipated or unknown issues encountered.
+/device/name/{name}/opstate/{opState}:
     displayName: Op state by name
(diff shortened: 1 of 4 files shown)


tsconn23:
recheck

tsconn23:
recheck

api/raml/core-command.raml
```diff
@@ -169,7 +169,49 @@ schemas:
                 description: if no device exists by the name provided.
             "500": 
                 description: for unanticipated or unknown issues encountered.
-/device/name/{name}/opstate/{opState}: 
+/device/name{name}/command/{command}:
```
tsconn23: Missing '/'. Should be `/device/name/{name}/command/{command}`

internal/core/data/errors/types.go
```diff
@@ -150,3 +150,15 @@ func (e ErrInvalidId) Error() string {
 func NewErrInvalidId(id string) error {
 	return &ErrInvalidId{id: id}
 }
+
+type ErrInvalidName struct {
```
tsconn23: Since this new error type is used within core-command, it should be declared in the internal/core/command package. It's not a good practice to have imports straddle from one service to the other as found in restDevice.go
`import "github.com/edgexfoundry/edgex-go/internal/core/data/errors"`

internal/core/command/restDevice.go
```diff
@@ -29,12 +30,68 @@ func restPutDeviceCommandByCommandID(w http.ResponseWriter, r *http.Request) {
...
 func issueDeviceCommand(w http.ResponseWriter, r *http.Request, p bool) {
-	defer r.Body.Close()
 
 	vars := mux.Vars(r)
 	did := vars[ID]
 	cid := vars[COMMANDID]
+	issueDeviceCommandWork(w, r, did, cid, p)
+}
+
+func issueDeviceCommandWork(w http.ResponseWriter, r *http.Request, did string, cid string, p bool) {
```
tsconn23: It looks like this was created due to the variance in looking up devices/commands by name instead of ID. How else could you do this in a way that makes the intent obvious to later developers? `issueDeviceCommand` versus `issueDeviceCommandWork` doesn't really communicate a lot about why the logic is branched and at first I missed it.

internal/core/command/restDevice.go
```diff
@@ -15,6 +15,7 @@ package command
 
 import (
 	"encoding/json"
+	"github.com/edgexfoundry/edgex-go/internal/core/data/errors"
```
tsconn23: See my note below on this.

tsconn23:
Rebase the PR branch and that should clear the docs-verify failure above.

internal/core/command/restDevice.go
scsosna99: Following pattern used by REST methods by IDs, create two methods for REST methods by names, followed by the corresponding non-rest functions that extract the IDs or names, as appropriate.  Then the "work" method becomes plain-old issueDeviceCommand because that's all it does, no data grabbing or interpretation, get the IDs and act upon them.

scsosna99:
Cleaned up commit history after making a mess of things and should have consolidated changes, both original and PR comments, included in this one, if I didn't mess it up.

tsconn23:
recheck

tsconn23:
@scsosna99 Build is now failing due to a PR that was merged today, so you'll need to rebase again.
```
13:51:40 ?   	github.com/edgexfoundry/edgex-go	[no test files]
13:52:04 # github.com/edgexfoundry/edgex-go/internal/core/command
13:52:04 internal/core/command/restDevice.go:77:31: d.Profile.Commands undefined (type models.DeviceProfile has no field or method Commands)
```
On the other build failure you called out in the #devops channel, I believe that had to do with no ARM base image being available for a recent Go version upgrade that happened yesterday. When there's no obvious cause you can retrigger a build by typing 'recheck' in the comment.

scsosna99:
need to check field name change you made, will update my code this evening.

scsosna99:
recheck

tsconn23:
recheck

jamesrgregg:
recheck

tsconn23:
recheck

tsconn23:
Re-assigning to @difince for closure. A new PR will be opened.

PR #1236, opened 2019-04-08
Replace device.DeviceProfile with device.ProfileName

difince:
Remove the embedded DeviceProfile struct from the Device.
The Device struct will keep just the ProfileName as a reference.
The Device SDK have a cache of Device Profiles, so they will be retrieved from the cache.
All other services will make a call to metadata in order to get a Device Profile.

It decreases the Device SDK cache size, and descrease the debug and error log messages.

Resolves/Fixes edgexfoundry/go-mod-core-contracts#27

Signed-off-by: Diana Atanasova <dianaa@vmware.com>

### DIFF
--- a/internal/core/command/device.go
+++ b/internal/core/command/device.go
@@ -155,6 +155,7 @@ func putDeviceOpStateByName(dn string, as string, ctx context.Context) (int, err
 }
 
 func getCommands(ctx context.Context) (int, []models.CommandResponse, error) {
+	var cr []models.CommandResponse
 	devices, err := mdc.Devices(ctx)
 	if err != nil {
 		chk, ok := err.(*types.ErrServiceClient)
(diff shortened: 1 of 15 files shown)


tsconn23:
@difince It doesn't look like this is ready for review yet given the build errors. Changes in https://github.com/edgexfoundry/go-mod-core-contracts/pull/63 would need to be merged first before this PR is buildable anyway.

internal/core/command/device.go
```diff
@@ -164,9 +165,32 @@ func getCommands(ctx context.Context) (int, []models.CommandResponse, error) {
 			return http.StatusInternalServerError, nil, err
 		}
 	}
-	var cr []models.CommandResponse
+
+	if len(devices) == 0 {
+		return http.StatusOK, cr, err
+	}
+
+	dps, err := mdpc.DeviceProfiles(ctx)
```
tsconn23: Why do we need to load device profiles here when the command service itself knows what commands are available for a device?

`http://localhost:48082/api/v1/device/name/Random-Integer-Generator01`
`http://localhost:48082/api/v1/device/{UUID}`

internal/core/command/device.go
```diff
@@ -175,25 +199,33 @@ func getCommands(ctx context.Context) (int, []models.CommandResponse, error) {
...
-		chk, ok := err.(*types.ErrServiceClient)
-		if ok {
-			return chk.StatusCode, models.CommandResponse{}, chk
-		} else {
-			return http.StatusInternalServerError, models.CommandResponse{}, err
-		}
+		return returnCommandRespError(err)
 	}
-	return http.StatusOK, models.CommandResponseFromDevice(d, Configuration.Service.Url()), err
+	dp, err := mdpc.DeviceProfileForName(d.ProfileName, ctx)
```
tsconn23: Again, the core-command service should already know what commands a device has.

tsconn23:
Closing this until dependency with core-contracts is resolved

Issue #1231, opened 2019-04-05
Interlock -- Rename DeviceProfile Properties


tsconn23:
See https://github.com/edgexfoundry/go-mod-core-contracts/issues/59 for context.

References to the properties in the above issue need to be changed in edgex-go for integration.

PR #1238, opened 2019-04-08
Integrate DeviceProfile Resources / Commands Name Change

tsconn23:
#1231

### DIFF
--- a/internal/pkg/db/mongo/models/deviceprofile.go
+++ b/internal/pkg/db/mongo/models/deviceprofile.go
@@ -91,8 +91,8 @@ type DeviceProfile struct {
 	Model           string            `bson:"model"`
 	Labels          []string          `bson:"labels"`
 	DeviceResources []DeviceResource  `bson:"deviceResources"`
-	Resources       []ProfileResource `bson:"resources"`
-	Commands        []mgo.DBRef       `bson:"commands"`
+	DeviceCommands  []ProfileResource `bson:"resources"`
+	CoreCommands    []mgo.DBRef       `bson:"commands"`
 }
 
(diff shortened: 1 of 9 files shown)


internal/pkg/db/mongo/models/deviceprofile.go
```diff
@@ -91,8 +91,8 @@ type DeviceProfile struct {
 	Model           string            `bson:"model"`
 	Labels          []string          `bson:"labels"`
 	DeviceResources []DeviceResource  `bson:"deviceResources"`
-	Resources       []ProfileResource `bson:"resources"`
-	Commands        []mgo.DBRef       `bson:"commands"`
+	DeviceCommands  []ProfileResource `bson:"resources"`
+	CoreCommands    []mgo.DBRef       `bson:"commands"`
```
jduranf: Bson names shouldn't be changed?

internal/pkg/db/redis/device_profile.go
```diff
@@ -42,7 +42,7 @@ func marshalDeviceProfile(dp contract.DeviceProfile) (out []byte, err error) {
 		Model:           dp.Model,
 		Labels:          dp.Labels,
 		DeviceResources: dp.DeviceResources,
-		Resources:       dp.Resources,
+		DeviceCommands:       dp.DeviceCommands,
```
jduranf: Should be aligned

internal/pkg/db/mongo/models/deviceprofile.go
jduranf: There have been many important changes in DeviceProfile in this release cycle which force the Mongo database to be flushed, so I think should be more clear if all names are aligned.
But anyway, it's fine for me to leave as it is.

Issue #1219, opened 2019-04-02
Core-Command: Correlation-ID not propagated when issuing command


tsconn23:
The core-command service invokes commands on a remote addressable using a native http client.

https://github.com/edgexfoundry/edgex-go/blob/65d9a2a4033eb83af861f200106da1aaea6cff03/internal/core/command/device.go#L26

The correlation-id HTTP header is not being passed when these calls are made. The `issueCommand` function needs to have a Context passed into it from which a correlation-id can be obtained and set on the request. Callers to `issueCommand` need to be updated to provide the Context.

AnthonyMBonafide:
@tsconn23 , I am thinking about creating a unit test to cover this newly added functionality. Is this OK or is that getting too granular?

brandonforster:
@AnthonyMBonafide We need more tests, not less.

You're doing the best work.

tsconn23:
@AnthonyMBonafide Go for it

AnthonyMBonafide:
Opened PR #1240

PR #1240, opened 2019-04-09
#1219 Updated core-command to include a correlation ID in the request headers when sending HTTP(s) requests

AnthonyMBonafide:
Refactored device.go command execution functionality to allow for easier
testing

Added tests to cover newly added functionality

### DIFF
--- a/internal/core/command/put.go
+++ b/internal/core/command/put.go
@@ -1,0 +1,44 @@
+/*******************************************************************************
+ * Copyright 2019 Dell Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+package command
+
+import (
+	"context"
+	"github.com/edgexfoundry/edgex-go/internal"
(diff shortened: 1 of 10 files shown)


internal/core/command/put.go
```diff
@@ -0,0 +1,43 @@
...
+
+	correlationID := context.Value(clients.CorrelationHeader)
+	if correlationID != nil {
+		request.Header.Set(clients.CorrelationHeader, correlationID.(string))
+	}
+
+	return ServiceCommand{
+		HttpCaller: httpCaller,
+		Request:    request,
+	}, nil
```
tsconn23: Device property not set

internal/core/command/types.go
```diff
@@ -0,0 +1,63 @@
...
+
+// ErrDeviceLocked error indicating the locked state of a device
+type ErrDeviceLocked struct{}
+
+func (ErrDeviceLocked) Error() string {
+	return "The device is in admin locked state"
+}
+
+// ServiceCommand type which encapsulates command information to be sent to the command service.
+type ServiceCommand struct {
```
tsconn23: Recommend `serviceCommand` so that the type is not exported. In general, I find it enforces utilization of an interface if the implementing types are not exportable. Granted that has a more firm boundary in a more complex package structure, but good to get in the habit.

internal/core/command/put_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"context"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+	"testing"
+)
+
+func TestNewPutCommandWithCorrelationId(t *testing.T) {
+	expectedCorrelationIDHeaderValue := "Testing"
+	testContext := context.WithValue(context.Background(), clients.CorrelationHeader, expectedCorrelationIDHeaderValue)
+
+	getCommand, _ := NewGetCommand(testDevice, testCommand, testContext, nil)
```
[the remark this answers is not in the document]
AnthonyMBonafide: Great catch. I also had some bad test failure messages which I cleaned up.

Issue #862, opened 2018-11-28
Add Redis support for Notification service


andresrinivasan:

andresrinivasan:
Picking this up to merge the service implemented by IOTech.

PR #1247, opened 2019-04-16
PR for Issue 862 - Add Redis Support for Notification Service

andresrinivasan:
#862 

IOTech deserves all the kudos. I just merged.

Signed-off-by: Andre Srinivasan <andre@redislabs.com>

### DIFF
--- a/internal/pkg/db/redis/queries.go
+++ b/internal/pkg/db/redis/queries.go
@@ -15,6 +15,7 @@ package redis
 
 import (
 	"encoding/json"
+	"strconv"
 
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db/redis/models"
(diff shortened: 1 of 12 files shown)


internal/support/notifications/init.go
```diff
@@ -95,9 +94,6 @@ func Init(useRegistry bool) bool {
 		registerUpdates = make(chan interface{})
 		go listenForConfigChanges()
 	}
-
-	go telemetry.StartCpuUsageAverage()
```
anonymouse64: Why was this removed here?

andresrinivasan: My guess is the IOTech branch hadn't been rebased in a bit and I missed it in the merge. I will correct.

internal/pkg/db/redis/data.go
```diff
@@ -243,7 +243,7 @@ func (c *Client) ReadingsByDeviceAndValueDescriptor(deviceId, valueDescriptor st
 	objects, err := getObjectsByRangeFilter(conn,
 		db.ReadingsCollection+":device:"+deviceId,
 		db.ReadingsCollection+":name:"+valueDescriptor,
-		0, limit-1)
+		0, int64(limit-1))
```
tsconn23: Is this due to a concern that the `int` type may not supply a large enough value? There's no way we're going to send back 2.1 billion readings via a REST response, let alone more than that.

internal/pkg/db/redis/queries.go
```diff
@@ -128,7 +129,7 @@ func getObjectsByRange(conn redis.Conn, key string, start, end int) (objects [][
 // Return objects by a score from a zset
 // if limit is 0, all are returned
 // if end is negative, it is considered as positive infinity
-func getObjectsByRangeFilter(conn redis.Conn, key string, filter string, start, end int) (objects [][]byte, err error) {
+func getObjectsByRangeFilter(conn redis.Conn, key string, filter string, start, end int64) (objects [][]byte, err error) {
```
tsconn23: See comment above. I don't know that int64 buys us anything here.

internal/pkg/db/redis/data.go
andresrinivasan: I was reacting to changes that came out of the notification service support. I'll revisit and follow up.

internal/pkg/db/redis/queries.go
andresrinivasan: Ditto to the above.

internal/pkg/db/redis/data.go
andresrinivasan: Looks like it was propagation from the notification interface. I've corrected to an `int` implementaiton.

internal/pkg/db/redis/queries.go
andresrinivasan: Looks like it was propagation from the notification interface. I've corrected to an `int` implementaiton.

tsconn23:
recheck

Issue #1243, opened 2019-04-11
Correlation ID Logging Statements Should be at TRACE Level


tsconn23:
For now, the community has decided, in the interest of reducing the log content, that correlation ID logging should be done at the TRACE level. This is not at all ideal for a high availability cloud environment, but that is not EdgeX's primary deployment target at the moment. It is understood that with the default log level of INFO, these log statements will not be captured and so may not be available in the case of troubleshooting an error that has already occurred. One would have to modify the LogLevel setting in the configuration and then hope to capture a similar occurrence to see the service interaction accompanying the error.

This should be as simple as changing the calls to LoggingClient in the `OnResponseBegin` and `OnResponseComplete` functions [here](https://github.com/edgexfoundry/edgex-go/blob/master/internal/pkg/correlation/middleware.go).

brandonforster:
closed via #1250

PR #1250, opened 2019-04-16
Updated correlation ID logging to be at the TRACE level.

brandonforster:
Issue #1243 


cmd/config-seed/res/properties/go.mod
```diff
@@ -1 +1,3 @@
 module github.com/edgexfoundry/edgex-go/cmd/config-seed/res/properties
+
+go 1.12
```
anonymouse64: are we now building with go 1.12?

tsconn23: No. @brandonforster probably has it installed locally. This line should be removed.

tsconn23: No mods to this file should be included with this PR.

PR #1265, opened 2019-04-19
Pare down logging for the docker-compose reference implementation of sys-mgmt-executor.

akramtexas:
- This PR is to enable the Docker-in-Docker solution to to deploy and function properly. For this, we need to pare down logging for the docker-compose reference implementation of the system management executor (aka `sys-mgmt-executor`) because, while the facilities for logging are available to edgex-go services in general, `sys-mgmt-executor` is a standalone, bare-bones executable for which it does not make sense to initialize full-blown logging; instead, we are paring down to lean (built-in golang) logging 👍 
- As part of this, we have moved the field `ExecutorPath` to the top-level in `configuration.toml` (for sys-mgmt-agent docker) so it can be picked up by the runtime.
- The relevant Issue is: [Issue-1266](https://github.com/edgexfoundry/edgex-go/issues/1266)
- Crucially, and separately, the file `docker-compose.yml` (located in the `compose-files` folder in a separate repo, the `developer-scripts` repo) will need to be updated to start services, in particular the Docker-in-Docker SMA container in order to enable REST calls to the SMA to reach the Docker-in-Docker image container, since we need to have port-forwarding enabled, by incorporating a new directive (`/var/run/docker.sock:/var/run/docker.sock`), like so:
```
  system:
    image: edgexfoundry/sys-mgmt-agent-go:1.0.0-dev
    ports:
      - "48090:48090"
    container_name: edgex-sys-mgmt-agent
    hostname: edgex-sys-mgmt-agent
    networks:
      - edgex-network
    volumes:
      - db-data:/data/db
      - log-data:/edgex/logs
      - consul-config:/consul/config
      - consul-data:/consul/data
      - /var/run/docker.sock:/var/run/docker.sock
    depends_on:
      - logging
```

### DIFF
--- a/cmd/sys-mgmt-executor/main.go
+++ b/cmd/sys-mgmt-executor/main.go
@@ -14,7 +14,6 @@
 package main
 
 import (
-	"errors"
 	"flag"
 	"fmt"
 	"os"
@@ -22,39 +21,27 @@ import (
 	"strings"
 	"time"
 
-	"github.com/edgexfoundry/edgex-go/internal"
 	"github.com/edgexfoundry/edgex-go/internal/system/agent"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 )
 
 // Global variables
-var err error
-
+var LoggingClient logger.LoggingClient
 var usageStr = `
-Usage: ./main service operation		Start app with requested {service} and {operation}
+Usage: ./sys-mgmt-executor <service> <operation>		Start app with requested {service} and {operation}
        -h							Show this message
 `
 
 const (
-	START      = "start"
-	STOP       = "stop"
-	RESTART    = "restart"
-	AppOpenMsg = "This is the docker-compose-executor application!"
+	START                       = "start"
+	STOP                        = "stop"
+	RESTART                     = "restart"
+	SystemManagementExecutorKey = "docker-compose-executor"
+	AppOpenMsg                  = "This is the docker-compose-executor application!"
+	LoggingTarget               = "console"
+	EnableRemote                = false
+	LogLevel                    = "INFO"
 )
-
-// For use explicitly by this SMA executor, and given that Docker recognizes these
-// labels (e.g. "Notifications" and not "edgex-support-notifications", and likewise)
-// for other services, this simple remapping takes care of that wherby exec.Command
-// can work. Hence the following map definition.
-var services = map[string]string{
-	internal.SupportNotificationsServiceKey: "Notifications",
-	internal.CoreCommandServiceKey:          "Command",
-	internal.CoreDataServiceKey:             "CoreData",
-	internal.CoreMetaDataServiceKey:         "Metadata",
-	internal.ExportClientServiceKey:         "Export",
-	internal.ExportDistroServiceKey:         "Distro",
-	internal.SupportLoggingServiceKey:       "Logging",
-	internal.SupportSchedulerServiceKey:     "Scheduler",
-}
 
 // usage will print out the flag options for the app.
 // This function is based on usage.go (in internal / pkg / usage)
@@ -71,10 +58,13 @@ func main() {
 	flag.Usage = HelpCallback
 	flag.Parse()
 
-	agent.LoggingClient.Info(AppOpenMsg)
+	// Setup Logging
+	LoggingClient = logger.NewClient(SystemManagementExecutorKey, EnableRemote, LoggingTarget, LogLevel)
+
+	LoggingClient.Info(AppOpenMsg)
 
 	// Time it took to start service
-	agent.LoggingClient.Info("Application started in: " + time.Since(start).String())
+	LoggingClient.Info("Application started in: " + time.Since(start).String())
 
 	var service = ""
 	var operation = ""
@@ -86,9 +76,9 @@ func main() {
 		err := ExecuteDockerCommands(service, operation)
 
 		if err != nil {
-			agent.LoggingClient.Error(fmt.Sprintf("error performing  %s on service %s: %v", operation, service, err.Error()))
+			LoggingClient.Error(fmt.Sprintf("error performing  %s on service %s: %v", operation, service, err.Error()))
 		} else {
-			agent.LoggingClient.Info(fmt.Sprintf("success performing %s on service %s", operation, service))
+			LoggingClient.Info(fmt.Sprintf("success performing %s on service %s", operation, service))
 		}
 	}
 }
@@ -102,7 +92,7 @@ func findDockerContainerStatus(service string, status string) bool {
 	cmdName := "docker"
 	cmdArgs := []string{"ps"}
 	if cmdOut, err = exec.Command(cmdName, cmdArgs...).CombinedOutput(); err != nil {
-		agent.LoggingClient.Error(err.Error())
+		LoggingClient.Error(err.Error())
 		os.Exit(1)
 	}
 
@@ -114,18 +104,18 @@ func findDockerContainerStatus(service string, status string) bool {
 
 			if status == "Up" {
 				if strings.Contains(line, "Up") {
-					agent.LoggingClient.Info(fmt.Sprintf("container for service %s started: %s", service, line))
+					LoggingClient.Info(fmt.Sprintf("container for service %s started: %s", service, line))
 					return true
 				} else {
-					agent.LoggingClient.Warn(fmt.Sprintf("container for service %s NOT started", service))
+					LoggingClient.Warn(fmt.Sprintf("container for service %s NOT started", service))
 					return false
 				}
 			} else if status == "Exited" {
 				if strings.Contains(line, "Exited") {
-					agent.LoggingClient.Info(fmt.Sprintf("container for service %s stopped: %s", service, line))
+					LoggingClient.Info(fmt.Sprintf("container for service %s stopped: %s", service, line))
 					return true
 				} else {
-					agent.LoggingClient.Warn(fmt.Sprintf("container for service %s NOT stopped", service))
+					LoggingClient.Warn(fmt.Sprintf("container for service %s NOT stopped", service))
 					return false
 				}
 			}
@@ -135,20 +125,19 @@ func findDockerContainerStatus(service string, status string) bool {
 }
 
 func ExecuteDockerCommands(service string, operation string) error {
-	_, knownService := services[service]
 
-	if knownService {
-		err := runDockerCommands(service, services[service], operation)
-		agent.LoggingClient.Error(fmt.Sprintf("service %s ran into error while running Docker command: %v", service, err))
+	if agent.IsKnownServiceKey(service) {
+		err := runDockerCommands(service, operation)
+		LoggingClient.Error(fmt.Sprintf("service %s ran into error while running Docker command: %v", service, err))
 		return err
 	} else {
-		err := errors.New("is an unknown service for which request was made to run Docker command")
-		agent.LoggingClient.Error(fmt.Sprintf("the service %s: %v", service, err))
+		err := fmt.Errorf("the service %s is an unknown service for which request was made to run Docker command", service)
+		LoggingClient.Error(err.Error())
 		return err
 	}
 }
 
-func runDockerCommands(service string, dockerService string, operation string) error {
+func runDockerCommands(service string, operation string) error {
 
 	var (
 		err    error
@@ -157,7 +146,7 @@ func runDockerCommands(service string, dockerService string, operation string) e
 
 	cmdName := "docker"
 
-	cmdArgs := []string{operation, dockerService}
+	cmdArgs := []string{operation, service}
 
 	// Validate that a known operation was requested.
 	if operation == START || operation == STOP || operation == RESTART {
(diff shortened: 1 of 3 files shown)


tsconn23:
I guess I don't understand why the compile time dependencies to EdgeX have been removed. The executor currently lives in the `edgex-go` repo and so has easy access to everything in the repo as well as what it imports. What's the issue with the logging? Is isn't the content or the verbosity since that hasn't really changed. If it's the mechanism, and logging to a file is undesirable, you can initialize the LoggingClient with a `logTarget` value of `console` and the statements should just write to the console.

akramtexas:
@anonymouse64 and @tsconn23 (Gents, I was going to invite your review comments.) Now that I'm back at my desk, and seeing a comment each from you, I thank you. Let me respond right away 👍

akramtexas:
> While overall I am +1 on removing the dependency on edgex-go, I am -1 on this PR as it stands because this executable no longer performs any verification on the service name provided to it, which to me is a big security vulnerability (even if it would be difficult to exploit). The issue with not verifying it is an attacker could be allowed to start or stop _any_ docker container with this. That's a vulnerability as it potentially makes it much easier for an attacker to control an arbitrary container and thus get access to the docker socket and perform privilege escalation. Note that I don't think there is currently a tangible exploit here since the executable is only ever run from HTTP requests sent that are themselves checked to be valid all it would take is a bug in that validation code to get access here to run (close to) arbitrary docker commands.
> 
> Also I think there should be (even if there isn't currently) some mechanism for a service to get the names of all other services that doesn't require importing internal edgex-go packages?

@anonymouse64 Good feedback 👍 Yes, I too didn't see a tangible exploit here since the executable is strictly run from HTTP requests. And yes, _very_ much along the lines of your latter point (i.e. _Also I think there should be (even if there isn't currently) some mechanism for a service to get the names of all other services that doesn't require importing internal edgex-go packages?_), I had grappled with striking a balance: Having deliberated this more, I believe that we should start simple (reminding myself of our project dictum of "crawl, walk, and then run") and retain the validation of at least the core services. As a follow up (in Fuji), we can revisit this and explore some mechanism whereby we discover the names of all other services in a way that does _not_ require importing internal edgex-go packages (since the executor is intended to be standalone and lean.)

anonymouse64:
Re: removing logging dependency, I don't really care one way or the other if this package uses the edgex logger from edgex-go or not. I commented in the associated issue there may be an alternative way to solve the panic that @akramtexas is seeing sometimes 

> since the executor is intended to be standalone and lean.)

I guess I also don't see how including edgex-go makes this not lean, since it's in edgex-go? Also, it isn't really standalone since it only is meant to be run from inside the SMA docker container

akramtexas:
> I guess I don't understand why the compile time dependencies to EdgeX have been removed. The executor currently lives in the `edgex-go` repo and so has easy access to everything in the repo as well as what it imports. What's the issue with the logging? Is isn't the content or the verbosity since that hasn't really changed. If it's the mechanism, and logging to a file is undesirable, you can initialize the LoggingClient with a `logTarget` value of `console` and the statements should just write to the console.

@tsconn23  Good feedback as well 👍 I had meant to circle back with you and check whether my roundup of the rationale, as captured in the associated issue, i.e. [Issue-1266](https://github.com/edgexfoundry/edgex-go/issues/1266) was convincing... Allow me to elaborate: 
- My motivation here (in replacing _LoggingClient_ and associated initialization machinery, and therefore removing compile-time dependencies on EdgeX) is purely pragmatic. 
- As you'll recall, when starting out work on the executor, I had used golang's built-in (barebones) logging as a first approximation, driven pointedly by how we were considering for the executor to live in its _own_ repo (i.e. outside of `edgex-go`).
- But once we subsequently decided as a community that the `edgex-go` repo would be the home for the executor, I had taken the feedback at that time that (given how the executor has access to the facilities for logging that are available to `edgex-go` services in general) the executor should _also_ make use of _LoggingClient_ and associated initialization machinery.
- The issue (and my use of and growing familiarity with Portainer has been invaluable in allowing me to make headway in this area) is a purely pragmatic one, one that I uncovered and investigated recently: Specifically, when an attempt is made to use _LoggingClient_-based logging, the (un-initialized) instances of _LoggingClient_ naturally lead to golang runtime panic. Some details captured in the associated issue, i.e. [Issue-1266](https://github.com/edgexfoundry/edgex-go/issues/1266) 
- But if the initialization machinery (to enable the use of _LoggingClient_-based logging) isn't onerous on the simplicity aspects of the executor (i.e. doesn't make it complex), then I'm willing to bring it back in (Exploring that at the moment, as also keeping in mind your thoughtful suggestion regarding how, if logging to a file is undesirable, I could initialize the _LoggingClient_ with a `logTarget` value of console and the statements should just write to the console.)

akramtexas:
> Re: removing logging dependency, I don't really care one way or the other if this package uses the edgex logger from edgex-go or not. I commented in the associated issue there may be an alternative way to solve the panic that @akramtexas is seeing sometimes
> 
> > since the executor is intended to be standalone and lean.)
> 
> I guess I also don't see how including edgex-go makes this not lean, since it's in edgex-go? Also, it isn't really standalone since it only is meant to be run from inside the SMA docker container

@anonymouse64 What I had in mind was this: Will the initialization machinery (to enable the use of _LoggingClient_-based logging) be onerous on the _keep-it-simple_ aspects of the executor (i.e. I want to avoid making the executor code elaborate and complex). In this context, I've made a note of (and replied to) the comment by @tsconn23 that I'm willing to bring back the use of _LoggingClient_-based logging (and exploring that at the moment) 👍

cmd/sys-mgmt-executor/main.go
```diff
@@ -86,13 +78,17 @@ func main() {
...
+			LoggingClient.Error(fmt.Sprintf("error performing  %s on service %s: %v", operation, service, err.Error()))
 		} else {
-			agent.LoggingClient.Info(fmt.Sprintf("success performing %s on service %s", operation, service))
+			LoggingClient.Info(fmt.Sprintf("success performing %s on service %s", operation, service))
 		}
 	}
 }
 
+func setLoggingTarget() string {
+	return LoggingTarget
```
anonymouse64: since this is just a constant is there any possibility that this will become configurable at some point? If not, you should just use the `LoggingTarget` constant in the call to `logger.NewClient` instead

cmd/sys-mgmt-executor/main.go
```diff
@@ -135,20 +131,19 @@ func findDockerContainerStatus(service string, status string) bool {
...
-		err := runDockerCommands(service, services[service], operation)
-		agent.LoggingClient.Error(fmt.Sprintf("service %s ran into error while running Docker command: %v", service, err))
+	if agent.IsKnownServiceKey(service) {
+		err := runDockerCommands(service, operation)
+		LoggingClient.Error(fmt.Sprintf("service %s ran into error while running Docker command: %v", service, err))
 		return err
 	} else {
 		err := errors.New("is an unknown service for which request was made to run Docker command")
-		agent.LoggingClient.Error(fmt.Sprintf("the service %s: %v", service, err))
+		LoggingClient.Error(fmt.Sprintf("the service %s: %v", service, err))
```
anonymouse64: this flow seems off to me, I think it should instead look like:

```go
err := fmt.Errorf("the service %s is an unknown service for which request was made to run Docker command",service)
return err
```

cmd/sys-mgmt-executor/main.go
```diff
@@ -17,45 +17,33 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
```
anonymouse64: please run `gofmt` on this file so that the github imports are separate from the std library imports and are alphabetized. `gofmt` will do this for you

cmd/sys-mgmt-executor/main.go
```diff
@@ -167,31 +162,31 @@ func runDockerCommands(service string, dockerService string, operation string) e
...
 
 		default:
-			agent.LoggingClient.Warn("unknown operation %s was requested", operation)
+			LoggingClient.Warn("unknown operation %s was requested", operation)
 			break
 		}
 	} else {
 		err := errors.New("an unknown operation")
-		agent.LoggingClient.Error(fmt.Sprintf("system management was requested to perform %v %s on the %s service", err.Error(), operation, service))
+		LoggingClient.Error(fmt.Sprintf("system management was requested to perform %v %s on the %s service", err.Error(), operation, service))
```
anonymouse64: same comment as above here, I think it would be better to use `fmt.Errorf` to create the full `error` object that contains the service name, then log that with `err.Error()` using `LoggingClient.Error` and then finally return the error.

akramtexas:
@tsconn23 @anonymouse64 
- I have implemented and tested (using Dockerized containers) the initialization of the LoggingClient with a logTarget value of console and thereby bring back the use of LoggingClient-based logging. 
- The initialization machinery (to enable the use of LoggingClient-based logging) is simple enough, and isn't onerous on the simplicity aspects of the executor (i.e. doesn't make it complex).
- I did _also_ explore (and experiment with) making that initialization machinery configurable (through the use of configuration.toml) the way it is for full-blown services. But decided against incorporating that: The initialization machinery becomes onerous, and introduces a lag into the execution speed. Hypothetically speaking, should there be an opinion on this particular aspect, I might be _persuaded_ about making that initialization machinery configurable, but I'm not _convinced_ that would be worthwhile or advisable.
- Gents, great feedback 👍

akramtexas:
`Example responses via a REST client (Postman, in this case) when testing (using deployment to Dockerized containers) as follows:
`
- Done. Stopped the requested services.
- Done. Started the requested services.
- Done. Restarted the requested services.

`Example logging when testing (using deployment to Dockerized containers) as follows:
`
> 
level=INFO ts=2019-04-22T14:55:57.2730103Z app=edgex-sys-mgmt-agent source=services.go:40 msg="invoking operation",
level=INFO ts=2019-04-22T14:55:57.3810284Z app=edgex-sys-mgmt-agent source=execution_engine.go:20 msg="invocation of executor succeeded",
level=INFO ts=2019-04-22T14:55:59.6691395Z app=edgex-sys-mgmt-agent source=services.go:40 msg="invoking operation",
level=INFO ts=2019-04-22T14:55:59.7968746Z app=edgex-sys-mgmt-agent source=execution_engine.go:20 msg="invocation of executor succeeded",
level=INFO ts=2019-04-22T14:55:59.7990705Z app=edgex-sys-mgmt-agent source=services.go:40 msg="invoking operation",
level=INFO ts=2019-04-22T14:55:59.9215242Z app=edgex-sys-mgmt-agent source=execution_engine.go:20 msg="invocation of executor succeeded",
level=INFO ts=2019-04-22T14:56:07.0028584Z app=edgex-sys-mgmt-agent source=services.go:40 msg="invoking operation",
level=INFO ts=2019-04-22T14:56:07.8329342Z app=edgex-sys-mgmt-agent source=execution_engine.go:20 msg="invocation of executor succeeded",
level=INFO ts=2019-04-22T14:56:07.8352618Z app=edgex-sys-mgmt-agent source=services.go:40 msg="invoking operation",
level=INFO ts=2019-04-22T14:56:08.6270739Z app=edgex-sys-mgmt-agent source=execution_engine.go:20 msg="invocation of executor succeeded",
level=INFO ts=2019-04-22T14:56:59.1596093Z app=edgex-sys-mgmt-agent source=services.go:40 msg="invoking operation",
level=INFO ts=2019-04-22T14:57:12.3241514Z app=edgex-sys-mgmt-agent source=execution_engine.go:20 msg="invocation of executor succeeded",
level=INFO ts=2019-04-22T14:57:12.3265597Z app=edgex-sys-mgmt-agent source=services.go:40 msg="invoking operation",
level=INFO ts=2019-04-22T14:57:25.3365935Z app=edgex-sys-mgmt-agent source=execution_engine.go:20 msg="invocation of executor succeeded",

akramtexas:
> Thanks for addressing my comments re: checking the service. Code looks good, few minor things that would be nice to have changed

@anonymouse64 👍 I will shortly address those minor points as well ✅

anonymouse64:
> I did also explore (and experiment with) making that initialization machinery configurable (through the use of configuration.toml) the way it is for full-blown services. But decided against incorporating that: The initialization machinery becomes onerous, and introduces a lag into the execution speed. Hypothetically speaking, should there be an opinion on this particular aspect, I might be persuaded about making that initialization machinery configurable, but I'm not convinced that would be worthwhile or advisable.

I don't think it's worthwhile or necessary for that level of complexity in this executor, so just go with logging to console for now.

akramtexas:
> > I did also explore (and experiment with) making that initialization machinery configurable (through the use of configuration.toml) the way it is for full-blown services. But decided against incorporating that: The initialization machinery becomes onerous, and introduces a lag into the execution speed. Hypothetically speaking, should there be an opinion on this particular aspect, I might be persuaded about making that initialization machinery configurable, but I'm not convinced that would be worthwhile or advisable.
> 
> I don't think it's worthwhile or necessary for that level of complexity in this executor, so just go with logging to console for now.

- OK good 👍  @anonymouse64 And I believe this aligns with the comment by @tsconn23 earlier to log to `console` to keep things lean 💯

cmd/sys-mgmt-executor/main.go
akramtexas: Good point: Done ✅ (That was a carry-over from my exploring (and experimenting with), earlier on, how to make the initialization machinery configurable (and subsequently deciding against doing so)).

akramtexas:
All feedback comments have been addressed. A final review and approval is requested @anonymouse64 @tsconn23 👍

PR #1280, opened 2019-04-23
Query Notification before marking PROCESSED

### LINKED ISSUE #1279: Created field of critical Notification is always 0
reproduce step:

1. create a critical notification in Support Notifications Service
2. get the notification by Id or check the database. and the created field is always 0

Expected:  created field is the timestamp of the creation time
Actual: created field is 0

cloudxxx8:
Created field of critical Notification is always 0,
because the field is updated to 0 when marking its status to PROCESSED
fix #1279 

Signed-off-by: Cloud Tsai <cloudxxx8@gmail.com>

### DIFF
--- a/internal/support/notifications/rest_notification.go
+++ b/internal/support/notifications/rest_notification.go
@@ -53,6 +53,13 @@ func notificationHandler(w http.ResponseWriter, r *http.Request) {
 
 		if n.Severity == models.NotificationsSeverity(models.Critical) {
 			LoggingClient.Info("Critical severity scheduler is triggered for: " + n.Slug)
+			n, err = dbClient.GetNotificationById(n.ID)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				LoggingClient.Error(err.Error())
+				return
+			}
+
 			err := distributeAndMark(n)
 			if err != nil {
 				return


cloudxxx8:
@akramtexas , please help review this PR.  Thanks.

internal/support/notifications/rest_notification.go
```diff
@@ -53,6 +53,13 @@ func notificationHandler(w http.ResponseWriter, r *http.Request) {
 
 		if n.Severity == models.NotificationsSeverity(models.Critical) {
 			LoggingClient.Info("Critical severity scheduler is triggered for: " + n.Slug)
+			n, err = dbClient.GetNotificationById(n.ID)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
```
akramtexas: - I was debating over the choice made here for indicating which HTTP error it was, but agree that this is the best choice possible: `http.StatusInternalServerError` 👍

akramtexas:
@cloudxxx8 LGTM 👍 Please rebase and I'm ready to approve.

internal/support/notifications/rest_notification.go
cloudxxx8: In this line, since the record has been persisted successfully, it would be some unknown error.  That's why I choose StatusInternalServerError

cloudxxx8:
@akramtexas I've rebased.  Thanks for the review.

internal/support/notifications/rest_notification.go
akramtexas: Yep, I'm good with that 👍

Issue #1256, opened 2019-04-17
Interlock -- Integrate Model Validation From Core-Contracts


tsconn23:
This issue is to track any mods to the edgex-go services that may be necessary to integrate core-contract modifications for internal validation per https://github.com/edgexfoundry/go-mod-core-contracts/issues/66.

I expect these will be minor.

tsconn23:
Welp, not gonna be minor. Some of these export tests...hooo, doggies....

PR #1284, opened 2019-04-24
[Interlock core-contracts] Integrate model validation, fix tests

tsconn23:
#1256

- The encapsulation of validation inside the contracts/models required
some unit tests to be reworked. In particular, unit tests that
previously used literal JSON strings as the basis from which to un-
marshal have been replaced with an actual test instance of the model.
This means that if during the course of a test a model is marshaled to
bytes and then unmarshaled, the internal model validation logic will
fire. This was particularly prevalent in the way the export tests are
implemented.
- The LogLevels constants were moved in core-contracts to the models
package and out of the clients package. Had to address that.
- Reworked the relationship of Get/Put properties on Commands to no
longer be pointers. Checking for nil is a poor way to see if a given
instance is empty, plus its error prone in cases where the check isn't
even done.
- Changes related to the removal of the contracts/models/Service struct.
This struct was only utilized by DeviceService and contained most of
that type's properties. It was removed due to lack of re-use.
- The above cleaned up property references on the Device (e.g. We no
longer have to reference something like device.Service.Service.Id)
- Controller validation that previously called anonymous functions to
validate OperatingState and AdminState values have been removed since
those checks are now part of the models themselves.


internal/export/distro/format_test.go
```diff
@@ -38,7 +38,8 @@ func TestJson(t *testing.T) {
 	if err := json.Unmarshal(out, &eventOut); err != nil {
 		t.Fatalf("Error unmarshalling event: %v", err)
 	}
-	if !reflect.DeepEqual(eventIn, eventOut) {
+	//if !reflect.DeepEqual(eventIn, eventOut) {
```
jduranf: Why this commented code?

internal/pkg/db/mongo/models/deviceservice.go
```diff
@@ -25,21 +27,82 @@ type deviceServiceTransform interface {
...
+	Description    string                  `bson:"description"`
+	Id             bson.ObjectId           `bson:"_id,omitempty"`
+	Uuid           string                  `bson:"uuid,omitempty"`
+	Name           string                  `bson:"name"`           // time in milliseconds that the device last provided any feedback or responded to any request
+	LastConnected  int64                   `bson:"lastConnected"`  // time in milliseconds that the device last reported data to the core
+	LastReported   int64                   `bson:"lastReported"`   // operational state - either enabled or disabled
+	OperatingState contract.OperatingState `bson:"operatingState"` // operational state - ether enabled or disableddc
+	Labels         []string                `bson:"labels"`         // tags or other labels applied to the device service for search or other identification needs
+	Addressable    mgo.DBRef               `bson:"addressable"`    // address (MQTT topic, HTTP address, serial bus, etc.) for reaching the service
 	AdminState contract.AdminState `bson:"adminState"` // Device Service Admin State
```
jduranf: This line is not aligned with others

internal/core/command/get_test.go
```diff
@@ -32,24 +32,23 @@ var testDevice = contract.Device{
 	Id:         TestDeviceId,
 	AdminState: contract.Unlocked,
 	Service: contract.DeviceService{
-		Service: contract.Service{
-			Addressable: contract.Addressable{
+			Addressable: contract.Addressable {
```
jduranf: One tabulation should be removed

internal/export/distro/format_test.go
tsconn23: I'll remove the comment. I replaced it with the comparison on the line underneath.
`if eventIn.String() != eventOut.String()`
For more information, see the second bullet point in the following comment https://github.com/edgexfoundry/go-mod-core-contracts/issues/66#issuecomment-482732387

tsconn23:
I have spoken with @cloudxxx8 here at the Seoul face to face. He's aware of the fact that this will break the blackbox test results via the automated execution. @weichou1229 is working on https://github.com/edgexfoundry/blackbox-testing/issues/205 in order to resolve the failures.

PR #1287, opened 2019-04-25
Fixed HTTP Response Codes

AnthonyMBonafide:
This commit resolves issue 1285

Updated `restPutDeviceStateByDeviceId` and
`restPutDeviceStateByDeviceName` functions to propegate the underlying
HTTP response code to the caller. Also, updated the HTTP response from
"Internal Server Error(500)" to "Bad Request(400)" when unable to decode
the request body.


tsconn23:
recheck

PR #1290, opened 2019-04-27
NOISSUE - Simplify Makefile and scripts

drasko:

This patch:
- Simplifies `Makefile` by introducing functions to replace repetitive commands
- Moves `./bin` to `./scripts` - because these are scripts, and renames them to `run.sh`
- Simplifies `run.sh` script
- ~Adds `package.sh` script which collects binaries and configs in a single dir, suitable for their transfer to a gateway~

This patch also adds appropriate variables in the `Makefile` needed for ARM cross-compilation.


drasko:
This might break something, so if someone could pull the branch and test, it would be great.

BTW, I have a successfull ARM cross-compile now on my PC (very fast). Tricky part was ZMQ lib, which also must be cross-compiled. I will document this as soon as I catch some time.

drasko:
Ah, I saw one potential miss, I re-atttached WIP lablel. You can review, but plase hold the merge until label is removed.

drasko:
WIP label removed, should be OK now.

scripts/package.sh
```diff
@@ -0,0 +1,33 @@
...
+# Copyright (c) 2019
+# Mainflux
+#
+# SPDX-License-Identifier: Apache-2.0
+
+
+# Collects EdgeX Go binaries (must be previously built)
+# and packages them into a dir alongside with necessary config files.
+# This facililtets transfer of binaries to otehr machine
+# (for example copy of cross-compiled binaris to ARM gateway)
```
anonymouse64: `binaris` -> `binaries`

scripts/package.sh
```diff
@@ -0,0 +1,33 @@
...
+#
+# Copyright (c) 2019
+# Mainflux
+#
+# SPDX-License-Identifier: Apache-2.0
+
+
+# Collects EdgeX Go binaries (must be previously built)
+# and packages them into a dir alongside with necessary config files.
+# This facililtets transfer of binaries to otehr machine
```
anonymouse64: `facililtets` -> `facilitates`

scripts/package.sh
```diff
@@ -0,0 +1,33 @@
...
+  core-data export-distro)
+
+PKG_DIR=${1:-edgex}
+
+# Copy binaries and config
+mkdir -p $PKG_DIR
+for i in "${SERVICES[@]}"; do
+  mkdir -p $PKG_DIR/$i
+  cp $CMD_DIR/$i/$i $PKG_DIR/$i
+  cp $CMD_DIR/$i/res/configuration.toml $PKG_DIR/$i 2>/dev/null || :
```
anonymouse64: won't this copy the configuration.toml files inside `$PKG_DIR/core-data/configuration.toml` instead of the standard `$PKG_DIR/core-data/res/configuration.toml`? I think this should probably be consistent with our other distribution layouts and go inside `res`

anonymouse64: this also doesn't copy over all of the properties files for config-seed for example

Makefile
```diff
@@ -1,155 +1,80 @@
...
+
+define clean_services
+	for svc in $(SERVICES); do \
+		rm -rf cmd/$$svc/$$svc; \
+	done
+endef
+
+define make_docker
+	docker build \
+		--no-cache \
```
tsconn23: This line needs to be removed. It is eliminating the current caching we use between assembling Docker images which greatly improves build times in the CI/CD pipeline.

scripts/package.sh
drasko: Yes, my idea of the "package" was something simple that I can easily copy on a gateway - for example put it on a USB stick, or send it via `scp`.

I do not see the benefit of having `./res` here - I would rather see a config file right next to the binary then have to go one level more in the `res` directory.

Moreover, these `res` directories today contain `docker` dir, which is completely not needed when we have these native binaries - so we would be creating `res` directory only to put one `config.toml` in it. I find this unnecessary complexity and overengineering.

drasko: I do not know well this `config-seed`, it is not clear completely to me how these https://github.com/edgexfoundry/edgex-go/tree/master/cmd/config-seed/res/properties are loaded. I see some Java configs. Additionally, I think this is Consul related - do we really need Consul in the single-instance on-gateway deployment scenario, where there is no service discovery and all the configs can be set manually for a given binary? I would say no...

anonymouse64: Yes, config-seed is used with consul, though in the future it will probably support other registries but today the only implementation of the registry abstraction is Consul. 

I don't think we should be making assumptions about how folks will deploy EdgeX even if they are running the binaries locally. If this was simply a packaging script used for your own purposes in a different repo sure that's fine to leave out `config-seed`, but since this is in upstream edgex-go and could have many possible deployment scenarios we should include all of what is "default" EdgeX even if a particular deployment may not use it (and they can always delete `config-seed` before doing a deployment - and note that this script doesn't even produce a deployment archive it just produces a directory with the files).

I'm more concerned about leaving out default functionality than I am with how many files are copied into a directory when running `scripts/package.sh`. If you feel strongly about needing to leave out config-seed, I'd instead like to see it implemented as an option to the script. For example, maybe some deployments don't need `support-notifications`, so having a generic option to leave out certain services seems more useful than just assuming that one service won't be used.

anonymouse64: FWIW, I agree with you and I don't understand why we have all these `res` directories, perhaps @tsconn23 can comment on why that is.
I do still think that this packaging script should follow the pattern of edgex-go itself, so perhaps the thing to do here is to also remove all of the `res` directories in the edgex-go.

Lastly, we follow the same pattern in the snap, simply because that's what edgex-go does and so we try to follow that.

Makefile
```diff
@@ -1,66 +1,66 @@
 #
```
anonymouse64: Also, I think that the packaging script here should also be a target from the Makefile so you can just run `make package` and it builds a release archive of the binaries (so basically just run the script you have and then zip up the directory)

drasko:
@anonymouse64 I removed packagins script from this PR. I think it will be simpler to clean this PR and make it merged, then add packaging in subsequent PR.

This PR is not only important because it simplifies Makefile, but also because it adds an easy way to do cross-compilation for ARM and other architectures (I tested on ARM and MIPS).

drasko:
@tsconn23 @jamesrgregg any info about CI and how it can be changed to be aligned with this Makefile?

I used `make` instead of `make build`, and same for docker - I used `make dockers` to build all dockers, or make `docker_<service>` to build a docker for a given service. I think that `build` word is redundant, usually whan you say `make` you actually want it to build your program.

anonymouse64:
@drasko the jenkins job definition will need to be changed here: https://github.com/edgexfoundry/ci-management/blob/master/jjb/edgex-go/edgex-go.yaml#L12

scripts/package.sh
tsconn23: The `./res` directories have been with the project longer than I have. Their purpose has always been to hold configuration information for the default / docker profiles as well as any legacy java *.properties files (see config-seed/res/properties). If you would like to change the way configuration files are organized and loaded, file an issue.

drasko: I will open an issue once I finish my tests, but basically idea is to pick up only non-docker config in some easy way when creating a binary package.

anonymouse64:
@drasko any interest in refreshing this PR or should we close it out?

tsconn23:
Closing due to lack of activity

Issue #1292, opened 2019-04-29
Interlock -- Consume Latest Core-Contracts for Unmarshal Bug Fix


tsconn23:
See original issue here
https://github.com/edgexfoundry/go-mod-core-contracts/issues/80

There's a bug in the unmarshaling of event readings whereby the value is being dropped. Once the relevant PR is merged for the above bug fix, it should be consumed ASAP in edgex-go.

brandonforster:
closed via #1293

PR #1293, opened 2019-04-29
Bug Fix -- Reading Value Mis-Assigned During Unmarshal

tsconn23:
#1292 

Consumption of new core-contracts to pick up this bug fix.


tsconn23:
recheck

Issue #1210, opened 2019-03-28
Set up Go Fmt


tsconn23:
Add a go fmt ./... step to the make test target in the makefile. Run it, and then check in the makefile addition with the results.

jduranf:
Related with #501

brandonforster:
closed via #1298

PR #1298, opened 2019-04-30
Set up make test to use gofmt and run it on codebase

brandonforster:
Issue #1210

This will also take care of #501 in a very simple, basic way. Golinter might still be an integration we want to add, however.


brandonforster:
recheck

Issue #1271, opened 2019-04-19
Add ProfileName to Device Persistence Models


tsconn23:
In the interest of removing the [linkage](https://github.com/edgexfoundry/edgex-go/blob/713baa536c1563875d9ec61d0096859740ca004d/internal/pkg/db/mongo/models/device.go#L49) from a Device to its DeviceProfile, we should add a property to the Device stored in the database. The value of this property should contain the device profile name.
```
ProfileName string
```

This will be used later in order to find devices associated with a given device profile when that profile is either updated or is requested to be deleted.

Mongo currently has a full compliment of [models](https://github.com/edgexfoundry/edgex-go/tree/master/internal/pkg/db/mongo/models) that support variance at the persistence layer but Redis does not. There IS a [package](https://github.com/edgexfoundry/edgex-go/tree/master/internal/pkg/db/redis/models) into which Redis models can be placed however. Once the Redis Device model is created, utilize a similar pattern to that found in Mongo whereby it is mapped to/from the contract representation.

** NOTE ** Adding the Redis Device model -- The simplest approach for now would be to simply create a Go embedded type that provides the extra property above while embedding the contract.Device within it. The new model should support JSON marshaling / unmarshaling and can participate in the existing logic dealing with Redis-specific operations. If you need help with questions related to Redis itself, ping @andresrinivasan

PR #1309, opened 2019-05-03
Add ProfileName to Device Persistence Models

difince:
Issue: https://github.com/edgexfoundry/edgex-go/issues/1271


difince:
@tsconn23 How do you test it ?  I see it persisted with value

Issue #1306, opened 2019-05-03
Services should support stopping gracefully with SIGTERM


anonymouse64:
SIGTERM is a standard Unix signal that is sent to daemons to tell them to stop gracefully, i.e. when the system is shutting down, or an upgrade is about to be performed, etc. Currently these signals are entirely ignored by all EdgeX services, it would be nice if they could be handled the same way as `os.Interrupt` is currently. 

Note that windows has no such concept so it would be Linux and macOS specific behavior.

anonymouse64:
Interestingly enough the services only ignore the SIGTERM when in the snap, if I just run the services natively not as part of a systemd service then they handle SIGTERM properly.

brandonforster:
This might be a regression introduced with my work on #995.

michaelestrin:
@anonymouse64 is this still an issue (and needs to be worked) or has it been resolved by changes merged to master since this issue was created?

anonymouse64:
I need to re-evaluate this, unclear if it now works or not

michaelestrin:
Ask is to test using one of the core services that has been updated to use the new common bootstrap package (i.e. core-data, core-metadata, system-management-agent, etc.).

michaelestrin:
> I need to re-evaluate this, unclear if it now works or not

@anonymouse64 when will you have an opportunity to test this and determine if this issue is still valid?

anonymouse64:
Probably not anytime soon. You can close this if you can't reproduce

michaelestrin:
Closing without testing.  If this is still a problem that causes someone heartache, I expect they will reopen this issue.

PR #1297, opened 2019-04-29
Update snap to core18 base, remove snap build scripts

anonymouse64:
This PR mainly switches from not defining a base (and thus using "legacy" snapcraft) to defining a base and thus using the newer version of snapcraft (3.X) directly. Additionally, this fully switches from having the build scripts defined here to using the build definitions defined in the jenkins job instead.

TODO: 
- [x] ci-management PR (https://github.com/edgexfoundry/ci-management/pull/357) merged
- [x] deploy verify job for edgex-go to jenkins sandbox to test this PR - see https://jenkins.edgexfoundry.org/sandbox/view/All/job/edgex-go-snap-master-verify-snap
- [x] ~~open back-port PR to delhi~~ - actually unnecessary as we are able to specify to use job definitions per build stream, so delhi continues to use the old `build_script`
- [x] test upgrading from core16 implicit base snap (current snap) to this core18 snap - yes works no snapd bugs AFAICT here

This PR also:
* Fixes #1204 
* Fixes #1103 by using the archive download links
* Adds a work-around for #1306 
* Limits RAM usage by cassandra to something more reasonable, on a 16 GB amd64 machine:

before (approx 770 MB in non-swappable memory):
```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root     23764  4.7  2.9 2440648 474816 ?      SLl  17:51   3:11 /snap/edgexfoundry/x1/usr/lib/jvm/java-1.8.0-openjdk-amd64/bin/java -Dcassandra.config=file:///var/snap/edgexfoundry/x1/cassandra/etc/cassandra.ya
```
after (approx 60 MB in non-swappable memory):
```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root      3547 60.0 27.3 6231000 4443968 ?     SLl  19:02   1:13 /snap/edgexfoundry/857/usr/lib/jvm/java-1.8.0-openjdk-amd64/bin/java -Dcassandra.config=file:///var/snap/edgexfoundry/857/cassandra/etc/cassandra.

```
* Invalidates #1202 (because since we're now using bases in the snapcraft.yaml remote parts aren't available to us and so we can't use them anymore)
* Switches to downloading the go release binaries from upstream directly and hashing them
* Removes erroneous daemon-notify plug from `sys-mgmt-agent` service
[...]

Issue #1227, opened 2019-04-04
Make shell scripts conform to shellcheck


anonymouse64:
There are a number of shell scripts in the repo that have issues as reported by [shellcheck](https://github.com/koalaman/shellcheck).

We should:

1. Fix these issues
2. Integrate the shellcheck program into CI pipelines (probably under `make test` to prevent new issues from being introduced to the project.

anonymouse64:
Note that if we are still considering SonarQube there is a plugin for this: https://github.com/emerald-squad/sonar-shellcheck-plugin

tonyespy:
A lot of these errors were cleaned up recently as part of https://github.com/edgexfoundry/edgex-go/pull/2534, however there are still some remaining warnings. They don't however warrant fixing for Geneva, so I'm removing the release milestone.

jpwhitemn:
Jim Wang has done some shellcheck work with security scripts - but probably needs to be formalized and reviewed globally.

Issue #959, opened 2018-12-20
Switch to redis as the database provider for the snap


anonymouse64:
Once all of the relevant services that currently use mongo are able to use redis instead, we should switch to use redis in the snap so that we can build/run the snap on arm 32-bit platforms such as the Raspberry Pi.

@tonyespy apparently has started work on this, so assigning to him...

anonymouse64:
As per my discussion today with @tonyespy we agreed to first add redis into the snap so we can experiment with it and then try to make the decision regarding switching to it as the default afterwards.

anonymouse64:
We will consider whether to change for Fuji as we've hit the limit on Edinburgh changes

anonymouse64:
Unassigning myself to let Tony decide on the default that should be used for the snap

anonymouse64:
Let's do this for Geneva

tonyespy:
This was already handled in this PR:

https://github.com/edgexfoundry/edgex-go/pull/2426

However it's not yet possible to switch back and forth between mongo and redis in the snap yet due to the config seed changes. As such, I'm leaving this open for now as there's a little more work to left remaining.

tonyespy:
Logic to switch back and forth was added in https://github.com/edgexfoundry/edgex-go/pull/2534, so closing this out.

PR #1319, opened 2019-05-08
Update snap to core18, other miscellaneous Edinburgh fixes

### LINKED ISSUE #799: Specify epochs for the snap to prevent automatic delhi->edinburgh refresh
Currently all setup is done once, during the `install` hook which installs everything into $SNAP_DATA, however this is problematic on refreshing ~~because the $SNAP_DATA dir is cleaned out~~ (see below https://github.com/edgexfoundry/edgex-go/issues/799#issuecomment-476341118 ) so it can be restored on a revert, so the post-refresh hook needs to set everything up again like the install hook.

EDIT: see later comments

### LINKED ISSUE #888: Define a base and use snapcraft 3.X for the snap

### LINKED ISSUE #922: Update snap to use new config-seed "overwrite" flag
With https://github.com/edgexfoundry/edgex-go/pull/902 there's now another option to config-seed `-overwrite` that allows for conditionally writing config values from configuration.toml files into consul. I'm not sure yet what the snap _should_ do, but we have a few options:

1. We could always have config-seed run with overwrite always, this will ensure that the configuration.toml files in $SNAP_DATA are the "canonical" source of truth for what config values should be used
2. We could have the default config-seed service run without overwrite so that it will only initially populate the configuration.toml files from $SNAP_DATA into consul on the first run, and after that the files are "not used", however we could also introduce a manual command in the snap that folks could run that does run config-seed with `-overwrite`, so that then if folks modify the configuration.toml files, those won't go into affect unless they also run the `update-config` command (straw-man name, could be anything).
3. We could do something else with `snap set` and friends to determine when we should/shouldn't use `-overwrite`. Not sure what this would be though...

I kind of like 2 the best because it 
1. makes it obvious when configuration.toml files will be introduced into consul (on the very first run of config-seed after install, and after a user manually runs `update-config`)
2. it would provide a better "user experience" for updating config files, as with 1 (and today) you would effectively need to re-start config-seed by doing `snap set edgexfoundry config-seed=off && snap set edgexfoundry config-seed=on`, whereas there's just a single command to run which doesn't disable/enable any of the services
3. It would be faster than restarting config-seed or restarting all of the services to run
4. We could also hook up all the configuration.toml files to a `snap set edgexfoundry SVC.KEY=...` so that the config hook would read SVC.KEY and write KEY into the configuration.toml for SVC, and then after this call the `update-config` app so that the configuration is automatically updated into consul. (but note that this last part doesn't need to be done to implement `update-config` app.

### LINKED ISSUE #1103: Implement more resilient method of downloading cassandra for the snap

### LINKED ISSUE #1202: Switch back to using go remote part for the snap
This is a followup to #1109 - the snapcraft `go` remote part works with modules again, so we should switch back to using that instead of building it here.

### LINKED ISSUE #1204: Use start-timeout for snap services
With snapd 2.38, a new `start-timeout` option can be used with snap.yaml (not yet snapcraft.yaml though it seems) to control how long systemd should wait before giving up on starting a snap daemon. We should at least specify this option in the snap to accommodate running on low power devices (such as the arm64 dragonboard 410c). We could also use this to finally get rid of some of the hackisk solutions we currently have, such as having all the go services be forking daemons and just have them all immediately fork to decrease the time until systemd thinks the service is ready. 

Some systems such as dragonboards have had failure to install the snap due to cassandra timing out and then kong will fail to start and then the whole installation is aborted.

### LINKED ISSUE #1262: Add new Go-based device-virtual to the snap
The new [device-virtual-go](https://github.com/edgexfoundry/device-virtual-go) should be added to the snap.

### LINKED ISSUE #1282: Update snap build documentation
```
Fetched 43.1 MB in 6s (4167 kB/s)                                                                                                                    
Failed to fetch stage packages: Error downloading packages for part 'cassandra': The package 'openjdk-8-jdk' was not found..
```

### LINKED ISSUE #1363: Remove device-mqtt, device-modbus from the snap
# 🚀 Feature Request

### Relevant Package
This feature request is to remove device-mqtt and device-modbus from the edgexfoundry snap, as they will now be in separate snaps for Edinburgh release.

anonymouse64:
This PR does a lot of things, including defining a base (and thus using "legacy" snapcraft) and thus using the newer version of snapcraft (3.X) directly which will unbreak the daily jobs such as https://jenkins.edgexfoundry.org/view/Snap/job/edgex-go-snap-master-stage-snap/347. Additionally, this fully switches from having the build scripts defined here to using the build definitions defined in the jenkins job instead.

Tasks for devops / ci-management: 
- [x] ci-management PR (https://github.com/edgexfoundry/ci-management/pull/357) merged
- [x] deploy verify job for edgex-go to jenkins sandbox to test this PR - see https://jenkins.edgexfoundry.org/sandbox/view/All/job/edgex-go-snap-master-verify-snap
- [x] ~~open back-port PR to delhi~~ - actually unnecessary as we are able to specify to use job definitions per build stream, so delhi continues to use the old `build_script`

This PR also:
* Fixes #1204 
* Fixes #1282
* Fixes #1262
* Fixes #888 
* Fixes #922
* Fixes #1363
* Partially addresses #959 by adding Redis to the snap, but doesn't enable it by default
* Fixes #799 by specifying `epoch: 1` for the edinburgh snap to block refreshes from delhi to edinburgh
* Fixes #1103 by using the archive download links
* Adds a work-around for #1306 
* Addresses #1227 for the snap scripts at least
* Limits RAM usage by cassandra to something more reasonable, on a 16 GB amd64 machine:

before (approx 770 MB in non-swappable memory):
```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root     23764  4.7  2.9 2440648 474816 ?      SLl  17:51   3:11 /snap/edgexfoundry/x1/usr/lib/jvm/java-1.8.0-openjdk-amd64/bin/java -Dcassandra.config=file:///var/snap/edgexfoundry/x1/cassandra/etc/cassandra.ya
```
after (approx 60 MB in non-swappable memory):
```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root      3547 60.0 27.3 6231000 4443968 ?     SLl  19:02   1:13 /snap/edgexfoundry/857/usr/lib/jvm/java-1.8.0-openjdk-amd64/bin/java -Dcassandra.config=file:///var/snap/edgexfoundry/857/cassandra/etc/cassandra.

```
* Switches to downloading the go release binaries from upstream directly and hashing them, which closes #1202 (because it is no longer needed to build go from source during the build)
* Removes erroneous daemon-notify plug from `sys-mgmt-agent` service
* Attempts to work around snapd bug https://bugs.launchpad.net/snapd/+bug/1818306 by not only disabling services with systemd when we initially install but also set the snap config items for those services to be off so that after a refresh when (due to the bug) snapd erroneously starts every service we will at least disable/turn off services that should have been left disabled and stopped during the `configure` hook
* Uses command-chain with passthrough for the hooks so that we have a wrapper generated for the hooks with proper environment variables set
* Eliminates the mongod wrapper and just runs `mongod` executable directly from the app
* Dynamically determines the perl version when running kong as we now have a different version of perl with the core18 base
* Removes calls to `make prepare` as that's obselete with go modules
* Removes all of the build scripts for the snap from here as all of that now lives in the ci-management repo [here](https://github.com/edgexfoundry/ci-management/blob/master/shell/edgexfoundry-snapcraft.sh)
* Eliminated the `fork-wrapper` and `go-services-wrapper` to just call the binaries directly and use `adapter: none` for static pure go binaries and `adapter: full` to use command-chain instead of snapcraft wrappers
* Specify profile directories explicitly in configuration.toml files for the device-* services so that we don't need to run the device services with a working directory of `$SNAP_DATA/config/device-*/`

To justify some of the changes made with respect to the adapter and the scripts we're using, previously with the delhi version of the snap we had 27 snapcraft generated shell wrappers (1 for each app) and 13 shell wrappers for the services, and now we have 1 snapcraft generated wrapper in `/snap/command-chain/snapcraft-runner` and 12 shell wrappers for the services (though I think we should have 11, the go-build-helper.sh is creeping in from somewhere and it shouldn't be). Associated with this change, the snap.yaml grew in size from 207 lines to 343, but that's being generated by snapcraft so I think it's okay to have it get more complex.

Originally opened as #1297 

I now have tests updated that should test all the additional new features in this branch available in the checkbox repo under branch `feature/edinburgh-tests`. The refresh tests are currently skipped because now all the snap revisions that have been released are `epoch: 0` and since this is `epoch: 1` it's not refresh-able. To run the tests:

```
$ # install snapcraft and multipass to build the snap
$ snap install snapcraft --classic
$ snap install multipass --classic --beta
$ # build the snap somewhere say $GOPATH/src/github.com/edgexfoundry/edgex-go/edgexfoundry_*.snap
$ git clone ssh://$YOUR_LP_USERNAME@git.launchpad.net/checkbox-provider-edgex
$ cd checkbox-provider-edgex
$ git checkout feature/edinburgh-tests
$ # install tools to run the tests
$ snap install jq remarshal
$ snap alias remarshal.toml2json toml2json
$ snap alias remarshal.json2toml json2toml
$ # run the tests against the local snap - note the script needs to be run as root now
$ sudo bin/run-all-tests-locally.sh $GOPATH/src/github.com/edgexfoundry/edgex-go/edgexfoundry_*.snap --verbose --ignorefail
```

The only tests that currently fail are the refresh tests and the security services, but the security services will need to be fixed up anyways and I don't want to hold back this PR for the security services.


anonymouse64:
Adding @mhall119 as a reviewer for the changes to snap/README.md

bnevis-i:
This PR should also remove a roadblock to building behind a proxy since Cassandra is fetched from Apache archive instead of built from source.  (FYI to future readers of this PR).

snap/snapcraft.yaml
```diff
@@ -12,256 +13,318 @@ description: |
...
-      after:
-        - core-config-seed
+    plugs: [hardware-observe, network, network-bind, system-observe]
+    stop-command: bin/mongod --shutdown --dbpath $SNAP_DATA/mongo/db
   cassandra:
+    adapter: full
     command: bin/cassandra-wrapper.sh
     daemon: forking
+    passthrough:
+      start-timeout: 15m
```
tonyespy: 15m seems like a *really* long timeout?

snap/snapcraft.yaml
```diff
@@ -12,256 +13,318 @@ description: |
...
+    after:
+      - consul
+      - pkisetup
+      - core-config-seed
+    command: bin/vault server --config $VAULT_CONFIG
+    daemon: simple
     environment:
-      VAULT_CONFIG: "$SNAP_DATA/config/security-secret-store/vault-config.json"
+      VAULT_CONFIG: "$SNAP_DATA/config/security-secret-store/vault-config.hcl"
       VAULT_ADDR: "https://localhost:8200"
```
tonyespy: Hmm, this means vault is always bound to port 8200 in the snap?

snap/snapcraft.yaml
```diff
@@ -12,256 +13,318 @@ description: |
...
     plugs: [network, network-bind]
-    passthrough:
-      after:
-        - core-config-seed
-        - mongo-worker
-        - edgexproxy
+    # for now, specify stop-timeout as really short until services learn how
+    # to handle SIGTERM properly with as per 
+    # https://github.com/edgexfoundry/edgex-go/issues/1306
+    stop-timeout: 1s
```
tonyespy: This seems short for a kill timeout for a process that deals with a db. Any chance this could lead to data corruption if core-data were in the midst of flushing writes to the db?

snap/snapcraft.yaml
```diff
@@ -12,256 +13,318 @@ description: |
...
-        - core-config-seed
-        - mongo-worker
-        - edgexproxy
-  device-mqtt:
-    command: bin/go-services-wrapper.sh device-mqtt
+  device-random:
+    adapter: none
+    after:
+      - core-config-seed
+      - mongo-worker
```
tonyespy: Is this a no-op if redis is enabled?

snap/snapcraft.yaml
```diff
@@ -12,256 +13,318 @@ description: |
...
+      - edgexproxy
+      - support-logging
```
tonyespy: See previous comment about device services and support-logging...

snap/snapcraft.yaml
```diff
@@ -297,30 +360,49 @@ parts:
...
+    # build after kong to prevent staging go binaries into the snap as they
+    # are large - see the TODO on the kong part
+    after: [go-build-helper, kong]
 
   consul:
     after: [go]
     plugin: make
     source: https://github.com/hashicorp/consul.git
     source-tag: v1.1.0
+    source-depth: 1
```
tonyespy: You didn't really explain *why* specifying source-depth: 1 was an improvement. What's the current default?

anonymouse64: I have updated the commit message to explain this

snap/snapcraft.yaml
anonymouse64: Yes it is a no-op if `mongo-worker` has been disabled. Specifically, `after` in snap.yaml and snapcraft.yaml corresponds to the `After` directive in a systemd unit which doesn't do anything if the specified service is disabled.

snap/snapcraft.yaml
anonymouse64: It could lead to data corruption. What do you think is more reasonable? Without this, systemd will use the snapd default of 30 seconds, and due to the referenced bug the service doesn't respond to SIGTERM at all in that time meaning that we always end up waiting at least 30 seconds to remove the snap or stop any of the services.

snap/snapcraft.yaml
anonymouse64: Yes but this is no different than it was before. There's not currently a clean way to modify environment variables for services so if we wanted to expose the port number here to say `snap set`, we would need to implement our own wrapper which doesn't use the `environment` section here which is what I was trying to avoid with many of these changes.

snap/local/runtime-helpers/bin/start-consul.sh
```diff
@@ -1,44 +1,23 @@
...
-#
-# This script includes some configuration copied from the
-# core-config-seed's Dockerfile, and is otherwise based
-# on two shell scripts which exist in the same directory.
-#
-#  - launch-consul-config.sh
-#  - docker-entrypoint.sh
-#
-set -e
+#!/bin/bash -e
```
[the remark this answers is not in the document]
anonymouse64: For consistency's sake and I think `bash` is a more usable scripting language than `sh`.

snap/local/runtime-helpers/bin/start-consul.sh
```diff
@@ -1,44 +1,23 @@
...
-     $CONSUL_ARGS | tee "$LOG_DIR"/core-consul.log
-     
+# to actually test if consul is ready, we simply check to see if consul 
+# itself shows up in it's service catalog
+# also note we don't have a timeout here because we use start-timeout for this
+# daemon so systemd will kill us if we take too long waiting for this
+CONSUL_URL=http://localhost:8500/v1/catalog/service/consul
+until [ -n "$(curl -s $CONSUL_URL | jq -r '. | length')" ] && 
+    [ "$(curl -s $CONSUL_URL | jq -r '. | length')" -gt "0" ] ; do
+    sleep 1
```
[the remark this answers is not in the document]
anonymouse64: Yes

snap/snapcraft.yaml
anonymouse64: Well there are a lot of services in the snap and especially on a refresh when all of the services get automatically enabled (before getting disabled when the configure hook runs to work around https://bugs.launchpad.net/snapd/+bug/1818306) it may be 15 minutes before a Raspberry Pi 3B+ running Ubuntu Server arm64 with only 1 GB can successfully start all the services. I don't think it's actually quite that high IIRC it's closer to 5-10 minutes, but I wanted to be safe.

snap/local/runtime-helpers/bin/kong-daemon.sh
```diff
@@ -1,23 +1,14 @@
...
-until $KONG_SNAP migrations up --yes --conf $KONG_CONF; do
-    sleep 10
-    # increment number of tries
-    num_tries=$((num_tries+1))
-    if (( num_tries > MAX_KONG_UP_TRIES )); then
-        echo "max tries attempting to bring up kong"
-        exit 1
-    fi
+until $KONG_SNAP migrations up --yes --conf "$KONG_CONF"; do
+    sleep 5
```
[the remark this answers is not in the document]
anonymouse64: Done

.gitignore
```diff
@@ -35,3 +35,4 @@ prime/*
 stage/*
 snap/.snapcraft/*
 go.sum
+squashfs-root/
```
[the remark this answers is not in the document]
anonymouse64: Added

snap/hooks/configure
```diff
@@ -84,14 +110,112 @@ for key in $ALL_SERVICES; do
...
+        fi
+
+        # save the database provider as mongodb for next invocation
+        echo "mongodb" > "$PREV_DB_PROVIDER_FILE"
+        ;;
+    *)
+        echo "invalid setting for dbtype: $dbProvider"
+        exit 1
+        ;;
+esac
```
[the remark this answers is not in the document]
anonymouse64: Added

snap/hooks/pre-refresh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -e
+
+# save this revision for when we run again in the post-refresh
+snapctl set lastrev="$SNAP_REVISION"
+
+snapctl set release="edinburgh"
```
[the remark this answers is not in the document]
anonymouse64: Added

snap/hooks/configure
```diff
@@ -54,23 +58,45 @@ handle_svc () {
...
+    status=$(snapctl get "$key")
     case $key in 
+        device*)
+            # the device services are all using the device-sdk-go which waits
+            # for support-logging to come online, so if we are turing one
+            # of the device services on, we should also turn support-logging
+            # on
+            # also turn on core-data and core-metadata on too
+            if [ "$status" = "on" ]; then
+                handle_svc "support-logging" "on"
```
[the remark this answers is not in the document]
anonymouse64: I'm not really sure, I just have noticed that sometimes device services using the go SDK want support-logging enabled as well before finishing startup.

snap/local/build-helpers/patches/0001-Remove-ARCH-expansion-from-the-Makefile.patch
```diff
@@ -0,0 +1,40 @@
...
+$(ARCH) variable in the Makefile gets erroneously interpreted
+as an argument to the `-ggdb' command line option during the
+compilation stage or as an input file name during the linking
+stage.
+
+This patch removes $(ARCH) expansions from the Makefile. This
+doesn't harm cross-compilation, the latter goes fine with the
+properly assigned CC environment variable. For native builds,
+this patch does not imply any changes.
+
```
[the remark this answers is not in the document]
anonymouse64: It's added to enable future cross-compilation, I added this as part of my work trying to get the snap building on armhf using the arm64 builders we have.
I can drop it if you like

anonymouse64:
recheck

snap/hooks/configure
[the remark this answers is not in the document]
tonyespy: I'm not going to block the PR, but we should figure this out. I'm pretty sure that most (if not all) of the device services formerly included in the snap had remote logging disabled. If the device-sdk-go still waits for support-logging in this case, then it's a bug that should get fixed. If the device-sdk-go doesn't wait for support-logging, then by enabling it here, we've got an idle, non-required service running...

snap/local/build-helpers/patches/0001-Remove-ARCH-expansion-from-the-Makefile.patch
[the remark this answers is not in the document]
tonyespy: I'd prefer to drop...

snap/snapcraft.yaml
tonyespy: Yes, I realize this wasn't change...  Might be worth an issue to track.

snap/snapcraft.yaml
tonyespy: 10s? Honestly until the services are fixed to respond properly to SIGTERM, that's still a blind guess, and corruption could still possibly occur, but it seemed like 1s was too short...

snap/hooks/configure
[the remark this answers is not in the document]
anonymouse64: I think this was the behavior under Delhi, as I just tested with the new `edgex-device-mqtt` snap with this branch and device-mqtt didn't block waiting for `support-logging`. I will remove the dependency of device services on support-logging. If a user wants the device services to use remote logging they will need to turn on support-logging manually.

snap/snapcraft.yaml
anonymouse64: Filed as https://github.com/edgexfoundry/edgex-go/issues/1387

snap/snapcraft.yaml
anonymouse64: Actually though this case is slightly different than the above case - I agree we shouldn't automatically start support-logging if a user starts a device service, but if both a device service is enabled and support-logging is enabled, then we should order the device service after support-logging on the change that the device service is configured to use remote logging as this will avoid the device service having to wait for support-logging to be started by a concurrent systemd task and instead will just be ordered afterwards.
As such, I will leave this bit in as it has no real cost.

snap/snapcraft.yaml
anonymouse64: Okay, I changed it to 10s

snap/local/build-helpers/patches/0001-Remove-ARCH-expansion-from-the-Makefile.patch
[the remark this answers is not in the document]
anonymouse64: Okay, I have filed https://github.com/edgexfoundry/edgex-go/issues/1388 to track cross-compiling as a feature for future development

anonymouse64:
Relabeling for Fuji since we missed the branch cut for this PR, however since this is still needed for Edinburgh, I have filed #1389 which will actually implement these changes for Edinburgh proper

snap/local/runtime-helpers/bin/start-consul.sh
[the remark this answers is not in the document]
tonyespy: OK, I asked as on a lot of minimal Ubuntu systems dash (sh) is all that's available, and thus is typically the shell of choice for system jobs. That said, I think we're fine with bash here. Thanks for the explanation.

PR #1324, opened 2019-05-08
Updated redis and mongo providers to fill in Device ID from event

### LINKED ISSUE #1311: When Persisting Readings, Set Device Name from the Event
Follows from device-sdk-go issue https://github.com/edgexfoundry/device-sdk-go/issues/269

In the interest of smaller request sizes and efficiency, the Device SDK team would like to not set the Device property on every reading for a given Event. The reason is that the value is always identical to the Device property on the Event itself. We do still need to set this value in the persistence layer because it is possible to [query readings](https://docs.edgexfoundry.org/core-data.html#reading_device__deviceId___limit_) separately from events.

In the [Mongo](https://github.com/edgexfoundry/edgex-go/blob/c67086fe10c4d34caeefffaee5379490103dd63f/internal/pkg/db/mongo/data.go#L56) and [Redis](https://github.com/edgexfoundry/edgex-go/blob/c67086fe10c4d34caeefffaee5379490103dd63f/internal/pkg/db/redis/data.go#L792) providers, please make the necessary change to set the Device property on a Reading from the Device property on its parent Event if the Reading.Device value is empty.

brandonforster:
Fix #1311.

### DIFF
--- a/internal/pkg/db/mongo/data.go
+++ b/internal/pkg/db/mongo/data.go
@@ -62,6 +62,10 @@ func (mc MongoClient) AddEvent(e contract.Event) (string, error) {
 
 			r.TimestampForAdd()
 
+			if reading.Device == "" {
+				r.Device = e.Device
+			}
+
 			ui = append(ui, r)
 
 			e.Readings[i].Id = id
--- a/internal/pkg/db/redis/data.go
+++ b/internal/pkg/db/redis/data.go
@@ -780,6 +780,12 @@ func addEvent(conn redis.Conn, e contract.Event) (id string, err error) {
 		return "", err
 	}
 
+	for i, reading := range e.Readings {
+		if reading.Device == "" {
+			e.Readings[i].Device = e.Device
+		}
+	}
+
 	_ = conn.Send("MULTI")
 	_ = conn.Send("SET", e.ID, m)
 	_ = conn.Send("ZADD", db.EventsCollection, 0, e.ID)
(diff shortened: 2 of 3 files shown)


internal/pkg/db/mongo/data.go
```diff
@@ -77,6 +77,12 @@ func (mc MongoClient) AddEvent(e contract.Event) (string, error) {
 		return "", err
 	}
 
+	for i, reading := range e.Readings {
+		if reading.Device == "" {
+			e.Readings[i].Device = e.Device
+		}
+	}
```
tsconn23: Reading.Device is still empty. The reason is because we have already mapped the event `e` by this point. On line 88 below, you can see that `mapped` is what is being saved to the database, not `e`. The above logic needs to run as part of the loop starting on line 54:
```
//Add the readings
if len(e.Readings) > 0 {
	var ui []interface{}
	for i, reading := range e.Readings {
                if reading.Device == "" {
                     reading.Device = e.Device
                }
		var r models.Reading
		id, err := r.FromContract(reading)
		if err != nil {
			return "", err
		}

		r.TimestampForAdd()

		ui = append(ui, r)

		e.Readings[i].Id = id
	}
	if err := s.DB(mc.database.Name).C(db.ReadingsCollection).Insert(ui...); err != nil {
		return "", errorMap(err)
	}
}
```

internal/pkg/db/redis/data.go
```diff
@@ -780,6 +780,12 @@ func addEvent(conn redis.Conn, e contract.Event) (id string, err error) {
 		return "", err
 	}
 
+	for i, reading := range e.Readings {
+		if reading.Device == "" {
+			e.Readings[i].Device = e.Device
+		}
+	}
```
tsconn23: This one works because Redis is operating on the contract model directly and not mapping it to a provider specific model.

PR #1325, opened 2019-05-08
Feature/issue 1216 cbor event handling

### LINKED ISSUE #1216: Core-Data: CBOR Event Handling
New events coming into core-data are handled by the POST section of `eventHandler` found in [router.go](https://github.com/edgexfoundry/edgex-go/blob/cd039ab618167a822852876435a45f7869f5901f/internal/core/data/router.go#L184). For this task we'll need to branch the logic according to the incoming `ContentType` header value. JSON is the default, but calling clients can also indicate CBOR. If the event is being passed in CBOR, we need to do the following:

1.) Unmarshal the incoming request body from CBOR to an Event and associated Readings using the `github.com/ugorji/go/codec` package
2.) Persist the event into the given database
3.) Publish the event to the message bus

Steps 2 and 3 above currently happen [here](https://github.com/edgexfoundry/edgex-go/blob/cd039ab618167a822852876435a45f7869f5901f/internal/core/data/event.go#L85)

There are a few points to how this should work
- When persisting an event originating as CBOR, the [BinaryValue](https://github.com/edgexfoundry/go-mod-core-contracts/blob/e53159cfbe2cc8832483252ada48f1a5475584f6/models/reading.go#L37) of a given reading should NOT be persisted to the database because of space considerations. This means that whatever custom logic we need to write to unmarshal an Event from CBOR should ignore the BinaryValue property.
- The bytes of the original CBOR request will therefore need to be retained as they constitute the only copy we will have of the binary data. This is important when we [publish the Event](https://github.com/edgexfoundry/edgex-go/blob/cd039ab618167a822852876435a45f7869f5901f/internal/core/data/event.go#L222).
- During the act of publishing the event, we need to indicate a ContentType of CBOR versus JSON. If JSON, the existing mechanism will work to re-marshal the given event. If CBOR, we should pass the original bytes.

AnthonyMBonafide:
fix #1216

Added BodyReader interface which abstracts the process handling the
request. This follows a more chain-of-command flow.

Updated the API for adding a new event to return an Accepted(202) status
for requests with CBOR contents. This is because the persistence
processing happens asynchronously and not guaranteed to exist when the
client receives the response

Created two implementations of BodyReader which handle JSON and CBOR
content types

Added original CBOR content to context for later publishing event to
avoid re-encoding to obtain the same data

Updated to use latest versions of go-mod-core-contracts and
go-mod-messaging. go-mod-messaging was updated to include the checksum
information in the MessageEnvelope which was needed when publishing
messages over the message bus.

Added tests to cover new Readers and Extractor functionality.

Added new Event endpoint which accepts the PUT HTTP method. The new
endpoint updates the pushed information for an event which can be
identified by checksum value. This new endpoint will also remove the
checksum information once it has been marked pushed. Also, all events
with a matching checksum will be updated.

Updated Mongo and Redis implementations to store checksum information

Update the DBClient interface to have functionality for persisting new
events with checksum information. Also, added functionality for
retrieving events by checksum


AnthonyMBonafide:
I manually verified the changes for Mongo and Redis.

tsconn23:
@AnthonyMBonafide Just FYI that I've started looking at this. We may need to dial the solution back a bit. I'm going to create a branch that we can review together rather than itemize comments here.

internal/core/data/handler.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	w.Write([]byte("true"))
+}
+
+func publishEventData(checksum string, event models.Event, ctx context.Context) {
```
tobiasmo1: Just an observation regarding API consistency.. 
I see you've introduced two methods for persisting an event with/without checksum **persistEventToDb** and **persistEventToDbWithChecksum** 
However, the **publishEventData** method is overloaded to publish event data with and without checksum, requiring caller to supply empty string if not in scenario where they'll apply checksum.

internal/core/data/interfaces/mocks/DBClient.go
```diff
@@ -78,20 +101,6 @@ func (_m *DBClient) CloseSession() {
 	_m.Called()
 }
 
-// Connect provides a mock function with given fields:
```
tobiasmo1: Is this removal of Connect() method intended to be part of CBOR event handling PR?

internal/core/data/event.go
```diff
@@ -192,6 +206,27 @@ func getEventById(id string) (contract.Event, error) {
 	return e, nil
 }
 
+// updateEventPushDateByChecksum updates the pushed dated for all events with a matching checksum which have not already been marked pushed
+func updateEventPushDateByChecksum(checksum string, ctx context.Context) error {
+	evts, err := dbClient.EventsByChecksum(checksum)
```
tobiasmo1: To maintain a 1:1 event -> push activity, is it safe to assume earliest single event with the matching checksum is what caller intends to mark 'pushed'?

If so, we could order the events by e.createTime ascending and update just the earliest matched; i.e., index 0 as the one which was pushed.

tsconn23:
@tobiasmo1 @AnthonyMBonafide Here's my branch based on this PR
https://github.com/tsconn23/edgex-go/tree/cbor_ingest

Anthony's got some time off between now and Tuesday next week. If you get some cycles, feel free to review. I would like to schedule some time for a 3 way call to at least discuss what I am proposing in this branch. Ping me on Slack w/r/t your respective availability.

cmd/core-data/res/configuration.toml
```diff
@@ -5,6 +5,7 @@ PersistData = true
 ServiceUpdateLastConnected = false
 ValidateCheck = false
 LogLevel = 'INFO'
+ChecksumAlgo = 'xxHash'
```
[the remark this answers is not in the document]
AnthonyMBonafide: No, but it should be. I will need to document this.

internal/core/data/handler.go
AnthonyMBonafide: For the case where there is no checksum information I was intending on using the [putEventOnQueue](https://github.com/edgexfoundry/edgex-go/blob/d8b36499b9ebed8c04bf5a46aea7dce45159b10f/internal/core/data/event.go#L245) function which provides the functionality of `publishEventData` but without providing the checksum information. For consistency I can rename `publishEventData` to `publishEventDataWithChecksum` so it is more clear to the reader.

internal/core/data/interfaces/mocks/DBClient.go
AnthonyMBonafide: This was done by the mockery tool. Since I updated the `DBClient` interface I had to use mockery to regenerate the `mock/DBClient`. The `Connect` function was removed from the `DBClient` in [this commit](https://github.com/edgexfoundry/edgex-go/commit/963a78a17580fc1390acdd718979617b762ec255), which causes the code generation tool to not include it. We can add it back if it is needed.

cmd/core-data/res/configuration.toml
```diff
@@ -42,16 +43,14 @@ File = './logs/edgex-core-data.log'
   Host = 'localhost'
   Name = 'coredata'
   Password = ''
-  Port = 27017
+  Port = 6379
   Username = ''
   Timeout = 5000
-  Type = 'mongodb'
+  Type = 'redisdb'
```
tsconn23: Please switch this back to Mongo.

internal/core/data/event.go
```diff
@@ -124,7 +127,35 @@ func addNewEvent(e contract.Event, ctx context.Context) (string, error) {
 	return e.ID, nil
 }
 
-func updateEvent(from contract.Event, ctx context.Context) error {
+func validateEvent(e contract.Event, ctx context.Context) error {
+	err := checkDevice(e.Device, ctx)
```
tsconn23: This function is only used by tests, not by actual business logic. Needs to be removed. Looks like the tests also only exist to exercise this function, so they can also conceivably be removed.
- TestAddEventWithValidationValueDescriptorExistsAndIsInvalid
- TestAddEventWithValidationValueDescriptorNotFound
- TestEventValidationValueDescriptorDBError

internal/core/data/event.go
tsconn23: See comment below w/r/t expectation of checksum uniqueness. Right now the way the logic is written, all events for a given checksum will be marked pushed. If folks feel strongly, we can change but then there could possibly be orphans.

internal/core/data/event.go
```diff
@@ -192,37 +225,50 @@ func getEventById(id string) (contract.Event, error) {
...
+		err = updateEvent(models.Event{Event: e}, ctx)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 func updateEventPushDate(id string, ctx context.Context) error {
+	// TODO: This is going to break. Probably need DBClient to return Correlation event model
```
tsconn23: Please remove this comment based on our follow-up review.

internal/core/data/event_test.go
```diff
@@ -21,16 +21,17 @@ import (
...
-	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 	"github.com/edgexfoundry/go-mod-core-contracts/models"
-
 	"github.com/globalsign/mgo/bson"
 	"github.com/stretchr/testify/mock"
+
+	"github.com/edgexfoundry/edgex-go/internal/core/data/errors"
+	dbMock "github.com/edgexfoundry/edgex-go/internal/core/data/interfaces/mocks"
+	correlation "github.com/edgexfoundry/edgex-go/internal/pkg/correlation/models"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
```
tsconn23: Please move this block of imports under line 24 above.

internal/pkg/db/mongo/models/event.go
```diff
@@ -14,11 +14,14 @@
...
 import (
-	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 	"github.com/globalsign/mgo"
 	"github.com/globalsign/mgo/bson"
 	"github.com/pkg/errors"
+
+	correlation "github.com/edgexfoundry/edgex-go/internal/pkg/correlation/models"
+
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
```
tsconn23: Please group the edgexfoundry imports -- up toward line 17.

internal/pkg/db/redis/data.go
```diff
@@ -14,11 +14,13 @@
...
 
 import (
-	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 	"github.com/gomodule/redigo/redis"
 	"github.com/google/uuid"
 	"github.com/imdario/mergo"
+
+	correlation "github.com/edgexfoundry/edgex-go/internal/pkg/correlation/models"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
```
tsconn23: Please group the edgexfoundry imports -- up toward line 17.

internal/pkg/db/redis/event.go
```diff
@@ -16,23 +16,27 @@ package redis
 import (
 	"encoding/json"
 
-	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
+	correlation "github.com/edgexfoundry/edgex-go/internal/pkg/correlation/models"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 	"github.com/gomodule/redigo/redis"
+
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
```
tsconn23: Organize imports

internal/core/data/handler.go
tsconn23: Resolving conversation as this logic has been removed entirely.

tsconn23:
Folks, Anthony is on vacation this coming week so I'll be taking over driving this PR to completion. Because of that, I can't participate as a reviewer. Let's get this closed out so that @lenny-intel and @rsdmike can start work on handling published events, and we can resolve any issues we find during integration.

internal/core/data/event.go
```diff
@@ -192,37 +197,49 @@ func getEventById(id string) (contract.Event, error) {
...
 }
 
 // Put event on the message queue to be processed by the rules engine
-func putEventOnQueue(e contract.Event, ctx context.Context) {
+func putEventOnQueue(evt models.Event, ctx context.Context) {
 	LoggingClient.Info("Putting event on message queue")
-	//	Have multiple implementations (start with ZeroMQ)
-	evt := models.Event{}
-	evt.Event = e
+	// 	Have multiple implementations (start with ZeroMQ)
```
lenny-intel: Why is this comment needed. This code is independent of the message bus implementation.

internal/core/data/interfaces/mocks/DBClient.go
```diff
@@ -2,8 +2,10 @@
 
 package mocks
 
-import "github.com/stretchr/testify/mock"
-import "github.com/edgexfoundry/go-mod-core-contracts/models"
+import go_mod_core_contractsmodels "github.com/edgexfoundry/go-mod-core-contracts/models"
```
lenny-intel: Use consistent aliasing to resolve the name conflicts. i.e. leave this as modes and alias models below as "correlation"

internal/core/data/interfaces/mocks/DBClient.go
```diff
@@ -220,16 +208,39 @@ func (_m *DBClient) Events() ([]models.Event, error) {
...
+	var r0 []go_mod_core_contractsmodels.Event
+	if rf, ok := ret.Get(0).(func(string) []go_mod_core_contractsmodels.Event); ok {
+		r0 = rf(checksum)
+	} else {
+		if ret.Get(0) != nil {
+			r0 = ret.Get(0).([]go_mod_core_contractsmodels.Event)
+		}
+	}
+
+	var r1 error
```
lenny-intel: can we use standard name like "err".

internal/core/data/interfaces/mocks/DBClient.go
```diff
@@ -220,16 +208,39 @@ func (_m *DBClient) Events() ([]models.Event, error) {
 	return r0, r1
 }
 
+// EventsByChecksum provides a mock function with given fields: checksum
+func (_m *DBClient) EventsByChecksum(checksum string) ([]go_mod_core_contractsmodels.Event, error) {
+	ret := _m.Called(checksum)
+
+	var r0 []go_mod_core_contractsmodels.Event
```
lenny-intel: Use meaningful name like "events"?

internal/core/data/event.go
tobiasmo1: @tsconn23 For the db.ErrNotFound to be meaningful (which I believe it could be) we should facilitate at least a 1:1 event received/pushed.

So I am only feeling strongly it should be one or the other: 
a) Mark all matching events as pushed (as it is) + do not produce error in cases where checksum is not found.  
b) Mark one matching event pushed (e.g., earliest in case time-sensitive but could be arbitrary) + produce error if/when checksum is not found.

internal/core/data/io.go
```diff
@@ -0,0 +1,100 @@
...
+		return event, err
+	}
+
+	switch Configuration.Writable.ChecksumAlgo {
+	case ChecksumAlgoxxHash:
+		event.Checksum = fmt.Sprintf("%x", xxhash.Checksum64(bytes))
+	default:
+		event.Checksum = fmt.Sprintf("%x", md5.Sum(bytes))
+	}
+	c = context.WithValue(c, checksumContextKey, event.Checksum)
```
lenny-intel: Are we still expecting the receiver to recalculate the checksum? If so we also need to add the name of the checksum algo used to the Context and to the message envelope so the receiver can calculate appropriately.

internal/core/data/interfaces/mocks/DBClient.go
tsconn23: This mock is generated code via `mockery`. Would prefer to not change things like this.

internal/core/data/interfaces/mocks/DBClient.go
tsconn23: This mock is generated code via mockery. Would prefer to not change things like this.

internal/core/data/io.go
tsconn23: No. The Checksum is provided as part of the MessageEnvelope specifically to avoid having the receiver perform this operation.

internal/core/data/event.go
tsconn23: Removed

internal/core/data/interfaces/mocks/DBClient.go
tsconn23: Actually I changed this alias to `contracts` which removes the need to alias the correlation model. This has to be regenerated every time the mock changes anyway.

tsconn23:
recheck

internal/core/data/event.go
tsconn23: @tobiasmo1 Well we have to return a 404 if an invalid checksum is supplied (like "abcdef"). I don't see what that has to do with updating all matching records versus just one though. We're also talking about a slim edge case here.

tobiasmo1: I suppose that when a device emits two events with same checksum... so long as we still push the 2nd event to subscribed apps (even if both were already marked pushed), it satisfies expected functionality.  I see a 404 as expected in this edge case, and it will occur 100% of the time when 2nd event gets marked as 'pushed'. . So if we update all matching records wouldn't the caller need to ignore such 404 errors instead of determining how/why the second event was 'lost'?

How about if we return 203 in response of first event being marked 'pushed' if/when there are > 1 matched events?

tsconn23: > I see a 404 as expected in this edge case, and it will occur 100% of the time when 2nd event gets marked as 'pushed'

Not true. The query to fetch events by checksum looks like this
`mc.getEvents(bson.M{"checksum": checksum})`
And that's it. There's no criteria around whether it's been pushed or not, so those events will still be returned.

If you want a safeguard to ensure the Pushed property is only set if it's 0, then I can add that.

tobiasmo1: Ah got it. Okay then we won't see a 404 and also handles case where caller retries to mark pushed for same event. Anyway, I am satisfied with current implementation so can resolve this thread.

PR #1327, opened 2019-05-09
Specify epoch 0 for Delhi, add post-refresh script to fix config file paths

### LINKED ISSUE #799: Specify epochs for the snap to prevent automatic delhi->edinburgh refresh

anonymouse64:
This is the first step in resolving #799 (for Delhi).
The crux of the issue is that currently we process configuration files at install time to use snap specific paths such as $SNAP, $SNAP_DATA, $SNAP_COMMON, and use the environment variables for that. However, those environment variables include the revision number of the snap explicitly such as `/var/snap/edgexfoundry/329` for $SNAP_DATA from the stable channel, which is on revision 329. This is an issue when we push an update to the snap, because the configuration files currently wouldn't be changed and so the services would be trying to use files from a previous revision (say 329) when the current revision and actual data is now in (for example) 330. 
We can remedy this by simply using the `current` symlink instead (i.e. `/var/snap/edgexfoundry/current` which is a symlink to `/var/snap/edgexfoundry/330`).
This patch implements solution by doing 3 things:
* On install, we don't use the revision number, we use `current`
* Before a refresh, we set config items that can be used to check in future revisions if the snap has already been migrated to use the `current` symlink
* After a refresh, we check if that config item is set, and if not then we migrate the config files to use the `current` symlink instead

Finally, this also sets the snap to have `epoch: 0`, so that when we release Edinburgh we can use `epoch: 1` and block automatic refreshes from Delhi to Edinburgh. Read more about epochs [here](https://docs.snapcraft.io/snap-epochs).

## Testing

To test this, I have made some useful scripts in our certification repo to use with checkbox. To run the tests, build the snap from this repo locally somewhere, then clone the repo and run the tests locally with the `run-all-tests-locally.sh`:

```bash
$ # build the snap somewhere say $GOPATH/src/github.com/edgexfoundry/edgex-go/edgexfoundry_*.snap
$ git clone ssh://$YOUR_LP_USERNAME@git.launchpad.net/checkbox-provider-edgex
$ cd checkbox-provider-edgex
$ git checkout feature/more-tests
$ # run the tests against the local snap
$ bin/run-all-tests-locally.sh $GOPATH/src/github.com/edgexfoundry/edgex-go/edgexfoundry_*.snap
```

This currently will run 5 tests:
1. Test that the snap installs without failure
2. Test that the snap services are still running/in expected states after 120 seconds
3. Test that the security services are running, that the generated cert is presented by Kong and doesn't get regenerated when the snap is restarted
4. Test that the snap config files don't contain the revision number path anywhere after installing
5. Test that the snap config files don't contain the revision number path anywhere after refreshing

There is an additional test for the snap service states after a refresh but that test is currently broken by https://bugs.launchpad.net/snapd/+bug/1818306, so that test result is set to not fail when the services are in incorrect state.

### DIFF
--- a/snap/hooks/install
+++ b/snap/hooks/install
@@ -21,6 +21,11 @@ esac
 export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/$MULTI_ARCH_PATH:$SNAP/usr/lib/$MULTI_ARCH_PATH"
 export LD_LIBRARY_PATH=$SNAP_LIBRARY_PATH:$LD_LIBRARY_PATH
 
+# get the values of $SNAP_DATA and $SNAP using the current symlink instead of
+# the default behavior which has the revision hard-coded, which breaks after
+# a refresh
+SNAP_DATA_CURRENT=${SNAP_DATA/%$SNAP_REVISION/current}
+
 # install all the config files from $SNAP/config/SERVICE/res/configuration.toml 
 # into $SNAP_DATA/config
 mkdir -p ${SNAP_DATA}/config
@@ -28,8 +33,10 @@ for service in security-api-gateway security-secret-store core-command config-se
     if [ ! -f "${SNAP_DATA}/config/${service}/res/configuration.toml" ]; then
         mkdir -p "${SNAP_DATA}/config/${service}/res"
         cp ${SNAP}/config/${service}/res/configuration.toml "${SNAP_DATA}/config/${service}/res/configuration.toml"
-        # do replacement of the SNAP_DATA and SNAP_COMMON environment variables in the config files
-        sed -i -e "s@\$SNAP_COMMON@$SNAP_COMMON@g" -e "s@\$SNAP_DATA@$SNAP_DATA@g" "${SNAP_DATA}/config/${service}/res/configuration.toml"
+        # do replacement of the $SNAP, $SNAP_DATA, $SNAP_COMMON environment variables in the config files
+        sed -i -e "s@\$SNAP_COMMON@$SNAP_COMMON@g" \
+            -e "s@\$SNAP_DATA@$SNAP_DATA_CURRENT@g" \
+            "${SNAP_DATA}/config/${service}/res/configuration.toml"
     fi
 done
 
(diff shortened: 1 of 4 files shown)


snap/hooks/post-refresh
```diff
@@ -0,0 +1,138 @@
...
+    sed -i \
+        -e "s@$SNAP_DATA_DIR/[0-9]*/vault/pki@$SNAP_DATA_CURRENT/vault/pki@" \
+        -e "s@$SNAP_DATA_DIR/[0-9]*/config/security-secret-store/res@$SNAP_DATA_CURRENT/config/security-secret-store/res@" \
+        "$SECRET_STORE_CONFIG"
+fi
+
+SYS_MGMT_AGENT_CONFIG="$SNAP_DATA/config/sys-mgmt-agent/res/configuration.toml"
+if [ -f "$SYS_MGMT_AGENT_CONFIG" ]; then
+    sed -i \
+        -e "s@$SNAP_DIR/[0-9]*/bin/sys-mgmt-agent-snap-executor.sh@$SNAP_CURRENT/bin/sys-mgmt-agent-snap-executor.sh@" \
```
tonyespy: Wasn't the executor added after Delhi was released?

snap/hooks/install
```diff
@@ -21,15 +21,24 @@ esac
 export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/$MULTI_ARCH_PATH:$SNAP/usr/lib/$MULTI_ARCH_PATH"
 export LD_LIBRARY_PATH=$SNAP_LIBRARY_PATH:$LD_LIBRARY_PATH
 
+# get the values of $SNAP_DATA and $SNAP using the current symlink instead of
+# the default behavior which has the revision hard-coded, which breaks after
+# a refresh
+SNAP_DATA_CURRENT=${SNAP_DATA/%$SNAP_REVISION/current}
+SNAP_CURRENT=${SNAP/%$SNAP_REVISION/current}
```
tonyespy: I couldn't find any instances of $SNAP being used in the delhi configurations files?

snap/hooks/post-refresh
anonymouse64: Yes, you're right I misapplied the commit from edinburgh where I originally wrote this commit. I will remove it.

snap/hooks/install
anonymouse64: Yes, again this was from my misapplied commit. $SNAP only needs to be replaced with the system management agent executor, which as you pointed out is an Edinburgh feature

anonymouse64:
@tonyespy no, the tests also test refreshing from both the stable and delhi/stable channels, see `test-refresh-services.sh` and `test-refresh-config-paths.sh` in the checkbox git repo

Issue #1326, opened 2019-05-09
Issues with core-data ReadMaxLimit


anonymouse64:
There are a few issues with the ReadMaxLimit configuration.toml config setting:

* 100 is way too low of a default it should be increased to something realistic
* after the limit is reached, core-data shouldn't fail to respond to GET endpoints like `/api/v1/reading` with `Error, exceeded the max limit as defined in config` - rather it should fail on adding new things with POST endpoints, but still allow accessing data that is in core-data with GET requests

tsconn23:
- 100 is an arbitrary value and can be reconfigured by an end user at deployment. It's meant to provide an example of functionality. If we want to change the default, that's fine but this is the first time I've heard a complaint given that it's an example.
- The idea of the `ReadMaxLimit` is to provide a page of results, not to enforce a capacity limit with regard to ingestion. The way I think of it, we could get rid of the error and simply return a number of results via GET that is capped at the `ReadMaxLimit` value. It should not limit our ability to ingest events. It has never been related to ingestion, only querying.

tobiasmo1:
I ran into this error on my first or second EdgeX query for core-data events, support-log entries and similar GET requests. So I understand Ian's perception that the default value is low. 
Ex: Attempt to get > 100 events and it will also error:
http://localhost:48080/api/v1/event/device/<_devicename_>/200

At first it may seem odd that this represents a hard limit for what is perceived as an explicit request for # elements; e.g., why does core-data not simply honor the requested 200 in this case. However, the server institutes such constraints to avoid spikes in I/O made by serving client requests that can potentially result in intentional/unintended denial of service.

I've seen similar handling by web servers in the data center, often controlling this by limiting the permitted Content-Length (# bytes) of a response; e.g., [IIS7](https://support.microsoft.com/en-us/help/942074/error-message-when-you-visit-a-web-site-that-is-hosted-on-a-server-tha) uses `maxAllowedContentLength` and limits to 30MB by default. Pointing out that size of payload controlled by # elements may have significant variance; hmm, could readings with string value be > 1MB each?

Instead of changing the default value, should we make it more obvious [where/how a person updates](https://github.com/edgexfoundry/edgex-go/blob/9a89dd08bf0d3cc841851472a5d4626c380280f5/cmd/core-data/res/configuration.toml#L16) the configuration (improve the error message)?

anonymouse64:
I think there are a few things here:

[...]

PR #1330, opened 2019-05-10
many: rename ReadMaxLimit to MaxResponseObjectLength, set default to 50k

anonymouse64:
This patch fixes part of what is discussed in issue #1326, namely that the name of the configuration item isn't clear and that the default is too low.

Now, the config item's name more accurately reflects what it is limiting, and
setting it higher helps with deploying results, where you can easily get
more than 100 items in a few seconds with the example device services.

I set the default here to 50000, but can easily change it if folks think that's too high, but as I mentioned on the issue.


tsconn23:
I personally don't know about the name `MaxResponseObjectLength`. It's not clear to me what that means. If I were someone without much history on this, it could mean the maximum size of a single object in the response...or it could mean the maximum size in bytes of the response. I would propose something like `MaxResultCount`. @jduranf @mhall119 Do you have an opinion?

anonymouse64:
How about `MaxResponseObjectsCount`? Then `Objects` is plural implying multiple objects in a response, and Count also suggests this

tobiasmo1:
My 2 cents. I agree using length in the name implies it limits byte size of object or overall response itself, and MaxResponseObjectsCount is an improvement. I wish it was simpler name like MaxResponseCount.

Are **all** responses to requests that involve database queries governed by this limit (across REST calls to any/all core services)?

anonymouse64:
I agree with @tobiasmo1 that it would be nice to have it shorter, but the issue I see with `MaxResponseCount` is that it doesn't make it clear that the thing we're limiting is the number of things _in a response_, not the number of _responses_.

tsconn23:
> the thing we're limiting is the number of things _in a response_

Results, right? `MaxResultCount`

tobiasmo1:
It does seem `MaxResultCount` is best so far. It characterizes the # limit top level core-contract data model objects returned in response to GET query. MaxRowCount would be too db centric.

anonymouse64:
Yeah I think `MaxResultCount` is better. I'll update the PR with that

tsconn23:
recheck

PR #1331, opened 2019-05-10
Fix Export-Distro Ignoring ^C

### LINKED ISSUE #1307: export-distro fails to be killed with ctrl-c after make run
To reproduce on current master build the binaries with `make build`, DON'T run mongodb (i.e. turn it off or disable it), and do:

```bash
make run
# wait a few seconds for things to try to connect and go into a connection loop
# interrupt with ctrl-c
```

You will continue to see export-distro show errors:

```
^Clevel=WARN ts=2019-05-03T02:05:40.613376466Z app=edgex-sys-mgmt-agent source=main.go:71 msg="terminating: interrupt"
level=WARN ts=2019-05-03T02:05:40.613389706Z app=edgex-support-notifications source=main.go:75 msg="terminating: interrupt"
level=WARN ts=2019-05-03T02:05:40.613503134Z app=edgex-core-command source=main.go:71 msg="terminating: interrupt"
Makefile:70: recipe for target 'run' failed
make: *** [run] Interrupt

me@localhost:~/go/src/github.com/edgexfoundry/edgex-go$ ^C
me@localhost:~/go/src/github.com/edgexfoundry/edgex-go$ ^C
me@localhost:~/go/src/github.com/edgexfoundry/edgex-go$ level=ERROR ts=2019-05-03T02:05:41.527711086Z app=edgex-export-distro source=client.go:28 msg="Error getting all registrations: http://localhost:48071/api/v1/registration. Error: Get http://localhost:48071/api/v1/registration: dial tcp 127.0.0.1:48071: connect: connection refused"
level=INFO ts=2019-05-03T02:05:41.535249365Z app=edgex-export-distro source=registrations.go:287 msg="Waiting for client microservice"
^C
me@localhost:~/go/src/github.com/edgexfoundry/edgex-go$ level=ERROR ts=2019-05-03T02:05:42.539137539Z app=edgex-export-distro source=client.go:28 msg="Error getting all registrations: http://localhost:48071/api/v1/registration. Error: Get http://localhost:48071/api/v1/registration: dial tcp 127.0.0.1:48071: connect: connection refused"
level=INFO ts=2019-05-03T02:05:42.544166385Z app=edgex-export-distro source=registrations.go:287 msg="Waiting for client microservice"
```

tsconn23:
Fix #1307 

In the case where export-client was not running when export-distro was
brought up, distro would get into a loop polling for export-client
connectivity and would ignore a signal to quit from the command line.

### DIFF
--- a/internal/export/distro/registrations.go
+++ b/internal/export/distro/registrations.go
@@ -292,6 +292,9 @@ func Loop() {
 				LoggingClient.Error(fmt.Sprintf("with error: %s", err.Error()))
 			}
 			return
+		case <-processStop:
+			LoggingClient.Info(fmt.Sprintf("received term signal"))
+			return
 		case <-time.After(time.Second):
 		}
 		allRegs, err = getRegistrations()
@@ -311,14 +314,14 @@ func Loop() {
 		select {
 		case e := <-messageErrors:
 			// kill all registration goroutines
-			for k, reg := range registrations {
-				if !reg.deleteFlag {
-					// Do not write in channel that will not be read
-					reg.chRegistration <- nil
-				}
-				delete(registrations, k)
-			}
+			stop(registrations)
 			LoggingClient.Error(fmt.Sprintf("exit msg: %s", e.Error()))
+			return
+
+		case <-processStop:
+			// kill all registration goroutines
+			stop(registrations)
+			LoggingClient.Info(fmt.Sprintf("received term signal"))
 			return
 
 		case update := <-registrationChanges:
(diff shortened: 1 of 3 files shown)


internal/export/distro/registrations.go
```diff
@@ -352,3 +355,13 @@ func Loop() {
 		}
 	}
 }
+
+func stop(registrations map[string]*registrationInfo) {
+	for k, reg := range registrations {
+		if !reg.deleteFlag {
```
anonymouse64: I'm not super familiar with this part of the codebase, but should if the `deleteFlag` is false, then we write something to the channel, then delete it? Or should we just write something to the channel and then continue to the next one? I.e. should this code actually be:

```go
	for k, reg := range registrations {
		if !reg.deleteFlag {
			// Do not write in channel that will not be read
			reg.chRegistration <- nil
                        continue
		}
		delete(registrations, k)
        }
```

tsconn23: Well, there's no testing or documentation on this. It looks harmless as the logic is just cleaning out the registrations map before returning. I'm hesitant to make a change here if there isn't a reported issue leading to it. If a message comes through that `messageErrors` channel above, the service dies, so that's the only path that would come through here.

Issue #1296, opened 2019-04-29
Operating State PUT Requests failing


AnthonyMBonafide:
With the newly added model validation, PUT requests for Operating state in core-command are unexpectedly returning a `400 (Bad Request)`. The reason is the logic first attempts to unmarshal the request as an `admin.UpdateRequest`(see [code here](https://github.com/edgexfoundry/edgex-go/blob/e2651e465714af2da6dbbf37abfa36a7555c727a/internal/core/command/restDevice.go#L104)), then performs the validation. The validation for `admin.UpdateRequest`(see [code here](https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/models/adminstate.go#L51)) will fail since the `adminState` is not provided. Previous to the validation functionality being added, the expected HTTP request body would only have the `operatingState` and not `adminState` information. As I see it, we have a few options:

1. We update the logic to better reflect the fact that we want both admin and operating state specified in the update request. This means possibly updating the request model so that it is one consolidated request struct.
2. Update the logic to handle both admin and operating state updates independently so that either one or both can be submitted in the request and the data will be persisted properly.
3. Add separate URI/APIs for admin and operating state so that the APIs are more deterministic.

I personally am a fan of option 3. I do not like that one URI is possibly processing requests differently based on the request body(admin state takes precedence over operating state). I think it makes more sense to have a seperate URI/APIs for each state and the more top level URI `/api/v1/device/name/:name` accepts all the device data. The question here would be would we expect all the data to be present in the body of the request or would we just update the information that is passed and leave the unspecified data as is.

brandonforster:
#1244 in my reading requires that a single endpoint be used, but I'll be the first to say as the author that I think my code is a good starting point and not a good solution. If we can make the the marshalling more robust I'd prefer that solution but I'm not conceptually opposed to going back to four endpoints.

tsconn23:
There's a fourth option to consider -- It could be useful for the validation logic in the core models to return a specific error type when validation fails, similar to those found [here](https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/clients/types/errors.go). Right now we just throw the basic error from the standard library. A type check could then be used [here](https://github.com/edgexfoundry/edgex-go/blob/a7ba7128f4b2664e47889ac3597732d19c924f7e/internal/core/command/restDevice.go#L105) and since we know there could be two different states being updated, we'd then proceed to the next decode operation if the first fails to validate.

[...]

PR #1333, opened 2019-05-11
fix update of command.OperatingStatus

### LINKED ISSUE #1332: Fix update of command.OperatingState
When user try to update command.OperatingState, the function decoding the status value, first try to decode it as an AdminStatus.This try fails and the function immediately returns and never give a chance to decode it as an OperStatus - > the update of Operating status always fails. 
The fix is just to remove the first [return](https://github.com/edgexfoundry/edgex-go/blob/2be1c3e67619ce95e0187e371a122be82c842a2d/internal/core/command/restDevice.go#L107)

difince:
fix #1332


tsconn23:
This also appears to have fixed #1296

PR #1334, opened 2019-05-12
Return New Notification ID Instead of Slug

### LINKED ISSUE #1329: Post notification API should return id when 202 Accept
###  Description
Currently the post notification/subscription API will return the `slug` value after successfully created a new notification/subscription.

### Expected
The **`id`** value should be returned instead of the **`slug`** as the user already has the slug value, also to be in consistent with other POST API in the EdgeX services.

tsconn23:
Fix #1329

When entities are created in EdgeX the normal practice is to return the
ID of the new record. Previously we were returning the "slug" value
when a notification entry was created, which is something that is provided
by the application creating the notification. Thus it wasn't very
useful. Standardized POST of new notification to return the new ID and
updated the RAML.

### DIFF
--- a/api/raml/support-notifications.raml
+++ b/api/raml/support-notifications.raml
@@ -863,10 +863,10 @@ traits:
           }
     responses:
       202:
-        description: If Alerts and Notifications Service is available, it always returns 202 Accepted to clients with the slug to indicate the Notification has been received.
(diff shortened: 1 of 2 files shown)


cloudxxx8:
Lindsey has verified it

PR #1336, opened 2019-05-13
Transmission resend entries created with 0 CreateDate

### LINKED ISSUE #1315: Created field of transmission has updated "0" after status changing to "ESCALATED"
OS version: ubuntu 18.04
EdgeX version: 1.0.0
Environment details: Docker container
Steps to reproduce:

1. Startup EdgeX
2. Create a subscription.
3. Create a critical notification by following API.
Post /api/v1/notification.

> Method: POST
> URL: http://localhost:48060/api/v1/notification
> {
>     "slug": "notice-0012",
>     "sender": "System Management",
>     "category": "HW_HEALTH",
>     "severity": "CRITICAL",
>     "content": " [ WARNING ] Room's temperature is 103 °F , it's over upper limit 100 °F.",
>     "labels": [
>         "room",
>         "temperature"
>     ]
> }

Actual results: 2 transmisstion records are created with failed status, and the created field displayed correctly. For a while, the status of both records are updated to ESCALATED, then created field is updated to 0
Expected results: The created field should not be changed.

tsconn23:
Fix #1315

If a notification comes in and transmittal to a subscriber fails, a
retry can occur. In this case, the retry transmission record was being
created with a 0 CreateDate value. This was because dates are managed
at the database provider level and the return value on an ADD operation
is simply the ID of the newly inserted record.  In order to fix this,
I had to add a new function to the DBClient and supporting providers
that would fetch the newly inserted Transmission record so that the
date values could be retained for further opertations.

### DIFF
--- a/internal/pkg/db/mongo/notifications.go
+++ b/internal/pkg/db/mongo/notifications.go
@@ -369,6 +369,14 @@ func (mc MongoClient) DeleteTransmission(age int64, status contract.Transmission
 	return mc.deleteAll(bson.M{"modified": bson.M{"$lt": end}, "status": status}, TRANSMISSION_COLLECTION)
 }
 
+func (mc MongoClient) GetTransmissionById(id string) (contract.Transmission, error) {
+	query, err := idToBsonM(id)
+	if err != nil {
+		return contract.Transmission{}, err
+	}
+	return mc.getTransmission(query)
+}
+
 func (mc MongoClient) GetTransmissionsByNotificationSlug(slug string, resendLimit int) ([]contract.Transmission, error) {
 	return mc.getTransmissionsLimit(bson.M{"resendcount": bson.M{"$lt": resendLimit}, "notification.slug": slug})
 }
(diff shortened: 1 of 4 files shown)


internal/pkg/db/mongo/notifications.go
```diff
@@ -369,6 +369,14 @@ func (mc MongoClient) DeleteTransmission(age int64, status contract.Transmission
 	return mc.deleteAll(bson.M{"modified": bson.M{"$lt": end}, "status": status}, TRANSMISSION_COLLECTION)
 }
 
+func (mc MongoClient) GetTransmissionById(id string) (contract.Transmission, error) {
+	query, err := idToBsonM(id)
+	if err != nil {
+		return contract.Transmission{}, err
```
akramtexas: I was initially unsure about the use of this type-less object (`Transmission{}`), but I see what looks (the following)  like a corresponding typecast (i.e. `transmission.ToContract()`) when the type is reified, in the 
func `getTransmissionsLimit(...){...}`:
```
	c = []contract.Transmission{}
	for _, transmission := range transmissions {
		c = append(c, transmission.ToContract())
	}
```
So I think we should be Okay 👍

tsconn23:
@cherrycl @cloudxxx8 Waiting for confirmation that this addresses the issue. If no input by tomorrow, will assume I can merge.

PR #1340, opened 2019-05-14
JSON Event ID Lost When Published to Bus

### LINKED ISSUE #1339: BUG : EventID is Lost When JSON Event Published from Core-Data
While working on the export-distro side of the CBOR effort I found a bug whereby the Event ID is lost in that case of a JSON event being published through ZeroMQ. We receive the event in core-data, it gets saved to the database and the ID is applied to it. Because of the way we handle our message bytes right now, the ID is getting lost. Will fix.

tsconn23:
Fix #1339

It's not feasible to set the bytes for a JSON event when the event is
read into core-data. This is because those bytes won't contain the new
event ID obtained during persistence. Marshal the event at the time of
publication if the Content-Type is application/json.

### DIFF
--- a/internal/core/data/event.go
+++ b/internal/core/data/event.go
@@ -16,8 +16,10 @@ package data
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
+	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 	msgTypes "github.com/edgexfoundry/go-mod-messaging/pkg/types"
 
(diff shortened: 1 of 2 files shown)


tsconn23:
recheck

internal/core/data/event.go
```diff
@@ -236,6 +238,15 @@ func updateEventPushDate(id string, ctx context.Context) error {
...
+	//Re-marshal JSON content into bytes.
+	if clients.FromContext(clients.ContentType, ctx) == clients.ContentTypeJSON {
+		data, err := json.Marshal(evt)
+		if err != nil {
+			LoggingClient.Error(fmt.Sprintf("error marshaling event: %s", evt.String()))
+			return
+		}
+		evt.Bytes = data
+	}
 	evt.CorrelationId = correlation.FromContext(ctx)
```
rsdmike: Do we wanna move this line above the remarshalling in the case that it fails the correlationId will be in the error output?

tsconn23: @rsdmike I will include this with the subsequent export-distro PR so I don't have to re-acquire an approval.

PR #1348, opened 2019-05-16
Add Config-Seed Override to Populate Redis Config

### LINKED ISSUE #1302: Update config-seed help message
Config-seed now has more options than the current message says:

```
$ cmd/config-seed/config-seed --help

Usage: cmd/config-seed/config-seed [options]
Server Options:
    -p, --profile <name>            Indicate configuration profile other than default
    -r, --props <dir>               Provide alternate location for legacy application.properties files
    -c, --cmd <dir>                 Provide absolute path to "cmd" directory containing EdgeX services
Common Options:
    -h, --help                      Show this message
```

Neither `-overwrite`, or `-confdir` are documented in the help message

### LINKED ISSUE #1347: Implement Redis Override in Config-Seed
For the Edinburgh release, we need a way to seed Consul with Redis connectivity configuration via docker-compose. The proposal is to add a cmd line flag such as 

```
--database=redis
-d=redis
```

The default value would be `mongo` for now. If this flag is set then right before config-seed puts the configuration into Consul, the Port and Type properties of the primary database configuration should be set to

```
Port=6379
Type=redisdb
```

Adding this flag will allow us to override the default Docker entrypoint for the config-seed from a docker-compose file using a `command` parameter. ([see here](https://docs.docker.com/compose/compose-file/#command))

Example usage from our security docker-compose file is [here](https://github.com/edgexfoundry/developer-scripts/blob/605ce7474e3372a562427fb7bb28f74a6e79d262/compose-files/security/docker-compose-delhi-0.7.0.yml#L106)

tsconn23:
Fix #1347
Fix #1302

If the new --database flag is provided to the config-seed on startup
and if the value is "redis" then the database configurations will be
overridden prior to Consul population with the Redis type and port.

Also cleaned up help message from command line usage.

### DIFF
--- a/internal/pkg/usage/usage.go
+++ b/internal/pkg/usage/usage.go
@@ -23,7 +23,7 @@ Usage: %s [options]
 Server Options:
     -r, --registry                  Indicates service should use Registry
     -p, --profile <name>            Indicate configuration profile other than default
-    -confdir                        Specify local configuration directory
+    --confdir                        Specify local configuration directory
 Common Options:
     -h, --help                      Show this message
 `
@@ -31,9 +31,14 @@ Common Options:
 var configSeedUsageStr = `
 Usage: %s [options]
 Server Options:
+	-c, --cmd <dir>                 Provide absolute path to "cmd" directory containing EdgeX services
+    -d, --database <name>			Indicate which database to use -- mongo or redis. Mongo is default.
+	-o, --overwrite					Indicates service should overwrite any entries already present in the configuration
     -p, --profile <name>            Indicate configuration profile other than default
     -r, --props <dir>               Provide alternate location for legacy application.properties files
-    -c, --cmd <dir>                 Provide absolute path to "cmd" directory containing EdgeX services
+	--confdir                       Specify local configuration directory
+    
+
 Common Options:
     -h, --help                      Show this message
 `
--- a/cmd/config-seed/main.go
+++ b/cmd/config-seed/main.go
@@ -30,6 +30,7 @@ func main() {
 	var useProfile string
 	var dirCmd string
 	var dirProperties string
+	var useDatabase string
 	var overwriteConfig bool
 
 	flag.StringVar(&useProfile, "profile", "", "Specify a profile other than default.")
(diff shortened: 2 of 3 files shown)


tsconn23:
recheck

tsconn23:
Non-arm verify build has failed twice now with
```
16:37:17 [0m[91mgo: github.com/ugorji/go/codec@v0.0.0-20190316192920-e2bddce071ad: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /go/pkg/mod/cache/vcs/4c4ce012b2736486e99bf427f8d15912b07cd69667f52e5aac2af9a9f3c09a9e: exit status 128:
16:37:17 	error: RPC failed; curl 56 OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 104
16:37:17 	fatal: the remote end hung up unexpectedly
16:37:17 	fatal: early EOF
16:37:17 	fatal: index-pack failed
```

tsconn23:
> Overall the code looks ok, I didn't test it but I think I am -1 on this approach as I think that this is adding too much complexity to config-seed. I would rather see one of the following approaches:
> 
> * Have a single docker container with two sets of configuration files, (one for mongo and one for redis) and in the docker-compose for redis override which directory to use
> * Have a more generic mechanism in config-seed for overriding values of keys in the configuration files that's not specific to redis (but could be used for this redis)

https://github.com/edgexfoundry/edgex-go/issues/1347#issuecomment-493479610

internal/pkg/usage/usage.go
```diff
@@ -23,17 +23,22 @@ Usage: %s [options]
...
 
 var configSeedUsageStr = `
 Usage: %s [options]
 Server Options:
+	-c, --cmd <dir>                 Provide absolute path to "cmd" directory containing EdgeX services
+    -d, --database <name>			Indicate which database to use -- mongo or redis. Mongo is default.
+	-o, --overwrite					Indicates service should overwrite any entries already present in the configuration
     -p, --profile <name>            Indicate configuration profile other than default
     -r, --props <dir>               Provide alternate location for legacy application.properties files
-    -c, --cmd <dir>                 Provide absolute path to "cmd" directory containing EdgeX services
```
lenny-intel: Why removing when this is still in the option defined in main?

internal/pkg/usage/usage.go
```diff
@@ -23,17 +23,22 @@ Usage: %s [options]
...
+	--confdir                       Specify local configuration directory
```
lenny-intel: Where is this option added? Not in the options defined in main.

tsconn23:
Closed with merge of #1359

Issue #1346, opened 2019-05-16
Remove Legacy Endpoints from Core-Command and Core-Metadata


tsconn23:
Discussed during the Core Working Group call 16-May-19. Please see the meeting notes for this session recorded on the[WIki](https://wiki.edgexfoundry.org/display/FA/Core+Working+Group)for more context.

**Core-command**
Removal of the following endpoints
- PUT /device/{id} (Used to set admin/operating State on device)
- PUT /device/name/{name}(Used to set admin/operating State on device)

**Core-metadata**
Removal of the following endpoints
- POST & PUT /command (adds or updates a command)
- DELETE /command/id/{id}(deletes a command by its ID)

Also adjust the RAML for the respective services.

PR #1352, opened 2019-05-17
Remove Legacy Endpoints from Core-Command and Core-Metadata

difince:
The raml files also updated
Issue: https://github.com/edgexfoundry/edgex-go/issues/1346

### DIFF
--- a/internal/core/metadata/rest_deviceprofile.go
+++ b/internal/core/metadata/rest_deviceprofile.go
@@ -362,7 +362,12 @@ func deleteDeviceProfile(dp models.DeviceProfile, w http.ResponseWriter) error {
 		http.Error(w, err.Error(), http.StatusConflict)
 		return err
 	}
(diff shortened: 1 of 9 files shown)


internal/core/metadata/rest_deviceprofile.go
```diff
@@ -362,7 +362,12 @@ func deleteDeviceProfile(dp models.DeviceProfile, w http.ResponseWriter) error {
 		http.Error(w, err.Error(), http.StatusConflict)
 		return err
 	}
-
+	for _, command := range dp.CoreCommands {
+		if err := dbClient.DeleteCommandById(command.Id); err != nil {
+			http.Error(w, err.Error(), http.StatusServiceUnavailable)
+			return err
+		}
+	}
```
tsconn23: Nice. Good catch adding this.

PR #1356, opened 2019-05-17
Adding CBOR support to export-distro

### LINKED ISSUE #1217: Export-Distro: CBOR Event Publication
When an event is [received in export-distro](https://github.com/edgexfoundry/edgex-go/blob/5d2583504758ac23e9d64ed3d763ea3189c02836/internal/export/distro/registrations.go#L334) via the internal pub/sub it is then exported to registered clients. For the purpose of supporting CBOR, this export will need to take the Content-Type into account.

Currently the Event is removed from the MessageEnvelope and then published through the `reg.chEvent` channel for receipt and processing by interested subscribers. This should probably be refactored so that the whole MessageEnvelope is sent through this channel.

On the subscribing end, the handler should know how to unpack the MessageEnvelope and what to do, if anything, with the ContentType. For example, the HTTP publisher will know to create a request and set the ContentType from the MessageEnvelope. The body of the request will be the `Payload` property of the MessageEnvelope.

The [processEvent()](https://github.com/edgexfoundry/edgex-go/blob/5d2583504758ac23e9d64ed3d763ea3189c02836/internal/export/distro/registrations.go#L168) function within export-distro has additional logic for formatting, compression, transformation which we will presumably want to keep. These operations are applied to the event before it is distributed, so unmarshaling the Event itself from the `MessageEnvelope.Payload` using CBOR/JSON as appropriate needs to happen here. In this case, we need to preserve the BinaryValue data which is opposite to issue #1216.

After going through the processEvent operations, the event needs to be remarshaled to the appropriate format for use as the body of the soon to be published message. In the case of an HTTP export, the `MessageEnvelope.ContentType` is used to set the header and the bytes of the marshaled Event comprise the request body. After this, the event's `PushedDate` can be [updated]((https://github.com/edgexfoundry/edgex-go/blob/5d2583504758ac23e9d64ed3d763ea3189c02836/internal/export/distro/registrations.go#L199)) in core-data.

tsconn23:
Fix #1217

- For all channels involved in handling of incoming events via bus,
  the type has been changed to MessageEnvelope to provide ease of access
  to ContentType, Checksum and so-forth
- Added handling of CBOR encoded events. JSON handling continues as
  previously defined.
- CBOR events are not processed in any way. The payload is simply sent
  out to the relevant subscriber, per decisions in Core WG
- Unit tests have been adjusted and extended to process at least one
  CBOR event.
- CBOR events will be sent to the Rules Engine whereupon the RE will
  log an error. This does not cause a crash, just a log message on the
  RE side. Pending decision from @jpwhitemn as to whether RE should be
  filtered from export-distro or possible code change to RE.

### DIFF
--- a/internal/core/data/event.go
+++ b/internal/core/data/event.go
@@ -238,6 +238,7 @@ func updateEventPushDate(id string, ctx context.Context) error {
 func putEventOnQueue(evt models.Event, ctx context.Context) {
 	LoggingClient.Info("Putting event on message queue")
 
+	evt.CorrelationId = correlation.FromContext(ctx)
 	//Re-marshal JSON content into bytes.
 	if clients.FromContext(clients.ContentType, ctx) == clients.ContentTypeJSON {
 		data, err := json.Marshal(evt)
@@ -247,7 +248,6 @@ func putEventOnQueue(evt models.Event, ctx context.Context) {
 		}
 		evt.Bytes = data
 	}
-	evt.CorrelationId = correlation.FromContext(ctx)
 
 	msgEnvelope := msgTypes.NewMessageEnvelope(evt.Bytes, ctx)
 	err := msgClient.Publish(msgEnvelope, Configuration.MessageQueue.Topic)
--- a/internal/export/distro/registrations.go
+++ b/internal/export/distro/registrations.go
@@ -23,8 +23,9 @@ import (
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+	msgTypes "github.com/edgexfoundry/go-mod-messaging/pkg/types"
 
-	"github.com/edgexfoundry/edgex-go/internal/pkg/correlation/models"
+	"github.com/pkg/errors"
 )
 
 const (
@@ -44,7 +45,7 @@ type registrationInfo struct {
 	filter       []filterer
 
 	chRegistration chan *contract.Registration
-	chEvent        chan *models.Event
+	chMessages     chan msgTypes.MessageEnvelope
 
 	deleteFlag bool
 }
@@ -58,7 +59,7 @@ func newRegistrationInfo() *registrationInfo {
 	reg := &registrationInfo{}
 
 	reg.chRegistration = make(chan *contract.Registration)
-	reg.chEvent = make(chan *models.Event)
+	reg.chMessages = make(chan msgTypes.MessageEnvelope)
 	return reg
 }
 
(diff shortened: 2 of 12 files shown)


internal/core/data/event.go
```diff
@@ -238,6 +238,7 @@ func updateEventPushDate(id string, ctx context.Context) error {
 func putEventOnQueue(evt models.Event, ctx context.Context) {
 	LoggingClient.Info("Putting event on message queue")
 
+	evt.CorrelationId = correlation.FromContext(ctx)
```
tobiasmo1: I appreciate this emerging pattern (assuring CorrelationId to potentially include in error log).
To better ensure consistency across data models (fail fast earlier, etc)... should this be enforced when validating schema [to occur during json.Marshal] or would that open a can of worms?

internal/export/distro/registrations.go
```diff
@@ -163,21 +164,41 @@ func (reg *registrationInfo) update(newReg contract.Registration) bool {
 	return true
 }
 
-func (reg registrationInfo) processEvent(event *models.Event) {
-	// Valid Event Filter, needed?
+func (reg registrationInfo) processMessage(msg msgTypes.MessageEnvelope) {
```
tobiasmo1: This method (and others) should be able to confirm the message contains an `event` payload and not simply `application/json` or `application/cbor` type. It seems to me that the `msgTypes.MessageEnvelope` need not be so opaque. Are there plans to add a property to the envelope at a later time?

internal/export/distro/registrations.go
```diff
@@ -163,21 +164,41 @@ func (reg *registrationInfo) update(newReg contract.Registration) bool {
...
+	var err error
+	ctx := context.WithValue(context.Background(), clients.CorrelationHeader, msg.CorrelationID)
+	if msg.ContentType == clients.ContentTypeCBOR {
+		err = reg.handleCBOR(msg, ctx)
+	} else {
+		err = reg.handleJSON(msg, ctx)
```
tobiasmo1: Opportunity to be more strict about expected inputs:
`else if msg.ContentType == clients.ContentTypeJSON {`
`. . . `
`else { // new error 'Unknown content type'`

internal/export/distro/registrations.go
```diff
@@ -187,30 +208,32 @@ func (reg registrationInfo) processEvent(event *models.Event) {
...
-
-	if reg.sender.Send(encrypted, event) && Configuration.Writable.MarkPushed {
-		id := event.ID
-		err := ec.MarkPushed(id, context.Background())
-
-		if err != nil {
-			LoggingClient.Error(fmt.Sprintf("Failed to mark event as pushed : event ID = %s: %s", id, err))
-		}
+	if reg.sender.Send(bytes, ctx) && Configuration.Writable.MarkPushed {
+		return ec.MarkPushed(event.ID, ctx)
```
tobiasmo1: Should we add comment here for consistency with handleCBOR? 
`// JSON content type is not needed when calling back to core-data`

tsconn23: The JSON ContentType is added by the EventClient `ec` if one isn't already specified in the context.

internal/export/distro/registrations.go
tsconn23: We can discuss, but I don't think the intent has ever been to make the MessageEnvelope aware of what it contains. In theory it could contain anything, our only current use case is events.

internal/export/distro/registrations.go
tsconn23: Modified to use switch. Default case handles unexpected `Content-type`

internal/core/data/event.go
tsconn23: I think it depends. Like we're discussing below, in this use case the payload is an event but it could be something else. If over time we find that we're only using this mechanism for one thing, then we can harden it as our discussions warrant.

One thing I have in the back of my mind is the discussion in Seoul about notifying northside partners about changes to the devices in the environment. The Messaging module could be a way to solve that, but I suggest we wait and see.

tobiasmo1: I largely agree. If we wait for CorrelationId to become a required property across a large percentage of contract data models, my concern was that defers a potentially significant # of breaking changes.

internal/export/distro/registrations.go
tobiasmo1: Sounds good. Yes, since it is only use case I agree this is beyond the scope of this issue.

internal/export/distro/registrations.go
tobiasmo1: Ah got it, thanks for the explanation.

internal/export/distro/registrations_test.go
```diff
@@ -203,6 +231,9 @@ func BenchmarkProcessEvent(b *testing.B) {
 	event := models.Event{}
 	event.Device = "dummyDev"
 
+	msg := msgTypes.MessageEnvelope{}
+	msg.Payload, _ = json.Marshal(event)
+
```
lenny-intel: Need to set content type?

tobiasmo1: +1 
There remains an opportunity to add a separate test that intentionally does not set ContentType. Trevor implemented [processMessage ](https://github.com/edgexfoundry/edgex-go/pull/1356/files/c5f9cc9d00e837fbb0269e0bf973f0439e3bd61f#diff-0d5a195fe75fb296358cc16865eb31d3R175) that responds to this condition accordingly.

tsconn23: Added content-type to test.

PR #1360, opened 2019-05-21
Add API for querying transmissions by slug

### LINKED ISSUE #1335: Get Transmission by slug REST API enhancement suggestion
#### Suggest to add  new `/transmission/slug/{slug}/start/{start}/end/{end}/{limit}` API
The `/transmission/slug/{slug}/{limit}` GET API can only fetch the MaxLimit number of transmissions by specific Notification slug, which is unable to fetch large scale transmission data for a given slug.

Suggest to add a new `/transmission/slug/{slug}/start/{start}/end/{end}/{limit}` API so that user can query transmission data for a given slug by time range, to support large scale transmission data query for the specific Notification slug

weichou1229:
Fix #1335 

Add API for querying limited transmissions with specified slug and created date range

### DIFF
--- a/internal/support/notifications/rest_transmission.go
+++ b/internal/support/notifications/rest_transmission.go
@@ -17,6 +17,7 @@ package notifications
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"strconv"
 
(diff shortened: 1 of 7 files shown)


internal/support/notifications/rest_transmission.go
```diff
@@ -90,6 +90,52 @@ func transmissionBySlugHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+	end, err := strconv.ParseInt(vars["end"], 10, 64)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		LoggingClient.Error("Error converting the end to an integer")
+		return
+	}
+	limitNum, err := strconv.Atoi(vars["limit"])
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
tsconn23: @weichou1229 These errors should all be `http.StatusBadRequest` since they would be thrown as the result of a type conversion failure based on incorrect URL values. I see, however, that this change is consistent with the rest of the logic in this controller -- so really all similar cases should return `http.StatusBadRequest`. Will changing this have a disruptive effect on tests?

Separately, please included `err.Error()` in the logged error message. Change error text to 
`fmt.Sprintf("failed to parse start %s %s", vars["start"], err.Error())`

andresrinivasan:
Should the body of `GetTransmissionsByNotificationSlugAndStartEnd` be moved into a new function `getTransmissionsByNotificationSlugAndStartEnd` so that `GetTransmissionsByNotificationSlug` can call it rather than duplicate the code?

```go
func (c Client) GetTransmissionsByNotificationSlugAndStartEnd(slug string, start int64, end int64, limit int) (transmissions []contract.Transmission, err error) {
	conn := c.Pool.Get()
	defer conn.Close()

        return getTransmissionsByNotificationSlugAndStartEnd(conn, slug, start, end, limit)
	}

func getTransmissionsByNotificationSlugAndStartEnd(conn redis.Conn, slug string, start int64, end int64, limit int) (transmissions []contract.Transmission, err error) {
	objects, err := getObjectsByScore(conn, db.Transmission+":slug:"+slug, start, end, limit)
	if err != nil {
		return transmissions, err
	}

	transmissions, err = unmarshalTransmissions(objects)
	if err != nil {
		return transmissions, err
	}

	return transmissions, nil
}

weichou1229:
> Should the body of `GetTransmissionsByNotificationSlugAndStartEnd` be moved into a new function `getTransmissionsByNotificationSlugAndStartEnd` so that `GetTransmissionsByNotificationSlug` can call it rather than duplicate the code

Yes! refactored.

internal/support/notifications/rest_transmission.go
weichou1229: > Will changing this have a disruptive effect on tests?

No, there are no such tests. But you remind me, I need to add this test case.

weichou1229: Changed, please help review.

PR #1393, opened 2019-05-30
Set snap release metadata to fuji

anonymouse64:
This was missed in PR #1319 since the PR was merged after edinburgh was
branched, but for master now we want fuji as the release.

Also set epoch to 2 for fuji to prevent automatic edinburgh -> fuji
upgrades until we have a migration strategy ready that will use an epoch
of 2* (to read epoch 1 and write epoch 2).

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -14,7 +14,7 @@ description: |
 icon: snap/local/assets/edgex-snap-icon.png
 
 # delhi is epoch 0, edinburgh epoch 1, etc.
-epoch: 1
+epoch: 2
 
 # TODO: add armhf here when that's supported
 architectures:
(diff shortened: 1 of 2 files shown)


anonymouse64:
recheck

snap/snapcraft.yaml
```diff
@@ -14,7 +14,7 @@ description: |
 icon: snap/local/assets/edgex-snap-icon.png
 
 # delhi is epoch 0, edinburgh epoch 1, etc.
-epoch: 1
+epoch: 2
```
tonyespy: Would it be worth adding a comment here stating that if it's decided that edinburgh can safely upgrade to fuji, then this should be roll this epoch back to '1'?

Issue #1270, opened 2019-04-19
Store CoreCommands as Denormalized Part of DeviceProfile


tsconn23:
When adding a DeviceProfile, the CoreCommands are stored in a manner similar to a relational database. Rather than remaining part of the DeviceProfile document, they are persisted into their own collection (Mongo) or index (Redis). We need to stop this behavior. CoreCommands should be stored on the DeviceProfile itself and then serve as a template for the commands supported by Devices that inherit from the DeviceProfile.

**MONGO**
Looking at the [AddDeviceProfile](https://github.com/edgexfoundry/edgex-go/blob/713baa536c1563875d9ec61d0096859740ca004d/internal/pkg/db/mongo/metadata.go#L343) function, you can see that Commands are saved first before the DeviceProfile itself. This results in a list of DBRef entries that are associated to the [CoreCommands property](https://github.com/edgexfoundry/edgex-go/blob/713baa536c1563875d9ec61d0096859740ca004d/internal/pkg/db/mongo/models/deviceprofile.go#L95) of the Mongo-specific data model, similar to a foreign key. Note that these commands are the only ones stored in this manner.

Eliminating this relationship will potentially allow us to remove code related to the [commandTransform](https://github.com/edgexfoundry/edgex-go/blob/713baa536c1563875d9ec61d0096859740ca004d/internal/pkg/db/mongo/models/command.go#L24) interface and its implementation in the [metadata MongoClient](https://github.com/edgexfoundry/edgex-go/blob/713baa536c1563875d9ec61d0096859740ca004d/internal/pkg/db/mongo/metadata.go#L885).

**REDIS**
A [similar approach](https://github.com/edgexfoundry/edgex-go/blob/713baa536c1563875d9ec61d0096859740ca004d/internal/pkg/db/redis/metadata.go#L535) is taken in Redis, which simply follows the pattern established in Mongo.

In both cases what we should wind up with once this work is done, given a fresh database, is a DeviceProfile saved in JSON that has all of its related data stored in one document without the need for additional lookups based on a relationship. The current `command` collection (Mongo) or index (Redis) where commands are stored should be empty.

[...]

Issue #1268, opened 2019-04-19
Add Device Information to Persisted Commands


tsconn23:
Currently, as described on the Core Working Group call 18-Apr, commands are saved in their own collection within the database. However these records do not currently contain any information about the devices to which they belong. We need to add this information about the device so that lookups can eventually done by core-command going straight to the database as opposed to calling through core-metadata. Commands are inserted into the database when we add a DeviceProfile.

There are two database integrations we have to support - Redis and Mongo. Their respective approach to persistence is very different. Mongo uses a model [struct type](https://github.com/edgexfoundry/edgex-go/blob/master/internal/pkg/db/mongo/models/command.go), mapped from the contract. Redis does not and is instead persisting the contract representation directly into its store (see [here](https://github.com/edgexfoundry/edgex-go/blob/713baa536c1563875d9ec61d0096859740ca004d/internal/pkg/db/redis/metadata.go#L1259)). Working this issue will require that Redis map the incoming contract.Command from the supplied DeviceProfile to a models.Command that contains Device information.

There are already supporting packages for persistence models in both [Mongo](https://github.com/edgexfoundry/edgex-go/tree/master/internal/pkg/db/mongo/models) and [Redis](https://github.com/edgexfoundry/edgex-go/tree/master/internal/pkg/db/redis/models). In the case of Mongo, we just have to extend the existing model. In the case of Redis, we have to create a new one.

It's a common practice in EdgeX to allow lookups on data according to an entity's ID or Name. I therefore recommend the following properties should be added to the Command struct that is persisted.
```
DeviceId string
DeviceName string
```
We will later be providing lookup capabilities to fetch commands accordingly. When commands are returned to the caller via the contract.Command model, these properties should NOT be included. They are for lookup purposes only.

The real heart of this issue is that Commands should be populated when a Device is provisioned. 
1.) The flow that starts from core-metadata/[AddNewDevice](https://github.com/edgexfoundry/edgex-go/blob/713baa536c1563875d9ec61d0096859740ca004d/internal/core/metadata/rest_device.go#L55) should go get the DeviceProfile for the device being added. 
2.) The DeviceProfile's CoreCommands need to be passed into the [AddDevice()](https://github.com/edgexfoundry/edgex-go/blob/713baa536c1563875d9ec61d0096859740ca004d/internal/core/metadata/interfaces/db.go#L41) DBClient function as a new parameter.
[...]

PR #1395, opened 2019-05-31
Store CoreCommands as Denormalized Part of DeviceProfile

difince:
Command.uuid is still kept

Mongo: commands collection stays empty. All functions that read commands from the database (getallCommmands, getcommandById, getCommmandByName) are changed. This change is temporary one.
It will evolve with https://github.com/edgexfoundry/edgex-go/issues/1268 where another collection of commands will be added.
Redis: nevertheless the commands are saved inside the deviceProfile, all indexes and command data are kept. It will also change with issue 1268

Fix https://github.com/edgexfoundry/edgex-go/issues/1270


internal/pkg/db/mongo/metadata.go
```diff
@@ -931,26 +898,30 @@ func (mc MongoClient) getCommandById(id string) (models.Command, error) {
...
-	var command models.Command
-	if err = s.DB(mc.database.Name).C(db.Command).Find(query).One(&command); err != nil {
+	query := bson.M{"commands.uuid": id}
+	var dp models.DeviceProfile
+	if err := s.DB(mc.database.Name).C(db.DeviceProfile).Find(query).Select(bson.M{"commands": 1}).One(&dp); err != nil {
 		return models.Command{}, errorMap(err)
 	}
-	return command, nil
+	if len(dp.CoreCommands) > 1 {
+		return models.Command{}, fmt.Errorf("not unique result found for command with %v", id)
```
tsconn23: Testing locally, the function is returning from this statement, which then returns a 500.
`GET http://localhost:48081/api/v1/command/ab29bd48-8ce5-4e74-bc5c-0e27031d4d9b`

In order to test I've started all of the edgex-go services and then also started the device-random service in order to populate deviceprofile, device, commands, etc. So that command ID is coming from device-random. When I look at the raw JSON of the device profile, the command ID `ab29bd48-8ce5-4e74-bc5c-0e27031d4d9b` is unique, so something about the above query isn't working.

Blackbox tests also fail
Screen Shot 2019-06-04 at 4 19 03 PM

internal/pkg/db/mongo/metadata.go
```diff
@@ -931,26 +898,30 @@ func (mc MongoClient) getCommandById(id string) (models.Command, error) {
...
-	return mapCommands(commands, err)
+	var dps []models.DeviceProfile
+	if err := s.DB(mc.database.Name).C(db.DeviceProfile).Find(bson.M{"commands.name": n}).Select(bson.M{"commands": 1}).All(&dps); err != nil {
+		return nil, errorMap(err)
+	}
+	commands := make([]models.Command, 0)
+	for _, dp := range dps {
+		commands = append(commands, dp.CoreCommands...)
+	}
+	return mapCommands(commands, nil)
```
tsconn23: This function is similar to the one above and works fine.

internal/pkg/db/mongo/metadata.go
```diff
@@ -391,12 +384,8 @@ func (mc MongoClient) UpdateDeviceProfile(dp contract.DeviceProfile) error {
 
 // Get the device profiles that are currently using the command
 func (mc MongoClient) GetDeviceProfilesByCommandId(id string) ([]contract.DeviceProfile, error) {
-	command, err := mc.getCommandById(id)
-	if err != nil {
-		return []contract.DeviceProfile{}, err
-	}
-
-	dps, err := mc.getDeviceProfiles(bson.M{"commands.$id": command.Id})
+	query := bson.M{"commands.uuid": id}
```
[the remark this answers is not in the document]
difince: Removed

internal/pkg/db/mongo/metadata.go
difince: I have reproduced it from the first time. Thanks
Probably, I have tested it with multiple deviceProfiles that contain only single command .....hm

Issue #1198, opened 2019-03-25
Update Device SDK C Guide with Edinburgh changes


mhall119:
TODO: Identify changes between Delhi and Edinburgh

iain-anderson:
@mhall119  can you have a look at #1397 please

PR #1397, opened 2019-05-31
Update C SDK documentation for Edinburgh

iain-anderson:
This accounts for changes in version names and numbers, device profiles, device locations (protocol properties), device service scheduling (automatic events) and the 'template' example supplied in the SDK.
Addresses #1198


anonymouse64:
The snap jobs here are erroneously required, I've requested they be removed from the Required checks for the edinburgh branch and at that point they should disappear from this PR

PR #1389, opened 2019-05-29
Edinburgh #1319 back-port 

anonymouse64:
This is a back-port of #1319 on top of Edinburgh since we missed the branch cut.

In addition to the changes from #1319 it also:
* Updates all the `source-branch` spec for parts to use `edinburgh` instead of master
* Sets the `grade` to `stable` so we can promote any releases to the `edinburgh/edge` channel to candidate or stable when we are ready

PR #1398, opened 2019-05-31
snap/many: update kong to 1.0.3

### LINKED ISSUE #1303: Update kong to 1.0.3 in the snap
The snap should be updated to match the docker-compose file for the security-api-gateway and use kong 1.0.3. A few outstanding issues to handle when doing so:

* Apparently kong has never worked properly on arm64, silently failing when run with:
```bash
ubuntu@ubuntu:~$ sudo kong migrations up --v --yes --conf /var/snap/kong/current/config/kong.conf
PANIC: unprotected error in call to Lua API (bad light userdata pointer)
```
(more details available in https://github.com/edgexfoundry/blackbox-testing/issues/185)
* Kong now has new dependencies such as a yaml library etc. that need to be installed. For reference, looking at the build tools repo here will be useful: https://github.com/Kong/kong-build-tools/tree/feat/multi-arch
* The version of perl in the snap after we migrate to core18 is different now so we need a better way to handle running perl inside the snap and setting up those environment variables
* Doing the trivial thing and just updating the `source-tag` somehow breaks the symlinks setup in the snap currently such that somewhere along the line kong tries (and fails) to execute:
```
ubuntu@ubuntu:~$ kong
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
ERROR: failed to run command "/build/kong/parts/openresty/install/nginx/sbin/nginx -p /tmp/resty_VzlrgHvfEw/ -c conf/nginx.conf": No such file or directory
```

- [ ] Disable security-api-gateway on arm64, including kong and cassandra
- [x] Use `lua5.1` and `luarocks` deb packages with `stage-packages` instead of building from source
- [x] Fix kong part to not build inside `$SNAPCRAFT_STAGE`
- [x] Debug hard-coded path to nginx when running kong perl module (shown above)
- [x] Dynamically determine perl version before running kong

anonymouse64:
* Upgrade to Kong 1.0.3
* Use `kong migrations bootstrap` instead of `kong migrations up`
* Remove lua and luarocks parts and use the bionic ubuntu archive pkgs
  lua5.1 and luarocks instead
* Remove kong build dependency on openresty part as they no longer need
  to be built after each other
* Build kong inside $SNAPCRAFT_PART_INSTALL
* Upgrade to OpenResty 1.13.6.2 and build with kong's patches
* Replace hard-coded location of nginx in "resty" binary from OpenResty
  to use /snap/edgexfoundry/current based path instead of
  $SNAPCRAFT_PART_INSTALL
* Fix the nginx and openresty symlinks inside openresty's
  $SNAPCRAFT_PART_INSTALL/bin dir to be relative instead of absolute
* Use dpkg --print-architecture instead of uname -m for better
  compatibility when cross compiling
* Add layout for /etc/luarocks to view into $SNAP/etc/luarocks using a
  bind mount such that luarocks doesn't trigger AppArmor denials every
  time it starts up via kong. This path is supposedly configurable with
  a build time flag but I haven't been able to figure out which flag is
  to be used for this.

Fixes #1303 for Fuji
This will need a backport to edinburgh after #1389 is merged

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -43,6 +43,13 @@ passthrough:
       command-chain:
       - snap/command-chain/snapcraft-runner
 
+# kong runs things through luarocks and luarocks expects it's configuration to
+# be located here and we can't override this at runtime, so map what's in 
+# $SNAP to the expected location
+layout:
+  /etc/luarocks:
+    bind: $SNAP/etc/luarocks
+
 # TODO: upgrade to stable before releasing to beta/candidate/stable
 grade: devel
 confinement: strict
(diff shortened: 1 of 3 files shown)


anonymouse64:
Forgot to add code to disable security-api-gateway and kong on arm64... Will push up a new commit implementing this after testing is done

anonymouse64:
Alright this is ready for review again, I added a snippet to disable the security-api-gateway services (which includes kong) on initial install as well as made the kong-daemon script exit 0 when running on arm64. The latter was necessary to work around long standing snapd bug https://bugs.launchpad.net/snapd/+bug/1818306

anonymouse64:
recheck

anonymouse64:
recheck

anonymouse64:
recheck

snap/snapcraft.yaml
```diff
@@ -800,100 +781,93 @@ parts:
     stage-packages:
       - perl
     override-build: |
+      cd $SNAPCRAFT_PART_SRC/bundle
+      # apply patches from openresty-kong-patches
+      for i in $SNAPCRAFT_STAGE/openresty-kong-patches/1.13.6.2/*.patch; do patch -p1 < $i; done
```
tonyespy: This would be easier to read if you split into multiple lines....

anonymouse64: Done

anonymouse64:
Commit message adjusted, let me know if it's still not clear the workarounds we are implementing

PR #1400, opened 2019-06-03
Improve delete event performance

### LINKED ISSUE #1286: Delete Device API causes performance issue when lots of events associated to the Device
When Delete Device API is called, all the associated Events and Readings in core-data are also deleted, but they are deleted one by one.  It causes performance issue if there are lots of events exist.

Delete Events one by one:
https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/data/event.go#L263-L270

Delete Readings one by one, too:
https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/data/event.go#L168-L172

AnthonyMBonafide:
Fix #1286

Relocate the responsibility of cascading Reading deletion when deleting
an Event.

Add a new function to the DBClient interface as well as the Mongo and
Redis implementations which abstract the process of deleting all
readings associated with a Device.

Update the business logic to leverage the newly added function for more
effectively deleting readings associated with the specified device.

The Mongo implementation has gone from O(n) to O(1) where n is the
number of readings associated with an Event. However, the Redis
implementation still remains at O(n) for event deletion process as a
whole where n is the number of readings associated with the event. The
deletion functionality is split into two processing units. One which
renames the reading Redis key, this is functionaly the same as a delete
from the perspective of other clients as they cannot retrieve the
original data. The other process cleans up the lingering data. The API
has been updated to rename the Redis reading keys which are to be
deleted, asynchronously delete the lingering data, and respond to the
user. Therefore, the API executes at O(1), but the deletion process as a
whole executes at O(n). The reason is that the process as a whole
executes at O(n) is because we must retrieve the full object that is
persisted so that we can remove all the data(Sorted sets) that reside in
the data-store. For example, there exists sorted sets for time
created(reading:created), and name(reading:name:{r.Name}) associated
with the reading. Given only the device information we are not be able
to remove the data from Redis(ZREM) without constructing the Redis key
which is compromised of the data which is stored.


AnthonyMBonafide:
The changes should be merged after the updates to Mongo have been merged to eliminate any performance issues related to the lack of indexes on the fields needed by the newly added queries. The related PRs can be found here:

1. https://github.com/edgexfoundry/docker-edgex-mongo/issues/19
1. https://github.com/edgexfoundry/developer-scripts/issues/116

internal/core/data/event.go
```diff
@@ -172,11 +172,10 @@ func deleteEventById(id string) error {
 
 // Delete the event and readings
 func deleteEvent(e contract.Event) error {
-	for _, reading := range e.Readings {
-		if err := deleteReadingById(reading.Id); err != nil {
-			return err
-		}
+	if err := dbClient.DeleteReadingsByDevice(e.Device); err != nil {
```
tsconn23: This should not delete readings for the event's device. The function's purpose is to delete the event indicated by the incoming parameter. If there are ten events from the same device, and the request is to delete the first one, then this logic will wipe out readings for all ten of them. See utilized code paths here, none of them relate to the device.
Screen Shot 2019-06-04 at 5 11 48 PM

internal/core/data/init.go
```diff
@@ -166,7 +166,17 @@ func newDBClient(dbType string) (interfaces.DBClient, error) {
...
+		redisClient, err := redis.NewClient(dbConfig) // TODO: Verify this also connects to Redis
+		if err != nil {
+			return nil, err
+		}
+
+		// Background process for deleting device readings and events.
+		// This only needs to be running for core-data since this is the service responsible for handling the deletion
+		// of events
+		go redisClient.AsyncDeleteReadings(LoggingClient)
+		go redisClient.AsyncDeleteEvents(LoggingClient)
```
tsconn23: Manage these inside of the redisClient itself.

```
// Client represents a Redis client
type Client struct {
    Pool *redis.Pool // A thread-safe pool of connections to Redis
    deleteReadingsChannel chan string
    deleteEventsChannel chan string
}

// Return a pointer to the Redis client
func NewClient(config db.Configuration) (*Client, error) {
    once.Do(func() {
    ...
        currClient = &Client{
            ...
            deleteReadingsChannel = make(chan string)
            deleteEventsChannel = make(chan string)
        }
        go currClient.AsyncDeleteReadings()
        go currClient.AsyncDeleteEvents()
    ...
    }
}
```
You should then close these channels in the RedisClient as part of its `CloseSession` function. Init.go calls `dbClient.CloseSession()` [here](https://github.com/edgexfoundry/edgex-go/blob/63c99229ab73133cb40a9f2e3e8640625d2de825/internal/core/data/init.go#L122) As it is now, the channels are never closed.

internal/pkg/db/redis/data.go
```diff
@@ -14,15 +14,29 @@
...
 )
 
+// deleteReadingsChannel channel used to delete readings asynchronously
+var deleteReadingsChannel = make(chan string, 50)
+var deleteEventsChannel = make(chan string, 50)
+
+const (
+	// Batch size used when pipelining asynchronous deletions
+	AsyncDeleteBatchSize = 1000
+)
```
tsconn23: Get rid of all of these global variables. See comment above w/r/t how the channels can be managed as members of the Redis client. `AsyncDeleteBatchSize` could even be a configuration setting that is fed to Redis.NewClient()

internal/pkg/db/redis/data.go
```diff
@@ -203,6 +217,122 @@ func (c *Client) DeleteEventById(id string) (err error) {
...
+	err = conn.Send("EXEC")
+	deleteEventsChannel <- deviceId
+
+	return len(ids), nil
+}
+
+// AsyncDeleteEvents Handles the deletion of device events asynchronously. This function is expected to be running in
+// a go-routine and works with the "DeleteEventsByDevice" function for better performance.
+func (c *Client) AsyncDeleteEvents(loggingClient logger.LoggingClient) {
+	loggingClient.Debug("Starting background event deletion process")
```
tsconn23: The use of loggingClient is tricky here. Nothing in the Redis client has used logging prior to this PR. Rather than passing it into this method, if we must use it, then it's probably better to hand it to Redis.NewClient and make it a member of the RedisClient itself. Then you don't need the parameter. You'd just call `c.logger.Debug("blah")`

internal/pkg/db/redis/data.go
```diff
@@ -203,6 +217,122 @@ func (c *Client) DeleteEventById(id string) (err error) {
...
+	for {
+		select {
+		case device, ok := <-deleteEventsChannel:
+			if ok {
+				loggingClient.Debug("Deleting event data for device: " + device)
+				startTime := time.Now()
+				c.deleteRenamedEvents(device, loggingClient)
+				loggingClient.Debug(fmt.Sprintf("Deleted events for device: '%s', elapsed time: %s", device, time.Since(startTime)))
+			}
```
tsconn23: See the comment about closing the channel above. If a read happens when the channel is closed, then ok will be false and this method should explicitly return -- which will, in turn, free up resources assigned to the goroutine.

internal/pkg/db/redis/data.go
```diff
@@ -203,6 +217,122 @@ func (c *Client) DeleteEventById(id string) (err error) {
...
+		_ = conn.Send("ZREM", db.EventsCollection, e.ID)
+		_ = conn.Send("ZREM", db.EventsCollection+":created", e.ID)
+		_ = conn.Send("ZREM", db.EventsCollection+":device:"+e.Device, e.ID)
+		_ = conn.Send("ZREM", db.EventsCollection+":pushed", e.ID)
+		if e.Checksum != "" {
+			_ = conn.Send("ZREM", db.EventsCollection+":checksum:"+e.Checksum, 0)
+		}
+
+		queriesInQueue++
+		if queriesInQueue >= AsyncDeleteBatchSize {
```
tsconn23: Suggest `c. AsyncDeleteBatchSize` driven from configuration as mentioned in comment above -- tune-able as performance needs / resources warrant.

internal/pkg/db/redis/data.go
```diff
@@ -458,6 +588,114 @@ func (c *Client) DeleteReadingById(id string) error {
...
+	loggingClient.Debug("Starting background event deletion process")
+	for {
+		select {
+		case device, ok := <-deleteReadingsChannel:
+			if ok {
+				loggingClient.Debug("Deleting reading data for device: " + device)
+				startTime := time.Now()
+				c.deleteRenamedReadings(device, loggingClient)
+				loggingClient.Debug(fmt.Sprintf("Deleted readings for device: '%s', elapsed time: %s", device, time.Since(startTime)))
+			}
```
tsconn23: See the comment about closing the channel above. If a read happens when the channel is closed, then ok will be false and this method should explicitly return -- which will, in turn, free up resources assigned to the goroutine.

internal/pkg/db/redis/data.go
```diff
@@ -458,6 +588,114 @@ func (c *Client) DeleteReadingById(id string) error {
...
+			loggingClient.Error("Unable to marshal reading: " + err.Error())
+		}
+		_ = conn.Send("UNLINK", db.DeletedReadingsCollection+":"+r.Id)
+		_ = conn.Send("ZREM", db.ReadingsCollection, r.Id)
+		_ = conn.Send("ZREM", db.ReadingsCollection+":created", r.Id)
+		_ = conn.Send("ZREM", db.ReadingsCollection+":device:"+r.Device, r.Id)
+		_ = conn.Send("ZREM", db.ReadingsCollection+":name:"+r.Name, r.Id)
+		queriesInQueue++
+
+		if queriesInQueue >= AsyncDeleteBatchSize {
```
tsconn23: Suggest `c. AsyncDeleteBatchSize` driven from configuration as mentioned in comment above -- tune-able as performance needs / resources warrant.

internal/pkg/db/db.go
```diff
@@ -26,7 +26,9 @@ const (
 
 	// Data
 	EventsCollection          = "event"
+	DeletedEventsCollection   = "gc:" + EventsCollection
 	ReadingsCollection        = "reading"
+	DeletedReadingsCollection = "gc:" + ReadingsCollection
```
tsconn23: I suggest putting these constants in the actual db/redis package since they aren't common to all DB platforms.

internal/core/data/init.go
AnthonyMBonafide: Trevor and I discussed these changes via Zoom. We talked about the undesired effects these changes would have. If we implement these changes other services that use the Redis client will also spawn these go-routines which handle the deletion of readings and events, which will do nothing and just sit idle since other services are not responsible for that functionality. We talked about the possibility of refactoring the `NewClient` function to also accept an interface, lets call it `ChannelHandler` which will create the necessary channels and go-routines, there is some challenges and design decisions that will need to happen if we go down that road. I am going to take a stab at implementing what we discussed

andresrinivasan:
I finally caught up on this thread and one thing to consider is all the work being done to avoid doing this in Lua. For the same batch operation, Lua will always be the fastest implementation. The tradeoff is when the Lua script is running, all other commands will be queued so there would be a need to chunk the key deletions.

AnthonyMBonafide:
As I see it, an implementation completely in Lua will not be as effective as you cannot pipeline the Redis calls from a script, which will then result in the client being blocked. On the other hand, we can do a hybrid approach which we have a Lua script which handles the deletion and the client makes "batched" calls to the server by invoking the script with smaller chunks of IDs which to delete. I am assuming the client portion which is doing the invoking would do this in the background to reduce the wait time for the user, which is similar to what is implemented here only minus the Lua script which we can add for the go-routine to leverage for the background deletion. Should we introduce a Lua script for the background deletion process?

andresrinivasan:
You don't need to pipeline from the Lua script because you can't get run any faster than a Lua script; the script and the server are running in the same memory space.

What I'm recommending is investigating a strategy to avoid creating the complexity that I'm seeing arise from using go routines. I'll pull in one of my colleagues and we'll noodle on this side.

tsconn23:
I'm personally wary of using Lua here because it's an additional skillset that will be necessary to support the Redis provider. In either case -- Mongo or Redis -- some knowledge of querying capabilities and syntax are required, but now for Redis we're talking about Lua as well. I'm not aware of anyone outside of our Redis team members who has familiarity with Lua. I'm not sure asking anyone who wants to extend a feature that touches the database layer to potentially learn Lua is a viable option.

AnthonyMBonafide:
I spoke briefly with @andresrinivasan yesterday and it seems like we can merge this and possibly work on improvements in a follow up PR. Based on @tsconn23 previous comment, it seems like we would ideally like to find an improvement that does not introduce a Lua script for the sake of logical complexity.

PR #1403, opened 2019-06-04
Update configuration and device profiles used in the docs

### LINKED ISSUE #1158: Remove ScheduleEvents from docs
Docs search has the following matches:

3.1.3. Metadata
3.2.3. Scheduling
3.6.1. Virtual Device
4.4. Working in a Hybrid Environment
6.2. Modbus - Adding a Device to EdgeX
6.3. SNMP - Adding a Device to EdgeX
6.5. Random Integer Device Service Example
7.7. APIs - Supporting Services - Scheduling

https://docs.edgexfoundry.org/search.html?q=scheduleevent&check_keywords=yes&area=default

Also there is a configuration.toml in the docs which is used by both the Go and C Device SDKs which defines a schedule and scheduled event readings that would need to be updated.

### LINKED ISSUE #1179: Update documentation to reflect Addressable/ProtocolProperties change
The Addressable object has been removed from the device services (#1101) and replaced with ProtocolProperties map. Any documentation that used Addressable in this way will need to be changed to show the use of ProtocolProperties instead.

mhall119:
* Fixes #1158
* Fixes #1179


docs/examples/EdgeX_ExampleMQTT_configuration.toml
```diff
@@ -4,7 +4,7 @@ Port = 49982
 ConnectRetries = 3
 Labels = []
 OpenMsg = "device mqtt started"
-MaxResultCount = 50000
+ReadMaxLimit = 256
```
anonymouse64: This needs to be changed back, the new key name is `MaxResultCount`. Also please don't set the example to something low like 256, the default of 50000 is a reasonable default

docs/examples/EdgeX_ExampleModbus_configuration.toml
```diff
@@ -4,7 +4,7 @@ Port = 49991
 ConnectRetries = 3
 Labels = []
 OpenMsg = "device modbus started"
-MaxResultCount = 50000
+ReadMaxLimit = 256
```
anonymouse64: see previous comment for mqtt example configuration.toml

docs/examples/EdgeX_ExampleMQTT_configuration.toml
mhall119: I pulled this from https://github.com/edgexfoundry/device-mqtt-go/blob/master/cmd/res/example/configuration.toml at @cloudxxx8's recommendation. Cloud, can you confirm whether I should make the change that Ian is suggesting?

anonymouse64: Hmm I suppose this is the configuration.toml for a device service and not for the device services. Perhaps we need to propogate https://github.com/edgexfoundry/edgex-go/issues/1326 to the device service SDKs as well

anonymouse64: I asked about this on the #device-services channel on slack for clarification

mhall119: I've removed this line as recommended by @cloudxxx8

docs/examples/EdgeX_ExampleModbus_configuration.toml
mhall119: I've removed this line as recommended by @cloudxxx8

cloudxxx8:
@mhall119 the modbus.test.device.profile.yml is wrong.  sorry for missing this update.
@weichou1229 please send the correct profile for your Modbus example to Michael

cloudxxx8:
@mhall119 please use this profile for the modbus example
[DENT.Mod.PS6037.profile.yaml.zip](https://github.com/edgexfoundry/edgex-go/files/3279565/DENT.Mod.PS6037.profile.yaml.zip)

mhall119:
Recheck

tsconn23:
recheck

Issue #1410, opened 2019-06-06
Command indexes left over, after deviceProfile is deleted


difince:
# 🐞 Bug Report

### Affected Services

 The issue is located in: Metadata 
Applicable with Redis Database only.
Minor Bug for edinburgh and master branches

### Is this a regression?

 No

### Description and Minimal Reproduction
After Issue: https://github.com/edgexfoundry/edgex-go/issues/1392 have been fixed, the DeviceProfile and other objects now can be deleted, but the _command indexes_ remains forever into the database.  

The expected behavior is after DeviceProfile is deleted, all commands and _their indexes_ to be deleted as well.

How to Reproduce: 
- Create DeviceProfile that contains some commands. 
- Delete the Profile 
- Observe that command indexes  remains into the db.

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍  Your Environment
**Deployment Environment:**

**EdgeX Version:**

**Anything else relevant?**

PR #1411, opened 2019-06-06
Wipe out command indexes when delete DeviceProfile

difince:
Fix: https://github.com/edgexfoundry/edgex-go/issues/1410

When Deleting DeviceProfile clean command indexes too.
Applicable only for Redis database.


tsconn23:
It appears the bug I called out may be addressed by https://github.com/edgexfoundry/edgex-go/pull/1409. Will re-test after merge.

tsconn23:
recheck

PR #1417, opened 2019-06-07
Minor tweaks to work with SecrectClient updates

AnthonyMBonafide:
Updated configuration toml and keys used to obtain database secrets

### DIFF
--- a/internal/core/data/init.go
+++ b/internal/core/data/init.go
@@ -17,12 +17,13 @@ package data
 import (
 	"errors"
 	"fmt"
-	"github.com/edgexfoundry/go-mod-secrets/pkg/providers/vault"
 	"os"
 	"os/signal"
 	"sync"
 	"syscall"
 	"time"
+
+	"github.com/edgexfoundry/go-mod-secrets/pkg/providers/vault"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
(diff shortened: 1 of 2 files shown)


AnthonyMBonafide:
After doing some manual acceptance tests, I found some issues that came up due to last minute cleanup changes I made with the client. Let me know if there are some changes that can be made to the client to make user's interaction more ergonomic.

internal/core/data/init.go
```diff
@@ -231,12 +232,12 @@ func connectAndPollSecrets() error {
 		return err
 	}
 
-	username, err := secretsClient.GetSecret("coredata")
+	username, err := secretsClient.GetSecret("User")
```
tsconn23: Should `User` be a constant exposed from the SecretClient?

internal/core/data/init.go
```diff
@@ -231,12 +232,12 @@ func connectAndPollSecrets() error {
...
 	if err != nil {
 		return err
 	}
 
-	password, err := secretsClient.GetSecret("coredatapasswd")
+	password, err := secretsClient.GetSecret("Passwd")
```
tsconn23: Should `Passwd` be a constant exposed from the SecretClient? (spell it out please)

AnthonyMBonafide: > Should `Passwd` be a constant exposed from the SecretClient? (spell it out please -- Password)

Good suggestion. @tingyuz, is there something I need to do to update the key that was stored in the Vault instance I was testing the client against? Currently the key is `Passwd` for your test instance.

Issue #1273, opened 2019-04-20
When a DeviceProfile is Deleted, Delete Devices and Commands


tsconn23:
The idea is that when a DeviceProfile is deleted, there should not be any lingering devices and their associated commands. In practice, we will probably want to ensure that a DeviceProfile cannot be deleted until all Devices have been deleted through a separate call or process. The effect of cascading a DeviceProfile delete is too significant. 

Still we want this as a requirement.

michaelestrin:
Closed.  Believe this to be addressed by https://github.com/edgexfoundry/edgex-go/pull/1496.

Issue #1272, opened 2019-04-20
Delete Commands When a Device is Deleted


tsconn23:
If a Device is deleted, then its related commands also need to be cleaned up.

Issue #1274, opened 2019-04-20
Implement Persistence Layer in Core-Command


tsconn23:
Once all of the above issues have been addressed, we need to then implement the Persistence layer in the Core-Command service. This will replace the current calls to core-metadata with a direct database call, eliminating what are effectively pass-through hops.

1.) Create an `internal/core/command/interfaces` package. A db.go file inside of that should define the necessary function signatures for the persistence layer. It's possible that these will overlap with functions [already found](https://github.com/edgexfoundry/edgex-go/blob/713baa536c1563875d9ec61d0096859740ca004d/internal/core/metadata/interfaces/db.go#L90) in the core-metadata DBClient. 
** TODO: We can discuss what to do with these when the time comes. **
2.) It will be necessary to add bootstrapping of the appropriate database client as found in [other services](https://github.com/edgexfoundry/edgex-go/blob/713baa536c1563875d9ec61d0096859740ca004d/internal/core/metadata/init.go#L220). This will also require configuration changes to include connection information (host/port) and database type, as well as Registry support.
3.) Functions that [previously utilized](https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/command/device.go) the metadata client to obtain their data should now use the DBClient instance.

The format of the data returned to the external caller of the API should be the same. That is, we preserve the `contract` struct return types.

difince:
@tsconn23 
Currently "device commands" (i call them this way, because they appears when the device is added, based on the device.Profile) are saved in "metadata" database. 

Implementing Persistent Layer in Core-Command service.. does it mean that a new database with name "command" should be added? 

It make sense each service to have each own database, but the way the device commands are created is tidy connected with metadata service - add/ delete device. 
If we add new database, we should add new endpoints for managing device commands in command service(not only reading as it is nowadays).These endpoints will be called from metadata service when the device is added/deleted. (note: these endpoints need to be only for internal use )

tsconn23:
@difince Would recommend for this effort that the existing core-command endpoints call through to the metadata db and the command collection it contains. I agree with all of your points above and if we were building this as a new service, that's how I'd proceed. However, I'd like to minimize impacts to the API for now -- at least until Fuji is out the door. If we decide a more thorough re-org of the responsibilities would make sense for Geneva, we can discuss at that time.

difince:
@tsconn23 
What I just realized is that if we force command service to directly query the metadata database, this means: 
[...]

PR #1423, opened 2019-06-11
Persist Commands containg device info

difince:
PR https://github.com/edgexfoundry/go-mod-core-contracts/pull/110 need to merged first.

Device Profile have commands saved in denormalized way inside the Profile itself.
When Device is added, DeviceProfile commands are used as template for creating for
new commands. For easy, fast search these commands have reference to the device they belong
via deviceName and deviceId properties.
When the device is deleted, these commands are cleaned up. 
Command service does not anymore get commands from device.DeviceProfile.commands
It use a new function in metadata service: restGetCommandsByDeviceId.
It is added via PR https://github.com/edgexfoundry/go-mod-core-contracts/pull/110
Command service will be rewriten and requests optimized once it has its
own persistence layer.

Fix: https://github.com/edgexfoundry/edgex-go/issues/1268, https://github.com/edgexfoundry/edgex-go/issues/1272, https://github.com/edgexfoundry/edgex-go/issues/1273

### DIFF
--- a/docs/getting-started/Ch-GettingStartedSDK-C.rst
+++ b/docs/getting-started/Ch-GettingStartedSDK-C.rst
@@ -12,20 +12,21 @@ To build a device service using the EdgeX C SDK you'll need the following:
  * libmicrohttpd
  * libcurl
  * libyaml
+ * libcbor
 
 You can install these on Ubuntu by running::
 
-    sudo apt install libcurl4-openssl-dev libmicrohttpd-dev libyaml-dev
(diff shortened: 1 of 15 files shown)


difince:
The problem is fixed by adding new metadata endpoint restGetCommandsByDeviceId as well as adding new func in CommandClient interface (go-mod-core-contracts up updated) 
I thought to update command service to use the new device.commands  when working on https://github.com/edgexfoundry/edgex-go/issues/1274. But because of a missing test in blackbox testing (nevertheless the issue for missing  test is opened and  assigned to me) I haven't realized it is an immediate work and without Id, no one could issue a command towards a device.

PR #1431, opened 2019-06-12
Restructured code to use retry logic in all services.

brandonforster:
Also incorporated changes from recent API changes and feedback from Trevor.

### DIFF
--- a/internal/core/metadata/init.go
+++ b/internal/core/metadata/init.go
@@ -29,6 +29,9 @@ import (
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
 	registryTypes "github.com/edgexfoundry/go-mod-registry/pkg/types"
 	"github.com/edgexfoundry/go-mod-registry/registry"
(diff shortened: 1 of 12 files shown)


internal/core/metadata/init.go
```diff
@@ -146,6 +159,28 @@ func initializeConfiguration(useRegistry bool, useProfile string) (*Configuratio
 	return configuration, nil
 }
 
+func connectAndPollSecrets() error {
```
AnthonyMBonafide: It looks like this function and its contents are duplicated in other go files. Is it possible to relocate the function and reuse it rather than duplicating?

brandonforster: Of course, but that's kind of the story of these init.go files that all do 90% of the same thing. We're definitely going deeper in tech debt here with this but I don't think it's invalid.

That being said, I'm open to starting to digging us out. It's just not obvious where it should go because the services are supposed to be "independent."

AnthonyMBonafide: Good point, this is probably a bigger concern regarding where to move all common logic.

PR #1437, opened 2019-06-13
Remove unnecessary switch cases in REST API handlers

### LINKED ISSUE #1435: Notifications code clean-up Switch with a single case
See 

https://github.com/edgexfoundry/edgex-go/blob/master/internal/support/notifications/rest_notification.go
has a bunch of Switch statements with a single case.
Not even a default to fall through and say "Error, invalid method call" or something to that effect.
it just silently calls off.
If we had used gorilla MUX at a higher level a single default reporting invalid rest call could have been reported, because corolla max lets you specify URL, parameters, and access method.

nkstanchev:
Fix #1435.

### DIFF
--- a/internal/support/notifications/rest_notification.go
+++ b/internal/support/notifications/rest_notification.go
@@ -30,47 +30,45 @@ func notificationHandler(w http.ResponseWriter, r *http.Request) {
 		defer r.Body.Close()
 	}
 
-	switch r.Method {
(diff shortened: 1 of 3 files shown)


akramtexas:
@nkstanchev As I began reviewing this PR, I notice that there are some conflicts... Could you please take a look and resolve?

PR #1439, opened 2019-06-13
Sorted events returned by getEventsByDeviceIdLimit.

### LINKED ISSUE #1413: Getting oldest data using Core data API
**http://localhost:48080/api/v1/event/device/{deviceName}/1** is returning the oldest event posted to this device instead of the latest one.

In [core-data.raml](https://github.com/edgexfoundry/edgex-go/blob/master/api/raml/core-data.raml), its written 
"Newest events are at the top of list.", But when fetch 10 events of a device, it gives me the oldest events of the device and that to oldest are at the top.

Is it a bug or an expected behaviour?

brandonforster:
Fixes #1413 

### DIFF
--- a/internal/core/data/event.go
+++ b/internal/core/data/event.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"sort"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
@@ -265,6 +266,10 @@ func getEventsByDeviceIdLimit(limit int, deviceId string) ([]contract.Event, err
 		LoggingClient.Error(err.Error())
 		return nil, err
 	}
+
+	sort.Slice(eventList, func(i, j int) bool {
+		return eventList[i].Created < eventList[j].Created
+	})
 
 	return eventList, nil
 }
--- a/internal/core/data/event_test.go
+++ b/internal/core/data/event_test.go
@@ -473,9 +473,29 @@ func TestGetEventsByDeviceIdLimit(t *testing.T) {
 	reset()
 	myMock := &dbMock.DBClient{}
 
+	testEvent2 := models.Event{
+		ID:     "2",
+		Device: testDeviceName,
+		Readings: []models.Reading{
+			{Id: "1"},
+			{Id: "2"},
+		},
+		Created: testEvent.Created + 2000,
+	}
+
+	testEvent3 := models.Event{
+		ID:     "3",
+		Device: testDeviceName,
+		Readings: []models.Reading{
+			{Id: "1"},
+			{Id: "2"},
+		},
+		Created: testEvent.Created + 3000,
+	}
+
 	myMock.On("EventsForDeviceLimit", mock.MatchedBy(func(deviceId string) bool {
 		return deviceId == "valid"
-	}), mock.Anything).Return([]models.Event{testEvent}, nil)
+	}), mock.Anything).Return([]models.Event{testEvent2, testEvent3, testEvent}, nil)
 
 	dbClient = myMock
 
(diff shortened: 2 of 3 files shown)


brandonforster:
> Is this the only place where sort order is wrong (only events)?

I went narrow with this since this is explicitly what the issue was describing and I wanted to make sure we quickly addressed a known defect, but I'll explore the rest of the API as part of this work. Should I close this PR and come back when I have full results?

> I think we should also add blackbox test to confirm order of events is operating correctly going forward.

Absolutely.

jpwhitemn:
@brandonforster - per the issue #1413 and @tsconn23 comment, what did you see with regard to this issue before submitting the PR?  Can you verify the issue or confirm Trevor's observation?

brandonforster:
> @brandonforster - per the issue #1413 and @tsconn23 comment, what did you see with regard to this issue before submitting the PR? Can you verify the issue or confirm Trevor's observation?

I saw that nothing in our implementation independent code guaranteed that events would be sorted.  From that, my idea with the PR was to have the sorting take place at the business logic level so that whether Mongo or Redis or whatever future implementation returned, we knew it would be sorted. I see that Trevor has verified that our Mongo implementation sorts the returned list, and I haven't run the Redis code locally but I don't see any obvious sorting happening with the caveat that I'm really unfamiliar with it. I'll run Trevor's scenario with Redis tomorrow and see what happens with it.

jpwhitemn:
Thanks @brandonforster - look forward to hear what you find tomorrow.  I want to make sure we are actually fixing a bug during a freeze period.

andresrinivasan:
...

> I see that Trevor has verified that our Mongo implementation sorts the returned list, and I haven't run the Redis code locally but I don't see any obvious sorting happening with the caveat that I'm really unfamiliar with it. I'll run Trevor's scenario with Redis tomorrow and see what happens with it.

For Redis, the implementation is intended to return the events sorted by their creation timestamp. The sorted set `event:device:{id}` is used to create an index back to the event object.

brandonforster:
Here's the output from Redis with a limit of `3`:

```
[
    {
        "id": "34bf1d5c-527f-43b6-b3cb-bb40c7e2e833",
        "device": "Random-Integer-Generator01",
        "created": 1560531048433,
        "origin": 1560531048432959000,
        "readings": [
            {
                "id": "3673d866-788b-4e33-9b9f-82ec1c198766",
                "created": 1560531048433,
                "origin": 1560531048432745000,
                "device": "Random-Integer-Generator01",
                "name": "RandomValue_Int32",
                "value": "427902791"
            }
        ]
    },
    {
        "id": "8a92bced-f3c0-46e6-9b18-aa9369c8db4a",
        "device": "Random-Integer-Generator01",
        "created": 1560531048434,
        "origin": 1560531048433032000,
        "readings": [
            {
                "id": "3ebc00b2-3b24-47f9-9388-21e6665907ab",
                "created": 1560531048434,
                "origin": 1560531048432729000,
                "device": "Random-Integer-Generator01",
                "name": "RandomValue_Int8",
                "value": "-38"
            }
        ]
    },
    {
        "id": "fb91f960-d210-4c1f-9bbc-7e6061373c53",
        "device": "Random-Integer-Generator01",
        "created": 1560531048435,
        "origin": 1560531048432986000,
        "readings": [
            {
                "id": "6ea0dff1-a898-4e9d-a85c-72cda913667e",
                "created": 1560531048435,
                "origin": 1560531048432732000,
                "device": "Random-Integer-Generator01",
                "name": "RandomValue_Int16",
                "value": "31630"
            }
        ]
    }
]
```

And a limit of `1`:
```[
    {
        "id": "34bf1d5c-527f-43b6-b3cb-bb40c7e2e833",
        "device": "Random-Integer-Generator01",
        "created": 1560531048433,
        "origin": 1560531048432959000,
        "readings": [
            {
                "id": "3673d866-788b-4e33-9b9f-82ec1c198766",
                "created": 1560531048433,
                "origin": 1560531048432745000,
                "device": "Random-Integer-Generator01",
                "name": "RandomValue_Int32",
                "value": "427902791"
            }
        ]
    }
]
```
It seems like Redis is returning the lowest creation timestamps and this is a bug. I think my solution is valid because we don't rely on implementation specific details to enforce our API constraints.

jpwhitemn:
Brandon, from our test results, it would suggest no error exists on Redis.  Did you run it also with Mongo?  If so, do we see the same consistent results?  I think Trevor's test suggested there was no error on Mongo either.  Can you confirm Trevor's results?  If so, then I don't see this as a bug and no PR is needed.

brandonforster:
@jpwhite I can confirm Trevor's results and we can close this.

PR #1440, opened 2019-06-13
Adds missing command and fixes format errors

alexcourouble:
I found some formatting errors in the document which broke the indentation and list numbering. 

I also fixed two typos and added a missing command to make it easier for developer to copy and paste each command from this document.

Signed-off-by: Alexandre Courouble <acourouble@vmware.com>


alexcourouble:
@tsconn23 just rebased

PR #1446, opened 2019-06-14
Introduce a shared encode function for all services.

### LINKED ISSUE #1445: Shared encode function
Currently, different services have an encode function which is identical

https://github.com/edgexfoundry/edgex-go/blob/0a656c12f3aab84787b909f781cd57e1c1c6030c/internal/core/data/utils.go#L23-L35

https://github.com/edgexfoundry/edgex-go/blob/0a656c12f3aab84787b909f781cd57e1c1c6030c/internal/core/command/rest.go#L87-L99

I propose to move the encode function definition out of the services into internal/pkg and use it that way

nkstanchev:
Fix #1445.
Currently, the different services have their own identical definition
of the encode function. This commit moves that definition to a
shared utils package in internal/pkg as to remove duplicate code
and improve maintainability.


tsconn23:
recheck

internal/pkg/utils/utils.go
```diff
@@ -0,0 +1,21 @@
+package utils
```
tsconn23: So `utils` isn't a preferred package name. Scroll down to the section on "Bad Package Names" in [this article](https://blog.golang.org/package-names).

```
Packages named util, common, or misc provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused.
```

I would be fine if you just put this function in an `encoding.go` file right in the `internal/pkg` folder. No need for a sub-package. Also add the appropriate copyright block to the top of this new file.

Also, for future reference when you have a package and then a Go file in that package with the same name -- `utils/utils.go` -- this is known as "stutter" and should be avoided.

tsconn23:
recheck

Issue #1458, opened 2019-06-19
Kong in snap doesn't use correct certificate anymore


anonymouse64:
With the recent changes to the security-api-gateway configuration.toml file the following sed expression no longer replaces the tokenpath correctly:

https://github.com/edgexfoundry/edgex-go/blob/9316aae57bf09fc6e089f7042539513ea2da8a54/snap/snapcraft.yaml#L976-L977

Because the upstream edinburgh branch configuration.toml file for security-api-gateway changed to no longer contain the `admin-token.json` we are trying to replace:

https://github.com/edgexfoundry/security-api-gateway/blob/96482b5a8dbee3213916b39e23d37dd61c08bdda/core/res/configuration.toml#L43

This means that `edgexproxy` is unable to load the certificates we generate for Kong from `pkisetup` for Kong to use when hosting HTTPS endpoints

I will send a patch shortly for this, we should include it in Edinburgh

anonymouse64:
There is an additional issue here in that Kong seems to have some kind of race condition when starting up that it sometimes ignores the SNIs setting when set via Form values in a POST. Using JSON seems to fix the issue for me in security-api-gateway, but unfortunately that means we will also need to land a patch to security-api-gateway in addition to #1459

PR #1459, opened 2019-06-19
Fix security-api-gateway config file

anonymouse64:
The upstream configuration.toml default value changed such that this sed expression changing the tokenpath to the correct one for the snap no longer applied, leading to edgexproxy being unable to use the token to load the certificates from vault into Kong for Kong to use when serving HTTPS endpoints.

Fixes: #1458 for Edinburgh, will need a forward port to master as well


anonymouse64:
checkbox

anonymouse64:
recheck

anonymouse64:
Unfortunately, this is only part of the problem, there is also an issue with Kong upstream, which can be worked around by switching from Form values to JSON when POST'ing from edgexproxy in security-api-gateway. That is fixed in https://github.com/edgexfoundry/security-api-gateway/pull/70

tingyuz:
LGTM.

Issue #1450, opened 2019-06-18
Override Configuration Using Environment Variables


michaelestrin:
# 🚀 Feature Request

### Relevant Package

This feature request is for all core services.

### Description

Enable ability to override service configuration using environment variables.  This is the generally accepted mechanism to inject variation when instantiating a container.

My specific use case -- a Helm Chart to support the management of multiple instances of EdgeX in a Kubernetes cluster -- precludes use of the containerized `configuration.toml` files and registry-based configuration.  Each instance of EdgeX must have its own uniquely named Kubernetes service to facilitate inter-instance communication.   Any EdgeX service that communicates with another would need to be updated with that unique instance/service name.

### Describe the solution you'd like

- In internal/pkg/config, implement a reflection-based traversal of an empty interface typed `ConfigurationStruct`.  
- Environment variable keys would be formatted using underscore separators.  For example, `MessageQueue_Host` would refer to core-data's `ConfigurationStruct.MessageQueue.Host` field.
- Environment variables would supersede both file- and registry-based configuration values.

### Describe alternatives you've considered

- Considered and discarded conversion of service-specific `ConfigurationStruct` (required for marshaling toml and registry data) into generic key/value configuration contract/implementation to be used by service to retrieve configuration.

iain-anderson:
Could the configuration structure be retained by replacing
`Clients      map[string]config.ClientInfo`
with
`Clients      ClientStruct`
and
```
type ClientStruct struct {
  MetadataClient  config.ClientInfo
  LoggingClient   config.ClientInfo
// etc
}
```
(and similar for databases)

michaelestrin:
@iain-anderson Indeed.  Verified.  Updated issue description to reflect.

michaelestrin:
Based on Core WG feedback, I modified the proposed solution as follows:

- Support environment variable overrides for config-seed to the `configuration.toml` content it uploads into registry.  An environment variable can be service-specific or generic.  

    A service-specific environment variable override would include the targeted service's name and would only affect configuration for that service; i.e. `core-metadata_Clients_logging_Host=<someAddr>`.  

    A generic environment variable override would not include a service name and would affect configuration across all services config-seed handles; i.e. `Registry_Host=<someAddr>`.

- Support environment variable override of a core service's RegistryInfo fields (i.e. host, port, and type) to allow a containerized service instance to be pointed at an alternate registry location.

See #1461 for additional details.

Issue #1405, opened 2019-06-05
Standartize json response function in Notifications


nkstanchev:
In Notifications Service two different functions are used for json response - encode and encodeWithUTF8. They are used interchangeably without any purpose or benefit.

https://github.com/edgexfoundry/edgex-go/blob/e930d0c25674590ac0bf23ecca9c268fe054f0c8/internal/support/notifications/utils.go#L27-L53

I propose to remove the encodeWithUTF8 function and replace it wherever it is called with the encode function.

PR #1359, opened 2019-05-21
Add redis config files into config-seed container

anonymouse64:
This is my proposal to solve #1347 (in contrast with #1348).

Specifically this PR is different in that in order for the config-seed container to support Redis, we just make copies of all the config files that are currently under `/edgex/cmd` at `/edgex/cmd-redis`, then use some simple shell tools (`toml2json`, `jq` and `json2toml`) to modify some of the configuration.toml files in the `cmd-redis` directory with Redis configuration. 

I originally had thought that the increased total file size would be 9 KB, but I measured it as actually 16 KB, which I still think is very minimal. You can measure this by building the config-seed container and getting the size with:
```
$ echo "$(docker image inspect redis-config-seed --format='{{.Size}}') - $(docker image inspect mongo-config-seed --format='{{.Size}}')" | bc
16968
```
(where `mongo-config-seed` is a local image built from of current master config-seed Dockerfile and `redis-config-seed` is a local image built from this branch's config-seed Dockerfile).

## Justification

I think this PR is better than #1348 for a few reasons:
* A slight increase in complexity of the docker build process offsets a slight increase in complexity of the go code because it has smaller scope
* Less maintenance burden for the config-seed options handling - less options means config-seed is simpler to use and simpler to maintain
* Lets us handle the larger problem for overwriting config keys from the command line using config-seed in a more general way for Fuji or Geneva

## Testing

To test this, I first modified the redis docker-compose file for delhi to delete the device-virtual and portainer containers, and then reference the local `1.0.0-dev` tags of the containers with this patch to the developer-scripts repo:

<details>

```patch
diff --git a/compose-files/docker-compose-redis-delhi-0.7.1.yml b/compose-files/docker-compose-redis-delhi-0.7.1.yml
index e75f299..1013ad9 100644
--- a/compose-files/docker-compose-redis-delhi-0.7.1.yml
+++ b/compose-files/docker-compose-redis-delhi-0.7.1.yml
@@ -57,7 +57,7 @@ services:
       - volume  
 
   config-seed:
-    image: edgexfoundry/docker-core-config-seed-go:0.7.1
+    image: edgexfoundry/docker-core-config-seed-go:1.0.0-dev
     container_name: edgex-config-seed
     hostname: edgex-core-config-seed
     networks:
@@ -103,7 +103,7 @@ services:
       - volume
 
   logging:
-    image: edgexfoundry/docker-support-logging-go:0.7.1
+    image: edgexfoundry/docker-support-logging-go:1.0.0-dev
[...]

PR #1461, opened 2019-06-19
Override Configuration Using Environment Variables

michaelestrin:

#1450

### DIFF
--- a/internal/core/command/init.go
+++ b/internal/core/command/init.go
@@ -118,6 +118,7 @@ func initializeConfiguration(useRegistry bool, useProfile string) (*Configuratio
 	if err != nil {
 		return nil, err
 	}
+	configuration.Registry = config.OverrideFromEnvironment(configuration.Registry)
 
 	if useRegistry {
 		err = connectToRegistry(configuration)
--- a/internal/pkg/config/environment.go
+++ b/internal/pkg/config/environment.go
@@ -1,0 +1,42 @@
+/*******************************************************************************
+ * Copyright 2019 Dell Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package config
+
+import (
+	"os"
+	"strconv"
+)
+
+const (
+	envKeyPrefix = "edgex_registry_"
+	envKeyHost   = envKeyPrefix + "host"
+	envKeyPort   = envKeyPrefix + "port"
+	envKeyType   = envKeyPrefix + "type"
+)
+
+func OverrideFromEnvironment(registry RegistryInfo) RegistryInfo {
+	if env := os.Getenv(envKeyHost); env != "" {
+		registry.Host = env
+	}
+	if env := os.Getenv(envKeyPort); env != "" {
+		if v, err := strconv.ParseInt(env, 10, 0); err == nil {
+			registry.Port = int(v)
+		}
+	}
+	if env := os.Getenv(envKeyType); env != "" {
+		registry.Type = env
+	}
+	return registry
+}
--- a/internal/pkg/config/environment_test.go
+++ b/internal/pkg/config/environment_test.go
@@ -1,0 +1,80 @@
+/*******************************************************************************
+ * Copyright 2019 Dell Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package config
+
+import (
+	"github.com/stretchr/testify/assert"
+	"os"
+	"strconv"
+	"testing"
+)
+
+const (
+	envValueString = "envValue"
+	envValueInt    = 123456789
+
+	defaultHostValue = "defaultHost"
+	defaultPortValue = 987654321
+	defaultTypeValue = "defaultType"
+)
+
+func initializeTest(t *testing.T) RegistryInfo {
+	os.Clearenv()
+	return RegistryInfo{
+		Host: defaultHostValue,
+		Port: defaultPortValue,
+		Type: defaultTypeValue,
+	}
+}
+
+func TestEnvVariableUpdatesRegistryInfo(t *testing.T) {
+	var tests = []struct {
+		name              string
+		envKey            string
+		envValue          string
+		expectedHostValue string
+		expectedPortValue int
+		expectedTypeValue string
+	}{
+		{"host", envKeyHost, envValueString, envValueString, defaultPortValue, defaultTypeValue},
+		{"port", envKeyPort, strconv.Itoa(envValueInt), defaultHostValue, envValueInt, defaultTypeValue},
+		{"type", envKeyType, envValueString, defaultHostValue, defaultPortValue, envValueString},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			registryInfo := initializeTest(t)
+			if err := os.Setenv(test.envKey, test.envValue); err != nil {
+				t.Fail()
+			}
+
+			registryInfo = OverrideFromEnvironment(registryInfo)
+
+			assert.Equal(t, registryInfo.Host, test.expectedHostValue)
+			assert.Equal(t, registryInfo.Port, test.expectedPortValue)
+			assert.Equal(t, registryInfo.Type, test.expectedTypeValue)
+		})
+	}
+}
+
+func TestNoEnvVariableDoesNotUpdateRegistryInfo(t *testing.T) {
+	registryInfo := initializeTest(t)
+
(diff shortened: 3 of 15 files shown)


internal/core/command/init.go
```diff
@@ -143,6 +143,8 @@ func initializeConfiguration(useRegistry bool, useProfile string) (*Configuratio
 		}
 	}
 
+	config.OverwriteFromEnvironment(configuration)
```
tsconn23: I admit I haven't looked at this really closely -- been taken up with other things. What I have looked at doesn't show a way to toggle this behavior. Maybe just point me in the right direction. We definitely don't want to do this by default.

michaelestrin: The behavior can't be toggled per-se.  

If you set an environment variable named after a valid configuration field, the variable's value will overwrite the value already in configuration.  If you don't set appropriately named environment variables, the configuration will not be overridden.

Not sure if the concern is overhead to run the function.  It runs once during start up and time consumption is negligible.

tonyespy:
A couple of comments...

* During the last TSC F2F we said we weren't going to be working on any k8s related features... 

* There's a single commit for ~500 lines of diff, and the message just says "Override Configuration Using Environment Variables".  There's no commit message body which explains why this change was done. The body just contains the string "#1405". I think it's totally fine to such shorthand in github PR descriptions, however someone looking at this git repository on a local machine doesn't provide anyone any real context. I would suggest in commit messages that we always use full URLs for links to github issues.

 * While "."s in environment variables are technically legal, most shells don't support them

 * If I understand the proposal, environment vars take precedence over both local configuration and the registry. How will this impact the new work happening in the SMA for configuration?

akramtexas:
@tonyespy Good point, regarding your mentioning `"If I understand the proposal, environment vars take precedence over both local configuration and the registry. How will this impact the new work happening in the SMA for configuration?"` ➡️ @michaelestrin had reached out to me (in general, as in where the work per this PR "interfaces" with `System Management`). So yes, we need to have a conversation, going forward 👍

michaelestrin:
> There's no commit message body which explains why this change was done.

Absolutely correct.  Pushed reworded commit message.

lenny-intel:
I see you have refactored the config structs to remove maps for this to work via reflection for those settings. Why not have the override code handle maps via the data contained in them rather than ignoring them or refactoring them out?

michaelestrin:
> Why not have the override code handle maps via the data contained in them rather than ignoring them or refactoring them out?

Reflection doesn't support in-place modification of map elements (or their content).  Support for maps would have made the code significantly more complex.

@lenny-intel

lenny-intel:
@michaelestrin , I wasn't suggesting modification of the maps, just traversing them based on the contents. I think that complexity is worth not having the refactor out all the use of maps in the configs or ignoring them when left in the config.

michaelestrin:
@lenny-intel My original implementation did traverse them but I found that reflection doesn't support in-place modification of map elements _(**or their content**)_.   That is, I found it wasn't possible -- for instance -- to modify by reflection a string field in a sub-struct that was a field on a map element's struct.  While I could recreate the overall service-specific ConfigurationStruct and return it, I didn't feel the complexity involved in a generic solution was worth the effort.

FWIW, based on the Core WG feedback I received yesterday I've moved on to a different implementation -- one that does not use reflection -- and will be updating this PR with a detailed description shortly.

michaelestrin:
Based on the feedback received in the June 20th Core WG call, I modified my approach to solving this issue.  In addition to the feedback, I was influenced by the following:

1. The project's published Docker images use the registry by default.

1. The existing `Endpointer` contract and implementation (edgexfoundry/go-mod-core-contracts@v0.1.0/clients/endpointer.go) populates a service's address from configuration.  It then periodically polls the registry to retrieve that service's address and updates its local copy.  There is no clean way to extend the existing implementation to support environment variable overrides of service addresses as the service-instance level.

1. The existing `listenForConfigChanges()` implementation across the core services which listens for and updates the service's `ConfigurationStruct.Writable` struct when changes occur.  There is no clean way to extend the existing implementation to support environment variable overrides of `Writable` content changes made via the registry and pushed to a running service.  

1. Future system management agent work is planned to enable system management to push updated configuration to an individual service.  This is a variation of the problems described by the prior two points.  

1. Outside of a development environment, the registry service is likely to be a key system component. 

This PR has been updated to:

- Support environment variable overrides for config-seed to the `configuration.toml` content it uploads into registry.  An environment variable can be service-specific or generic.  

    A service-specific environment variable override would include the targeted service's name and would only affect configuration for that service; i.e. `core-metadata_Clients_logging_Host=<someAddr>`.  

    A generic environment variable override would not include a service name and would affect configuration across all services config-seed handles; i.e. `Registry_Host=<someAddr>`.

- Support environment variable override of a core service's RegistryInfo fields (i.e. host, port, and type) to allow a containerized service instance to be pointed at an alternate registry location.

@tsconn23 @tonyespy @lenny-intel @akramtexas

lenny-intel:
@michaelestrin , ahh not modify the map, but the field of the struct via the  map. Got it! :thumbsup:

internal/system/agent/init.go
```diff
@@ -129,6 +129,7 @@ func initializeConfiguration(useRegistry bool, useProfile string) (*Configuratio
 	if err != nil {
 		return nil, err
 	}
+	configuration.Registry = config.OverwriteFromEnvironment(configuration.Registry)
```
akramtexas: Consider renaming the func `OverwriteFromEnvironment()` to `OverrideFromEnvironment()`.

akramtexas:
> ...
> 4. Future system management agent work is planned to enable system management to push updated configuration to an individual service.  This is a variation of the problems described by the prior two points.
> ... 
> @tsconn23 @tonyespy @lenny-intel @akramtexas

Referring to a point noted in the PR (i.e. `4. Future system management agent work is planned to enable system management to push updated configuration to an individual service. This is a variation of the problems described by the prior two points.`), please note that the SMA _will_, on occasion, though not always, be called upon to update configuration for _multiple_ service (e.g. update the setting with the key "Logging.EnableRemote" to take on the value "false").

internal/pkg/config/environment.go
```diff
@@ -0,0 +1,42 @@
...
+
+package config
+
+import (
+	"os"
+	"strconv"
+)
+
+const (
+	envKeyPrefix = "edgex_registry_"
```
akramtexas: - As far as interacting with the Registry is concerned, please note that the SMA will (in all likelihood) use the go-mod-registry PutConfigurationValue() since it will not have the services config struct the is passed to `PutConfiguration`. And since the SMA will go directly to the Registry, rather than the service, the need exists for knowing the proper path in Consul for the (given) key.
- The Registry Client is created for a specific service with that services service key. That takes care of the base path.
- So we need to know the (trailing) `sub-path` for the specific configuration value.
- Here, in the context of the logic encoded in `environment.go`, would this logic have responsibility for handling that aspect? (i.e. "figure out the (trailing) `sub-path` for the specific configuration value").

michaelestrin:
> SMA _will_, on occasion, though not always, be called upon to update configuration for _multiple_ service

I see this as an orthogonal concern.  

config-seed overrides are there for system start up.  Once the system is underway, changes to the configuration stored in the registry (if it's being used) are outside the scope of the changes in this PR.

Do you concur @akramtexas ?

akramtexas:
> > SMA _will_, on occasion, though not always, be called upon to update configuration for _multiple_ service
> 
> I see this as an orthogonal concern.
> 
> config-seed overrides are there for system start up. Once the system is underway, changes to the configuration stored in the registry (if it's being used) are outside the scope of the changes in this PR.
> 
> Do you concur @akramtexas ?

@michaelestrin That was very helpfully-articulated. Yep, makes sense, and I concur 👍

internal/pkg/config/environment.go
michaelestrin: The changes in this PR override the settings in a configuration.toml file.  They know nothing of the mapping from configuration.toml to registry.  The referenced source `internal/pkg/config/environment.go` simply host, port and type in a RegistryInfo struct that's contained within a service's ConfigurationStruct struct.

akramtexas: That makes sense 👍 Thanks for confirming. I'm good with it.

anonymouse64:
> * Support environment variable overrides for config-seed to the `configuration.toml` content it uploads into registry.  An environment variable can be service-specific or generic.

This kind of change is exactly what I was referring to in my second point here: https://github.com/edgexfoundry/edgex-go/pull/1348#pullrequestreview-238957299

This should enable us to drop the duplicated Redis configuration files from the docker image (see #1347 and #1359) and just override the necessary configuration.toml changes to use Redis with environment variables specified in the docker-compose.yml. 

I haven't reviewed this code at all (and don't intend to), I just wanted to chime in that I am +1 on this approach

akramtexas:
@michaelestrin I want to spend just a bit more time playing with the updated code in this PR, run the unit tests, etc., and will then approve the PR. Overall, so far, LGTM 👍

internal/system/agent/init.go
michaelestrin: Done.

michaelestrin:
Updated to make @akramtexas suggested function name change.  Also corrected two instances where OverrideFromEnvironment() should have been called but wasn't.

internal/pkg/config/environment_test.go
```diff
@@ -0,0 +1,96 @@
...
+	assert.Equal(t, registryInfo.Port, defaultPortValue)
+}
+
+func TestNoTypeEnvVariableDoesNotUpdateRegistryInfo(t *testing.T) {
+	registryInfo := initializeTest(t)
+
+	registryInfo = OverrideFromEnvironment(registryInfo)
+
+	assert.Equal(t, registryInfo.Type, defaultTypeValue)
+}
```
tsconn23: @michaelestrin Question -- Have you looked at table driven tests at all in Go? It looks like these methods could utilize that technique, and you could thereby collapse these methods into one with a table containing a series of testable criteria and mocks. I think it's a little more concise. Example: https://github.com/edgexfoundry/edgex-go/blob/2cd11714e75ea178d29e77b27c6dfbe4b8618a15/internal/core/metadata/operators/device/add_test.go#L28

michaelestrin: Recognize they're idiomatic.  I shy away from them because they tend to be harder for me to read/grok than the old school "one method, one test" approach.  Refactored per your implied request.

tsconn23: Yeah, I had that same cognitive clash. Look at them enough and you get used to it though.

PR #1464, opened 2019-06-20
Use constants in Logging service API

### LINKED ISSUE #1463: Use constants in the API of Logging Service
The API definition of Logging service uses strings instead of constants

https://github.com/edgexfoundry/edgex-go/blob/5052c17bac2189cdb9e67c3b37c177105c7a05d7/internal/support/logging/server.go#L259-L272

nkstanchev:
Fix #1463.


tsconn23:
recheck

PR #1468, opened 2019-06-20
Use constants in System Agent API

### LINKED ISSUE #1467: Use constants in the API of System Agent service
The API definition of System Agent service uses strings instead of constants

https://github.com/edgexfoundry/edgex-go/blob/5052c17bac2189cdb9e67c3b37c177105c7a05d7/internal/system/agent/router.go#L36-L46

nkstanchev:
Fix #1467.

### DIFF
--- a/internal/system/agent/const.go
+++ b/internal/system/agent/const.go
@@ -1,0 +1,23 @@
+/*******************************************************************************
+ * Copyright 2019 VMware Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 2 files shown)


nkstanchev:
@michaelestrin 
I started incorporating the use of constants in the APIs of the different services after a discussion on this with @tsconn23.
I agree that at this point it doesn't make that much sense but it can maybe be useful for future endpoints. @tsconn23 What is your opinion on this?

AnthonyMBonafide:
Even though we are not using those strings in multiple places I think it is nice to have a consistency with how we are building the URIs. Also, if we decide to add more APIs then the work is already done.

PR #1470, opened 2019-06-20
Fix security-api-gateway config file (for master)

### LINKED ISSUE #1458: Kong in snap doesn't use correct certificate anymore

anonymouse64:
Forward-port to master of #1459 

Fixes #1458 for master/Fuji

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -967,13 +967,13 @@ parts:
       cd $SNAPCRAFT_PART_SRC
       make build
 
-      install -DT core/edgexproxy "$SNAPCRAFT_PART_INSTALL/bin/edgexproxy"
+      install -DT cmd/edgexproxy/edgexproxy "$SNAPCRAFT_PART_INSTALL/bin/edgexproxy"
 
       # also modify the configuration file to use "/" instead of "\" in the file
       # and set the snis by default to be localhost so that https works when using the certificate authority 
       # file generated and a url like "https://localhost:8443/{svc}" etc. for kong
       mkdir -p $SNAPCRAFT_PART_INSTALL/config/security-api-gateway/res
-      cp core/res/configuration.toml $SNAPCRAFT_PART_INSTALL/config/security-api-gateway/res/configuration.toml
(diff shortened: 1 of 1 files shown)


tsconn23:
recheck

anonymouse64:
recheck

anonymouse64:
Ah there is actually a breakage here with the snap (but nothing to do with this PR):
```
16:41:03 install: cannot stat 'core/edgexproxy': No such file or directory
```

I assume this is because of the newly merged changes to security-api-gateway on master.

@tsconn23 @mkbhanda how soon is the transition to edgex-go going to happen? If it's not going to happen for a few weeks I'll fix it here in the meantime, but if it's gonna switch soon I'll just let it be  because we will need to rewrite it for the transition to edgex-go anyways

mkbhanda:
I encountered this issue too blackbox testing .. was wondering if I pulled before merge, but pulled after Jim sent out message of release.
Tingyu

tsconn23:
@anonymouse64 
> @tsconn23 @mkbhanda how soon is the transition to edgex-go going to happen?
Could take some time, Ian. Work hasn't even started yet.

anonymouse64:
This PR now handles the new master branch layout of security-api-gateway and, in addition to fixing the config file for #1458 will also unbreak all the currently broken stage-snap master jobs in Jenkins.

anonymouse64:
recheck

PR #1496, opened 2019-06-28
Persist Commands containg device info

difince:
Currently the Device Profile save its commands in denormalized way.
When Device is added, DeviceProfile's commands are used as template for creating for
new commands. For easy and fast search these commands have reference to the device they belong
via deviceName and deviceId properties.
When the device is deleted, these commands are cleaned up.

New metadata endpoint added: restGetCommandsByDeviceId.
Core-Command use this new endpoint to retrieve Commands by deviceId.
Core-Command service will be rewritten and requests optimized once it has its
own persistence layer.

Fix: https://github.com/edgexfoundry/edgex-go/issues/1268, https://github.com/edgexfoundry/edgex-go/issues/1272, https://github.com/edgexfoundry/edgex-go/issues/1273


internal/pkg/db/mongo/metadata.go
```diff
@@ -146,8 +145,13 @@ func (mc MongoClient) AddDevice(d contract.Device) (string, error) {
 	}
 
 	mapped.TimestampForAdd()
+	if err = errorMap(col.Insert(mapped)); err != nil {
+		return id, err
+	}
 
-	return id, errorMap(col.Insert(mapped))
+	//add commands based on DeviceProfile.CommandProfile
+	err = mc.addCommands(commands, mapped.Uuid, mapped.Name)
```
AnthonyMBonafide: It looks like if we have a problem persisting the commands in the `addCommands` function the database will be in an undesirable state since we will have already persisted the device, but not the commands. Should we add some additional protection against this?

internal/core/metadata/rest_command.go
```diff
@@ -82,3 +82,28 @@ func restGetCommandByName(w http.ResponseWriter, r *http.Request) {
...
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+		}
+
+		LoggingClient.Error(err.Error())
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	json.NewEncoder(w).Encode(commands)
+
```
tsconn23: Given that we are in the process of extending unit test coverage through the use of what we're calling the "operator" pattern, I'd like to see this new endpoint refactored according to that pattern. Otherwise it will be tech debt we have to go back and address. The check for `db.ErrNotFound` should be inside the operator and then the router layer should perform type checking to determine the appropriate response code. This example should illustrate what I'm looking for.
https://github.com/edgexfoundry/edgex-go/blob/a478f95df2952767d2a1b82a35f69e280022e99c/internal/core/metadata/rest_device.go#L38

Then, to really finish it off, add a test for the router logic as well. Example:
https://github.com/edgexfoundry/edgex-go/blob/a478f95df2952767d2a1b82a35f69e280022e99c/internal/core/metadata/rest_device_test.go#L30

internal/core/metadata/operators/device/add.go
```diff
@@ -75,7 +75,7 @@ func (op addDevice) Execute() (id string, err error) {
 	op.device.Profile = profile
 
 	// Add the device
-	id, err = op.database.AddDevice(op.device)
+	id, err = op.database.AddDevice(op.device, profile.CoreCommands)
```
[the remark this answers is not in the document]
tsconn23: @AnthonyMBonafide I used the blackbox test core-metadata-importer collection to do this. Verified against both Redis and Mongo.

internal/core/metadata/rest_command.go
```diff
@@ -82,3 +82,28 @@ func restGetCommandByName(w http.ResponseWriter, r *http.Request) {
...
+func restGetCommandsByDeviceId(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	did, err := url.QueryUnescape(vars[ID])
+	if err != nil {
+		LoggingClient.Error(err.Error())
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	commands, err := dbClient.GetCommandsByDeviceId(did)
```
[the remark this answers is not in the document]
tsconn23: I also added a comment to this effect.

internal/pkg/db/mongo/metadata.go
tsconn23: The point is valid but nothing in our Mongo layer supports transactions. That would be a system-wide refactoring. In this case if there were an error, then the addCommands would throw an error. That would bubble up to the router and be returned to the client. The client could then call AddDevice again and the subsequent request would return a new ID and be its own entity. There would be some manual housekeeping necessary to remove the orphaned device record, true.

internal/core/metadata/operators/device/add.go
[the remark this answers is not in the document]
difince: > Created a new device using the URI /api/v1/device and a body which contains deviceCommands and coreCommands

Both DeviceCommands and CoreCommands are part of the DeviceProfile structure, not from the Device. The Device contains reference to the Profile via DeviceProfiles Name. Probably the request was incorrect?

internal/core/metadata/rest_command.go
difince: Ok, I will rewrite it and ... will add the missing RAML

Issue #1494, opened 2019-06-27
Device Profile Update -- Check for Existing Devices


tsconn23:
Decided during Core WG 27-Jun-19

When a request is received by core-metadata to update a DeviceProfile, the service should first check to see if there are any devices associated with that profile. If so, then the service should return a `409 Conflict` HTTP status to the caller.

To be consistent with the DELETE of a DeviceProfile, we should probably add a similar check for the association with one or more Provision Watchers. See associated issue #1495.

Handler is here
https://github.com/edgexfoundry/edgex-go/blob/9421ecd6e689ba6c8274a1d2f5deb84c71282b73/internal/core/metadata/router.go#L94

AnthonyMBonafide:
Resolved via PR #1498

PR #1498, opened 2019-07-02
Add validation when update device profile

AnthonyMBonafide:
Fix issue #1494

Add checks before updating device profile and replace direct calls to the DBClient with the operator pattern to allow for easier testing.

Update the restUpdateDeviceProfile rest handler to check if there are
associated devices or provision watchers before performing an update. If
there are associations for the device profile a Conflict(409) is
returned to the user and the update is not persisted. Update the successful response to return a 204(No Content) with no body rather than a "true"

Add new `UpdateDeviceProfileExecutor` which provides functionality for
updating device profiles by using the operator pattern. Also, move all
business logic to the `UpdateDeviceProfileExecutor` for segregation of
responsibilities. The REST handler is responsible for HTTP/REST logic
like serializing the request body, where the
`UpdateDeviceProfileExecutor` is responsible for interacting with the
database and ensuring the device profile which is being updated is in
the correct state before performing the update. For example, no devices
nor provision watchers are associated with the device profile.

Update the `restUpdateDeviceProfile` rest handler to use the new
operator.

Add new error types for the different errors the
`UpdateDeviceProfileExecutor` may encounter during processing. This will
allow for type checking for more insight on the underlying error

Update the `restUpdateDeviceProfile` to handle the different types of
errors the `UpdateDeviceProfileExecutor` may return

Add new `ProfileIdExecutor` for obtaining devices by a profile ID using
the operator pattern.

Update `notifyProfileAssociates` function to use the new operator

Added necessary mocks for new interfaces

Remove dead code which checked for duplicate command names within
DeviceProfile. This is dead code due to the validation logic already
handling this scenario and will result in the marshaling process to
short-circuit and result in a 400(Bad Request) response to the user

Add tests for `restUpdateDeviceProfile` REST handler, `UpdateDeviceProfileExecutor`, and `UpdateDeviceProfileExecutor`


AnthonyMBonafide:
TLDR:
Added the necessary checks as per the associated issue. and moved all the business logic into an operator. Also added tests.

internal/core/metadata/rest_deviceprofile.go
```diff
@@ -97,119 +101,33 @@ func restUpdateDeviceProfile(w http.ResponseWriter, r *http.Request) {
...
-		to.CoreCommands = from.CoreCommands
-	}
-
-	return nil
-}
-
-// Check for duplicate names in device profiles
-func checkDuplicateProfileNames(dp models.DeviceProfile, w http.ResponseWriter) error {
-	profiles, err := dbClient.GetAllDeviceProfiles()
+	err = notifyProfileAssociates(dp, dbClient, http.MethodPut)
```
AnthonyMBonafide: I did not include tests to cover the `notifyProfileAssociates` function. This function is tightly coupled to the `http` client. I did not want to complicate this PR anymore by refactoring further to decouple the HTTP client, so I left those tests out.

tsconn23: Yep, that's fine. When I created the original test for notifications in the AddNewDevice path, I knew there would be other areas that would need a similar abstraction. Could be they'll all converge at some point.

internal/core/metadata/errors/types.go
```diff
@@ -106,3 +106,47 @@ func (e ErrItemNotFound) Error() string {
...
+	id   string
+}
+
+func (e ErrDeviceProfileNotFound) Error() string {
+	if e.id != "" {
+		return fmt.Sprintf("device profile with id '%s' not found", e.id)
+	} else if e.name != "" {
+		return fmt.Sprintf("device profile with name '%s' not found", e.name)
+	}
+	return "device profile not found"
```
tsconn23: Suggest reduce the body of this function to just be
```
return fmt.Sprintf("device profile not found -- id: '%s' name: '%s'", e.id, e.name)
```

internal/core/metadata/errors/types.go
```diff
@@ -106,3 +106,47 @@ func (e ErrItemNotFound) Error() string {
...
+	description string
+}
+
+func (e ErrDeviceProfileInvalidState) Error() string {
+	if e.id != "" {
+		return fmt.Sprintf("device profile with id '%s' is in an invalid state. %s", e.id, e.description)
+	} else if e.name != "" {
+		return fmt.Sprintf("device profile with name '%s' is in an invalid state. %s", e.name, e.description)
+	}
+	return "device profile invalid state"
```
tsconn23: Same as above, can be collapsed to one line.

internal/core/metadata/operators/device/get.go
```diff
@@ -15,10 +15,11 @@
 package device
 
 import (
-	"github.com/edgexfoundry/edgex-go/internal/core/metadata/errors"
-	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+
+	"github.com/edgexfoundry/edgex-go/internal/core/metadata/errors"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
```
tsconn23: Not sure why these were moved as `e` comes before `g`

internal/core/metadata/operators/device/get_test.go
```diff
@@ -15,25 +15,29 @@
 package device
 
 import (
-	"github.com/pkg/errors"
 	"testing"
 
-	"github.com/edgexfoundry/edgex-go/internal/core/metadata/operators/device/mocks"
-	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
+	"github.com/pkg/errors"
```
tsconn23: This should be last in the ordering, alphabetizing everything after github.com. Also edgex-go comes before go-mod-core-contracts.

internal/core/metadata/operators/device/profile/update.go
```diff
@@ -0,0 +1,141 @@
+package profile
+
+import (
+	"strconv"
+
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+
+	errors2 "github.com/edgexfoundry/edgex-go/internal/core/metadata/errors"
```
tsconn23: Shouldn't need the `errors2` alias here.

internal/core/metadata/operators/device/profile/update.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	d, err := n.database.GetDevicesByProfileId(to.Id)
+	if err != nil {
+		return contract.DeviceProfile{}, err
+	}
+
+	if len(d) > 0 {
+		// LoggingClient.Debug(strconv.Itoa(len(d))+"devices are associated with the device profile named: " + to.Name)
+		// http.Error(w, "Conflict", http.StatusConflict)
```
tsconn23: comments can be removed

internal/core/metadata/operators/device/profile/update.go
```diff
@@ -0,0 +1,141 @@
...
+func NewUpdateDeviceProfileExecutor(db DeviceProfileUpdater, dp contract.DeviceProfile) UpdateDeviceProfileExecutor {
+	return updateDeviceProfile{
+		database: db,
+		dp:       dp}
+}
+
+// Update the fields of the device profile
+// to - the device profile that was already in Mongo (whose fields we're updating)
+// from - the device profile that was passed in with the request
+func (n updateDeviceProfile) updateDeviceProfileFields(from contract.DeviceProfile, to *contract.DeviceProfile) error {
```
tsconn23: I know this is a previously defined function and you didn't write it, just moved it into the operator from the router. I _**cannot stand**_ functions that mutate their parameters. I cannot tell you how many bugs I've had to root cause because of code written this way. Suggest this be refactored to return the new DeviceProfile rather than mutate the `*to` parameter. The caller should be changed to expect the new return as well.

internal/core/metadata/operators/device/profile/update.go
```diff
@@ -0,0 +1,141 @@
...
+	if from.Origin != 0 {
+		to.Origin = from.Origin
+	}
+
+	if from.Name != "" {
+		to.Name = from.Name
+		// Names must be unique for each device profile
+		if err := n.checkDuplicateProfileNames(*to); err != nil {
+			return err
+		}
```
tsconn23: In my opinion, this check should be moved into Execute() and performed before this function is even called. Once we get to invoking this function, all checks should be complete and the function does nothing but assign values to the new instance.

_Something to think about for later -- the creation of a new instance from old could actually be a function of the DeviceProfile model itself..._

internal/core/metadata/rest_deviceprofile.go
```diff
@@ -20,10 +20,14 @@ import (
...
 
-	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 	"github.com/edgexfoundry/go-mod-core-contracts/models"
 	"github.com/gorilla/mux"
 	"gopkg.in/yaml.v2"
+
+	errors2 "github.com/edgexfoundry/edgex-go/internal/core/metadata/errors"
+	"github.com/edgexfoundry/edgex-go/internal/core/metadata/operators/device"
+	"github.com/edgexfoundry/edgex-go/internal/core/metadata/operators/device/profile"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
```
tsconn23: Sorting's a little weird here. See other comments as well. All github.com imports should be alphabetized. Ignore aliases.

AnthonyMBonafide:
> One thing I'd ask is to not create a `device/profile` directory.

Should I put everything in the `device` directory?

internal/core/metadata/operators/device/profile/update.go
AnthonyMBonafide: :+1: Excellent point and I agree

tsconn23:
> > One thing I'd ask is to not create a `device/profile` directory.
> 
> Should I put everything in the `device` directory?

No, create a `metadata/operators/device_profile` directory

AnthonyMBonafide:
Recheck

tsconn23:
@AnthonyMBonafide One last thing on this -- Can you add the new 409 status being returned for this route to the RAML please? Then it should be good to go.

https://github.com/edgexfoundry/edgex-go/blob/42b502f5bb4ec663984d8826656a4b84953494ce/api/raml/core-metadata.raml#L803

PR #1500, opened 2019-07-02
Migration of PKISetup Service from Security-Secret-Store

### LINKED ISSUE #1499: Migrate PKISetup From Security-Secret-Store
This issue is to track the work related to migrating the PKISetup (Vault) service out of the security-secret-store repo and into edgex-go.

tsconn23:
Fix #1499 

- Please see commit log on this branch for detailed information.
- This PR is based on the current state of `master` in the https://github.com/edgexfoundry/security-secret-store repository.
- There is a security-secret-store PR https://github.com/edgexfoundry/security-secret-store/pull/69 currently ongoing with 144 comments on it. I'm not going comb through all of that to see what needs to be applied here. Once that PR is merged (hopefully soon) I will review the diff. Subsequent edits to the PKISetup service can be made to this codebase, if applicable.


.gitignore
```diff
@@ -18,6 +18,7 @@ cmd/core-data/core-data
 cmd/core-metadata/core-metadata
 cmd/export-client/export-client
 cmd/export-distro/export-distro
+cmd/security-setup/pkisetup
```
anonymouse64: not sure how I feel about the `security-setup` name for this service, has this been discussed? I think it makes the most sense to split up the 3 services in the security services as they currently exist (currently pkisetup and vaultworker from security-secret-store, and edgexproxy from security-api-gateway) like this:
* security-secret-store/pkisetup
* security-secret-store/vault-worker
* security-api-gateway/edgexproxy

tsconn23: Whoops! I tagged the wrong Ian as a reviewer.

tsconn23: Looking at this a bit more closely -- The reason I did this was to avoid having a deeper directory level for these services as compared to the other services in the cmd directory. If I was going to strictly preserve the naming, it would look like this
```
- cmd
    - core-data
       main.go
    - core-metadata
       main.go
    - security-secret-store
        - pkisetup
            main.go
        - vault-worker
            main.go
    - security-api-gateway
        - edgexproxy
            main.go
```
I was trying to be consistent with the layout. My thought on the other service names were `security-vault` and `security-proxy`. I don't know that we have to necessarily preserve the names of the old repos when the services are moved.

anonymouse64: If that's the case I'd have the following suggestions:
1. make all of the security-secret-store and security-api-gateway binaries into binaries prefixed just by `security-*` so just like you could run support-scheduler without support-notifications you could run just some of the `security-*` services
2. Drop `edgex` prefix for edgexproxy since it's redundant
3. Rename `vault` in `vault-worker` to be generic secret-store since eventually I assume this will be like the registry client where it could be replaced by something other than vault

So maybe:
* `edgexproxy` -> `security-proxy-worker`
* `pkisetup` -> `security-secrets-setup`
* `vault-worker` -> `security-secrets-worker`
I'm also not sure how I feel about worker, I think I like agent better, but agent also implies it's persistent and none of these services are actually persistent, they all are oneshot services which run once to configure something else like config-seed.

AnthonyMBonafide:
Also, nice test coverage!

internal/security/setup/directory.go
```diff
@@ -0,0 +1,73 @@
...
+	// Is the CA there? (if nil then OK... but could be something else than a directory)
+	stat, err := os.Stat(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Errorf("CA PKI setup directory does not exist: %s", path)
+		}
+		return fmt.Errorf("CA PKI setup directory cannot be reached: %s (%s)", path, err)
+	}
+	if stat.IsDir() {
+		h.logger.Debug("Existing CA PKI setup directory: %s", path)
```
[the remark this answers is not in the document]
tsconn23: Good catch. FYI -- for all of these I took the approach of wrapping the test in `fmt.Sprintf()`. I'd prefer to use the formatting. But you're right, `Debug` doesn't support that by itself.

AnthonyMBonafide:
LGTM

.gitignore
anonymouse64: @tsconn23 @tingyuz thoughts on the above proposal for naming?

tingyuz: The naming suggestion looks good to me. we may need to review the dockercompose file, dockfile and other files that may have extra requirement regarding the names of the project. And Jenkins jobs may be changed too. We may need some extra input for the last part.

tsconn23: @anonymouse64 Are you just proposing to name the executables differently, or should the relevant cmd folder and executable be named as above?

anonymouse64: @tsconn23 I think the executables should be named differently and the cmd folder be reorganized to match that.
@tingyuz I don't think the jenkins jobs for edgex-go would need to be modified as long as `make build` also builds these services. The old jobs for security-secret-store repo wouldn't be needed anymore for Fuji (except for edinburgh) since the code would live in edgex-go now. The docker jobs would need to be moved to edgex-go but that work needs to happen anyways and won't be more complicated if we re-name things now. It would be more work to rename them after the jobs have been setup however

tsconn23: @anonymouse64 cmd directory for this service has been renamed to `security-secrets-setup`

tsconn23:
recheck

.gitignore
anonymouse64: @tsconn23 sorry if I was unclear on this, but I meant to make the directory for this service `security-secrets-setup/security-secrets-setup` so the cmd dir is the same name of the executables. This matches core-data, export-client, etc.

tsconn23: done

PR #1501, opened 2019-07-03
Core Metadata UpdateDevice refactor and unit testing

### LINKED ISSUE #1485: Unit Test -- Core-Metadata Update Device
Add unit testing to cover the usage of PUT to /api/v1/device route exposed by core-metadata service.

jdharms:
Fix #1485

Refactors Core Metadata UpdateDevice function to extract
logic into an "operator" function and adds corresponding unit tests.


internal/core/metadata/operators/device/update.go
```diff
@@ -0,0 +1,172 @@
...
+	evt.HttpMethod = http.MethodPost
+	evt.ServiceId = op.device.Service.Id
+
+	op.events <- evt
+
+	return nil
+}
+
+func updateDeviceFields(from contract.Device, to *contract.Device, dbClient DeviceUpdater) error {
+
```
tsconn23: As I said to @AnthonyMBonafide in my comments on his [PR](https://github.com/edgexfoundry/edgex-go/pull/1498#discussion_r300075041), I know you didn't write this function and it was just brought over from the existing router logic. Please refactor so that the `*to` parameter isn't provided by the caller. Instead, the function should return the new Device instance.

internal/core/metadata/operators/device/update.go
```diff
@@ -0,0 +1,172 @@
...
+		// Check if the new profile exists
+		// Try ID first
+		dp, err := dbClient.GetDeviceProfileById(from.Profile.Id)
+		if err != nil {
+			// Then try Name
+			dp, err = dbClient.GetDeviceProfileByName(from.Profile.Name)
+			if err != nil {
+				return errors.NewErrItemNotFound("Device profile not found for updated device")
+			}
+		}
```
tsconn23: Also, following from the comment above and the one I linked to on Anthony's PR, all of these database checks should be moved into Execute() prior to this function being called. When we call `updateDeviceFields` we should be confident that duplication checks and so on have been done already so we can just set the new values. As I said on the other PR, eventually the Device model should handle this itself. For that to occur, we don't want to burden the Device model with the need to connect to a database or some other service.

internal/core/metadata/operators/device/update_test.go
```diff
@@ -0,0 +1,158 @@
...
+package device
+
+import (
+	"fmt"
+	"github.com/edgexfoundry/edgex-go/internal/core/metadata/operators/device/mocks"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
+	"github.com/edgexfoundry/go-mod-core-contracts/models"
+	"sync"
+	"testing"
```
tsconn23: Organize imports, stdlib first then 3rd party

PR #1508, opened 2019-07-08
Unit Testability -- core-metadata/GetAllDeviceServices

### LINKED ISSUE #1507: Unit Test -- Core-metadata "Get All Device Services" Route
Refactor the implementation of Core-metadata/GetAllDeviceServices using the operator pattern to increase unit test coverage.

https://github.com/edgexfoundry/edgex-go/blob/42b502f5bb4ec663984d8826656a4b84953494ce/internal/core/metadata/rest_deviceservice.go#L30

tsconn23:
Fix #1507


internal/core/metadata/operators/device_service/get_test.go
```diff
@@ -0,0 +1,70 @@
...
+			op := NewDeviceServiceLoadAll(tt.cfg, tt.dbMock, logger.MockLogger{})
+			_, err := op.Execute()
+			if err != nil && !tt.expectError {
+				t.Error(err)
+				return
+			}
+			if err == nil && tt.expectError {
+				t.Errorf("error was expected, none occurred")
+				return
+			}
```
jdharms: Should we be checking for more than just the operator's return code? i.e., checking the number of services returned, checking an ID for sanity?

tsconn23: In this case, I personally don't think so because everything around persistence is mocked. I have an interface that I'm glomming some logic onto through the mockery extensions, and there really isn't any kind of calculation or complex operation going on. My approach has been to target a specific execution flow in the code and then set up a mock to exactly satisfy that flow -- like if I need a particular exception to be returned. If there were some result obtained from passing in some params and then operating on them, then yes I would make that `expectedValue` part of my test criteria. This is sort of the approach taken [here](https://github.com/edgexfoundry/edgex-go/pull/1508/files/50c35001a25b6e50ca63384178fdf3e7cbce7f21#diff-d49e364aa4cf09204647193cd6b5a9ef) in testing the router. I've got a particular status I'm expecting from the operator, and the dbClient mock is setup in combination withe Configuration to ensure that condition is returned.

jdharms: I can understand your point, but the fact that a `return nil, nil` would satisfy the positive cases still kind of bothers me.  I think at the bare minimum I'd prefer checking to make sure the slice returned is the expected length, but if you don't think it's necessary I understand.

brandonforster: It seems like there's nontrivial code happening between the mocks and the returns. We should be doing SOME sort of checking on the return value, whether that's checking if it's a nil slice in error cases or has an expected length in positive cases to ensure that inadvertent code changes don't slip through.

mkbhanda: @jdharms and @brandonforster - OK you convinced me.
BTW -- should we not have a test case to cover no device services registered without it being an error. That could happen at the beginning of time. Am I missing something?

PR #1510, opened 2019-07-08
Update reference to go-mod-messaging, fix race condition

### LINKED ISSUE #1509: Edinburgh: Update Messaging Module Reference due to Race Condition
Myself, @xmlviking and @brandonforster have all confirmed the possibility of an error due to a race condition in the go-mod-messaging module. As is the case with race conditions, it doesn't happen all the time but we have each experienced it using the following steps. All services below are running natively.

1.) Clean your database
2.) Build the Edinburgh branch of edgex-go
3.) Build device-random (at the current time `master` is fine)
4.) Bring up edgex-go services via `make run`
5.) Bring up device-random

In each case where you go through this scenario, you have a good chance of encountering the following error:

```
level=ERROR ts=2019-07-05T23:31:37.314722Z app=edgex-core-data source=event.go:255 msg="Unable to send message for event: {\"id\":\"e7dfcec9-0db7-45c3-bc20-1473d81046a5\",\"device\":\"Random-Integer-Generator01\",\"origin\":1562369497307028000,\"readings\":[{\"id\":\"a845adba-b390-42c3-ae6b-d00cf96e9bb7\",\"origin\":1562369497306850000,\"device\":\"Random-Integer-Generator01\",\"name\":\"RandomValue_Int16\",\"value\":\"6252\"}]} error: address already in use [tcp://*:5563]"
```

The "address already in use" is happening due to two goroutines going into the same block of logic to bind the message client to the ZeroMQ port. When this happens, it hoses the connection to ZeroMQ. It will look like core-data is processing subsequent messages just fine but in truth, they're getting dropped. 

A fix for this has already been made to the go-mod-messaging module via https://github.com/edgexfoundry/go-mod-messaging/pull/26

To fix this issue in Edinburgh, the above update must be consumed via the edgex-go `go.mod` in the Edinburgh branch.

tsconn23:
Fix #1509

- Also noticed that if an error were to occur during publishing, our
  logging was dropping that from the message. Added error contents to
  log output in this case.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/OneOfOne/xxhash v1.2.5
 	github.com/eclipse/paho.mqtt.golang v1.1.1
 	github.com/edgexfoundry/go-mod-core-contracts v0.1.0
-	github.com/edgexfoundry/go-mod-messaging v0.1.0
+	github.com/edgexfoundry/go-mod-messaging v0.1.5
 	github.com/edgexfoundry/go-mod-registry v0.1.0
 	github.com/globalsign/mgo v0.0.0-20181015135952-eeefdecb41b8
 	github.com/go-kit/kit v0.8.0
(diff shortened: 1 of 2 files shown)


tsconn23:
> LGTM - but did we really jump from messaging 0.1.0 to 0.1.5?

@jpwhitemn It appears so
https://github.com/edgexfoundry/go-mod-messaging/releases

PR #1516, opened 2019-07-09
Updated Consul to 1.5.2.

brandonforster:
Verified locally. Will update after changes are made to https://github.com/edgexfoundry/go-mod-registry/pull/19

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/eclipse/paho.mqtt.golang v1.1.1
 	github.com/edgexfoundry/go-mod-core-contracts v0.1.3
 	github.com/edgexfoundry/go-mod-messaging v0.1.5
-	github.com/edgexfoundry/go-mod-registry v0.1.0
+	github.com/edgexfoundry/go-mod-registry v0.1.5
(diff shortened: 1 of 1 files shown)


anonymouse64:
It would be great if this PR could also update the version of consul that is built in the snap: https://github.com/edgexfoundry/edgex-go/blob/master/snap/snapcraft.yaml#L407-L446

I'm not sure if the build steps for consul have significantly changed from 1.3 to 1.5 or not, but it's usually pretty straightforward with their Makefile. There may have upgrade to go modules however

brandonforster:
> It would be great if this PR could also update the version of consul that is built in the snap: https://github.com/edgexfoundry/edgex-go/blob/master/snap/snapcraft.yaml#L407-L446
> 
> I'm not sure if the build steps for consul have significantly changed from 1.3 to 1.5 or not, but it's usually pretty straightforward with their Makefile. There may have upgrade to go modules however

The only breaking changes are around the use of the `watch` package, and we don't use it directly anyway. The biggest change as a result of this upgrade is that the indirect module dependencies are significantly simplified.

I'll update the snap and take a look into that.

brandonforster:
@anonymouse64 Per @jpwhitemn, I've created #1531 to address the Snap story.

PR #1522, opened 2019-07-10
Add First Steps in Implementing Core-Command Persistent Layer

difince:
- Added direct connection from core-command service to "metadata" database
(should be change in future, core-command is supposed to have its own database)
- Added DBClient interface
- Add new GetCommandByNameAndDeviceId function in the BDClient
- Rewrite all functions that called metadata endpoints for quering commands to use
newly edded DBClient interface
- rename GetCommandByName function to GetCommandsByName (plural)

Fix: https://github.com/edgexfoundry/edgex-go/issues/1274

### DIFF
--- a/cmd/core-command/res/configuration.toml
+++ b/cmd/core-command/res/configuration.toml
@@ -31,3 +31,13 @@ File = './logs/edgex-core-command.log'
   Protocol = 'http'
   Host = 'localhost'
   Port = 48061
+
+[Databases]
+  [Databases.Primary]
+  Host = 'localhost'
+  Name = 'metadata'
+  Password = ''
+  Port = 27017
+  Username = ''
+  Timeout = 5000
+  Type = 'mongodb'
--- a/cmd/core-command/res/docker/configuration.toml
+++ b/cmd/core-command/res/docker/configuration.toml
@@ -31,3 +31,13 @@ File = '/edgex/logs/edgex-core-command.log'
   Protocol = 'http'
   Host = 'edgex-support-logging'
   Port = 48061
+
+[Databases]
+  [Databases.Primary]
+  Host = 'edgex-mongo'
+  Name = 'metadata'
+  Password = 'password'
+  Port = 27017
+  Username = 'meta'
+  Timeout = 5000
+  Type = 'mongodb'
--- a/internal/core/command/device.go
+++ b/internal/core/command/device.go
@@ -16,9 +16,11 @@ package command
 import (
 	"context"
 	"fmt"
+	"net/http"
+
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
-	"net/http"
 )
 
 func commandByDeviceID(deviceID string, commandID string, body string, isPutCommand bool, ctx context.Context) (string, int) {
@@ -40,12 +42,11 @@ func commandByDeviceID(deviceID string, commandID string, body string, isPutComm
 	}
 
 	//once command service have its own persistence layer this call will be changed.
-	commands, err := cc.CommandsForDeviceId(d.Id, ctx)
+	commands, err := dbClient.GetCommandsByDeviceId(d.Id)
 	if err != nil {
 		LoggingClient.Error(err.Error())
-		chk, ok := err.(*types.ErrServiceClient)
-		if ok {
-			return err.Error(), chk.StatusCode
+		if err == db.ErrNotFound {
+			return err.Error(), http.StatusNotFound
 		} else {
 			return err.Error(), http.StatusInternalServerError
 		}
@@ -79,33 +80,16 @@ func commandByNames(dn string, cn string, body string, isPutCommand bool, ctx co
 			return err.Error(), http.StatusInternalServerError
 		}
 	}
-	//once command service has its own persistence layer, we will directly read from the db getCommandByNameAndDeviceID
-	var commands []contract.Command
-	if commands, err = cc.CommandsForDeviceId(d.Id, ctx); err != nil {
+	command, err := dbClient.GetCommandByNameAndDeviceId(cn, d.Id)
+	if err != nil {
 		LoggingClient.Error(err.Error())
-		chk, ok := err.(*types.ErrServiceClient)
-		if ok {
-			return err.Error(), chk.StatusCode
+		if err == db.ErrNotFound {
+			return err.Error(), http.StatusNotFound
 		} else {
 			return err.Error(), http.StatusInternalServerError
 		}
 	}
-
-	var c contract.Command
-	for _, one := range commands {
-		if cn == one.Name {
-			c = one
-			break
-		}
-	}
-
-	if c.String() == (contract.Command{}).String() {
-		errMsg := fmt.Sprintf("Command with name '%v' not found.", cn)
-		LoggingClient.Error(errMsg)
-		return errMsg, http.StatusNotFound
-	}
-
-	return commandByDevice(d, c, body, isPutCommand, ctx)
+	return commandByDevice(d, command, body, isPutCommand, ctx)
 }
 
 func commandByDevice(device contract.Device, command contract.Command, body string, isPutCommand bool, ctx context.Context) (string, int) {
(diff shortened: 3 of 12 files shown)


cmd/core-command/res/configuration.toml
```diff
@@ -30,4 +30,14 @@ File = './logs/edgex-core-command.log'
   [Clients.Logging]
   Protocol = 'http'
   Host = 'localhost'
-  Port = 48061
\ No newline at end of file
+  Port = 48061
+
+[Databases]
+  [Databases."Metadata.Primary"]
```
tsconn23: This should just be `[Databases.Primary]`

cmd/core-command/res/docker/configuration.toml
```diff
@@ -30,4 +30,14 @@ File = '/edgex/logs/edgex-core-command.log'
   [Clients.Logging]
   Protocol = 'http'
   Host = 'edgex-support-logging'
-  Port = 48061
\ No newline at end of file
+  Port = 48061
+
+[Databases]
+  [Databases."Metadata.Primary"]
```
tsconn23: This should just be `[Databases.Primary]`

AnthonyMBonafide:
I am wondering if we should adopt the operator pattern here and put the database related functionality there. My thought is that it will not only allow us to test, but when we make changes in the future with the separation of databases we will be able to to it a little easier since it will be isolated and testable.

cmd/core-metadata/res/configuration.toml
```diff
@@ -31,6 +31,7 @@ File = './logs/edgex-core-metadata.log'
   Host = 'localhost'
   Port = 48060
 
+#same changes done in this section need to be applied in core-commands service configuration.toml file(temporary)
```
tsconn23: Comment isn't necessary. This could be said for any service requiring persistence, and now core-command is just another one of those. For example, if the customer wants to switch and use `Redis` they have to change the Type and Port in all services supporting persistence.

cmd/core-metadata/res/docker/configuration.toml
```diff
@@ -31,6 +31,7 @@ File = '/edgex/logs/edgex-core-metadata.log'
   Host = 'edgex-support-notifications'
   Port = 48060
 
+#same changes done in this section need to be applied in core-commands service configuration.toml file(temporary)
```
tsconn23: Same RE: Comment

internal/core/command/device.go
```diff
@@ -79,33 +80,17 @@ func commandByNames(dn string, cn string, body string, isPutCommand bool, ctx co
 			return err.Error(), http.StatusInternalServerError
 		}
 	}
-	//once command service has its own persistence layer, we will directly read from the db getCommandByNameAndDeviceID
-	var commands []contract.Command
-	if commands, err = cc.CommandsForDeviceId(d.Id, ctx); err != nil {
+	var command contract.Command
+	command, err = dbClient.GetCommandByNameAndDeviceId(cn, d.Id)
```
tsconn23: Why not just `command, err := dbClient.GetCommandByNameAndDeviceId(cn, d.Id)` without the `var` line?

Also, what do you think of turning the first line in this function into a DBClient call?
`d, err := mdc.DeviceForName(dn, ctx)`

internal/core/command/init.go
```diff
@@ -27,19 +27,24 @@ import (
...
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db/mongo"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db/redis"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/startup"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/telemetry"
 )
 
 var Configuration *ConfigurationStruct
 var LoggingClient logger.LoggingClient
 var mdc metadata.DeviceClient
```
tsconn23: Should we replace the above client with additional DBClient functions?

internal/core/command/init.go
```diff
@@ -237,3 +247,39 @@ func setLoggingTarget() string {
 	}
 	return Configuration.Logging.File
 }
+
+func connectToDatabase() error {
+	var err error
+
+	dbClient, err = newDBClient(Configuration.Databases["Metadata.Primary"].Type)
```
tsconn23: Just "Primary" please, not "Metadata.Primary". I should enter an issue to make this a constant.

internal/core/command/init.go
```diff
@@ -237,3 +247,39 @@ func setLoggingTarget() string {
...
+
+	return nil
+}
+
+// Return the dbClient interface
+func newDBClient(dbType string) (interfaces.DBClient, error) {
+	switch dbType {
+	case db.MongoDB:
+		dbConfig := db.Configuration{
+			Host:         Configuration.Databases["Metadata.Primary"].Host,
```
tsconn23: Same with these, just "Primary"

cmd/core-metadata/res/configuration.toml
difince: I assumed, that each service could use different database type. Some could use redis, others mongo :) That is why I put these comments in the tolm files and used [Databases."Metadata.Primary"] instead of just [Databases.Primary].
But if this is not that case, i am removing all them

difince:
@AnthonyMBonafide I agree with the adoption of operator pattern. I just thought this to be done in a separate PR.

internal/core/command/init.go
difince: what do you mean...  to remove _**var mdc metadata.DeviceClient**_  and replace all calls (getAllDevices, getDeviceById, getDeviceByName) made from Command service to Metadata service  with new functions in the local DBClient ?

internal/core/command/device.go
difince: Absolutely agree with your first question. 
About the second one, I am not sure what you mean , guess it is connected with 

> Should we replace the above client with additional DBClient functions?

difince:
@tsconn23 

-  > Would you mind ensuring these errors get logged? Also rather than empty string, returned response should be err.Error()

       Yes, I will do 

- >
 	responseBody, responseCode, err := ex.Execute()
        if err != nil {
		return "", http.StatusInternalServerError
	}

        Here, the only consideration could be if we want to hide from the end user the real error coming from another micro service (DS)? My opinion is always to show the real error, but probably the author was thinking differently ?   

In addition, I though to open another issue for rewriting some parts command service. 

- There are places like [this](https://github.com/difince/edgex-go/blob/0866d3f611b2cd8e37dc515ac70ab54bdab5d870/internal/core/command/restDevice.go#L94-L98), where whatever error appears, it is overriden with - _"not found err"_. 

- There are endpoints like this: **[/api/v1/device/{deviceId}/command/{commandId}}](https://github.com/difince/edgex-go/blob/0866d3f611b2cd8e37dc515ac70ab54bdab5d870/internal/core/command/rest.go#L58-L59)**  Why the client should be forced to provide both commandId and DeviceId ? Providing just the _commandId_ is enough. Currently we do not have a function that could give us the device by commandId, or even when we read the command (command.deviceId property is not returned), but we could handle it. I know that changing the endpoints signature is not wanted for this release. 

- adoption of operator pattern, as already mentioned
- I would change some of the functions names

internal/core/command/init.go
tsconn23: Yes. I'm asking what you think of doing that. On the one hand, I would like to do it to eliminate the cross-service hop. On the other hand, metadata is authoritative for device information so the call makes sense. I'm conflicted. Asking for your opinion.

tsconn23:
> In addition, I thought to open another issue for rewriting a bit some parts of command service.
>
> - There are places like this, where whatever error appears, it is overriden with - "not found err".
> - adoption of operator pattern, as already mentioned
> - I would change some of the functions names

The above can be three separate issues to keep the PR size manageable. Tag with "Fuji" and assign to yourself.

> There are endpoints like this: /api/v1/device/{deviceId}/command/{commandId}} Why the client should be forced to provide both commandId and DeviceId ? Providing just the commandId is enough. Currently we do not have a function that could give us the device by commandId, or even when we read the command (command.deviceId property is not returned), but we could handle it. I know that changing the endpoints signature is not wanted for this release.

Sure, go ahead and open an issue for this. Tag it for "Geneva" and we can discuss once we get to that release cycle.

difince:
> The above can be three separate issues to keep the PR size manageable. Tag with "Fuji" and assign to yourself. Please give the operator refactoring priority.

I will need an assistance for doing it, because I have permissions just to open issues (have no permissions to assign a issue or set a label)

internal/core/command/init.go
difince: I am conflicted as well, but ... by that moment my answer will be -  just for the sake of micro service architecture it is good to keep the cross-service calls. 
(nevertheless that with this PR the micro-service architecture pillars between core-command and metadata services will be broken.) 
On the other hand should  we blindly follow the the best-practice by providing own database per each service? Does it mean/is this enough condition to say that two service are really decoupled? (I still haven't get to the end of DDD book you recommenced )

tsconn23: OK, we'll leave this as is for now.

PR #1523, opened 2019-07-10
Add validation when deleting a device profile

### LINKED ISSUE #1495: Device Profile Delete -- Check for Existing Devices
Follows from https://github.com/edgexfoundry/edgex-go/issues/1494

When a request is received by core-metadata to delete a DeviceProfile, the service should first check to see if there are any devices associated with that profile. If so, then the service should return a 409 Conflict HTTP status to the caller. Note that this check is in addition to the existing check for ProvisionWatcher association.

The expectation is that this work will conform to the pattern for unit testing being spread throughout the core services. If you don't have background on that, please reach out to me on Slack.

Endpoints include
- https://github.com/edgexfoundry/edgex-go/blob/9421ecd6e689ba6c8274a1d2f5deb84c71282b73/internal/core/metadata/router.go#L98
- https://github.com/edgexfoundry/edgex-go/blob/9421ecd6e689ba6c8274a1d2f5deb84c71282b73/internal/core/metadata/router.go#L112

AnthonyMBonafide:
Fix #1495

Update the REST handlers responsible for deleting device profiles to
perform validation to ensure no devices nor provision watchers are
associated with the device profile before deletion.

Add validation to REST handlers `restDeleteProfileByProfileId` and
`restDeleteProfileByName`.

Update REST handlers to use new operator pattern.

Create new operator `DeviceProfileDeleter` which abstracts the deletion
process.

Add tests for `DeviceProfileDeleter` to cover functionality when
deleting by name and by ID.

Add tests for REST handler `restDeleteProfileByProfileId`

Add tests for REST handler `restDeleteProfileByName`.

Updated RAML to match the endpoints that have been updated.


api/raml/core-metadata.raml
```diff
@@ -546,10 +546,12 @@ schemas:
     delete:
         description: Remove the DeviceProfile designated by database generated id. This does not remove associated commands. Returns Internal Service Error (HTTP 500) for unknown or unanticipated issues. Returns NotFoundException (HTTP 404) if the device profile cannot be found by the identifier provided. Returns DataValidationException (HTTP 413) if devices still reference the profile.
         responses:
-            "200":
-                description: boolean indicating success of the remove operation
+            "204":
```
tsconn23: I'm sympathetic to changing these status codes to `204` but unfortunately that counts as an API change. If Edinburgh returns `200` for these operations and Fuji returns a `204` then Fuji isn't backward compatible. I realize both are "Success" codes but if a client is specifically looking for `200` instead of a code from the set indicating success, that client will break. Based on that, I would leave this (and the others like it) as a `200`.

Along with the `200` response we currently write out `true`. Unfortunately, that needs to be preserved as well.

RAML description updates are fine though.

api/raml/core-metadata.raml
```diff
@@ -546,10 +546,12 @@ schemas:
...
+                description: successful deletion of the device profile
             "404":
                 description: if the device profile cannot be found with the identifier provided
+            "409":
+                description: if devices or provision watchers still reference the profile
```
tsconn23: This change IS backward compatible because it's additive. It didn't exist at all in Edinburgh and has been added for Fuji so an Edinburgh client will continue to work.

internal/core/metadata/operators/device/db.go
```diff
@@ -32,3 +29,23 @@ type DeviceUpdater interface {
...
+	GetDeviceProfileById(id string) (contract.DeviceProfile, error)
+	GetDeviceProfileByName(n string) (contract.DeviceProfile, error)
+}
+
+type DeviceProfileDeleter interface {
+	DeleteDeviceProfileById(id string) error
+	DeviceProfileLoader
+	DeviceLoader
+	provsion_watcher.ProvisionWatcherLoader
+}
```
tsconn23: Should these DeviceProfile-related interfaces be defined instead in internal/core/metadata/operators/device_profile/db.go?

internal/core/metadata/operators/device_profile/delete.go
```diff
@@ -0,0 +1,107 @@
+package device_profile
+
+import (
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+
+	errors2 "github.com/edgexfoundry/edgex-go/internal/core/metadata/errors"
```
tsconn23: This alias can be removed.

internal/core/metadata/operators/device_profile/update_test.go
```diff
@@ -7,7 +7,8 @@ import (
 
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 
-	"github.com/edgexfoundry/edgex-go/internal/core/metadata/operators/device_profile/mocks"
+	"github.com/edgexfoundry/edgex-go/internal/core/metadata/operators/device"
+	mocks2 "github.com/edgexfoundry/edgex-go/internal/core/metadata/operators/device/mocks"
```
tsconn23: Don't need this alias.

internal/core/metadata/rest_deviceprofile.go
```diff
@@ -150,28 +150,25 @@ func restGetProfileByProfileId(w http.ResponseWriter, r *http.Request) {
...
 
-	// Check if the device profile exists
-	dp, err := dbClient.GetDeviceProfileById(did)
+	op := device_profile.NewDeleteByIDExecutor(dbClient, did)
+	err := op.Execute()
 	if err != nil {
-		if err == db.ErrNotFound {
+		LoggingClient.Error(err.Error())
+		switch err.(type) {
+		case errors2.ErrDeviceProfileNotFound:
```
tsconn23: Something to consider -- It looks to me like we have custom error types in `internal/core/metadata/errors/types.go` that we could use in this file instead of `errors.New("blah")` If that's the case, then the need for an `errors2` alias goes away.

For example, line 42
`err = errors.New("Max limit exceeded with request for profiles")`
We have an `ErrLimitExceeded` type we could use instead.

Your call as to whether to include such an audit (for this file only) in this PR. If not, please create an issue to follow-up.

internal/core/metadata/rest_deviceprofile_test.go
```diff
@@ -10,10 +10,12 @@ import (
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+	"github.com/gorilla/mux"
```
tsconn23: Move to the bottom of imports

api/raml/core-metadata.raml
AnthonyMBonafide: Ok, I think I made a similar change in my other PR for the update device profile. I will revert that in this PR.

internal/core/metadata/operators/device/db.go
AnthonyMBonafide: I moved them around due to cyclic dependencies.  The device-profile related interfaces require some of the device related interfaces. Is there a better way to organize these interfaces?

api/raml/core-metadata.raml
AnthonyMBonafide: It just hurt to see `application/json` as the content type and then return "true" :frowning_face:

internal/core/metadata/rest_deviceprofile.go
AnthonyMBonafide: I'd like to tackle this now while I am in here. Less overhead rather than creating an issue, new PR, etc.

internal/core/metadata/operators/device/db.go
AnthonyMBonafide: To be more clear, if I move the device-profile related interfaces I get the following cycle:
```
DeviceAdder -> DeviceProfileLoader -> DeviceLoader
```

This is one of the few cycles that will exist as a result of the move.

However, the `DeviceProfileLoader` interface is not dependant on any of the device interfaces only the `device` interfaces depend on the `DeviceProfileLoader` which we should be able to move. Sadly, this also causes a cycle with the tests as the tests in `delete_test.go` and `update_test.go` rely on the interfaces in the `device` package.

AnthonyMBonafide:
Recheck

internal/core/metadata/operators/device/db.go
tsconn23: Got it. Remember though that Go uses interfaces differently from other languages. So we can define interfaces scoped to a given package. They don't need to be centrally defined and the emphasis on re-use of a single, shared interface definition is lessened. I personally don't think there's anything wrong with redeclaring interfaces that are scoped to a different package. In fact, I think mixing the approaches toward interfaces will get us into trouble w/r/t circular deps. See the following branch
https://github.com/tsconn23/edgex-go/tree/pr_1523_tsc

tsconn23:
recheck

Issue #1527, opened 2019-07-11
Unit Tests -- Core-Metadata: Get All Commands


difince:
Related to: https://github.com/edgexfoundry/edgex-go/projects/11#card-23264347

difince:
this issue is supposed to be with label - enhancement ..but now i do not see a way to change the label.
Could you please assign it to me

PR #1529, opened 2019-07-11
Unit Tests -- Core-Metadata: Get All Commands

difince:
Fix: https://github.com/edgexfoundry/edgex-go/issues/1527


internal/core/metadata/operators/command/get.go
```diff
@@ -21,3 +23,30 @@ func NewDeviceIdExecutor(db CommandLoader, deviceId string) DeviceIdExecutor {
 func (op commandLoadByDeviceId) Execute() (commands []contract.Command, err error) {
 	return op.database.GetCommandsByDeviceId(op.deviceId)
 }
+
+//Get All commands
+type CommandAllExecutor interface {
```
AnthonyMBonafide: Can we add a test suite(test table) for this new operator?  I see that the tests in `rest_command_test.go` cover all the lines in this file, but I still think it would be nice to unit test this operator in isolation to get better functional coverage.

tsconn23: @difince To zero in on what @AnthonyMBonafide is referring to

1.) run `make test`
2.) run `go tool cover -html=coverage.out`
3.) Once the browser opens, use the drop-down at the top to view `internal/core/metadata/operators/command/get.go` 

It should have %20 coverage with this PR and the untested `commandLoadAll` logic is in red. Once the logic is covered, re-run the above steps and the lines will be green.

Issue #1526, opened 2019-07-11
Unit Test -- Core-Metadata: Get Command by Id


difince:
Related to: https://github.com/edgexfoundry/edgex-go/projects/11#card-23264363

difince:
Could you please assign it to me

PR #1543, opened 2019-07-16
Add Unit Test -- Core-Metadata: Get Command by Id

difince:
Rewrite coremetadata endpoint GetCommandById to use operator pattern
Add unit tests

Fix: https://github.com/edgexfoundry/edgex-go/issues/1526


internal/core/metadata/operators/command/get_test.go
```diff
@@ -48,9 +66,27 @@ func TestLoadAllCommands(t *testing.T) {
...
+			config.ServiceInfo{MaxResultCount: 1},
+			createCommandLoaderMock("GetAllCommands", []models.Command{testCommand}, nil, ""),
+			[]models.Command{testCommand},
+			false,
+		},
+		{
+			"GetAllFailMaxExceeded",
+			config.ServiceInfo{},
+			createCommandLoaderMock("GetAllCommands", []models.Command{testCommand}, nil, ""),
+			[]models.Command{testCommand},
```
jdharms: I know this isn't what this PR is about, but in the max result count exceeded case the operator should return `[]models.Command{}, errors.NewErrLimitExceeded(op.config.MaxResultCount)`, and not return the Commands returned by the database.

internal/core/metadata/operators/command/get_test.go
```diff
@@ -72,33 +108,71 @@ func TestLoadAllCommands(t *testing.T) {
...
+				return
+			}
+
+		})
+	}
 }
 
-func createAllCommandsLoaderMockFail() CommandLoader {
+func createCommandLoaderMock(methodName string, ret interface{}, err error, arg string) CommandLoader {
 	dbCommandMoch := &mock.CommandLoader{}
```
jdharms: Should be `dbCommandMock`

internal/core/metadata/operators/command/get_test.go
difince: I do not agree totally because createCommandLoaderMock define how the mocked function GetAllCommands should behave. 
[GetAllCommands](https://github.com/edgexfoundry/edgex-go/blob/aedf85076467d5f6e837d3bd8b59809f5e360a0f/internal/core/metadata/operators/command/get.go#L44) does not return NewErrLimitExceeded, the error is returned one layer [above](https://github.com/edgexfoundry/edgex-go/blob/aedf85076467d5f6e837d3bd8b59809f5e360a0f/internal/core/metadata/operators/command/get.go#L50) in the [commandLoadAll.Execute](https://github.com/edgexfoundry/edgex-go/blob/aedf85076467d5f6e837d3bd8b59809f5e360a0f/internal/core/metadata/operators/command/get.go#L43). 
But thank you for your comment, it help me realize that my return statement in case of NewErrLimitExceeded is wrong. (I am suppose to return empty list in case of NewErrLimitExceeded)

jdharms: I think we were in agreement :)

I didn't mean the mock db command should return that error--but that the Execute function should return `[]models.Command{}, errors.NewErrLimitExceeded(op.config.MaxResultCount)` and that the test should be expect that.

difince:
Despite my misunderstanding, we are in total agreement. Thank you :relaxed:

PR #1549, opened 2019-07-16
Unit Test -- Refactored add addressable using operator pattern and added tests.

### LINKED ISSUE #1517: Unit Test -- Core-Metadata Add Addressables
Related to https://github.com/edgexfoundry/edgex-go/projects/11#card-22883564

brandonforster:
Fix #1517 

Refactors the AddAddressable logic and adds unit test for that operator and executor.

### DIFF
--- a/internal/core/metadata/operators/addressable/add_test.go
+++ b/internal/core/metadata/operators/addressable/add_test.go
@@ -1,0 +1,109 @@
+package addressable
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/edgexfoundry/edgex-go/internal/core/metadata/errors"
+	"github.com/edgexfoundry/edgex-go/internal/core/metadata/operators/addressable/mocks"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
+
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+	"github.com/stretchr/testify/mock"
+)
+
+var missingName = contract.Addressable{
+	User:       "User1",
+	Protocol:   "http",
+	Id:         "address #1",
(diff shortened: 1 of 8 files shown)


internal/core/metadata/operators/addressable/add_test.go
```diff
@@ -0,0 +1,97 @@
...
+func TestExecutor(t *testing.T) {
+	success := SuccessfulDatabaseResult[0]
+
+	tests := []struct {
+		name             string
+		mockDb           AddressLoader
+		addr             contract.Addressable
+		expectedResult   string
+		expectedError    bool
+		expectedErrorVal error
```
jdharms: It looks like you're not using this field.  I'd use it to ensure you're getting the `ErrDuplicateName` you're expecting in the duplicate test case. See TestAllAddressablesExecutor  in metadata/operators/addressable/get_test.go for an example.

internal/core/metadata/operators/addressable/db.go
```diff
@@ -5,6 +5,7 @@ import contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 
 // AddressLoader provides functionality for obtaining Addressables.
 type AddressLoader interface {
+	AddAddressable(addressable contract.Addressable) (string, error)
```
jdharms: This needs to be in a different interface I believe, as it's not used "for obtaining Addressables."  I'd suggest creating an "AddressSaver" or something similar interface.

jdharms:
recheck

internal/core/metadata/operators/addressable/db.go
jdharms: Maybe `AddressAdder`

brandonforster:
recheck

PR #1550, opened 2019-07-16
Patch docker-compose timing issues in export/rules-engine

### LINKED ISSUE #1544: Edinburgh -- Address Docker-Compose Bootstrap Issues
There are timing issues in the Edinburgh docker-compose bootstrapping process that have been identified and root caused. These issues occur when no volumes are present and must be created for the first time via `docker-compose up`. To replicate use the Edinburgh "no security" [compose file](https://github.com/edgexfoundry/developer-scripts/blob/master/compose-files/docker-compose-edinburgh-no-secty-1.0.0.yml)

1.) `docker volume prune`
2.) `docker-compose -f docker-compose-edinburgh-no-secty-1.0.0.yml up`

You will observe two symptoms caused by three different issues. I will detail the symptoms here.

1.) Support-logging is the first service to come up. However there is additional latency for this to occur as the volume is being created. Other services start coming up and try to log their startup info remotely via support-logging. Since the actual API router in support-logging doesn't come up until the config is read and connectivity to the database is established, these other services will all throw "connection refused" errors when attempting to log for a period of time. This eventually corrects itself.
2.) The flow whereby the RulesEngine registers as an export-client, which should then trigger a notification to export-distro so that it will publish events is disrupted by the additional volume creation latency. This results in no events ever reaching the Rules Engine.

It should be noted that you can go through the scenario above and experience these errors and then shut the whole thing down via `docker-compose down`. If you then bring the services back up without pruning your docker volumes, the whole thing works just fine. I suspect most of us were using this latter approach without thinking to prune the underlying volumes as would be done on a brand new deployment.

I will also point out that the code that will be associated with this issue for the dot release is targeting these symptoms. There are more fundamental issues with regard to service bootstrapping in a containerized environment that will need to be addressed for a long term fix. I do not yet have a view on what these long term patterns should be, but the operability of this process is pretty essential for Edinburgh.

tsconn23:
Fix #1544

The primary issue is observed when one runs "docker-compose up" after
pruning local docker volumes.

The following problematic symptoms have been observed. The requisite
code changes in this PR are enumerated below. I hesitate to properly
call this a "fix" because I think more analysis is warranted for a long
term solution. However for the purpose of stabilizing the release, this
should suffice.

- Observed that population of configuration data in Consul and
  subsequent creation of DBs/Collections and credentials in Mongo were
  delayed when creating volumes for the first time. This lead to timing
  issues in the following areas
- I observed conditions where the Config-Seed was populating Consul for
  a given service at the same time that that service's configuration was
  being read into Consul by the config-seed. Thus services would come
  up and read their configuration from Consul when that config was only
  partially populated. In Delhi this would not be a problem because the
  listenForConfigChanges() function would listen for a change anywhere
  in the service's config in Consul. This would cause the entirety of
  the ConfigurationStruct to be populated when the last key was written.
  In Edinburgh, we only listen for changes to the "Writable" section,
  which is written first.
  -- This is why initializeConfiguration() has been modified to look at
     the last value of the service's configuration within Consul to see
     if it has been populated. If it is not, then we throw an error --
     forcing a retry one second later. If it is populated, the whole
     config has been written to Consul and we can proceed.
- All services write logging information when they are bootstrapping.
  If the service is configured to log remotely, those calls go to
  support-logging. Because of the observed latency (up to 5 seconds) in
  populating configuration and Mongo with new volumes, any service trying
  to send its startup messages to support-logging would throw an error.
  This lead to many, mnay errors in the starup process. Once support-
  logging came up, the remote logging would work. Understand that every
  service must receive its configuration information from Consul and,
  where applicable, establish DB connectivity before it activates its
  API handlers. If a service calls support-logging (or any other service)
  before the API is enabled, the caller will receive a "connection re-
  fused".
  -- This was also mitigated by the code change above.
- In the case where a service is delayed in its startup, it will not
  register its endpoint with Consul. Export-client does not require a
  database connection and so it will come up before export-distro. It
  was observed that the rules-engine registers with export-client in
  order to receive events. As a result of that, export-client attempts
  to POST a notification to export-distro. Because export-distro had not
  registered itself with Consul at the time export-client was being
  configured, the endpoint information would be blank, leading to an
  error like this (some fields omitted):
    level=ERROR app=edgex-export-client source=registration.go:310
    msg="error from distro: Put http://:0/api/v1/notify/registrations:
         dial tcp :0: connect: connection refused"
  Noticee the the host and port.
  -- The code change to address this can be found in
     internal/pkg/startup/endpoint.go
  -- This only works due to the non-HA mode in which we deploy from
     docker-compose. Right now, what's in the docker configuration is
     always the same as what is (or will be) in Consul.
  -- Note also that there is a circular dependency between distro/client
     w/r/t how they call each other. Because the result of a notification
     sent to distro in the above case is that distro simply calls back
     to export-client.


tsconn23:
@lenny-intel I want to make sure you saw the email I sent and have gone through the verification instructions therein. We need to be sure. Right now, AFAIK I'm the only one that has actually run the fix.

tsconn23:
> I still get “unhealthy” on the logging service via Portainer. Did you intend to leave that new indication in the logging service?

@jpwhitemn You must still be using that original docker-compose.yaml I sent you when we were troubleshooting then.

lenny-intel:
@tsconn23 , I was able to test this yesterday. After clean start (i.e. volume prune) the data is flowing to Rules Engine properly and all services are healthy except scheduler. If I then restart without prune the scheduler is healthy. When unhealthy, the only thing I see in the logs is:

```
level=ERROR ts=2019-07-17T21:59:36.673704006Z app=edgex-support-scheduler source=main.go:80 msg="edgex-support-scheduler: Service bootstrap failed!"
```
I reran this scenario multiple times and get the same result each time.

jpwhitemn:
@xmlviking - can you help take a look

    On Thursday, July 18, 2019, 9:23:42 AM CDT, Lenny Goodell <notifications@github.com> wrote:  
 
 
@tsconn23 , I was able to test this yesterday. After clean start (i.e. volume prune) the data is flowing to Rules Engine properly and all services are healthy except scheduler. If I then restart without prune the scheduler is healthy. When unhealthy, the only thing I see in the logs is:
level=ERROR ts=2019-07-17T21:59:36.673704006Z app=edgex-support-scheduler source=main.go:80 msg="edgex-support-scheduler: Service bootstrap failed!"

I reran this scenario multiple times and get the same result each time.

lenny-intel:
@tsconn23 , @jpwhitemn , the support schedule issue occurs every time do a clean start. I resolved the issue by adding `restart: unless-stopped` to scheduler in the compose file. Adding this to all the appropriate service would make bootstrap more robust.

xmlviking:
I'm testing this on my PC as well with the Edinburgh release versions of the compose files.  Trying to to trouble shoot any timing related issues with  Scheduler or potentially any of  default interval/intervalActions  or configured interval, intervalActions it's loading at boot strap.

Issue #318, opened 2018-06-23
Logging submodule under support module missed a Attribution.txt file


yanghua:
will add soon~

PR #1557, opened 2019-07-17
Relocate ValueDescriptor creation responsibility

### LINKED ISSUE #1471: Manage Value Descriptors Within Core-Metadata
Currently, value descriptors are [created](https://github.com/edgexfoundry/device-sdk-go/blob/9c16751eb33bbe0a682fe4b2aeae4f8a0e2e1fef/internal/provision/profiles.go#L138) within core-data by the Device Service SDK when a Device Profile is either [created](https://github.com/edgexfoundry/device-sdk-go/blob/9c16751eb33bbe0a682fe4b2aeae4f8a0e2e1fef/managedprofiles.go#L45) or [updated](https://github.com/edgexfoundry/device-sdk-go/blob/9c16751eb33bbe0a682fe4b2aeae4f8a0e2e1fef/managedprofiles.go#L118). This functionality should be moved out of the Device SDK and should be moved to core-metadata. Thus, when core-metadata is asked to [create](https://github.com/edgexfoundry/edgex-go/blob/6b4a598816fcfb16ccd31072baaf3109d13e93f7/internal/core/metadata/rest_deviceprofile.go#L48) or [update](https://github.com/edgexfoundry/edgex-go/blob/6b4a598816fcfb16ccd31072baaf3109d13e93f7/internal/core/metadata/rest_deviceprofile.go#L90) a DeviceProfile, it should also manage the ValueDescriptors in the same manner.

Creation of the Value Descriptors should happen through the same Client currently in use by the Device SDK. That way we are moving the responsibility while preserving the current API signature. 

_** NOTE: There is still a gap here w/r/t the deletion of a DeviceProfile. Currently, ValueDescriptors are not cleaned up when a DeviceProfile is deleted. This causes problems because names of ValueDescriptors are required to be unique and if they're left hanging around, they could cause collisions later when a new profile tries to re-use one. This also occurs during an update of a DeviceProfile when an individual DeviceResource gets removed. The corresponding ValueDescriptor is orphaned. This should be solved in a subsequent issue **after** the above functionality is ported. **_

Related issue https://github.com/edgexfoundry/device-sdk-go/issues/318 The work on these two issues should be either closely timed, or else a configuration toggle should be added in order to turn this behavior on/off prior to its removal, pending integration with the Device Services.

AnthonyMBonafide:
Fix #1471

Add functionality in the `restAddDeviceProfile` REST handler to not only
create DeviceProfiles, but also create the necessary associated
ValueDescriptors. This functionality was previously expected to be in
the Device SDK. Now, we are expecting the SDK functionality to be
updated to only create DeviceProfiles and not ValueDescriptors as
Core-Metadata will now handle that responsibility. See issue #318 for
more details regarding the removal of this functionality from the SDK.

Add new operator `ValueDescriptorAdder` which handles the responsibility
of creating and persisting the necessary ValueDescriptors given a
DeviceResources

Add tests for `ValueDescriptorAdder` operator.


tsconn23:
recheck

AnthonyMBonafide:
Recheck

tsconn23:
FYI -- Pending review, please do not merge this PR until after the Device Services WG call on 22-Jul-2019. We need to coordinate this effort with changes to the SDK.

AnthonyMBonafide:
Recheck

AnthonyMBonafide:
Recheck

AnthonyMBonafide:
Recheck

internal/core/metadata/operators/device_profile/value_descriptor_test.go
```diff
@@ -0,0 +1,149 @@
...
+var TestValueDescriptor2 = createValueDescriptorFromDeviceResource(TestDeviceResource2)
+var TestValueDescriptorError = createValueDescriptorFromDeviceResource(TestDeviceResourceError)
+var TestDeviceResources = []contract.DeviceResource{
+	TestDeviceResource1,
+	TestDeviceResource2,
+}
+
+var TestContext = context.Background()
+
+func Test(t *testing.T) {
```
tsconn23: Please make the function name more descriptive, such as `TestAddValueDescriptors`

internal/core/metadata/operators/device_profile/value_descriptor_test.go
```diff
@@ -0,0 +1,149 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			op := NewAddExecutor(TestContext, createMockValueDescriptorAdder(), logger.MockLogger{}, test.dr...)
+			err := op.Execute()
+
+			if test.expectError && err == nil {
+				t.Error("We expected an error but did not get one")
+			}
+
+			if !test.expectError && err != nil {
+				t.Errorf("We do not expected an error but got one. %s", err.Error())
```
tsconn23: `We did not expect...`

internal/core/metadata/operators/device_profile/value_descriptor.go
```diff
@@ -0,0 +1,72 @@
...
+	logger logger.LoggingClient
+}
+
+// Execute creates the necessary ValueDescriptors for a set of DeviceResources.
+func (a addValueDescriptor) Execute() error {
+	for _, dr := range a.drs {
+
+		desc := createValueDescriptorFromDeviceResource(dr)
+		id, err := a.client.Add(&desc, a.ctx)
+		if err != nil {
```
tsconn23: You might consider integrating the [ErrServiceClient](https://github.com/edgexfoundry/go-mod-core-contracts/blob/5f45a867aa5be00d1f023143816e99a17fae2b1b/clients/types/errors.go#L36) here so that the error returned from core-data is bubbled out to the caller adding the DeviceProfile.

AnthonyMBonafide:
As per my discussion with @tsconn23, we need to also do the following in this PR

1. Add a configuration setting so that we can enable/disable this newly added functionality.

1. Add similar functionality for the Update device profile API

internal/core/metadata/operators/device_profile/value_descriptor.go
AnthonyMBonafide: Won't the `Add` method return an `ErrServiceClient`? If my assumption is wrong then what should we set the `StatusCode` and/or `body` information?

cmd/core-metadata/res/configuration.toml
```diff
@@ -30,6 +30,11 @@ File = './logs/edgex-core-metadata.log'
   Protocol = 'http'
   Host = 'localhost'
   Port = 48060
+  [Clients.Data]
```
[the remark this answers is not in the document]
AnthonyMBonafide: I also forgot to add this to the Docker profile's `configuration.toml`. I will address that in my next iteration as well.

internal/core/metadata/operators/device_profile/value_descriptor.go
tsconn23: That's just one of several different error types you might expect from this call. Drilling down to [PostRequest](https://github.com/edgexfoundry/go-mod-core-contracts/blob/5f45a867aa5be00d1f023143816e99a17fae2b1b/clients/request.go#L103) I see two different error types that could be returned, and higher up the stack there are others.

Looking at this again, this area is fine. It's probably in `restAddDeviceProfile` where we need type checking in order to return the appropriate http status. The following `err != nil` check
```
op := device_profile.NewAddExecutor(r.Context(), vdc, LoggingClient, dp.DeviceResources...)
err = op.Execute()
if err != nil {
	return
}
```
For now I'd be satisfied if this just did a type check w/r/t `ErrServiceClient`. If the error is of that type, then return the status code and message wrapped by the error. Otherwise return a 500 as it does now.

internal/core/metadata/operators/device_profile/value_descriptor.go
```diff
@@ -0,0 +1,196 @@
...
+	for _, d := range deleted {
+		err = u.client.DeleteByName(d.Name, u.ctx)
+		if err != nil {
+			return err
+		}
+
+	}
+
+	for _, up := range update {
+		v, err := u.client.ValueDescriptorForName(up.Name, u.ctx)
```
AnthonyMBonafide: I really do not like doing this. If there is another way of doing this I would like to remove this. I thought about updating the `usage` endpoint in core-data to include the IDs in the response.

internal/core/metadata/operators/device_profile/value_descriptor.go
```diff
@@ -0,0 +1,196 @@
...
+	logger logger.LoggingClient
+}
+
+// UpdateValueDescriptorExecutor updates a value descriptor.
+type UpdateValueDescriptorExecutor interface {
+	Execute() error
+}
+
+// Execute updates a value descriptor with the provided information.
+func (u updateValueDescriptor) Execute() error {
```
AnthonyMBonafide: Normally when I see something simple require a lot of leg work in a micro-service architecture, like I am doing here, I look to see if we should either relocate a responsibility or create a new micro-service to handle it. I do not know if either of these are good solutions, but I am not the biggest fan of doing all this work for a simple update. This gets even more complicated if we try to ensure the state of the system during an error, like if we update ValueDescriptors and then fail to update the DeviceProfile. I feel like we cannot handle that situation elegantly with the current layout. Just a thought a had while doing this work.

PR #1559, opened 2019-07-17
Unit Test -- update addressable now uses operator pattern and has tests.

### LINKED ISSUE #1554: Unit Test -- Core-Metadata Update Addressables
Related to https://github.com/edgexfoundry/edgex-go/projects/11#card-22883576

brandonforster:
Fix #1554


jdharms:
recheck

internal/core/metadata/operators/addressable/update_test.go
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+package addressable
+
+import (
+	"github.com/edgexfoundry/edgex-go/internal/core/metadata/errors"
```
jdharms: Imports are slightly out of order

internal/core/metadata/rest_addressable.go
```diff
@@ -104,10 +106,9 @@ func restUpdateAddressable(w http.ResponseWriter, r *http.Request) {
 		LoggingClient.Error(err.Error())
 		return
 	}
-
-	w.Header().Set("Content-Type", "application/json")
```
jdharms: You're missing the content-type header in your change.

jdharms: Fails blackbox-test 333 in core-metadata/addressable collection.

jdharms:
recheck

brandonforster:
recheck

tsconn23:
recheck

brandonforster:
recheck

internal/core/metadata/operators/addressable/update.go
```diff
@@ -0,0 +1,106 @@
...
+
+	if err := op.writer.UpdateAddressable(dest); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// This factory method returns an executor used to update an addressable.
+func NewUpdateExecutor(loader AddressLoader, writer AddressWriter, addressable contract.Addressable) UpdateExecutor {
```
AnthonyMBonafide: Do you think we should create a new interface in the `db.go` file which encapsulates both the `AddressLoader` and the `AddressWriter`, that way we only need to pass in one type? You can see an example of this in the `device` operator for updating:
https://github.com/edgexfoundry/edgex-go/blob/73854f625f543dcd78e33e0be13d55dc32395032/internal/core/metadata/operators/device/db.go#L23-L29


Here we have created a new `DeviceUpdater` which has the `DeviceServiceLoader` and `DeviceProfileLoader` embedded within.

PR #1562, opened 2019-07-18
Removes ./docs/ folder which now resides in the edgex-docs repo

### LINKED ISSUE #1528: Remove 'docs' from edgex-go
Per the 10-Jul-19 TSC call, it sounds like we have confirmation that the content in the new [edgex-docs](https://github.com/edgexfoundry/edgex-docs) repo have been approved and are considered complete. In the 11-Jul-19 DevOps call, @ernestojeda confirmed that the docs-verify job has been removed from edgex-go builds. That means we can delete the entire `docs` directory from this project.

@mhall119 for visibility. Please +1 if you agree this can be worked.

mhall119:
Fixed #1528


mhall119:
I couldn't find any reference in the rest of edgex-go to the docs folder or build scripts, so it should be as simple as just removing this folder

tsconn23:
Wonderful! This will also make the `git clone` footprint so much smaller!

tsconn23:
@mhall119 Can you resolve the conflict on this? Unfortunately there was a PR we had to pull in for the dot release that has docs changes. Same will happen on master.

Issue #1504, opened 2019-07-05
Add support for username-password authentication and invalid (self-signed) certs


rudderfree:
Adding the relevant comment chain from PR of issue #1469  
AnthonyMBonafide 14 days ago • Member

I think we need to make the Username portion of the mail.PlainAuth function parameters configurable so that users can specify the credentials for their SMTP server. I think it is valid for someone to require a username and password in order to send mail. This would mean updating the SmtpInfo to include a Username field and then leveraging it here.

Just a side note. In the issue related to this pull request you asked about SMTP servers to use for testing. I have and still use a postfix docker container which you can find here. Under the usage section you can find commands to start the server. The nice thing is that using that container will result in an email actually being sent. I used this to verify your changes. I hope this helps

rudderfree:
https://github.com/edgexfoundry/edgex-go/pull/1469#issuecomment-507827960
https://github.com/edgexfoundry/edgex-go/pull/1469#issuecomment-508720894

rudderfree:
@AnthonyMBonafide Looks like we might have to do our own version of smtp.SendMail() to accomplish this, because there seems to be no way of passing InsecureSkipVerify or leveraging any other parameters that are being passed to pass the same information to smtp.SendMail()

AnthonyMBonafide:
I think you are correct and we will have to connect and/or send the mail via another mechanism. I poked around for a bit while verifying PR #1469 and was not able to find anything that would allow for this with the way we are doing things now. This will require some investigation to see how we can accomplish this and what is the best way to seamlessly incorporate it into our existing implementation.

rudderfree:
I have done a version of SendMail function, the function declaration looks like this.

`func sendMail(addr string, a mail.Auth, from string, to []string, msg []byte, skipVerify bool) error {`

I am in the process of pushing the branch, so that rest of the code can be seen.
I am using it as such in notifications.smtpSend
```
       err := sendMail(smtp.Host+":"+strconv.Itoa(smtp.Port),
                mail.PlainAuth("", smtp.Sender, smtp.Password, smtp.Host),
                smtp.Sender, addressees, []byte(buf.String()), false)
```
Notice the last parameter, that can be passed as the configuration parameter for skipping the verification, it is false to keep the existing functionality intact. also smtp.Sender is actually fulfilling the function of username

rudderfree:
@AnthonyMBonafide I have added the commit at the link below containing the code for the new sendMail function, please take a look, Thx!
[...]

PR #1565, opened 2019-07-19
Enable skipping verification of certs for emails

rudderfree:
Adding revised version of smtp.SendMail to enable support
for skipping invalid (self-signed) certificates.

Also addes ability to add username and configure self-signed
enablement

Issue Number: #1504

Signed-off-by: Amit Kumar Chaudhary <a.rudder.free@gmail.com>


rudderfree:
@AnthonyMBonafide 

Originally I had thought of submitting the infrastructure functions first and then bring in the rest of the feature in the next commit, with that in view I think it is at a good stage for PR. I am still working on understanding the support-notifications better from the db perspective and hope to ascertain what other parts may be relevant or not, to this feature.

Requesting the review, much appreciate your comments, Thanks!

internal/support/notifications/sending_service.go
```diff
@@ -141,3 +143,70 @@ func handleFailedTransmission(t models.Transmission) {
...
+	}
+	c, err := mail.Dial(addr)
+	if err != nil {
+		return errors.New("Notifications: Address dialing error")
+	}
+	serverName, _, err := net.SplitHostPort(addr)
+	if err != nil {
+		return err
+	}
+	defer c.Close()
```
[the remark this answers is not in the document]
rudderfree: Yes, done! great catch.

internal/support/notifications/sending_service.go
```diff
@@ -141,3 +143,70 @@ func handleFailedTransmission(t models.Transmission) {
...
+	if err = c.Hello(addr); err != nil {
+		return err
+	}
+	if ok, _ := c.Extension("STARTTLS"); ok {
+		config := &tls.Config{ServerName: serverName}
+		config.InsecureSkipVerify = skipVerify
```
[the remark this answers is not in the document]
rudderfree: Thanks, yes, I was trying to make it so that enabling this feature is an explicit activity.

internal/support/notifications/sending_service.go
```diff
@@ -141,3 +143,70 @@ func handleFailedTransmission(t models.Transmission) {
 		}
 	}
 }
+
+// validateLine checks to see if a line has CR or LF as per RFC 5321
+func validateLine(line string) error {
```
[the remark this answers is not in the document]
rudderfree: Yes, it makes sense to have this check at entry point so that if the mistake occurs it can be reported back right away, and can be correct at the source, but OTOH it serves well as a secondary check, will take a look at it. Thanks!

internal/support/notifications/sending_service.go
```diff
@@ -92,12 +95,12 @@ func smtpSend(message string, addressees []string) models.TransmissionRecord {
 	buf.WriteString("\r\n")
 	buf.WriteString(message)
 	if smtp.Password != "" {
```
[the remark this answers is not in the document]
rudderfree: Looking at it.

internal/support/notifications/sending_service.go
```diff
@@ -141,3 +143,70 @@ func handleFailedTransmission(t models.Transmission) {
...
+
+// validateLine checks to see if a line has CR or LF as per RFC 5321
+func validateLine(line string) error {
+	if strings.ContainsAny(line, "\n\r") {
+		return errors.New("smtp: A line must not contain CR or LF")
+	}
+	return nil
+}
+
+func sendMail(addr string, a mail.Auth, from string, to []string, msg []byte, skipVerify bool) error {
```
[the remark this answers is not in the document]
rudderfree: Working on it.

internal/support/notifications/sending_service.go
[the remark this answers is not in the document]
rudderfree: Thanks for pointing this, I had wondered about it. Created a new function called deduceAuth, this takes into account all the four combination of password and username and returns accordingly.

It is called directly by the function, now called smtpSend (earlier called sendMail, swapped the names, will comment about it separately). I am not sure what exactly should be done for the condition where username is present but not the password, I have taken it as an error to be on the safe side.

internal/support/notifications/sending_service.go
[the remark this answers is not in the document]
rudderfree: Agreed, but first I have re-named this function to smtpSend, and the original smtpSend to sendMail, basically swapped the names, because the smtpSend sounded like the name of a low level function, while sendMail sounded like a bit higher level function, I was constantly getting confused. I could equally keep them as is, but I liked them this way better.

I have moved SmtpInfo related processing to now called (smtpSend) by passing the SmtpInfo as a parameter, that led to shortening of parameter list. The processing of message content is still in the sender function (sendMail) because I thought it would be good to keep protocol communication separate from the message that is being sent. smtpSend handles the protocol part, while sendMail handles message processing and other things related to support-notification feature. Auth is now completely handled within the (now) smtpSend function.

internal/support/notifications/sending_service.go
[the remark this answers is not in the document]
rudderfree: Also, this time I have checked in the configuration.toml with EnableSelfSignedCert set to false.

internal/support/notifications/sending_service.go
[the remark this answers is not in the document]
rudderfree: Thanks, this commit moves it up.

internal/support/notifications/sending_service.go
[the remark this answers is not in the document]
rudderfree: I have moved this check up to the source, that is, to subscribe api, subscriptionHandler. If a CRLF pair is found the api responds back with the message that email addresses contains CRLF, and returns. It is in both PUT and POST methods. The function that implements it (validateEmail) probably can be enhanced further later to add more email validation. I have not been able to get to the check for addresses read from .toml file, code is more complicated there. I am inclining to handle that as a separate PR.

rudderfree:
@AnthonyMBonafide Thanks for the detailed review, really appreciate it. I have pushed the branch with the changes from the comments and a few more. Thanks!

internal/support/notifications/sending_service.go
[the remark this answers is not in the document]
rudderfree: I forgot to update the comment, it had only a nominal comment, expanded it to a more detailed one and updated the branch.

internal/support/notifications/rest_subscription.go
```diff
@@ -283,6 +303,15 @@ func splitVars(vars string) []string {
 	return strings.Split(vars, ",")
 }
 
+func validateEmail(c models.Channel) bool {
```
AnthonyMBonafide: Can we rename this function to `validateEmailAddresses` for better clarity? Also, could we make this return an `error` rather than a `bool` so that the caller does not have to construct an an `error` like what is happening on lines `107-108`

internal/support/notifications/sending_service.go
```diff
@@ -141,3 +135,101 @@ func handleFailedTransmission(t models.Transmission) {
...
+// first, a blank line, and then the message body. The lines of msg
+// should be CRLF terminated. The msg headers should usually include
+// fields such as "From", "To", "Subject", and "Cc".  Sending "Bcc"
+// messages is accomplished by including an email address in the to
+// parameter but not including it in the msg headers.
+//
+// The smtpSend provide no support for DKIM signing, MIME
+// attachments (see the mime/multipart package), or other mail
+// functionality.
+
```
AnthonyMBonafide: Can we remove this extra blank line?

AnthonyMBonafide:
@rudderfree where you able to manually verify these changes?

internal/support/notifications/rest_subscription.go
rudderfree: yes, good point, its duplication and there may be future users.

rudderfree:
> @rudderfree where you able to manually verify these changes?
Yes, I subscribed with an email address with CRLF and the api returned with the error message, that the email address contains CRLF pair.
For the email sending part, I was able to send to postfix with enabled self-signed certs, and to the local sendmail, where there is no authentication.

rudderfree:
> Added a few minor comments. My apologies for the delay in providing a review.

No problem, I know it takes time to catch up. Thanks for the review.

internal/support/notifications/rest_subscription.go
rudderfree: Updated the branch, I think validateEmailAddresses has attained its final form, in order to get the errors in I had to make few more changes, which makes it looks cleaner and more in harmony with the rest of the code.

internal/support/notifications/sending_service.go
rudderfree: Removed part of comment also, thought that is more suitable for an exported function.

rudderfree:
I have a lingering thought with this PR, that is, it is borrowing code from Go lang package. I have explained the reason and credited the source in the comment on top of the (code) function smtpSend. But in case there is something that may be needed done additionally in terms of crediting the source, I will be happy to do. Thanks!

rudderfree:
@AnthonyMBonafide Made the changes, fixed two bugs as well. Please review when you get a chance, TIA!

internal/support/notifications/rest_subscription.go
```diff
@@ -284,6 +298,20 @@ func splitVars(vars string) []string {
 	return strings.Split(vars, ",")
 }
 
+func validateEmailAddresses(s models.Subscription) error {
+	err := errors.NewErrMailAddrContainsCRLF(s.Slug)
+	for _, c := range s.Channels {
+		if c.Type == models.ChannelType(models.Email) {
+			for _, m := range c.MailAddresses {
+				if strings.ContainsAny(m, "\n\r") {
+					return err
```
AnthonyMBonafide: Could we return an error with all the email addresses which have failed the validation check. I think that would be helpful to both the client and during any troubleshooting that might take place while EdgeX is in use. For example, I created a notification with the following payload:
```
{
	"slug": "sys-admin3",
	"receiver": "System Administrator",
	"subscribedCategories": [
		"HW_HEALTH"
	],
	"subscribedLabels": [
		"temperature",
		"humidity"
    ],
	"channels": [
		{
			"type": "EMAIL",
			"mailAddresses": [
				"AnthonyMBonafide@gmail.com\r\n",
				"jack@example.com\n\r"
	    	]
		}
	]
}
```

and received the response body:

```
Subscription 'sys-admin3' mail addresses contain CRLF
```

Where a list of the emails would be nice. Maybe something like:

```
Subscription 'sys-admin3' mail addresses contain CRLF: [jack@example.com,...]
```

AnthonyMBonafide:
> I have a lingering thought with this PR, that is, it is borrowing code from Go lang package. I have explained the reason and credited the source in the comment on top of the (code) function smtpSend. But in case there is something that may be needed done additionally in terms of crediting the source, I will be happy to do. Thanks!

I think what you have is great. You expressed your intent and the source in your comments so that in the future we are aware of why this code exists. As for the crediting side of your comment, I do not think you need to do anything else. This looks great to me.

internal/support/notifications/rest_subscription.go
rudderfree: Yes, makes sense, will do.

rudderfree:
> > I have a lingering thought with this PR, that is, it is borrowing code from Go lang package. I have explained the reason and credited the source in the comment on top of the (code) function smtpSend. But in case there is something that may be needed done additionally in terms of crediting the source, I will be happy to do. Thanks!
> 
> I think what you have is great. You expressed your intent and the source in your comments so that in the future we are aware of why this code exists. As for the crediting side of your comment, I do not think you need to do anything else. This looks great to me.

Thanks! I wonder no more.

AnthonyMBonafide:
Recheck

rudderfree:
Recheck

internal/support/notifications/errors/types.go
```diff
@@ -39,3 +39,31 @@ func (e ErrSubscriptionNotFound) Error() string {
 func NewErrSubscriptionNotFound(slug string) error {
 	return ErrSubscriptionNotFound{slug: slug}
 }
+
+type ErrMailAddrContainsCRLF struct {
```
AnthonyMBonafide: Great call on making a specific error type for these types of errors. I like this approach. However, can we make this error type a little more general, so we can also use this for other email related errors we might encounter, like invalid format. We can possibly rename this to `ErrInvalidEmailAddresses`? If we do make this more general we can probably rename `slug` to `description` and in the case of the `CRLF` error we can state that the email contains those invalid characters.

internal/support/notifications/errors/types.go
```diff
@@ -39,3 +39,31 @@ func (e ErrSubscriptionNotFound) Error() string {
...
+	slug             string
+	errMailAddresses []string
+}
+
+func (e ErrMailAddrContainsCRLF) Error() string {
+	return fmt.Sprintf("Subscription '%s' mail addresses contain CRLF %v", e.slug, e.errMailAddresses)
+}
+
+func (e *ErrMailAddrContainsCRLF) AddErrMailAddrContainsCRLF(m string) {
```
AnthonyMBonafide: Can we remove this method so that this error is more consistent with the others in the project? I think we can expect the code that wants to construct this error to be able to construct a slice of the offending email addresses and simply provide that to the constructor function.

internal/support/notifications/rest_subscription.go
```diff
@@ -58,6 +58,13 @@ func subscriptionHandler(w http.ResponseWriter, r *http.Request) {
 		dec := json.NewDecoder(r.Body)
 		err := dec.Decode(&s)
 
+		// validate if email addresses contain CRLF pair
+		if err = validateEmailAddresses(s); err != nil {
```
AnthonyMBonafide: Should we add some error type checking here and return `http.StatusBadRequest` if the error is of type `ErrMailAddrContainsCRLF` otherwise return a `http.StatusInternalServerError`?

internal/support/notifications/rest_subscription.go
```diff
@@ -94,6 +101,13 @@ func subscriptionHandler(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+		// validate if email addresses contain CRLF pair
+		if err = validateEmailAddresses(s); err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
```
AnthonyMBonafide: Do we need do any type checking here as well(See my previous comment)?

rudderfree:
> Looking good overall. I think we just need to clean up the error types and add some error type checking and we should be good here

Agreed, I was thinking along similar lines, just didn't commit to it. Bringing in the changes. Thanks!

internal/support/notifications/errors/types.go
rudderfree: yes, done.

internal/support/notifications/errors/types.go
rudderfree: Thanks, I had spent couple minutes debating this myself. Done.

internal/support/notifications/rest_subscription.go
rudderfree: Yes, done.

internal/support/notifications/rest_subscription.go
rudderfree: Yes, both are identical in nature. Done.

rudderfree:
@AnthonyMBonafide pushed the branch with the above changes, please review when you get a chance. Thanks.

internal/support/notifications/rest_subscription.go
```diff
@@ -284,6 +310,29 @@ func splitVars(vars string) []string {
...
+		if c.Type == models.ChannelType(models.Email) {
+			for _, m := range c.MailAddresses {
+				if strings.ContainsAny(m, "\n\r") {
+					invalidAddrs = append(invalidAddrs, m)
+				}
+			}
+		}
+	}
+	if len(invalidAddrs) > 0 {
+		resp := "Subscription " + s.Slug + " mail addresses contain CRLF: ["
```
AnthonyMBonafide: If my previous comment is valid then we can eliminate this in favor of the `ErrInvalidEmailAddresses` type handling the error message formatting.

internal/support/notifications/errors/types.go
```diff
@@ -39,3 +39,15 @@ func (e ErrSubscriptionNotFound) Error() string {
 func NewErrSubscriptionNotFound(slug string) error {
 	return ErrSubscriptionNotFound{slug: slug}
 }
+
+type ErrInvalidEmailAddresses struct {
```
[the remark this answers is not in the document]
rudderfree: Thank you very much, I can see you had suggested this in one of your yesterday's comment, but I managed to miss and jumped to the conclusion that I had thought about earlier. Sorry about that.
This helps keep both the parts clean and easy to follow, instead of 'either-or' situation it was evolving into. Made the changes.

internal/support/notifications/rest_subscription.go
rudderfree: Yes, happy to remove. Done.

rudderfree:
@AnthonyMBonafide Thank you very much for your detailed reviews, it helps me a lot in getting the code into right shape.

AnthonyMBonafide:
@tsconn23 , This looks good to me and I think it is ready to be merged. Any objections?

cmd/support-notifications/res/configuration.toml
```diff
@@ -39,11 +39,10 @@ File = './logs/edgex-support-notifications.log'
...
-
-
-
+  Host = 'smtp.gmail.com'
+  Username = 'username@mail.example.com'
+  Password = 'password'
+  Port = 587
+  Sender = 'jdoe@gmail.com'
+  EnableSelfSignedCert = false
+  Subject = 'EdgeX Notification'
```
tsconn23: This same modification needs to be made in the `res/docker/configuration.toml`

internal/support/notifications/sending_service.go
```diff
@@ -141,3 +135,87 @@ func handleFailedTransmission(t models.Transmission) {
...
+	if s.Username == "" && s.Password == "" {
+		return nil, errors.New("Notifications: Expecting username")
+	}
+	if s.Username != "" && s.Password == "" {
+		return nil, nil
+	}
+	if s.Username == "" && s.Password != "" {
+		return nil, errors.New("Notifications: Expecting username")
+	}
+	return mail.PlainAuth("", s.Username, s.Password, s.Host), nil
```
tsconn23: I have a problem with this function because it isn't fully backward compatible. The previous implementation used `s.Sender` as part of the mail.PlainAuth call and not `s.Username`. If `s.Username` is blank, ideally this should fall back to `s.Sender` since that's what the Edinburgh release uses.

Consider adding the following function to the `SmtpInfo` type and then using that in the checks and `PlainAuth` above.
```
func(s SmtpInfo) CheckUsername() string {
	if s.Username != "" {
		return s.Username
	}
	return s.Sender
}
```

AnthonyMBonafide: Good point.

@rudderfree , can you also add comments stating that the function exists for backwards compatibility, so that later on we are not confused as to the intent of the logic?

cmd/support-notifications/res/configuration.toml
rudderfree: Thanks for noticing, added.

internal/support/notifications/sending_service.go
rudderfree: Yes, agreed. Added the function and the comment. Thx!

Issue #1525, opened 2019-07-11
Unit Test - Core-Metadata: Get Command by Name


difince:
Related to https://github.com/edgexfoundry/edgex-go/projects/11#card-23264372

difince:
could you assign it to me

difince:
thanks, @tsconn23

PR #1566, opened 2019-07-19
Unit Test -- Metadata getCommandsByName use Operator Pattern

difince:
Fix: https://github.com/edgexfoundry/edgex-go/issues/1525


internal/core/metadata/operators/command/mocks/CommandsByNameExecutor.go
```diff
@@ -0,0 +1,34 @@
+// Code generated by mockery v1.0.0. DO NOT EDIT.
```
jdharms: Do we need a mock for the Executor?  I feel like this along with the other two mock Executor files can be removed from /operators/command/mocks/

difince: No, we don't . Removed

Issue #1555, opened 2019-07-17
core-metadata.raml causing edgexfoundry/edgex-docs build failure


michaelestrin:
# 📚 Docs or Wiki Bug Report

### Description

core-metadata.raml is consumed by the edgexfoundry/edgex-docs build pipeline.  #1523 introduced defects into the raml file that cause the edgex-docs build to fail.

michaelestrin:
Related pull request with failing build - https://github.com/edgexfoundry/edgex-docs/pull/12

PR #1568, opened 2019-07-19
add new raml verification target

ernestojeda:
This PR adds a `raml_verify` target for use within the ci-mangement Jenkins build. I did not add it directly to the `test` target due to the fact that it requires docker and that it downloads the edgex-docs-builder image which is quite large. Local users can manually invoke the `raml_verify` target if they have docker installed and feel the need to.

The `--privileged` flag is needed for Jenkins to access the bind mounted workspace.


ernestojeda:
Related https://github.com/edgexfoundry/edgex-go/issues/1555

Issue #1564, opened 2019-07-18
Able to send query parameters in GET command requests to devices


pregit:
# 🚀 Feature Request

### Relevant Package
core-command

### Description
Should be able to pass query parameters in command ***GET*** requests to devices.
Example: "/api/v1/device/{deviceID}/command/{commandID}?param1=value&param2=value"

lenny-intel:
FYI, @rsdmike  and I did a quick POC to see what the API impact would be. This can be accomplished without breaking REST API.   Device SDK would have to pass the query parameters to HandleReadCommands without changing the function signature, so we put the the query parameters in CommandRequest.Attributes. @rsdmike will submit the PR when he returns from vacation next week.

tsconn23:
What is a practical example for `param1=value&param2=value` that is not satisfied by the current API?

pregit:
command: "get_inventory"
facility=backstock&facility=fitting room

tsconn23:
> command: "get_inventory" where facility=backstock

That's not enough context to understand your use case. What do inventory and backstock mean in the case of an IOT solution?

antoniomtz:
Hello,

To expand more on this feature request. We have a device that exposes multiple methods through jsonrpc . We were able to register some of those methods that don't require parameters as _EdgeX commands_ and using the _mqtt-device-service_ to communicate with the device.

For example: 

**Request:**
```
{
 "jsonrpc" : "2.0",
 "method" : "sensor_get_device_ids"
}
```
**Response:**
```
{
 "jsonrpc" : "2.0",
 "result" : [ "sensor001", "sensor002"]
}
```
Using the current EdgeX Commands API, we are able to execute such command by making a request to the registered GET command endpoint:

`http://edgex-core-command:48082/api/v1/device/{deviceID}/command/{commandID}`

However, some of the exposed methods from our device require parameters, for example:
**Request:**
```
{
 "jsonrpc" : "2.0",
 "method" : "sensor_get_basic_info",
 "params" : {
   "device_id" : "sensor001"
 }
```
**Response:**
```
{
 "jsonrpc" : "2.0",
 "result" : {
    "device_id" : "sensor001",
    "status": "something",
    "x" : "y"
 }
```
This cannot be accomplished by the current EdgeX commands API since we are not able to pass either URL params or Query params to the Edgex GET command endpoints. Only PUT command endpoints support params in the body, but they do not return a response back. 

[...]

PR #1571, opened 2019-07-22
feature(query-params): Pass QueryParams through EdgeX to Device Services

rsdmike:
This pull request is an example/POC that @lenny-intel and I put together of how #1564 could be accomplished. This is only 1/2 of the equation as the device-sdk would need to be updated to handle the query parameters. The good thing with this is that it wouldn't be a breaking change as the device-sdk would simply ignore the query parameters that are sent on.

Signed-off-by: Mike Johanson <michael.johanson@intel.com>


tsconn23:
@ernestojeda I assume you're working on the ARM build?
```
13:31:26 standard_init_linux.go:190: exec user process caused "exec format error"
13:31:26 Makefile:171: recipe for target 'raml_verify' failed
13:31:26 make: *** [raml_verify] Error 1
13:31:26 Build step 'Execute shell' marked build as failure
```

PR #1572, opened 2019-07-22
add ARCH to makefile to pull correct image for RAML verify

ernestojeda:
For add ARCH to pull node architecture to support both ARM and x86_64 architectures for the RAML verification step.


ernestojeda:
recheck

Issue #1573, opened 2019-07-22
Eliminate Literals from API Endpoint Responses


tsconn23:
We currently have API endpoints that return string literals instead of a formal response object. Here are some examples:

- GET <any service>/api/v1/ping returns `pong`
- POST core-data/api/v1/event returns `id of new event`
- POST core-data/api/v1/valuedescriptor returns `id of new value descriptor`
- PUT core-metadata/api/v1/device returns `true`
- DELETE core-metadata/api/v1/device/id/{id} returns `true`
- Errors when they occur write out a literal string to the caller.

Instead, API responses should be well formatted types that can be unmarshaled by the caller. In cases where we need to indicate success for an operation that returns no content, rather than a literal "true" we should be using an HTTP status code to indicate that success.

I'm creating this issue to track the desire for a complete audit of the services in edgex-go w/r/t their responses. I would like to see these changes included for the Geneva release.

michaelestrin:
Closed.  Expect this to be addressed by the [Geneva v2 API](https://github.com/tsconn23/edgex-geneva-api) changes.

PR #1577, opened 2019-07-22
Update GetDeviceProfile APIs to use operators

### LINKED ISSUE #1530: Refactor the get device profile rest handlers to use operator pattern
# 🚀 Feature Request

### Relevant Package
Core-metadata

### Description

As part of the Unit test refactoring effort, refactor the REST handler endpoints responsible for getting device profiles to leverage the new operator pattern and include tests for all newly added operators.

Update the REST handler endpoints to use operators rather than the DBClient so that we can better leverage mocks to unit test the database components and handler function. 

N/A this effort was reviewed and accepted.

AnthonyMBonafide:
Fix #1530

Update all of the GetDeviceProfileBy* APIs to use operators.

Create new operators for retrieving DeviceProfiles for the various Get
APIs.

Add tests for the newly added operators and the updated REST handlers.


AnthonyMBonafide:
I want to make a few notes on some things I noticed:

1. These Get APIs do not seem consistent when it comes to error handling. For example, some will return a `500(Internal Server Error)` if no device profile can be found(restGetProfileByModel, restGetProfileByLabel, etc), while others will return a `404(Not Found)` under the same circumstances(restGetProfileByName, restGetProfileById, etc). 

1. These APIs do not consistently conform to the `MaxResultCount` set in the `ConfigurationStruct`. For example, `restGetAllDeviceProfiles` performs the check to ensure the limit is not exceeded, but other APIs which can return multiple results do not perform this check(restGetProfileByModel, restGetProfileByLabel, etc).

I know that even if these inconsistencies should be adjusted, we will not do this here as it is a change to the API. I am wondering if I should create an issue to fix not only these APIs but others to be more consistent so we can address this in a major release where backwards compatibility is less of an issue.

tsconn23:
> I know that even if these inconsistencies should be adjusted, we will not do this here as it is a change to the API. I am wondering if I should create an issue to fix not only these APIs but others to be more consistent so we can address this in a major release where backwards compatibility is less of an issue.

I made a similar issue yesterday w/r/t our API returning literals https://github.com/edgexfoundry/edgex-go/issues/1573. What I'd suggest is create an issue that targets one broad class of issues (such as `MaxResultCount`). Provide a description of what problem we're trying to solve and then as you find them, add items to a bullet list in the issue. Tag the issue with "geneva". I'm pretty confident the Geneva release will be another major version rev.

Issue #1558, opened 2019-07-17
Add security-pki-init into EdgeX-Go


jim-wang-intel:
To add component security-pki-init into edgex-go mono repo instead of adding it into security-secret-store repo/project.

tingyuz:
This issue is being tracked under PR #1747.

tingyuz:
PR #1747, #1794, #1830 addressed this issue. Close.

PR #1578, opened 2019-07-22
Add security pki-init 

jim-wang-intel:
This implements the PKI-Init component and address the issue #1558.  

The high level design document can be found here: https://github.com/bnevis-i/security-secret-store/blob/bnevis/threat-model-2/doc/threat-model/high_level_design.md#pki-initialization  

The usage and brief readme on this component can be found from README.md here: https://github.com/jim-wang-intel/edgex-go/blob/add-security-pki-init/internal/security/pkiinit/README.md


jim-wang-intel:
@tingyuz @mkbhanda  @tsconn23 please assign reviewers. Thanks.

Makefile
```diff
@@ -64,10 +64,13 @@ cmd/support-scheduler/support-scheduler:
...
 	$(GO) build $(GOFLAGS) -o ./cmd/security-secrets-setup/security-secrets-setup ./cmd/security-secrets-setup
 
+cmd/security-pki-init/security-pki-init:
+	$(GO) build $(GOFLAGS) -o ./cmd/security-pki-init/security-pki-init ./cmd/security-pki-init
+
 clean:
 	rm -f $(MICROSERVICES)
 
-test:
+test: cmd/security-secrets-setup/security-secrets-setup
```
tsconn23: Would it be feasible to refactor the code in order to have a mock PKI generator that does not require the unit tests to call an actual instance of this service? This would remove the need to add this to the build target.

jim-wang-intel: Yea, we can look into that.

internal/security/pkiinit/cert/x509Config.go
```diff
@@ -0,0 +1,129 @@
...
+	"path/filepath"
+)
+
+const (
+	skFileExt   = ".priv.key"
+	certFileExt = ".pem"
+)
+
+// RootCA parameters from JSON: x509_root_ca_parameters
+type RootCA struct {
```
tsconn23: These struct types are already declared here:
https://github.com/edgexfoundry/edgex-go/blob/master/internal/pkg/config/x509.go

You may be able to re-use some of the file access code from there as well.

internal/security/pkiinit/cert/x509Config_test.go
```diff
@@ -0,0 +1,128 @@
+//
```
tsconn23: Perhaps incorporate this file into https://github.com/edgexfoundry/edgex-go/blob/master/internal/pkg/config/x509_test.go

internal/security/pkiinit/option/cache.go
```diff
@@ -0,0 +1,96 @@
...
+
+func cachePkis(cacheca bool) (statusCode exitCode, err error) {
+	// generate a new one if pkicache dir is empty
+	pkiCacheDir := getPkiCacheDirEnv()
+	cachedCAPrivateKeyFile := filepath.Join(pkiCacheDir, caServiceName, tlsSecretFileName)
+	if empty, err := isDirEmpty(pkiCacheDir); err != nil {
+		return exitWithError, err
+	} else if empty {
+		// perform generate func
+		statusCode, err = generatePkis()
```
tsconn23: Abstracting this capability would allow for the `make test` target to not have to build the `security-secrets-setup` service

jim-wang-intel: Good call, I am thinking towards that direction of refactory as well- but not the whole `generatePkis()` because that could greatly reduce the test coverage due to the mocked one.  Maybe abstracting out the calling of real binary action part to remove that `make test` dependency issue as one solution that it can be used in unit tests.

internal/security/pkiinit/cert/x509Config_test.go
jim-wang-intel: Ok, we'll see how we can incorporate these tests.

internal/security/pkiinit/option/cache.go
jim-wang-intel: @tsconn23 Trevor, I have done the code re-factory regarding the pkisetup build dependency issue of `make test`.  Take a look and move to another issue you mentioned about re-uses some of the existing codes.

internal/security/pkiinit/cert/x509Config.go
jim-wang-intel: @tsconn23 done!

internal/security/pkiinit/cert/x509Config_test.go
jim-wang-intel: @tsconn23  yes, I have incorporated the tests into the one under pkg/config/x509_test.go and pushed the changes.  Please review it again.  Thanks for your feedback!

jim-wang-intel:
Good feedback everyone! @anonymouse64 Ian, I created an issue, issue #1635,  based on your feedback and we should approach this like we had discussion in the F2F meeting in SC.

anonymouse64:
Sounds great @jim-wang-intel, feel free to close this PR in that case and request me as a reviewer on followup PR's.

tsconn23:
Closing per discussion in the Core WG 8-Aug-2019. A redesigned approach has been decided upon that will be reflected in a series of new PRs.

PR #1584, opened 2019-07-24
add docker- to the sys-mgmt-agent-go image name

ernestojeda:


MightyNerdEric:
recheck

MightyNerdEric:
It looks like the verify job failed just due to a network blip that Go couldn't handle. This is an unfortunately common problem; hopefully it will pass on recheck.

PR #1591, opened 2019-07-25
Add New ValueDescriptor Usage Endpoint

### LINKED ISSUE #1583: Add Core-Data Endpoint to Check Whether ValueDescriptor Can be Deleted
In the case where core-metadata is responsible for managing value descriptors related to a device profile update, we do not have the luxury of doing reconciliation of value descriptors along with the updated profile in a transactional fashion. For example, an updated device profile may have a different set of deviceResources than the original. DeviceResources not found on the new definition of the profile should be removed. It is undesirable to first update the profile and then try to manage the descriptors only to have one of them throw an error because it is associated with readings that have not been purged from core-data. In this case the profile's state would have been updated and the descriptors will be out of alignment.

@AnthonyMBonafide and I discussed adding an endpoint to core-data that would allow for a pro-active check of whether or not prospective descriptors can be deleted before we start making changes to the data itself. The proposed endpoint and request/response format looks like this.

Request
`GET http://edgex-core-data:48080/api/v1/valuedescriptor/usage?names=abc,def,ghi`

Response
`[{"abc":true}, {"def":true}, {"ghi":false}]`

This will allow us to check all of the value descriptors associated to a given device profile that are candidates for deletion with one round trip to core-data. If any of the names are associated with a `true` value in the response, then core-metadata's update device profile response should be equivalent to the one found [here](https://github.com/edgexfoundry/edgex-go/blob/9a1996e0f0100ca36f26da262eda24085a84676e/internal/core/data/valuedescriptor.go#L306).

_** Don't forget to add this new endpoint to the core-data.raml **_

The current inconsistencies related to this scenario were discussed on [this issue](https://github.com/edgexfoundry/edgex-go/issues/1071) and also during several sessions of the Core WG calls.

AnthonyMBonafide:
Fix #1583

Add new endpoint to Core-Data which will check if ValueDescriptors are
in use by any readings.

Add new operators for retrieving ValueDescriptors and Readings

Add tests to cover new endpoint and new operators

Update RAML to include the new endpoint


internal/core/data/rest_value_descriptor.go
```diff
@@ -0,0 +1,439 @@
...
+}
+
+// restValueDescriptorsUsageHandler checks if value descriptors are currently being used.
+// This functionality is useful for determining if a value descriptor can be updated, or deleted.
+// This functionality does not provide any guarantee that the value descriptor will not be in use in the near future.
+// Any functionality using the check to perform updates or deletes is responsible for handling any race conditions which
+// may occur.
+// Returns a map[string]bool where the key is the ValueDescriptor Name and the value is a bool stating if the
+// ValueDescriptor is currently in use.
+func restValueDescriptorsUsageHandler(w http.ResponseWriter, r *http.Request) {
```
AnthonyMBonafide: I took the liberty of breaking up the `router.go` file to follow the other services. This function is the only thing that is new, everything else has been moved over from `router.go`.

internal/core/data/rest_value_descriptor.go
```diff
@@ -0,0 +1,439 @@
...
+		default:
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+		}
+
+		return
+	}
+
+	// Use this data structure so that we can obtain the desired JSON format. Please see RAML for response format
+	// information.
+	resp := make([]map[string]bool, 0)
```
AnthonyMBonafide: I had to do this in order to get the desired response format. I am not 100% a fan of this. If someone can suggest something else, I'd like to readdress this.

internal/core/data/router.go
```diff
@@ -1133,352 +1134,6 @@ func readingByValueDescriptorAndDeviceHandler(w http.ResponseWriter, r *http.Req
 	pkg.Encode(readings, w, LoggingClient)
 }
 
-// Value Descriptors
```
AnthonyMBonafide: See my other comment  regarding moving the REST handlers to another file.

internal/core/data/rest_value_descriptor.go
tsconn23: Yeah, I'm not %100 onboard with this. With this PR we've now got an inconsistency in core-data B/C value descriptors have their own implementation style that differs from events and readings. I see that you're going for a style similar to core-metadata, but now it's inconsistent until we create an issue to go back in and make it consistent. I wouldn't include this change in the PR.

internal/core/data/rest_value_descriptor.go
tsconn23: `var resp []map[string]bool`

internal/core/data/router.go
AnthonyMBonafide: Disregard this comment. These changes have been reverted as per @tsconn23 's comments.

internal/core/data/rest_value_descriptor.go
tsconn23: @AnthonyMBonafide You know what? The previous way you had this defined was better because now if there are no results a literal "null" is returned in the response instead of the empty map.

AnthonyMBonafide: Good catch! I missed that.

Issue #1596, opened 2019-07-29
Add new endpoint for obtaining service version


iain-anderson:
# 🚀 Feature Request

### Relevant Package

 This feature request is for...
All supported services

### Description

The earlier intention was to provide version information via the ping endpoint. However this did not land in Edinburgh and as Fuji will not change existing APIs, ping will continue to return "pong" until at least Geneva.
In lieu of the above, provide a new endpoint "version" which will return the version in a simple JSON response type.
`{"version":"v1.1.0"}`
This issue obsoletes #1449

tsconn23:
Question -- Should this endpoint be served from the root rather than as `/api/v1/version`?

The reason is that in v2, the v1 api endpoints will still be there for backward compatibility. We won't want to dual implement a version endpoint (e.g. `/api/v1/version` and `/api/v2/version` and so on) and it seems odd that `/api/v1/version` would return `{"version":"v2.1.3"}`.

So possibly this route should be `/api/version`

anonymouse64:
+1 for `/api/version`

PR #1599, opened 2019-07-30
Add 'version' REST endpoint for edgex-go services

iain-anderson:
Implements #1596


iain-anderson:
I've not updated the RAML for this. Not sure how best to do so given our baseUri usage.

tsconn23:
> I've not updated the RAML for this. Not sure how best to do so given our baseUri usage.

I'll put in an issue for this. The `baseUri` needs to be adjusted anyway if we're going to start adding /v2 endpoints for Geneva (depending on how the Swagger effort goes). Then that entails a larger edit for the service endpoints to include /v1. I'll create an issue and we'll take care of it there.

Issue #1486, opened 2019-06-26
Implement the design: SMA to get Metrics via Executor.


akramtexas:
- This issue tracks the _implementation_ of the `design` that was created previously per [Issue#1414](https://github.com/edgexfoundry/edgex-go/issues/1414)
- The goal here is to refactor metrics collection - moving metrics collection to the executor so that it can remain platform and even service implementation agnostic.
- Effectively, this is a direct follow-up to: [Issue #1414](https://github.com/edgexfoundry/edgex-go/issues/1414)

michaelestrin:
@anonymouse64 

I agree with your assessment (made on several abandoned PRs related to the system executor) that we should have a standard contract for metrics and, further, that we should also return the raw metrics for optional consumption.

The Fuji road map specifies:

> Refactor metrics collection - moving metrics collection to the executor so that it can remain platform and even service implementation agnostic.

I believe this implies the executor will become the source of metric collection and the existing service-based metric collection implementation and endpoints will be removed.  I discussed this with @akramtexas this morning and I'll be working to obtain clarity from @jpwhitemn on this point when he returns from his vacation next week.

anonymouse64:
@michaelestrin the discussion I had with Jim and Akram was that we would keep the service-based metric collection as an option for non-docker based implementations such as the snap or native process deployments, etc. So you can pick which implementation to use from the configuration.

michaelestrin:
@anonymouse64 

That makes sense to me.  Services in non-containerized environments would be the likeliest source for metrics.  You could obtain them by querying a separate OS-specific process (i.e. top) but I'm not sure why one would take on the added complexity.

Which raises a separate but related question -- why not continue to use the service-based metric collection implementation for Docker-based deployments (i.e. what is the value of having the executor handle metrics collection for Docker-based deployments)?  

I appreciate everyone's patience.  I'm not trying to be difficult or to change the existing agreements or design.  I'm just trying to understand why we're doing what we're doing the way we're doing it.

anonymouse64:
TBH, I don't recall the reasoning for creating a new docker-specific implementation... I think we should have 2 choices for collecting metrics:
- ask the service through the service endpoints, like what we do today
- call out to an arbitrary binary that responds in JSON or some other structured format

The second option allows folks running inside different distribution mechanisms or where they have more accurate or useful metrics to implement their own and use it instead.

michaelestrin:
[...]

PR #1609, opened 2019-07-31
Refactor the fetch metrics operation for services to use the Executor.

akramtexas:
Fix [#1486](https://github.com/edgexfoundry/edgex-go/issues/1486) 
- I fully expect making updates to the code submitted for review in this PR. The main thing is that I want to have this baseline code (i.e. per this PR) reviewed so we are on the same page, and thereby avoid optimizing  the implementation injudiciously.
- But first, and as brought up during the discussion following my demo of this functionality during the System Management Working Group meeting yesterday, what about RAML – how will the choice of metric be documented? We will need to provide some examples. 
- Please make note that the preceding is not just a _which_-executor issue. Different metrics could come from a _C_ SDK-based service vs a _Go_ SDK-based service, for example (Also make a note here of the detailed discussion on _Metrics Keys_ we had a while ago during another System Management WG meeting).
- To (hopefully) help in the area of RAML code-generation, I have added more comments to the TOML configuration files (for sys-mgmt-agent) than I normally would. In a nutshell, it would be helpful to look at some sample results from invoking the metrics API, when running with the configuration setting for the "MetricsMechanism" property with the respective value:
```
# The MetricsMechanism can be one of the following three options, proceeding which is a non-operational default:
MetricsMechanism = 'custom'
# MetricsMechanism = 'executor'
# MetricsMechanism = 'direct-service'
```
- To illustrate some of this variability,  I invite you to check out some sample results of calling the SMA's metrics API: `http://localhost:48090/api/v1/metrics/edgex-core-command,edgex-support-notifications`.

**[1]** A GET response per `docker stats`, i.e. when MetricsMechanism = 'executor':
```
{
    "Metrics": {
        "edgex-core-command": {
            "BLOCK I/O": "0B/0B",
            "CPU %": "0.00%",
            "MEM %": "0.21%",
            "MEM USAGE / LIMIT": "4.238MiB/1.952GiB",
            "NET I/O": "222kB/131kB",
            "PIDS": "14"
        },
        "edgex-support-notifications": {
            "BLOCK I/O": "0B/0B",
            "CPU %": "0.05%",
            "MEM %": "0.20%",
            "MEM USAGE / LIMIT": "3.973MiB/1.952GiB",
            "NET I/O": "126kB/111kB",
            "PIDS": "13"
        }
    }
}
```
**[2]** A GET response per the GeneralClient, i.e. when MetricsMechanism = 'direct-service':
```
{
    "Metrics": {
        "edgex-core-command": {
            "CpuBusyAvg": 2.766726943942134,
            "Memory": {
                "Alloc": 477584,
                "Frees": 67704,
                "LiveObjects": 2914,
                "Mallocs": 70618,
                "Sys": 72284408,
                "TotalAlloc": 19326672
            }
        },
        "edgex-support-notifications": {
            "CpuBusyAvg": 2.6968325791855206,
            "Memory": {
                "Alloc": 1139120,
                "Frees": 109434,
                "LiveObjects": 11182,
                "Mallocs": 120616,
                "Sys": 72284408,
                "TotalAlloc": 8023136
            }
        }
    }
}
```
**[3]** A GET response per snap, for example, i.e. when MetricsMechanism = 'custom':
```
{
    "Metrics": {
        "edgex-core-command": "the requested custom executor (e.g. snap) has not been integrated",
        "edgex-support-notifications": "the requested custom executor (e.g. snap) has not been integrated"
    }
}
```
- For snap, I believe that we should be able to use the following TOML setting: ExecutorPath = '/path/to/the/file' in conjunction with snap-specific directives such as:
```
      # for sys-mgmt-agent we also need to specify the operations type as "snap"
      cat "./cmd/sys-mgmt-agent/res/configuration.toml" | \
        sed -e s:./logs/edgex-sys-mgmt-agent.log:\$SNAP_COMMON/sys-mgmt-agent.log: \
            -e s:"ExecutorPath = '/path/to/the/file'":"ExecutorPath = '\$SNAP/bin/sys-mgmt-agent-snap-executor.sh'": \
            -e s:'http\://localhost\:48061/api/v1/logs':: > \
       "$SNAPCRAFT_PART_INSTALL/config/sys-mgmt-agent/res/configuration.toml"
```
- I extracted a number of helper functions out of code that processes the output return from an execution of the `docker stats` command. Those helper functions are, in turn, called from the following function: `func processOutput(bytes []byte, service string) ([]byte, error) { ... }`.
- With an eye on assembling _the simplest thing that works_, I want to add a (hopefully) clarifying comment about the following function: `func populateMap(fields [][]string) map[string]Stats { ... }`. The method to the madness in the following admittedly ungainly assignments...
```
func populateMap(fields [][]string) map[string]Stats {
  data := make(map[string]Stats)

  // Populate the map ("data") with elements of the console output that are
  // retrieved after running the "docker stats" command.
  size := len(fields)
  for i := 0; i < size; i++ {
    data[fields[i][1]] = Stats{
      fields[i][2],
      fields[i][3] + fields[i][4] + fields[i][5],
      fields[i][6],
      fields[i][7] + fields[i][8] + fields[i][9],
      fields[i][10] + fields[i][11] + fields[i][12],
      fields[i][13],
    }
  }
  return data
}
```
...is a reflection of how the individual data-bearing columns are returned from an execution of the `docker stats` command:
```
CONTAINER ID        NAME                          CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O           PIDS
```
...and which I was able to access as the following individual elements (note how even a single column is at times straddling three individual sub-fields; oh, the fun!), for, in turn, further processing to report back to the client in JSON format:
```
  0 = {string} "28764318732d"
  1 = {string} "edgex-core-command"
  2 = {string} "0.01%"
  3 = {string} "4.832MiB"
  4 = {string} "/"
  5 = {string} "1.952GiB"
  6 = {string} "0.24%"
  7 = {string} "8.68MB"
  8 = {string} "/"
  9 = {string} "5.04MB"
  10 = {string} "10.1MB"
  11 = {string} "/"
  12 = {string} "0B"
  13 = {string} "16"
  0 = {string} "64568e54fe61"
```
- At this time, an invocation of the SMA's metrics API is monumentally slow for the case when the underlying implementation behind the interface happens to involve an invocation, in turn, of the `docker stats` command, taking on the order of six entire seconds. 
- The bulk of the delay comes directly from the unavoidable delay that I had to introduce (`timer := time.NewTimer(time.Second * 3)`) in one of the functions (in the executor itself) that works with capturing stdout:
```
    // Create a timer that will kill the process
    timer := time.NewTimer(time.Second * 3)
    go func(timer *time.Timer, ticker *time.Ticker, cmd *exec.Cmd) {
      for range timer.C {
        //err := cmd.Process.Signal(os.Kill)
        cmd.Process.Signal(os.Kill)
        //LoggingClient.Error(fmt.Sprintf("an error occurred in terminating process: %v ", err.Error()))
        ticker.Stop()
      }
    }(timer, ticker, cmd)
    stderr, errStderr = copyAndCapture(os.Stderr, stderrIn)
```
- Also at this time, we programmatically run the command `docker stats`, and then selectively pick out the relevant services specified in the client request. It turns out that the command `docker stats` can also be run selectively, like so: `docker stats edgex-core-command edgex-export-client`
```
CONTAINER ID        NAME                  CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O           PIDS
466fa006eabe        edgex-core-command    0.30%               4.461MiB / 1.952GiB   0.22%               1.45MB / 843kB      10.2MB / 0B         14
a540da9c7a5c        edgex-export-client   1.70%               4.566MiB / 1.952GiB   0.23%               1.19MB / 854kB      12.2MB / 0B         15
```
- Using that selective approach (above) may help shave off a fraction of a second, or more; this needs to be investigated ⛏ 
- Note that an invocation of the `docker stats` command does not return immediately, and instead takes nearly a second to "stabilize" on stdout. With a view to this, I put together a mechanism to work with capturing stdout asynchronously.
- This leads me to make another observation: The way I terminate an invocation of the `docker stats` command is by creating a timer that will kill the spawned process via an O/S-level signal. The goroutine snippet is like so:
```
    timer := time.NewTimer(time.Second * 3)
    go func(timer *time.Timer, ticker *time.Ticker, cmd *exec.Cmd) {
      for range timer.C {
        cmd.Process.Signal(os.Kill)
        ticker.Stop()
      }
    }(timer, ticker, cmd)
```
- This results in the following, _expected_ ERROR:
```
level=ERROR ts=2019-07-31T14:48:37.349918Z app=docker-compose-executor source=main.go:227 msg=\"cmd.Run() failed with signal: killed\\n\"\n2019/07/31 09:48:37 cmd.Run() failed with signal: killed\n "
```
- In a nutshell, I had to work with treating this particular error as _expected_, for example in `func (ea *ExecuteApp) Metrics(ctx context.Context, service string) ([]byte, error) { ... }` like so: `if err.Error() != "exit status 1" { ... }`


akramtexas:
`An example of testing with the bits deployed to a (Docker) container:`
container-metrics

akramtexas:
- This is the associated sequence diagram (`Executor Gets Service Metrics (Redesign - 2.0)`) that has been reviewed during `Sys Mgmt WG` meeting(s).
[Executor Gets Service Metrics (Redesign - 2.0).pdf](https://github.com/edgexfoundry/edgex-go/files/3452772/Executor.Gets.Service.Metrics.Redesign.-.2.0.pdf)

anonymouse64:
> For snap, I believe that we should be able to use the following TOML setting: ExecutorPath = '/path/to/the/file' in conjunction with snap-specific directives such as:

From my memory, I think the snap would not use a separate executor, and instead just use the `direct-service` mechanism that you mentioned to operate like it currently does.

Also, FYI since I'm not up to date on the design for this, I'm just going to review the code abstraction and the like and not focus too much on the design choices (though if there's something that looks wrong I will still call it out as a comment).

cmd/sys-mgmt-executor/main.go
```diff
@@ -72,7 +76,11 @@ func main() {
 		service = os.Args[1]
 		operation = os.Args[2]
 
-		err := ExecuteDockerCommands(service, operation)
+		LoggingClient.Info(fmt.Sprintf("1. service: %s", service))
+		LoggingClient.Info(fmt.Sprintf("2. operation: %s", operation))
```
tsconn23: These should be `Debug` IMHO

cmd/sys-mgmt-executor/main.go
```diff
@@ -120,26 +128,36 @@ func checkDockerContainerStatus(service string, running bool) (bool, error) {
...
 }
 
-func ExecuteDockerCommands(service string, operation string) error {
+func ExecuteDockerCommands(service string, operation string) (string, error) {
 	// don't run commands for unknown services - could be an attack
 	if agent.IsKnownServiceKey(service) {
-		return runDockerCommands(service, operation)
+
+		LoggingClient.Info(fmt.Sprintf("4. IsKnownServiceKey() check a success for service %s", service))
+		LoggingClient.Info(fmt.Sprintf("5. operation: %s", operation))
```
tsconn23: These should be `Debug` IMHO. Also you don't have a #3 log entry,. It goes 1,2,4,5. Would do away with the numbers.

cmd/sys-mgmt-executor/main.go
```diff
@@ -120,26 +128,36 @@ func checkDockerContainerStatus(service string, running bool) (bool, error) {
...
+
+		_, err := runDockerCommands(service, operation)
```
tsconn23: `ExecuteDockerCommands` is effectively a pass-through method which then calls `runDockerCommands`. The check for the known service key can be moved up to `main` and the guts of `runDockerCommands` put into this function to eliminate this empty layer.

cmd/sys-mgmt-executor/main.go
```diff
@@ -120,26 +128,36 @@ func checkDockerContainerStatus(service string, running bool) (bool, error) {
 	}
 }
 
-func ExecuteDockerCommands(service string, operation string) error {
+func ExecuteDockerCommands(service string, operation string) (string, error) {
```
tsconn23: This function is currently exported. Does it need to be?

cmd/sys-mgmt-executor/main.go
```diff
@@ -120,26 +128,36 @@ func checkDockerContainerStatus(service string, running bool) (bool, error) {
...
 	}
-	return fmt.Errorf("the service %s is an unknown service for which request was made to run Docker command", service)
+	return "", fmt.Errorf("the service %s is an unknown service for which request was made to run Docker command", service)
 }
 
-func runDockerCommands(service string, operation string) error {
+func runDockerCommands(service string, operation string) (string, error) {
+
+	var outStr = ""
+	var errStr = ""
```
tsconn23: You don't use these variables until the end of the function. You could eliminate these declarations and just do the following below.
`outStr, errStr := string(stdout), string(stderr)`

cmd/sys-mgmt-executor/main.go
```diff
@@ -148,19 +166,99 @@ func runDockerCommands(service string, operation string) error {
...
+
+		wg.Wait()
+
+		err = cmd.Wait()
+		if err != nil {
+			LoggingClient.Error(fmt.Sprintf("cmd.Run() failed with %s\n", err.Error()))
+			log.Fatalf("cmd.Run() failed with %s\n", err)
+		}
+		if errStdout != nil || errStderr != nil {
+			log.Fatal("failed to capture stdout or stderr\n")
```
tsconn23: Why is `log.Fatal` used here? Are we unable to recover gracefully from the error which has been logged already through the LoggingClient?

cmd/sys-mgmt-executor/main.go
```diff
@@ -148,19 +166,99 @@ func runDockerCommands(service string, operation string) error {
...
+			}
+		}(timer, ticker, cmd)
+		stderr, errStderr = copyAndCapture(os.Stderr, stderrIn)
+
+		wg.Wait()
+
+		err = cmd.Wait()
+		if err != nil {
+			LoggingClient.Error(fmt.Sprintf("cmd.Run() failed with %s\n", err.Error()))
+			log.Fatalf("cmd.Run() failed with %s\n", err)
```
tsconn23: Why is `log.Fatal` used here? Are we unable to recover gracefully from the error which has been logged already through the LoggingClient?

cmd/sys-mgmt-executor/main.go
```diff
@@ -148,19 +166,99 @@ func runDockerCommands(service string, operation string) error {
...
+		}
+		if errStdout != nil || errStderr != nil {
+			log.Fatal("failed to capture stdout or stderr\n")
+			LoggingClient.Error(fmt.Sprintf("failed to capture stdout or stderr\n"))
+		}
+		outStr, errStr = string(stdout), string(stderr)
+		LoggingClient.Info(fmt.Sprintf("\nout:\n%s\nerr:\n%s\n", outStr, errStr))
```
tsconn23: `Debug`?

cmd/sys-mgmt-executor/main.go
```diff
@@ -148,19 +166,99 @@ func runDockerCommands(service string, operation string) error {
...
+		LoggingClient.Info("invocation of metrics executor in runDockerCommands() succeeded...")
+		return outStr, nil
+	} else {
+		return "", fmt.Errorf("system management was requested to perform an unknown operation %s on the service %s", operation, service)
+	}
+}
+
+func copyAndCapture(w io.Writer, r io.Reader) ([]byte, error) {
```
tsconn23: Would this not work? https://golang.org/pkg/io/#Copy

internal/system/agent/router.go
```diff
@@ -154,6 +174,7 @@ func metricsHandler(w http.ResponseWriter, r *http.Request) {
 	pkg.Encode(send, w, LoggingClient)
 	return
 }
+*/
```
tsconn23: Can be removed?

akramtexas:
> I'm going to stop here. The executor code comes up first in the diff and that's what I've primarily looked at. Overall I think the executor code is starting to look very procedural and would benefit from utilizing more granular operations organized within its own `internal/system/executor` package structure.
> 
> Also, there are no unit tests anywhere to be found in this PR. A primary goal for Fuji w/r/t the edgex-go repository is to increase unit test coverage. Indeed some of this code is not currently testable as written. What is the plan for making it so? I stopped my review once it got to a point where I realized we need to align on this.

- Good feedback on the executor code @tsconn23 👍 (I'll revisit it shortly to utilize more granular operations).
- Yep, I will also revisit the code to make it (more) testable. I will follow up with a plan for making it so 👍

cmd/sys-mgmt-executor/main.go
```diff
@@ -148,19 +166,99 @@ func runDockerCommands(service string, operation string) error {
...
 		}
 		if !correctStatus {
-			return fmt.Errorf("docker %s operation failed for service %s", operation, service)
+			return "", fmt.Errorf("docker %s operation failed for service %s", operation, service)
+		}
+		return "", nil
+	} else if operation == METRICS {
+
+		LoggingClient.Info(fmt.Sprintf("run the docker stats on service (%s) for operation (%s)", service, operation))
+		cmd := exec.Command("docker", "stats")
```
anonymouse64: this is a streaming command, which AFAICT will never actually exit. This is not what we should be using to get the output for a single metrics endpoint. Instead I'd recommend using at least `--no-stream` option to get a single snapshot. That would remove the need to do all of the complicated wait group, tickers, timers, and pipes below. Additionally, it would be better to use the `--format` option to get formatted output like what's shown in https://kylewbanks.com/blog/docker-stats-memory-cpu-in-json-format (you don't need to format in json specifically, that's just easy to parse from the stdlib. That also would greatly simplify your parsing code...

cmd/sys-mgmt-executor/main.go
anonymouse64: I agree with Trevor, why is this not just io.Copy?

internal/system/agent/execution_engine.go
```diff
@@ -23,7 +24,27 @@ func runExec(service string, operation string) error {
 	return err
 }
 
-func (ec *ExecuteApp) Start(service string) error {
+func execMetrics(service string) ([]byte, error) {
+
+	// Preparing the call to the executor app.
+	//var result []byte
```
anonymouse64: please delete unused code

internal/system/agent/execution_engine.go
```diff
@@ -23,7 +24,27 @@ func runExec(service string, operation string) error {
...
+	cmd := exec.Command(Configuration.ExecutorPath, service, METRICS)
+
+	out, err := cmd.CombinedOutput()
+	LoggingClient.Info(fmt.Sprintf("err.Error(): %s ", err.Error()))
+	if err.Error() != "exit status 1" {
+		LoggingClient.Error(fmt.Sprintf("an error occurred in calling executor (to fetch metrics) for service %s: %v ", service, err.Error()))
+		// TODO: Following line is DEBUG only.
+		LoggingClient.Info(fmt.Sprintf("out: %s ", string(out)))
+	} else {
+		LoggingClient.Info("invocation of execMetrics() succeeded")
```
anonymouse64: this message should be at debug level

internal/system/agent/execution_engine.go
```diff
@@ -23,7 +24,27 @@ func runExec(service string, operation string) error {
...
+func execMetrics(service string) ([]byte, error) {
+
+	// Preparing the call to the executor app.
+	//var result []byte
+
+	cmd := exec.Command(Configuration.ExecutorPath, service, METRICS)
+
+	out, err := cmd.CombinedOutput()
+	LoggingClient.Info(fmt.Sprintf("err.Error(): %s ", err.Error()))
+	if err.Error() != "exit status 1" {
```
anonymouse64: The logic here looks wrong to me. How this is written if err is `exit status 1`, the SMA will think it succeeded which seems wrong. Shouldn't `exit status 1` still be considered an error?

internal/system/agent/router.go
```diff
@@ -134,7 +154,7 @@ func configHandler(w http.ResponseWriter, r *http.Request) {
 	return
 }
 
-func metricsHandler(w http.ResponseWriter, r *http.Request) {
+/*func metricsHandler(w http.ResponseWriter, r *http.Request) {
```
anonymouse64: please delete commented out code

internal/system/agent/services.go
```diff
@@ -34,6 +37,193 @@ const (
...
+func InvokeMetrics(services []string, ctx context.Context) (MetricsRespMap, error) {
+
+	m := MetricsRespMap{}
+	m.Metrics = map[string]interface{}{}
+
+	// Loop through requested actions, along with (any) respectively-supplied parameters.
+	for _, service := range services {
+		LoggingClient.Info("invoke metrics")
+
+		if Configuration.MetricsMechanism == "direct-service" {
```
anonymouse64: this isn't quite how I imagined the abstraction would work. I think it should be organized like the service management, where we have a variable that is `interface{}` but might implement some other interfaces and we do a type cast to see if the variable implements any of the interfaces for gathering different kinds of metrics. For example, maybe the `direct-service` implementation only has CPU and Memory stats, so it only implements the interfaces `CPUStats` and `MemStats`, but the executor implementation also provides `NetworkIOStats` and `DiskIOStats` interfaces, so the executor implementation provides those stats as well. Again, I haven't been to many SMA WG meetings recently so I'm not sure how the design has progressed, but that's what I remember discussing.

Additionally, I think it's cleaner to have the inspection of the configuration value done during init, and assign the chosen implementation for the metrics gathering function, then this function doesn't need to read any information from the configuration struct.

internal/system/agent/services.go
```diff
@@ -34,6 +37,193 @@ const (
...
 )
 
+func InvokeMetrics(services []string, ctx context.Context) (MetricsRespMap, error) {
+
+	m := MetricsRespMap{}
+	m.Metrics = map[string]interface{}{}
+
+	// Loop through requested actions, along with (any) respectively-supplied parameters.
+	for _, service := range services {
+		LoggingClient.Info("invoke metrics")
```
anonymouse64: this should be debug level?

tsconn23:
> * Yep, I will also revisit the code to make it (more) testable. I will follow up with a plan for making it so 👍

My suggestion would be, then, to close this PR and open a new one later. From experience, I know that testable code will look very different from what's here currently.

internal/system/agent/services.go
```diff
@@ -34,6 +37,193 @@ const (
...
+
+	assembled, err := json.Marshal(data)
+	if err != nil {
+		LoggingClient.Error(err.Error())
+		return nil, err
+	}
+	return assembled, err
+}
+
+func selectRelevantStats(stats map[string]Stats, service string) map[string]Stats {
```
anonymouse64: I don't quite follow the purpose of this function - it seems like this function is just getting the specified service key from the map? I think this could be much more efficiently implemented by just accessing the map directly.

internal/system/agent/services.go
```diff
@@ -34,6 +37,193 @@ const (
...
+type Stats struct {
+	CpuPercent       string
+	MemUsageAndLimit string
+	MemPercent       string
+	Net_IO           string
+	Block_IO         string
+	Pids             string
+}
+
+func populateMap(fields [][]string) map[string]Stats {
```
anonymouse64: I'm not going to review these parsing functions because I don't think this should be implemented with all this manual parsing - I think it should be outputted in a formatted way that's easier to parse, see my previous comment

akramtexas:
> > * Yep, I will also revisit the code to make it (more) testable. I will follow up with a plan for making it so 👍
> 
> My suggestion would be, then, to close this PR and open a new one later. From experience, I know that testable code will look very different from what's here currently.

Gotcha @tsconn23 👍 Specifically, are you alluding to the the use of molding code into a shape whereby we can apply the unit test patterns you had shared lately (for which you have a Kanban board)?

internal/system/agent/services.go
```diff
@@ -172,86 +362,90 @@ func getConfig(services []string, ctx context.Context) (ConfigRespMap, error) {
...
+		if err != nil {
+			//m.Metrics[service] = fmt.Sprintf(err.Error())
+			LoggingClient.Error(err.Error())
+		} else {
+			//m.Metrics[service] = ProcessResponse(responseJSON)
+			out = []byte(responseJSON)
 		}
 	}
-	return m, nil
+	/*	out, err := execMetrics(service)
```
anonymouse64: please delete commented out code

internal/system/agent/services.go
```diff
@@ -172,86 +362,90 @@ func getConfig(services []string, ctx context.Context) (ConfigRespMap, error) {
...
 
-		// Check whether SMA does _not_ know of ServiceKey ("service") as being one for one of its ready-made list of clients.
-		if !IsKnownServiceKey(service) {
-			LoggingClient.Info(fmt.Sprintf("service %s not known to SMA as being in the ready-made list of clients", service))
+		// Service unknown to SMA, so ask the Registry whether `service` is available.
+		err := registryClient.IsServiceAvailable(service)
+		if err != nil {
+			out = []byte(fmt.Sprintf(err.Error()))
+			LoggingClient.Info(fmt.Sprintf(string(out)))
+			// m.Metrics[service] = fmt.Sprintf(err.Error())
```
anonymouse64: please delete commented out code

internal/system/agent/services.go
```diff
@@ -172,86 +362,90 @@ func getConfig(services []string, ctx context.Context) (ConfigRespMap, error) {
...
 
-			// Service unknown to SMA, so ask the Registry whether `service` is available.
-			err := registryClient.IsServiceAvailable(service)
+			// Since service is unknown to SMA, ask the Registry for a ServiceEndpoint associated with `service`
+			e, err := registryClient.GetServiceEndpoint(service)
 			if err != nil {
-				m.Metrics[service] = fmt.Sprintf(err.Error())
-				LoggingClient.Error(fmt.Sprintf(err.Error()))
+				out = []byte(fmt.Sprintf("on attempting to get ServiceEndpoint for service %s, got error: %v", service, err.Error()))
+				// m.Metrics[service] = fmt.Sprintf("on attempting to get ServiceEndpoint for service %s, got error: %v", service, err.Error())
```
anonymouse64: please delete commented out code

internal/system/agent/services.go
```diff
@@ -172,86 +362,90 @@ func getConfig(services []string, ctx context.Context) (ConfigRespMap, error) {
...
-					// TODO: [Need a manifest-like functionality in future...]
-					// Add the service key to the map where the value is the respective GeneralClient
-					generalClients[e.ServiceId] = general.NewGeneralClient(params, startup.Endpoint{RegistryClient: &registryClient})
+				// Preparing to add the specified key to the map where the value will be the respective GeneralClient
+				clientInfo := config.ClientInfo{}
+				clientInfo.Protocol = Configuration.Service.Protocol
+				clientInfo.Host = e.Host
+				clientInfo.Port = e.Port
+
+				// This code will evolve to take into account a manifest-like functionality in future. So
```
anonymouse64: I'm not sure I understand this comment - is this saying that if the registerClient is available then use the registry, otherwise don't?

Also I think large comments like this are prone to growing stale in that when the manifest functionality is eventually implemented, it's unfortunately often the case that updating/removing comments like this is forgotten and since it's already in the codebase when that functionality is implemented, reviewers may not know to look at this part of the code to see that it should be updated, so I'd suggest removing this. I think a more appropriate comment is  TODO: as that's easier for reviewers to see when the feature is implemented that the TODO should be removed

internal/system/agent/services.go
```diff
@@ -172,86 +362,90 @@ func getConfig(services []string, ctx context.Context) (ConfigRespMap, error) {
...
-					}
+				Configuration.Clients[e.ServiceId] = clientInfo
+				params := types.EndpointParams{
+					ServiceKey:  e.ServiceId,
+					Path:        "/",
+					UseRegistry: usingRegistry,
+					Url:         Configuration.Clients[e.ServiceId].Url() + clients.ApiMetricsRoute,
+					Interval:    internal.ClientMonitorDefault,
 				}
+				// TODO: The following note is related to future work:
```
anonymouse64: I'd suggest refactoring these comment to just be a single line comment TODO, I think it's clear by the TODO that it's future work, and it's not clear what is meant by manifest-like functionality. In addition to those points, there's already a comment above about future manifest work so this seems duplicative.

internal/system/agent/services.go
```diff
@@ -172,86 +362,90 @@ func getConfig(services []string, ctx context.Context) (ConfigRespMap, error) {
...
-				LoggingClient.Error(err.Error())
-			} else {
-				m.Metrics[service] = ProcessResponse(responseJSON)
-			}
+		responseJSON, err := generalClients[service].FetchMetrics(ctx)
+		if err != nil {
+			//m.Metrics[service] = fmt.Sprintf(err.Error())
+			LoggingClient.Error(err.Error())
+		} else {
+			//m.Metrics[service] = ProcessResponse(responseJSON)
```
anonymouse64: please delete commented out code

internal/system/agent/services.go
```diff
@@ -172,86 +362,90 @@ func getConfig(services []string, ctx context.Context) (ConfigRespMap, error) {
...
-			responseJSON, err := generalClients[service].FetchMetrics(ctx)
-			if err != nil {
-				m.Metrics[service] = fmt.Sprintf(err.Error())
-				LoggingClient.Error(err.Error())
-			} else {
-				m.Metrics[service] = ProcessResponse(responseJSON)
-			}
+		responseJSON, err := generalClients[service].FetchMetrics(ctx)
+		if err != nil {
+			//m.Metrics[service] = fmt.Sprintf(err.Error())
```
anonymouse64: please delete commented out code

internal/system/agent/services.go
```diff
@@ -172,86 +362,90 @@ func getConfig(services []string, ctx context.Context) (ConfigRespMap, error) {
...
+			LoggingClient.Error(err.Error())
```
anonymouse64: shouldn't we also return this error as well as log it?

internal/system/agent/utils.go
```diff
@@ -35,3 +35,12 @@ func ProcessResponse(response string) map[string]interface{} {
 	}
 	return rsp
 }
+
+func ProcessFetchedResponse(response []byte) map[string]interface{} {
```
anonymouse64: IMHO this function is largely unnecessary - it doesn't return an error so a caller won't know if the operation actually succeeded or if the response really was just an empty map, and in addition you don't really save many lines of code by having the error check from this function (if this function returned an error which it should) in addition to the error check inside this function.

anonymouse64:
Also as a naming point, if the sys-mgmt-executor now handles both metrics and service management in the same executable, but both using docker, I think the name of that executable should change to something that includes docker in the name. But renaming executables could be seen as a breaking change, so it's fine to leave this as a to-do for Geneva or another release where we can make such breaking changes

akramtexas:
> There are some smallish issues with implementation here that I mention in the comments, but as an overview I think that the design of the abstraction of metrics gathering needs some work. Specifically, I think that you should have interfaces for each type of metrics, i.e. a `CPUStatser` (this is a weird name, would need to think harder about a better name) interface, `MemStatser` interface, etc. and your various implementations (direct-service, executor, etc.) would just implement as many of those interfaces as they can. This would allow for easier growth adding new types of metrics to the direct-service implementation, for example today that probably doesn't implement anything like `NetworkIOStats`, but the executor implementation should support that because i.e. docker could support that (and maybe someday the direct-service implementation would learn how to do `NetworkIOStats`).
> 
> Also I think some more design needs to go into how the arbitrary executable executor feature could communicate back to the SMA what features the executable supports. For example, with service management, currently we expect that the executable supports starting, stopping and restarting services as all required, but for metrics, you will probably have lots of different metrics available to different implementations (and we don't want to exclude implementations like docker that can produce much more rich metrics than other executor implementations that don't provide as rich metrics). I think one way this could be done is by having the configuration.toml file specify which types of metrics the executor supports, this would allow more flexibility and better error handling (though if you do this you may need to get into using reflection to add interfaces to the struct based on what's in the configuration.toml, but that's probably not too hard). On this same topic, it would be really cool to see the service management executor code and the metrics gathering executor code share some common code towards accomplishing that flexibility.
> 
> Lastly, there are some unrelated changes in here that I think should be in separate PR's.

@anonymouse64 Thanks for the detailed feedback, much appreciated. I will follow up for next steps 👍

cmd/sys-mgmt-agent/res/docker/configuration.toml
```diff
@@ -15,7 +24,7 @@ LogLevel = 'INFO'
 BootTimeout = 30000
 ClientMonitor = 15000
 CheckInterval = '10s'
-Host = 'localhost'
+Host = 'edgex-sys-mgmt-agent'
```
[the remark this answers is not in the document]
akramtexas: Done. That PR has been reviewed and merged 👍

cmd/sys-mgmt-executor/main.go
```diff
@@ -72,7 +76,11 @@ func main() {
 		service = os.Args[1]
 		operation = os.Args[2]
 
-		err := ExecuteDockerCommands(service, operation)
+		LoggingClient.Info(fmt.Sprintf("1. service: %s", service))
```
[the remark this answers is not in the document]
akramtexas: The rationale was to provide debug info. Since then, such lines have been updated to `Debug` LEVEL 👍

akramtexas:
Recheck

akramtexas:
- @tsconn23 I will, much as I had indicated earlier, follow through on your note (i.e. `My suggestion would be, then, to close this PR and open a new one later.`)
- Meanwhile, I am wrapping up addressing the smaller comments so we can track them more easily; hopefully, these will assist you as well. 
- But yes, I will open a new PR to address the major points of (1) making the code more testable, and (2) making the executor code less procedural and it instead benefit from utilizing more granular operations organized within its own internal/system/executor package structure: The `executor` code (as in `main.go`) _has_ grown up quite a bit from where it used to be a mere two-dozen LOC to now where it could use some structure and restructuring ⛏ 
- @anonymouse64 I appreciated very much your feedback as well (e.g. `Also I think some more design needs to go into how the arbitrary executable executor feature could communicate back to the SMA what features the executable supports.` etc.)
- Gents, I will follow up, thanks 👍

cmd/sys-mgmt-executor/main.go
akramtexas: I compared the signatures of the two functions (i.e. the one you mention, and the one I've used) and notice some differences... The implementation I'm using comes from a Go Cookbook [Advanced command execution in Go with os/exec](https://blog.kowalczyk.info/article/wOYk/advanced-command-execution-in-go-with-osexec.html) and which certainly reflects that thought and care has been taken in putting it together. If possible, I, too, would prefer using a call to the Go standard library. Thanks for mentioning 👍

akramtexas:
> * @tsconn23 I will, much as I had indicated earlier, follow through on your note (i.e. `My suggestion would be, then, to close this PR and open a new one later.`)
> * Meanwhile, I am wrapping up addressing the smaller comments so we can track them more easily, and carry them forward into the subsequent PR; hopefully, doing so will assist you as well by making the reconciliation of updates easier to follow.
> * But yes, I will open a new PR to address the major points of (1) making the code more testable, and (2) making the executor code less procedural and it instead benefit from utilizing more granular operations organized within its own internal/system/executor package structure: The `executor` code (as in `main.go`) _has_ grown up quite a bit from where it used to be a mere two-dozen LOC to now where it could use some structure and restructuring ⛏
> * @anonymouse64 I appreciated very much your feedback as well (e.g. `Also I think some more design needs to go into how the arbitrary executable executor feature could communicate back to the SMA what features the executable supports.` etc.)
> * @jpwhitemn You are traveling on business so I don't expect you to have had the chance to take a look at this PR or our conversation here. Once you return, we can synch up.
> * Gents, I will follow up, thanks 👍

`Final Update:`
- Gents, at this time, I have addressed all of the (numerous, smaller) PR comments, and which I'll carry forward into the _subsequent_ proposed PR. 
- That leaves addressing the major points of (1) making the code more testable, (2) making the executor code less procedural and instead benefit from utilizing more granular operations organized within its own internal/system/executor package structure, and (3) look into some more design into how the arbitrary executable executor feature could communicate back to the SMA what features the executable supports.
- To that end, I will open the _subsequent_ proposed PR, _after_ having had the chance to synch with all reviewers of this PR (I've already got comments from two of the three reviewers) 👍

tsconn23:
Per Sys Mgmt Working Group call 6-Aug-19 and following on discussion above, closing this PR. @akramtexas will open a new one from the same feature branch when the relevant feedback has been incorporated.

PR #1612, opened 2019-07-31
Fix host config in TOML.

### LINKED ISSUE #1594: Wrong host configuration in System Management Agent docker image
# 🐞 Bug Report

### Affected Services
The issue is located in:  System Management Agent

### Is this a regression?
No

https://github.com/edgexfoundry/edgex-go/blob/8b5a3a86adb03d8d3708adf15906ed0d5873a114/cmd/sys-mgmt-agent/res/docker/configuration.toml#L18
It should not be `localhost`, but `edgex-sys-mgmt-agent` according to the docker-compose file in 
https://github.com/edgexfoundry/developer-scripts/blob/deac70c76d9a837a6e9b3f0168aa66f17acf8ed8/releases/edinburgh/compose-files/docker-compose-edinburgh-1.0.1.yml#L234

## 🔥 Exception or Error
health check will always fail on Consul if running in docker enviromnet

## 🌍  Your Environment
**Deployment Environment:**
Docker

**EdgeX Version:**
v1.0.1

akramtexas:
- Fix Issue-1594 (in `master`)
- This is to fix the inadvertent value of `localhost` being used instead of `edgex-sys-mgmt-agent` in the (res/docker) `configuration.toml` for sys-mgmt-agent.


akramtexas:
- Tag had been applied: [v1.0.1](https://github.com/edgexfoundry/edgex-go/releases/tag/v1.0.1)
- I will create separate PR to apply the fix to the tagged release (since that would be a different build job).

anonymouse64:
> I will create separate PR to apply the fix to the tagged release (since that would be a different build job).

What does this mean? You shouldn't be re-tagging things? For this fix to be "released to Edinburgh" I think you would need to rev up to 1.0.2

akramtexas:
> > I will create separate PR to apply the fix to the tagged release (since that would be a different build job).
> 
> What does this mean? You shouldn't be re-tagging things? For this fix to be "released to Edinburgh" I think you would need to rev up to 1.0.2

I'm not aware of any re-tagging plans... Yes, we will rev, and for that I have a separate PR open: [PR-1613](https://github.com/edgexfoundry/edgex-go/pull/1613)

PR #1613, opened 2019-07-31
Fix host config in config TOML.

akramtexas:
- Fix Issue-1594 (in `edinburgh`)
- This is to fix the inadvertent value of `localhost` being used instead of `edgex-sys-mgmt-agent` in the (res/docker) `configuration.toml` for sys-mgmt-agent.


akramtexas:
Note that the corresponding fix (for Issue-1594) for `master` has a separate PR: [PR-1612](https://github.com/edgexfoundry/edgex-go/pull/1612), which has a separate build job associated with it.

Issue #1246, opened 2019-04-15
Allow specifying sub-second schedule frequencies


anonymouse64:
AFAICT, there's no way to create a schedule that has a frequency larger than 1 Hz (i.e. the minimum duration time for scheduled events that are scheduled is 1 second).

This seems to just be a limitation of our implementation of handling ISO 8601 duration spec parsing where we only allow integer time values. The code _could_ be adapted to also allow specifying something like 

```PT0.1S```

which would have a duration of 0.1 seconds and correspond to a frequency of 10 Hz. 

The only part that would need to be changed for this to work would be this function:
https://github.com/edgexfoundry/edgex-go/blob/78ba8bd0017a61459c0f8cd4eb25efb4f917d6ab/internal/support/scheduler/schedulecontext.go#L109-L124

where instead of assuming that the scheduler components are all int64's, we would support parsing decimal fractions too.

anonymouse64:
I see that the device-sdk now supports AutoEvents with simple golang time duration parsing as provided by `time.ParseDuration` and that supports sub-second time durations. 
Perhaps support-scheduler should also switch to supporting simple durations like device-sdk-go rather than using ISO-8601?

xmlviking:
@anonymouse64 I agree I believe the scheduler move towards the ubiquitous time standard used throughout edgex-go and now the device-sdk-go.  ISO-8601 is the original standard we used in the Java version.  It was simply left behind.

xmlviking:
@anonymouse64  - I have moved the frequency validation to using the go lang time parseDuration.  I have a question regarding the Start and Stop DateTime formats.  We are currently using ISO 8601.  Go lang in general supports the RFC versions including straight RFC33339 non Nano.   We have a few options if we desire to change our DateTime format to something more precise.   

**ISO8601** (RFC33339)  "What were doing currently" but parsing it manually not using the Go fmt option with RFC33339. 
Example:
     **RFC3339**     = "2006-01-02T15:04:05Z07:00"

Hashicorp et. al. use the RFC33339Nano 
Example: 
 **RFC3339Nano** = "2006-01-02T15:04:05.999999999Z07:00"

My only issue with this is regarding reproduction of the value after persistence in a data store. 
[Slack Issue RFC3339Nano](https://stackoverflow.com/questions/55251584/can-a-golang-rfc3339nano-timestamp-be-reliably-converted-to-and-from-a-unix-time) 

[...]

Issue #1424, opened 2019-06-11
Add database versioning support


AnthonyMBonafide:
# 🚀 Feature Request

### Relevant Packages
This feature request is for core services.

### Description
Leverage a Go based database migration tool so that we can better manage changes to databases while also better facilitating backwards compatibility. See [migrate](https://github.com/golang-migrate/migrate) for possible library and features we could use.

#### Problems:

1. Database initialization - The initialization of database(s) are done via scripts, like `init_mongo.js`, which are copied across multiple repos (`docker-edgex-mongo` and `developer-scripts`). This makes standing up an EdgeX system complicated since the user has to first configure Mongo, then execute the `init_mongo.js` script(located in either of the two repos), then start the necessary services.

2. Database maintainability - Making updates in the code sometimes can lead to making changes in the database schemas. This results in multiple PRs across repos which then have to come together for the final product. This coordination adds some complexity to the maintainability of the project/system and can be error prone.

3. Backwards compatibility - With the current implementation, there is no way to provide backwards compatibility for situations which require moving around of data or re-structuring the database schema. For example, if we decide that --for some crazy reason-- we would like to merge data from two different collections into one. As of now we would make the necessary code changes, update the `init_mongo` script, and roll out the updates together. However, this would not be ideal for users that are already using the system and just want to update to the latest version as it would require them to lose data as they switch to the new schema by re-executing the initialization script.

4. Fault tolerance - We do not have a good way of ensuring the state of the database within EdgeX. Since the code and the database schema information are disjointed we don't have a reliable way of ensuring that the database and the code have been updated, shipped, and installed correctly. In some cases we may not see any errors until we actually try to access the problematic data.

#### Suggested Solution:

Use a Go based database migration tool, possibly [migrate](https://github.com/golang-migrate/migrate), and handle the database setup and versioning within the respective services.

The thought is that each service has its own scripts, specific to its data and underlying database, which will make the necessary changes to the database to bring the database to a state which the application can interact with. This should work with any previous database state which is valid.

For example, given the structure of the `edgex-go` as of this writing. The database files for `coredata` would reside in a folder with the different versions, lets assume `cmd/coredata/res`, and would look something like:
```
- 1_initalize_mongo.js
- 2_add_index_to_reading_device.js
- 3_rename_reading_collection.js
...
```

[...]

PR #1625, opened 2019-08-05
Changing interval frequency field to use go standard time format

xmlviking:
Issue #1246  - Adjust Interval frequency so that uses Go standard time format.
Should allow miliseconds, microseconds, as well as nonoseconds.

### DIFF
--- a/internal/support/scheduler/const.go
+++ b/internal/support/scheduler/const.go
@@ -16,22 +16,21 @@ package scheduler
 const (
 
 	/* -------------- Constants for Scheduler -------------------- */
-	ID                   = "id"
-	NAME                 = "name"
-	TARGETNAME           = "targetname"
-	INTERVALACTION       = "intervalaction"
-	INTERVAL             = "interval"
-	LABEL                = "label"
-	YAML                 = "yaml"
-	COMMAND              = "command"
-	KEY                  = "key"
-	VALUE                = "value"
-	UNLOCKED             = "UNLOCKED"
-	ENABLED              = "ENABLED"
-	SCHEDULER_TIMELAYOUT = "20060102T150405" /** Scheduler **/
-	TIMELAYOUT           = "20060102T150405" /** Scheduler **/
-	SCRUB                = "scrub"
-	TARGET               = "target"
+	ID             = "id"
+	NAME           = "name"
+	TARGETNAME     = "targetname"
+	INTERVALACTION = "intervalaction"
+	INTERVAL       = "interval"
+	LABEL          = "label"
+	YAML           = "yaml"
+	COMMAND        = "command"
+	KEY            = "key"
+	VALUE          = "value"
+	UNLOCKED       = "UNLOCKED"
+	ENABLED        = "ENABLED"
+	TIMELAYOUT     = "20060102T150405"
+	SCRUB          = "scrub"
+	TARGET         = "target"
 
 	/* ---------------- URL PARAM NAMES -----------------------*/
 	ContentTypeKey       = "Content-Type"
--- a/internal/support/scheduler/schedulecontext.go
+++ b/internal/support/scheduler/schedulecontext.go
@@ -9,9 +9,6 @@ package scheduler
 
 import (
 	"github.com/edgexfoundry/go-mod-core-contracts/models"
-
-	"regexp"
-	"strconv"
 	"time"
 )
 
@@ -70,7 +67,11 @@ func (sc *IntervalContext) Reset(interval models.Interval) {
 	//frequency and next time
 	nowBenchmark := time.Now().Unix()
 	if !sc.Interval.RunOnce {
-		sc.Frequency = parseFrequency(sc.Interval.Frequency)
+		frequency, err := parseFrequency(sc.Interval.Frequency)
+		if err != nil {
+			LoggingClient.Error("interval parse frequency error  %v", err.Error())
+		}
+		sc.Frequency = frequency
 	}
 
 	sc.NextTime = sc.StartTime
@@ -106,35 +107,4 @@ func (sc *IntervalContext) isComplete(time time.Time) bool {
 		(sc.NextTime.Unix() > sc.EndTime.Unix()) ||
 		((sc.MaxIterations != 0) && (sc.CurrentIterations >= sc.MaxIterations))
 	return complete
-}
-
-func parseFrequency(durationStr string) time.Duration {
-	durationRegex := regexp.MustCompile(`P(?P<years>\d+Y)?(?P<months>\d+M)?(?P<days>\d+D)?T?(?P<hours>\d+H)?(?P<minutes>\d+M)?(?P<seconds>\d+S)?`)
-	matches := durationRegex.FindStringSubmatch(durationStr)
-
-	years := parseInt64(matches[1])
-	months := parseInt64(matches[2])
-	days := parseInt64(matches[3])
-	hours := parseInt64(matches[4])
-	minutes := parseInt64(matches[5])
-	seconds := parseInt64(matches[6])
-
-	second := int64(time.Second)
-	minute := int64(time.Minute)
-	hour := int64(time.Hour)
-	day := int64(24 * hour)
-	month := int64(30 * day)
-	year := int64(365 * day)
-	return time.Duration(years*year + months*month + days*day + hours*hour + minutes*minute + seconds*second)
-}
-
-func parseInt64(value string) int64 {
-	if len(value) == 0 {
-		return 0
-	}
-	parsed, err := strconv.Atoi(value[:len(value)-1])
-	if err != nil {
-		return 0
-	}
-	return int64(parsed)
 }
--- a/internal/support/scheduler/schedulercontext_test.go
+++ b/internal/support/scheduler/schedulercontext_test.go
@@ -21,6 +21,7 @@ const (
 	TestUnexpectedMsgFormatStrForIntVal   = "unexpected result, active: '%d' but expected: '%d'"
 	TestUnexpectedMsgFormatStrForFloatVal = "unexpected result, active: '%f' but expected: '%f'"
 	TestUnexpectedMsgFormatStrForBoolVal  = "unexpected result, active: '%t' but expected: '%t'"
+	TestUnexpectedMsgFormatStrForInt64Val = "unexpected result, active: '%d' but expected: '%d'"
 )
 
 // Test Schedule model const fields
@@ -28,12 +29,11 @@ const (
 	TestIntervalName         = "midnight-1"
 	TestIntervalStart        = "20000101T000000"
 	TestIntervalEnd          = ""
-	TestIntervalFrequency    = "P1D"
+	TestIntervalFrequency    = "24h"
 	TestIntervalCron         = "This is a description"
 	TestIntervalRunOnce      = true
 	TestIntervalUpdatingName = "midnight-2"
-
-	TestBadFrequency = "423"
+	TestBadFrequency         = "423"
 )
 
 // Test IntervalAction model const fields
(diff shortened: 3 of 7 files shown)


cmd/config-seed/res/properties/go.mod
```diff
@@ -1 +1,3 @@
 module github.com/edgexfoundry/edgex-go/cmd/config-seed/res/properties
+
+go 1.12
```
brandonforster: Why do you need Go 1.12?

internal/support/scheduler/const.go
```diff
@@ -28,7 +28,6 @@ const (
 	VALUE                = "value"
 	UNLOCKED             = "UNLOCKED"
 	ENABLED              = "ENABLED"
-	SCHEDULER_TIMELAYOUT = "20060102T150405" /** Scheduler **/
 	TIMELAYOUT           = "20060102T150405" /** Scheduler **/
```
brandonforster: Is this comment still valid?

internal/support/scheduler/schedulecontext.go
```diff
@@ -70,7 +67,11 @@ func (sc *IntervalContext) Reset(interval models.Interval) {
 	//frequency and next time
 	nowBenchmark := time.Now().Unix()
 	if !sc.Interval.RunOnce {
-		sc.Frequency = parseFrequency(sc.Interval.Frequency)
+		frequency, err := parseFrequency(sc.Interval.Frequency)
+		if err != nil {
+			LoggingClient.Error("parse frequency error, the original frequency string is : " + sc.Interval.Frequency)
```
brandonforster: Right now, you're eating the error message.

internal/support/scheduler/utils.go
```diff
@@ -48,6 +49,30 @@ func msToTime(ms string) (time.Time, error) {
...
+// Hour = 60 * Minute
+func parseFrequency(durationStr string) (time.Duration, error) {
+
+	// default
+	var durationDefault = 24 * time.Hour
+
+	// check Frequency
+	timeDuration, err := time.ParseDuration(durationStr)
+	if err != nil {
+		LoggingClient.Error("parse frequency error, the original time string is : " + durationStr)
```
brandonforster: This is the same output that your error handling code is going to have. Your code will output the same error alert twice, both without any real insight into what exactly is unparsable.

internal/support/scheduler/utils.go
```diff
@@ -14,25 +14,26 @@
...
 
-const (
-	frequencyPattern = `^P(\d+Y)?(\d+M)?(\d+D)?(T(\d+H)?(\d+M)?(\d+S)?)?$`
-)
-
+// Frequency indicates how often the specific resource needs to be polled.
+// It represents as a duration string.
+// The format of this field is to be an unsigned integer followed by a unit which may be "ms", "s", "m" or "h"
+// representing milliseconds, seconds, minutes or hours. Eg, "100ms", "24h"
 func isFrequencyValid(frequency string) bool {
```
brandonforster: This function is simple enough that I think we should refactor it out. It's not adding a whole lot of value.

internal/support/scheduler/const.go
xmlviking: Yes :) that is ISO 8601 in general

brandonforster: What is "scheduler" telling us because I don't understand the point of the comment. These are all scheduler constants.

internal/support/scheduler/utils.go
xmlviking: I left it because it's being used by other potential functions.

brandonforster: Then remove it there too.

xmlviking: after our discussion I will be refactoring this util function out and just modify the code to use the parseDuration func :)

internal/support/scheduler/utils.go
xmlviking: fixed

mhall119:
Is this going to cause configuration files made for EdgeX 1.0.x break? Is breaking backwards compatibility for configuration files considered a major-version-changing break?

internal/support/scheduler/schedulecontext.go
xmlviking: fixed . logging error message as well.

xmlviking:
> Is this going to cause configuration files made for EdgeX 1.0.x break? Is breaking backwards compatibility for configuration files considered a major-version-changing break?

Michael as it sits right now this would break backward compatibility.  The request was to update scheduler interval's to use more prevalent go time format for frequency.

internal/support/scheduler/const.go
xmlviking: removed the comment :) . fixed

cmd/config-seed/res/properties/go.mod
xmlviking: removed!  :)

anonymouse64:
Is it possible to implement this in a backward compatible way? I'm not sure what the syntactical relationship between ISO-8601 and RFC3339Nano is, but if there's no overlap between a valid ISO-8601 date and a RFC3339Nano date, then we could implement "smart parsing" which supports both formats automatically. 

If there is overlap between the two formats, then perhaps for Fuji to not break backwards compatibility we add a new `FrequencyFormat` key which defaults to `"ISO-8601"` but also supports a `"RFC3339Nano"` value, then for services wishing to use RFC3339Nano, they could add that key and get the newer behavior.

xmlviking:
> Is it possible to implement this in a backward compatible way? I'm not sure what the syntactical relationship between ISO-8601 and RFC3339Nano is, but if there's no overlap between a valid ISO-8601 date and a RFC3339Nano date, then we could implement "smart parsing" which supports both formats automatically.
> 
> If there is overlap between the two formats, then perhaps for Fuji to not break backwards compatibility we add a new `FrequencyFormat` key which defaults to `"ISO-8601"` but also supports a `"RFC3339Nano"` value, then for services wishing to use RFC3339Nano, they could add that key and get the newer behavior.

Problem is the existing Intervals in an existing database.  Backward compatibility.

anonymouse64:
> Problem is the existing Intervals in an existing database. Backward compatibility.

Well I'm not too familiar with the persistence implementation, but if we are storing the format specified directly then all that would happen is logic working with that would learn to also check for the `FrequencyFormat` specifier, if it was absent or empty it would work the same, otherwise parsing would continue based on what format was selected with that key.

If the Interval is converted to a time.Duration, then it probably would just work? If the format is deserialized out of the database into a time.Duration (i.e. just used directly) then no parsing would be necessary and it wouldn't matter if a Fuji or Edinburgh client wrote the entry into the database as both the ISO-8601 and RFC3339Nano would be serialized into the same time.Duration. If the format is yet some other different serialization (i.e. not the string from configuration.toml and also not a time.Duration), then I assume it could still work the same as described initially, checking if a `FrequencyFormat` key exists and basing deserialization logic on that. 

Finally, I thought that @tsconn23 said we weren't going to support direct database migrations between releases even if those releases were backwards compatible?

tsconn23:
> Finally, I thought that @tsconn23 said we weren't going to support direct database migrations between releases even if those releases were backwards compatible?

@anonymouse64 You might be thinking of what I said in the fourth bullet point on [this comment](https://github.com/edgexfoundry/edgex-go/issues/1424#issuecomment-501488802). The gist is that transitory data like events and readings should be drained before doing a minor version upgrade. All other data should be considered long-lived for the life of the major version as we don't want someone to have to re-provision their whole environment going from Edinburgh --> Fuji

anonymouse64:
@xmlviking can you speak to my questions about how intervals get serialized / deserialized into databases?

> All other data should be considered long-lived for the life of the major version as we don't want someone to have to re-provision their whole environment going from Edinburgh --> Fuji

Understood, thanks for clarifying

xmlviking:
Ian,

The issue with backward compatibility relates to what you mentioned.
Existing installations with Scheduler "Intervals" with the previous version
of the frequency format.  The actually date time is fine and is ISO 8601 so
we will be sticking with that for now.
Again to ensure were not breaking anyone's existing EdgeX environment were
holding off on the frequency change for now and wait till Fuji.

Eric

anonymouse64:
Thanks @xmlviking, unfortunately I didn't quite understand your comment as it doesn't seem to doesn't answer my question about how the scheduler intervals are serialized to a database like Mongodb.

In the meantime, I took the liberty of looking at the code myself and it seems that we serialize intervals with the ISO-8601 string exactly, which means that we could still implement this change in a backwards compatible manner by doing the following:

- add a new FrequencyFormat string to the Interval spec (adding new strings doesn't break backwards compatibility for semver)
- if the FrequencyFormat key is defined for a given Interval, then use that value to deserialize the Frequency string into a time.Duration
- if the FrequencyFormat key is missing or not defined, then assume the format is ISO-8601 and deserialize appropriately

None of these changes would break an existing installation, though note I am assuming that mongo/redis can handle a new key added to a struct when serializing it.

xmlviking:
@anonymouse64 - Ian could you please re-review this change request.  We decided that not changing the schema per say to include the format would not be required at this point.  The code simply supports both formats legacy format as well as the goLang style.  Changes have been made and implemented in the go-mod-contracts to support this bifurcation in the acceptable frequency format.  TY :)

anonymouse64:
> We decided that not changing the schema per say to include the format would not be required at this point

I'm a little confused by this statement, did you change the schema or not? And is a format now included or not?

xmlviking:
> > We decided that not changing the schema per say to include the format would not be required at this point
> 
> I'm a little confused by this statement, did you change the schema or not? And is a format now included or not?

No change in the schema.   No format field.   Were just accepting both old and new formats. It's the least invasion option.

xmlviking:
> > We decided that not changing the schema per say to include the format would not be required at this point
> 
> I'm a little confused by this statement, did you change the schema or not? And is a format now included or not?
Ian,

Based on my analysis of the required areas in EdgeX-go I felt the path of least resistance might be a better approach.   Rather than include a new schema change to the type Interval I chose to enhance the go-mod-contracts serialize and deserialize functions.  I had to do this anyhow regardless of the approach I took.   I actually started with the full change you suggested but found I did not have to do that if I accounted for both the existing frequency time format and the proposed Go Lang standard way.   

In summary I made changes to the go-mod-core contracts interval in the frequency validation section.  Additionally, I updated the validation for the interval frequency in the scheduler as well.   I have not included a schema change for any type persistence.  Nor have I included a new format field

anonymouse64:
Great, thanks for the in depth response Eric. 

I personally won't be able to review this in depth this week, but it has my general +1 for the approach. If it's still open next week I will review, but feel free to merge without my approval this week

xmlviking:
> Great, thanks for the in depth response Eric.
> 
> I personally won't be able to review this in depth this week, but it has my general +1 for the approach. If it's still open next week I will review, but feel free to merge without my approval this week

still open requires rebase but requires your review Ian.   Let me know if you can't get around to this one and I can tap someone else.  Thanks :)

xmlviking:
@brandonforster can I assume your changes have been integrated satisfactory for this PR?  I just don't want to assume they have been.

internal/support/scheduler/schedulercontext_test.go
```diff
@@ -182,26 +182,62 @@ func TestIsComplete(t *testing.T) {
...
+	if err != nil {
+		t.Errorf(TestUnexpectedMsgFormatStrForInt64Val, duration.Nanoseconds(), 50)
+	}
+	if duration.Nanoseconds() != int64(50) {
+		t.Errorf(TestUnexpectedMsgFormatStrForInt64Val, duration.Nanoseconds(), 50)
+	}
+}
+
+// Note Time.Duration does not support milliseconds, or microseconds directly.
+func TestParseMicrosecondsFrequency(t *testing.T) {
```
akramtexas: - Making a recommendation for future unit-testing organization. While these tests are thorough, they could benefit from conciseness by the use of a `table-driven` format (it really helps cuts through repetitive testing steps), and an example of which would be the unit tests in [operation_test.go]( https://github.com/edgexfoundry/edgex-go/blob/master/internal/system/agent/executor/operation_test.go)
- The PR as it stands LGTM 👍 
- Approving the PR, pending building checks following the `rebase`.

xmlviking:
recheck

PR #1629, opened 2019-08-06
Apply Operator Pattern - Notification Get By End

nkstanchev:
Addresses #1628.

### DIFF
--- a/internal/support/notifications/operators/notification/get_test.go
+++ b/internal/support/notifications/operators/notification/get_test.go
@@ -67,7 +67,7 @@ func TestIdExecutor(t *testing.T) {
 			expectedError:  true,
 		},
 		{
-			name:           "Unsuccessful database call",
+			name:           "Notification not found",
 			mockDb:         createMockNotificiationLoaderStringArg("GetNotificationById", ErrorNotFound, contract.Notification{}, Id),
 			expectedResult: contract.Notification{},
 			expectedError:  true,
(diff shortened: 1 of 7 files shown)


internal/support/notifications/operators/notification/get_test.go
```diff
@@ -267,6 +267,55 @@ func TestStartExecutor(t *testing.T) {
...
+			expectedError:  false,
+		},
+		{
+			name:           "Unsuccessful database call",
+			mockDb:         createMockNotificiationLoaderStartStringArg("GetNotificationsByEnd", Error, []contract.Notification{}, End, Limit),
+			expectedResult: []contract.Notification{},
+			expectedError:  true,
+		},
+		{
+			name:           "Unsuccessful database call",
```
tsconn23: Not really unsuccessful DB call, more like successful but the item wasn't found.

internal/support/notifications/rest_notification.go
```diff
@@ -345,20 +345,16 @@ func notificationByEndHandler(w http.ResponseWriter, r *http.Request) {
...
-	n, err := dbClient.GetNotificationsByEnd(end, limitNum)
+	op := notification.NewEndExecutor(dbClient, end, limitNum)
+	results, err := op.Execute()
 	if err != nil {
-		if err == db.ErrNotFound {
-			http.Error(w, "Notification not found", http.StatusNotFound)
-		} else {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-		}
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
tsconn23: We can't stop returning 404 here since that's what the published v1.0.0 API does. In the Execute() implementation check to see if there are zero results. If so, return a db.ErrNotFound
```
if len(res) == 0 {
	return res, db.ErrNotFound
}
```
From there you can preserve the previous type checking and send the appropriate http code.

Issue #1635, opened 2019-08-07
security-secrets-setup docker image should not generate certs on image build time


jim-wang-intel:
The current security-secrets-setup docker image produces the certificates of root ca and signed vault and kong certs.  This creates the vulnerability of the same certificates for all docker images use these services.   We need to fix the docker images to not all have the same certificate setup.

Current CWE issues in `security-secrets-setup`:  
1. CWE-318: Cleartext Storage of Sensitive Information in Executable 
2. CWE-313: Cleartext Storage in a File or on Disk

jim-wang-intel:
This can be done in two phases: 1) through the script like @anonymouse64 Ian suggested in snap as the short-term solution and until then 2) integrating the full features of generation from Pki-Init services (which can be done piece meal, small chunk at a time for better review and tests)

tingyuz:
#1794 and #1830 addressed this issue, closed.

PR #1638, opened 2019-08-08
Fix security-secrets-setup vulnerability issue of sensitive TLS assets on docker image

jim-wang-intel:
- Remove the execution of security-secrets-setup inside Dockerfile to avoid generating TLS assets on the build time so that docker image does not contain sensitive information 

- Add docker entrypoint shell script to generate TLS assets of Vault/Kong on run-time if not exists yet

With above two it circumvents the issue CWE-318 of CA root private key although the whole process still running on the persistent storage (which should be addressed once features of Pki-Init Go-code are integrated into security-secrets-setup as one: using tmpfs running on the memory, not on the disk).  Nevertheless, with this step 0, it serves the purpose to pave the path to code-refactor security-secrets-setup with Pki-Init (which is much bigger effort) in terms of docker-image aspects.  It also unblocks the road to Fuji release due to vulnerability point-of-view.

Ref: issue #1635 

### DIFF
--- a/cmd/security-secrets-setup/Dockerfile
+++ b/cmd/security-secrets-setup/Dockerfile
@@ -1,6 +1,7 @@
 #  ----------------------------------------------------------------------------------
 #  Copyright 2018 ForgeRock AS.
 #  Copyright 2019 Dell Technologies, Inc.
+#  Copyright 2019 Intel Corp.
 #
 #  Licensed under the Apache License, Version 2.0 (the "License");
 #  you may not use this file except in compliance with the License.
@@ -14,11 +15,11 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 #
-#  @author: Alain Pulluelo, ForgeRock (created: June 14, 2018)
-#  @version: 1.0.0
+#  @author: Alain Pulluelo, ForgeRock (created: Jun 14, 2018)
 #
 #  @author: Trevor Conn, Dell Technologies, Inc. (created: July 2, 2019)
-#  @version: 1.0.1
+#
+#  @author: Jim Wang, Intel Corp. (modified: August 8, 2019)
 #
 #  SPDX-License-Identifier: Apache-2.0'
 #  ----------------------------------------------------------------------------------
@@ -50,11 +51,17 @@ LABEL license='SPDX-License-Identifier: Apache-2.0' \
 
 USER root
 
+# install necessary tools
+RUN apk update && apk add ca-certificates dumb-init jq \
+    && rm -rf /var/cache/apk/* 
+
+ENV BASE_DIR=/vault
+
 # Vault Config File
-WORKDIR /vault/config
+WORKDIR $BASE_DIR/config
 COPY --from=build-env /edgex-go/cmd/security-secrets-setup/res/local-tls.hcl ./local.hcl
 
-WORKDIR /vault
+WORKDIR $BASE_DIR
 RUN mkdir res
 
 # Vault PKI/TLS setup/config binary
@@ -66,15 +73,20 @@ COPY --from=build-env /edgex-go/cmd/security-secrets-setup/res/pkisetup-vault.js
 # Kong PKI/TLS materials
 COPY --from=build-env /edgex-go/cmd/security-secrets-setup/res/pkisetup-kong.json ./res
 
+# setup the entry point script
+# it uses `dumb-init` as the top-level process to reap any
+# zombie processes created by sub-processes
+COPY --from=build-env /edgex-go/cmd/security-secrets-setup/entrypoint.sh /usr/local/bin/
+RUN chmod +x /usr/local/bin/entrypoint.sh \
+    && ln -s /usr/local/bin/entrypoint.sh /
+
 # Create assets folder (needed for unseal key/s, root token and tmp)
 # Run CA/Vault and Kong PKI/TLS setups and peform housekeeping tasks
-RUN mkdir /vault/config/assets && \
-    chown -R vault:vault /vault && \
-    chmod 644 /vault/config/local.hcl && \
-    chmod 744 security-secrets-setup && \
-    /vault/security-secrets-setup --config /vault/res/pkisetup-vault.json && \
-    echo "" && \
-    /vault/security-secrets-setup --config /vault/res/pkisetup-kong.json && \
-    chown -R vault:vault /vault/config/pki
+RUN mkdir $BASE_DIR/config/assets && \
+    chown -R vault:vault $BASE_DIR && \
+    chmod 644 $BASE_DIR/config/local.hcl && \
+    chmod 744 security-secrets-setup
 
-VOLUME /vault/config
+VOLUME $BASE_DIR/config
+
+ENTRYPOINT ["entrypoint.sh"]
--- a/cmd/security-secrets-setup/entrypoint.sh
+++ b/cmd/security-secrets-setup/entrypoint.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/dumb-init /bin/sh
+#  ----------------------------------------------------------------------------------
+#  Copyright (c) 2019 Intel Corporation
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+#  SPDX-License-Identifier: Apache-2.0'
+#  ----------------------------------------------------------------------------------
+
+set -e
+
+# Use dumb-init as PID 1 in order to reap zombie processes and forward system signals to 
+# all processes in its session. This can alleviate the chance of leaking zombies, 
+# thus more graceful termination of all sub-processes if any.
+
+# runtime directory is set per user:
+export XDG_RUNTIME_DIR=${XDG_RUNTIME_DIR:-/run/user/$(echo $(id -u))}
+
+PKI_INIT_RUNTIME_DIR=${XDG_RUNTIME_DIR}/${PKI_INIT_DIR}
+
+# debug output:
+echo XDG_RUNTIME_DIR $XDG_RUNTIME_DIR
+echo PKI_INIT_RUNTIME_DIR $PKI_INIT_RUNTIME_DIR
+
+# configuration for TLS materials:
+PKI_CONFIG_JSON_DIR="res"
+PKI_SETUP_VAULT_FILE=${PKI_CONFIG_JSON_DIR}"/pkisetup-vault.json"
+PKI_SETUP_KONG_FILE=${PKI_CONFIG_JSON_DIR}"/pkisetup-kong.json"
+
+# check if files exists
+if [ ! -f "${PKI_SETUP_VAULT_FILE}" ]; then
+    echo "Error: certificate config file for Vault is missing"
+    exit 1
+fi
+
+if [ ! -f "${PKI_SETUP_KONG_FILE}" ]; then
+    echo "Error: certificate config file for Kong is missing"
+    exit 1
+fi
+
+# the working dir should be in vault dir based upon the current Docker image
+BASE_DIR="${BASE_DIR:-/vault}"
+cd $BASE_DIR
+CERT_DIR=$(jq -r '.working_dir' ${PKI_SETUP_VAULT_FILE})
+CERT_SUBDIR=$(jq -r '.pki_setup_dir' ${PKI_SETUP_VAULT_FILE})
+ROOT_NAME=$(jq -r '.x509_root_ca_parameters | .ca_name' ${PKI_SETUP_VAULT_FILE})
+CERT_EXEC="${CERT_EXEC:-./security-secrets-setup}"
+# check to see if the root certificate generate with security-secrets-setup already exists
+# if so then do not generate a new set of them
+if [ ! -f "$CERT_DIR/$CERT_SUBDIR/$ROOT_NAME/$ROOT_NAME.pem" ]; then
+    ${CERT_EXEC} --config ${PKI_SETUP_VAULT_FILE}
+    [ $? -eq 0 ] || (echo "failed to generate TLS assets for Vault" && exit 1)
+
+    ${CERT_EXEC} --config ${PKI_SETUP_KONG_FILE}
+    [ $? -eq 0 ] || (echo "failed to generate TLS assets for Kong" && exit 1)
+
+    # delete CA private key
+    rm "$PWD/$CERT_DIR/$CERT_SUBDIR/$ROOT_NAME/$ROOT_NAME.priv.key"
+    [ $? -eq 0 ] || (echo "failed to delete sensitive CA private key" && exit 1)
+
+    # take ownership
+    chown -R vault:vault $PWD/$CERT_DIR/$CERT_SUBDIR || true
+fi 
+
+# run the Vault's docker-entry script 
+source /usr/local/bin/docker-entrypoint.sh
--- a/cmd/security-secrets-setup/README.md
+++ b/cmd/security-secrets-setup/README.md
@@ -1,0 +1,27 @@
+# Testing Instruction for security-secrets-setup
+
+## Build
+
+Use Makefile in the base directory of this repository to build the docker image of security-secretes-setup:
+
+```sh
+make docker_security_secrets_setup
+```
+
+It should create a docker image with the name `edgexfoundry/docker-edgex-vault:<version>-dev` if sucessfully built.
+
+# Run
+
+To see what the TLS materials are generated inside the docker container, 
+one can run the built docker image above (assuming the version number is `1.1.0`) as:
+
+```sh
+docker run --rm -it edgexfoundry/docker-edgex-vault:1.1.0-dev /bin/sh
+```
+
+once that is run, one should see the execution console outputs and gives the interactive console prompt in the base directory `/vault`.  Based on the current JSON configuration, TLS materials should be able to found in the directory of `./config/pki/EdgeXFoundryCA/` showed as follows:
+
+```sh
+/vault # ls ./config/pki/EdgeXFoundryCA/
+EdgeXFoundryCA.pem    edgex-kong.pem        edgex-kong.priv.key   edgex-vault.pem       edgex-vault.priv.key
+```


jim-wang-intel:
@tingyu, @tsconn23 @anonymouse64 please review

jim-wang-intel:
recheck

cmd/security-secrets-setup/Dockerfile
```diff
@@ -66,15 +71,20 @@ COPY --from=build-env /edgex-go/cmd/security-secrets-setup/res/pkisetup-vault.js
...
 # Run CA/Vault and Kong PKI/TLS setups and peform housekeeping tasks
 RUN mkdir /vault/config/assets && \
     chown -R vault:vault /vault && \
     chmod 644 /vault/config/local.hcl && \
-    chmod 744 security-secrets-setup && \
-    /vault/security-secrets-setup --config /vault/res/pkisetup-vault.json && \
-    echo "" && \
-    /vault/security-secrets-setup --config /vault/res/pkisetup-kong.json && \
-    chown -R vault:vault /vault/config/pki
+    chmod 744 security-secrets-setup
```
bnevis-i: Granted that this is not currently broken since root is running the executable, but executables should generally be chmod 755.

cmd/security-secrets-setup/entrypoint.sh
```diff
@@ -0,0 +1,74 @@
...
+
+    # delete CA private key
+    rm "$PWD/$CERT_DIR/$CERT_SUBDIR/$ROOT_NAME/$ROOT_NAME.priv.key"
+    [ $? -eq 0 ] || (echo "failed to delete sensitive CA private key" && exit 1)
+
+    # take ownership
+    chown -R vault:vault $PWD/$CERT_DIR/$CERT_SUBDIR || true
+fi 
+
+exec "$@"
```
bnevis-i: I think we should be exec'ing /usr/local/bin/docker-entrypoint.sh here since that was the original entrypoint for the container.

cmd/security-secrets-setup/Dockerfile
```diff
@@ -50,6 +51,10 @@ LABEL license='SPDX-License-Identifier: Apache-2.0' \
 
 USER root
 
+# install necessary tools
+RUN apk update && apk add ca-certificates dumb-init jq \
```
bnevis-i: It is more efficient code to write
``
apk add --no-cache ca-certificates dumb-init jq
```

Instead of writing apk update before and rm /var/cache/apk after.

Source: https://stackoverflow.com/questions/49118579/alpine-dockerfile-advantages-of-no-cache-vs-rm-var-cache-apk

File: cmd/security-secrets-setup/Dockerfile [hunk 4]
```diff
@@ -66,15 +71,20 @@ COPY --from=build-env /edgex-go/cmd/security-secrets-setup/res/pkisetup-vault.js
 # Kong PKI/TLS materials
 COPY --from=build-env /edgex-go/cmd/security-secrets-setup/res/pkisetup-kong.json ./res
 
+# setup the entry point script
+# it uses `dumb-init` as the top-level process to reap any
+# zombie processes created by sub-processes
+COPY --from=build-env /edgex-go/cmd/security-secrets-setup/entrypoint.sh /usr/local/bin/
+RUN chmod +x /usr/local/bin/entrypoint.sh
+RUN ln -s usr/local/bin/entrypoint.sh /
```
bnevis-i: These should be combined:
```
RUN chmod +x /usr/local/bin/entrypoint.sh && ln -s /usr/local/bin/entrypoint.sh /
```

cmd/security-secrets-setup/Dockerfile
jim-wang-intel: 744 was there originally. I agree that an executable file usually is 755.

cmd/security-secrets-setup/entrypoint.sh
jim-wang-intel: This one is provisional for future binary/executable command arguments.  The `docker-entrypoint.sh` is always the one to be run as it is specified in the `ENTRYPOINT` of the Dockerfile. `$@` is a shell variable that means all command arguments, in which is specified in `CMD` of the Dockerfile if any.  I'll put off the creating vault config from env for now as it is not this step to do yet.

cmd/security-secrets-setup/Dockerfile
jim-wang-intel: I use `rm` literally since `--no-cache` is not working for me as expected locally here.  When I used `--no-cache` and then tried to run docker image exec, the cache folder still there not being cleaned up or deleted for some reason. Maybe you can try and see.

cmd/security-secrets-setup/Dockerfile
jim-wang-intel: ok

cmd/security-secrets-setup/entrypoint.sh
bnevis-i: Except that you changed the entrypoint here: https://github.com/edgexfoundry/edgex-go/blob/d679d553706f5fbbeb76b7501e36641dea75b722/cmd/security-secrets-setup/Dockerfile#L90

cmd/security-secrets-setup/Dockerfile
jim-wang-intel: so I guess, the original intention is that this executable is owned by `vault` user as the ownership is transferred to `vault` user and `vault` group in line 71 of `docker-entrypoint.sh`.  So I'll leave as it is for now unless there is a strong objection.

cmd/security-secrets-setup/entrypoint.sh
jim-wang-intel: yes, that is where you can specify the `ENTRYPOINT` to run inside the docker image and it is specified on the build time.

cmd/security-secrets-setup/Dockerfile
```diff
@@ -66,15 +71,20 @@ COPY --from=build-env /edgex-go/cmd/security-secrets-setup/res/pkisetup-vault.js
 # Kong PKI/TLS materials
 COPY --from=build-env /edgex-go/cmd/security-secrets-setup/res/pkisetup-kong.json ./res
 
+# setup the entry point script
+# it uses `dumb-init` as the top-level process to reap any
+# zombie processes created by sub-processes
+COPY --from=build-env /edgex-go/cmd/security-secrets-setup/entrypoint.sh /usr/local/bin/
+RUN chmod +x /usr/local/bin/entrypoint.sh \
+    && ln -s usr/local/bin/entrypoint.sh /
```
bnevis-i: Why drop the leading / from "/usr"?

jim-wang-intel: good catch, operator error :-)

cmd/security-secrets-setup/entrypoint.sh
jim-wang-intel: @bnevis-i got your point and changed accordingly

bnevis-i:
Latest comment addresses all of my concerns.

jim-wang-intel:
@tingyuz, @anonymouse64 , and @tsconn23 any more inputs before this can be merged?

anonymouse64:
I will try to get around to testing this today, but I will submit my re-review since there were a number of changes made since my initial review

cmd/security-secrets-setup/entrypoint.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/dumb-init /bin/sh
```
anonymouse64: it would be nice if instead of using dumb-init in the shebang we just use /bin/sh, and then call /usr/bin/dumb-init from the `ENTRPOINT` in the Dockerfile and use the `ENTRYPOINT` for the actual entrypoint.sh script as it currently is. This would allow us to try and use the same script in the snap, assuming that the environment variables are setup appropriately so we can override them from the snap.

However this is just a nice to have, since the snap already has a mostly working script, so I'm fine to push this off to a new issue if you just want to merge as-is to fix docker ASAP.

cmd/security-secrets-setup/entrypoint.sh
```diff
@@ -0,0 +1,75 @@
...
+    exit 1
+fi
+
+if [ ! -f "${PKI_SETUP_KONG_FILE}" ]; then
+    echo "Error: certificate config file for Kong is missing"
+    exit 1
+fi
+
+# the working dir should be in vault dir based upon the current Docker image
+cd /vault
```
anonymouse64: we should use an environment variable here rather than hard-coding the working dir to always be `/vault`

cmd/security-secrets-setup/entrypoint.sh
```diff
@@ -0,0 +1,75 @@
...
+CERT_DIR=$(jq -r '.working_dir' ${PKI_SETUP_VAULT_FILE})
+CERT_SUBDIR=$(jq -r '.pki_setup_dir' ${PKI_SETUP_VAULT_FILE})
+ROOT_NAME=$(jq -r '.x509_root_ca_parameters | .ca_name' ${PKI_SETUP_VAULT_FILE})
+CERT_EXEC="./security-secrets-setup"
```
anonymouse64: this variable should have a default, because as it is, this requires the CERT_EXEC to always be `/vault/security-secrets-setup`, when for example with the snap we might want to use the same script here but with different working directories and with a different location of the binary (i.e. the binary would be at `$SNAP/bin/security-secrets-setup` and the working dir would be somewhere like `$SNAP_DATA/vault/pki`)

anonymouse64: this variable should have a default, because as it is, this requires the CERT_EXEC to always be `/vault/security-secrets-setup`, when for example with the snap we might want to use the same script here but with different working directories and with a different location of the binary (i.e. the binary would be at `$SNAP/bin/security-secrets-setup` and the working dir would be somewhere like `$SNAP_DATA/vault/pki`)

cmd/security-secrets-setup/entrypoint.sh
bnevis-i: This is actually modeled after the `docker-entrypoint.sh` that ships with Vault.  Since we're replacing the ENTRYPOINT to include a preprocessing script we are doing it the same way Vault is doing it.

anonymouse64: Right I understand that's how it's done in the vault Dockerfile, but I think my suggestion still stands as a possible improvement. I'm not sure of how to do this by inheriting the default args in the vault upstream Dockerfile, but you could certainly do something like:

```dockerfile
ENTRYPOINT ["/usr/bin/dumb-init"]
CMD ["/vault/our-entrypoint.sh", "server","-dev"]
```

(since the upstream dockerfile for vault at 1.0.1 has `CMD ["server","-dev"]` see https://github.com/hashicorp/docker-vault/blob/6ca0fa239f6a7150dfab827a0852a0031a02e028/0.X/Dockerfile#L80)

anonymouse64: Another way this could be done is to use a chain of scripts where at the end of the script, it executes all of it's arguments a la:

```bash
#!/bin/sh -e

# do stuff

exec "$@"
```

which you could then implement the desired effect with:

```dockerfile
ENTRYPOINT ["/usr/bin/dumb-init"]
CMD ["/vault/our-entrypoint.sh", "/usr/local/bin/upstream-vault-entrpoint.sh","server","-dev"]
```

cmd/security-secrets-setup/entrypoint.sh
```diff
@@ -0,0 +1,75 @@
...
+set -e
+
+# Use dumb-init as PID 1 in order to reap zombie processes and forward system signals to 
+# all processes in its session. This can alleviate the chance of leaking zombies, 
+# thus more graceful termination of all sub-processes if any.
+
+# runtime directory is set per user:
+export XDG_RUNTIME_DIR=${XDG_RUNTIME_DIR:-/run/user/$(echo $(id -u))}
+
+PKI_INIT_RUNTIME_DIR=${XDG_RUNTIME_DIR}${PKI_INIT_DIR}
```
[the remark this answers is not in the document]
jim-wang-intel: sure, this sounds good

cmd/security-secrets-setup/entrypoint.sh
jim-wang-intel: I think the way Vault's implementation looks cleaner to me since it centralizes the implementation in the shell script instead of spreading the logic into two places.

cmd/security-secrets-setup/entrypoint.sh
jim-wang-intel: ok, we can create an env for this vault working dir

cmd/security-secrets-setup/entrypoint.sh
jim-wang-intel: so maybe we should also use an env for this `CERT_EXEC` then? That way, you in the snap you can set up this env to be different before run it.

cmd/security-secrets-setup/entrypoint.sh
jim-wang-intel: so, Ian, why in the SNAP process, we do not use dumb-init to reap left-over zombie processes? If in the stand-alone, still having the same issue, I think we should use dumb-init the same way here.

anonymouse64: In snaps, services are implemented using systemd, which is a real init process and doesn't have the same issue

cmd/security-secrets-setup/entrypoint.sh
anonymouse64: Yes, you could define it like so:

```bash
CERT_EXEC=${CERT_EXEC:-./security-secrets-setup}
```

cmd/security-secrets-setup/entrypoint.sh
jim-wang-intel: ok, so the initial assumption is different then- in order to share the shell script code or at the most of codes, maybe can introduce another shell script and then from certain point of process, it calls into that shared script then. IMHO, however, since this is small enough script, I am not sure if it worth the efforts to do that kind of refactor or not (just for the sake of small duplication of codes?)

anonymouse64: As I said, it's probably not worth blocking this PR on refactoring to handle the snap as well, but the benefit is that both docker and the snap are calling `security-secrets-setup` in the same manner, which maintains consistency between the two. If we find a bug or problem with how security-secrets-setup is being called, with a shared script like this we reduce the amount of work it takes to fix and update that and greatly reduce the risk that one distribution method works and the other one doesn't. But since the next rewrite of this code will likely be to just call security-secrets-setup directly instead, I think it's fine to have 2 different implementations for now.

bnevis-i: The goal that I am trying to steer us towards is that we run security-secrets-setup in its own container after merging it with pki-init.  We would then run a stock version of Vault and override its RUN command to block on PKI availability instead of making the custom version that we do now.

anonymouse64: Sure, I understand this PR is just meant to be temporary, if that's your plan and it's too much work then feel free to ignore this suggestion for the time being.

jim-wang-intel:
@anonymouse64 Ian, I added the README.md [here](https://github.com/jim-wang-intel/edgex-go/blob/fix-security-secrets-setup-cwe318/cmd/security-secrets-setup/README.md).  Please read and run it accordingly to see if that works for you.

jim-wang-intel:
Thanks, Ian @anonymouse64 .  So would you please, @tsconn23 merge this onto master if you feel good about it? I do not have the permission to do so.

anonymouse64:
@jim-wang-intel please rebase and then I will merge

jim-wang-intel:
@anonymouse64 Ian, ok, do you also want to squash to one commit or it dose not matter to you?

anonymouse64:
@jim-wang-intel the changes here are small enough that I think you could squash to one commit when you rebase, provided you update the commit description to match what the final changes ended up being.

jim-wang-intel:
@anonymouse64 Ian done rebase and squash!

anonymouse64:
@jim-wang-intel could you add more description to the commit message explaining how you resolved the issue? As is, the commit message doesn't say anything about why the change was necessary. See 
- https://chris.beams.io/posts/git-commit/
- https://github.com/erlang/otp/wiki/writing-good-commit-messages
- https://medium.com/compass-true-north/writing-good-commit-messages-fc33af9d6321

mkbhanda:
Good work on patch Jim. Echoing Ian’s comments here.

Suggested commit message:

Moves the creation of PKI keys and certificate from container build to launch time for uniqueness.

Originally the PKI key pair and certificate were created at container build time, resulting in
all deployments having the same key material. The key creation code is now executed as
part of docker container launch entrypoint code. Environment variable BASE_DIR was added to
allow custom storage location for keys, but defaults to vault. Environment variable CERT_EXEC ..
Absolute path support was added … (forward slash)
README now captures how to manually test functional correctness of unique key creation

Please skip the below individual commits, a full squash, blurring individual commits.

@jim-wang-intel<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fjim-wang-intel&data=02%7C01%7Cmbhandaru%40vmware.com%7Ce4d5ce6b04f644e8fe5508d722944880%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637015892578838372&sdata=W9f3ytQxmsJiEZSvP18mXMxgCtxnE7%2BNMD%2FKhda8bJI%3D&reserved=0> pushed 8 commits.

  *   6f04b36<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fedgexfoundry%2Fedgex-go%2Fcommit%2F6f04b3639930efa2aff3b05baee8d8ab1d055a12&data=02%7C01%7Cmbhandaru%40vmware.com%7Ce4d5ce6b04f644e8fe5508d722944880%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637015892578838372&sdata=fXdpX8oGt2haPmAF8TLe22hDK9WhP9NYyoADXlVcQb4%3D&reserved=0> Removal of config-seed/res/properties/go.mod
  *   b748eb6<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fedgexfoundry%2Fedgex-go%2Fcommit%2Fb748eb6a2f3427197be86f5339ccaca8c0f52719&data=02%7C01%7Cmbhandaru%40vmware.com%7Ce4d5ce6b04f644e8fe5508d722944880%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637015892578848368&sdata=fYnVpMaadpvAOjfxs%2FeUqdczOxwBBGLqG1ikLW5703E%3D&reserved=0> Merge pull request #1681 from tsconn23/issue_1680
  *   e80b802<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fedgexfoundry%2Fedgex-go%2Fcommit%2Fe80b802e59f11f313ecc6b1d79baa352bd7ac59b&data=02%7C01%7Cmbhandaru%40vmware.com%7Ce4d5ce6b04f644e8fe5508d722944880%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637015892578848368&sdata=De7XV%2B2uEOIwAx90vqp7cG8a1OCltnOLEfpnaGl%2B80M%3D&reserved=0> fix security-secrets-setup cwe318
  *   cc4e7e4<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fedgexfoundry%2Fedgex-go%2Fcommit%2Fcc4e7e442463e0d148cac4fa28cbc01a0603a696&data=02%7C01%7Cmbhandaru%40vmware.com%7Ce4d5ce6b04f644e8fe5508d722944880%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637015892578858364&sdata=2DdGpfiaFwVG0gPzV7GEoqJW5tUP3H%2F%2BXonFdRyAX94%3D&reserved=0> Address PR feedback and comments
  *   b551090<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fedgexfoundry%2Fedgex-go%2Fcommit%2Fb55109082e9fed8063f583faca9e5012e2faa945&data=02%7C01%7Cmbhandaru%40vmware.com%7Ce4d5ce6b04f644e8fe5508d722944880%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637015892578858364&sdata=%2B3SNuVBUFosHQSTOzWu6Jc2jWilQE7jGs3Id%2BoJCU7A%3D&reserved=0> Fix the missing leading / for symbolic link's target
  *   c192952<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fedgexfoundry%2Fedgex-go%2Fcommit%2Fc192952c69e204def835db0793a2876dc6a93a56&data=02%7C01%7Cmbhandaru%40vmware.com%7Ce4d5ce6b04f644e8fe5508d722944880%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637015892578868359&sdata=ZCwduCrPl4IDO6tMD8%2FispA%2FGAn9rmcluJTT2QdSXs0%3D&reserved=0> Address PR feedback regarding docker-entrypoint shell script
  *   729c0f9<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fedgexfoundry%2Fedgex-go%2Fcommit%2F729c0f9f7a131ad2c69d4046d322de94de736bc3&data=02%7C01%7Cmbhandaru%40vmware.com%7Ce4d5ce6b04f644e8fe5508d722944880%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637015892578868359&sdata=fkGSnN1yt3HfkYK3IAgRumrvg8Wv7atT4ZbWt2VCyr0%3D&reserved=0> Address PR feedback- adding envs for BASE_DIR and CERT_EXEC
  *   e06c2c9<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fedgexfoundry%2Fedgex-go%2Fcommit%2Fe06c2c9bfb08c31172b824e3b2ef559091eeb615&data=02%7C01%7Cmbhandaru%40vmware.com%7Ce4d5ce6b04f644e8fe5508d722944880%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637015892578868359&sdata=GqqbLj6i5GCn%2BCVqfcl%2Fg9f1HAtKPvQKoTw427IwvzA%3D&reserved=0> Add readme for instruction on how to run security-secrets-setup in docker

jim-wang-intel:
@anonymouse64 Ian, pushed my new commit message

Issue #1626, opened 2019-08-05
Unit Test -- Core Metadata Upload Device Profile


brandonforster:
This issue captures both uploading as raw and uploading as file.

https://github.com/edgexfoundry/edgex-go/projects/11#card-23265666

https://github.com/edgexfoundry/edgex-go/projects/11#card-23265680

PR #1640, opened 2019-08-08
Added unit tests for restAddDeviceProfile

### LINKED ISSUE #1639: Unit Test -- Core-Metadata Add Device Profile
https://github.com/edgexfoundry/edgex-go/projects/11#card-23265629

brandonforster:
Fix #1639 

Reused the operator and centralized logic from #1626


internal/core/metadata/rest_deviceprofile_test.go
```diff
@@ -123,6 +124,102 @@ func TestGetAllProfiles(t *testing.T) {
...
+	tests := []struct {
+		name           string
+		request        *http.Request
+		dbMock         interfaces.DBClient
+		vdcMock        MockValueDescriptorClient
+		expectedStatus int
+	}{
+		{
+			"OK",
+			createRequestWithBody(TestDeviceProfile),
```
tsconn23: Drilling into this function it always returns a "PUT" request. This seems odd given that `restAddDeviceProfile` is mapped to a POST in router.go. Would prefer these to align. Solution is to specify `http.Method` as this function does `createRequestWithPathParameters(http.MethodGet, map[string]string{})`

PR #1663, opened 2019-08-14
Add missing validation for DeviceProfile APIs

### LINKED ISSUE #1603: PR 1557 Follow-up (Value Descriptors)
Following up on this comment https://github.com/edgexfoundry/edgex-go/pull/1557#pullrequestreview-268660608

I will work with @AnthonyMBonafide when he returns from vacation to address the following

- Before making any changes to Value Descriptors (add, update, delete) we should first be sure we can perform the requested operation on the Device Profile. In the case of an add, we should perform the duplicate check. In the case of an update/delete, integrity checks regarding devices and readings.
- Right now, we're doing all the Value Descriptor operations before handling the Device Profile. If the Device Profile operation fails, we've already made the changes to the Value Descriptors. This results in the following bug --> If I have a Device Profile associated to a Device and I try to update it including Value Descriptor changes, the Value Descriptor changes are applied even though an error saying `Device Profile in Use` is returned. 
- From what I can see, Value Descriptors are not cleaned up when a Device Profile is deleted.

AnthonyMBonafide:
Fix #1603

Add missing validation checks for DeviceProfile's Update and Add APIs
which allowed the system to enter an undesirable state due to orphaned
ValueDescriptors.

Add validation check for DeviceProfile non-unique name during Add before
creating the associated ValueDescriptors.

Add validation check to ensure the DeviceProfile is not in use by a
Device before performing the update.

Add necessary tests to cover the newly added logic.

### DIFF
--- a/internal/core/metadata/operators/device_profile/value_descriptor.go
+++ b/internal/core/metadata/operators/device_profile/value_descriptor.go
@@ -3,11 +3,13 @@ package device_profile
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 
 	"github.com/edgexfoundry/edgex-go/internal/core/data/errors"
+	errors2 "github.com/edgexfoundry/edgex-go/internal/core/metadata/errors"
 )
 
 // ValueDescriptorAdder provides the necessary functionality for creating a ValueDescriptor.
@@ -51,8 +53,8 @@ func (a addValueDescriptor) Execute() error {
 	return nil
 }
 
-// NewAddExecutor creates a new ValueDescriptorAddExecutor.
-func NewAddExecutor(ctx context.Context, client ValueDescriptorAdder, loggingClient logger.LoggingClient, drs ...contract.DeviceResource) ValueDescriptorAddExecutor {
+// NewAddValueDescriptorExecutor creates a new ValueDescriptorAddExecutor.
+func NewAddValueDescriptorExecutor(ctx context.Context, client ValueDescriptorAdder, loggingClient logger.LoggingClient, drs ...contract.DeviceResource) ValueDescriptorAddExecutor {
 	return addValueDescriptor{
 		ctx:    ctx,
 		drs:    drs,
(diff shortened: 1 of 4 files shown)


AnthonyMBonafide:
Recheck

internal/core/metadata/operators/device_profile/value_descriptor.go
```diff
@@ -3,11 +3,13 @@ package device_profile
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 
 	"github.com/edgexfoundry/edgex-go/internal/core/data/errors"
+	errors2 "github.com/edgexfoundry/edgex-go/internal/core/metadata/errors"
```
tsconn23: Maybe what we need to consider is moving all error type definitions to internal/pkg/errors to avoid this kind of aliasing. Separate issue, of course.

internal/core/metadata/rest_deviceprofile.go
```diff
@@ -62,8 +62,18 @@ func restAddDeviceProfile(w http.ResponseWriter, r *http.Request) {
...
+		// will fail during the creation process later on.
+		nameOp := device_profile.NewGetProfileName(dp.Name, dbClient)
+		_, err := nameOp.Execute()
+		// The operator will return an ItemNotFound error if the DeviceProfile can not be found.
+		if err == nil {
+			http.Error(w, "Duplicate name for device profile", http.StatusConflict)
+			return
+		}
+
 		op := device_profile.NewAddExecutor(r.Context(), vdc, LoggingClient, dp.DeviceResources...)
```
tsconn23: This should be `device_profile.ValueDescriptorAddExecutor` because as it is now, it reads as if it's a DeviceProfile Add executor.

internal/core/metadata/rest_deviceprofile.go
```diff
@@ -102,6 +112,8 @@ func restUpdateDeviceProfile(w http.ResponseWriter, r *http.Request) {
 				http.Error(w, err.Error(), http.StatusNotFound)
 			case *errors2.ErrValueDescriptorsInUse:
 				http.Error(w, err.Error(), http.StatusConflict)
+			case errors.ErrDeviceProfileInvalidState:
+				http.Error(w, err.Error(), http.StatusConflict)
```
tsconn23: I also notice here the difference in referring to error types with pointers versus not. In truth, these don't have to be pointer references so I'm aligned to the logic that's been added. But we should have an issue to address this globally and standardize.

PR #1664, opened 2019-08-14
Unit Test -- DeviceService Update AdminState/ OpState

### LINKED ISSUE #1649: Unit Test -- Core-Metadata Update Device Service Admin/Operating State
https://github.com/edgexfoundry/edgex-go/projects/11#card-23264806

https://github.com/edgexfoundry/edgex-go/projects/11#card-23264780

https://github.com/edgexfoundry/edgex-go/projects/11#card-23264722

https://github.com/edgexfoundry/edgex-go/projects/11#card-23264738

brandonforster:
Fixes #1649 

Includes tests for REST endpoints.

The variable `TestURI` was renamed because I almost used it and I figured it was worth changing while the large scoped name with a narrow usage was still fresh in my head.


internal/core/metadata/operators/device_service/update_test.go
```diff
@@ -0,0 +1,307 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package device_service
+
+import (
+	goErrors "errors"
```
AnthonyMBonafide: I like this approach to aliasing the Go error import. I have been using `errors2` to alias our error packages, but I like this more as I think we should really be using our own errors types more often, which means aliasing the Go errors would make more sense.

internal/core/metadata/operators/device_service/update.go
```diff
@@ -0,0 +1,154 @@
...
+package device_service
+
+import (
+	"github.com/edgexfoundry/edgex-go/internal/core/metadata/errors"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
+
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+)
+
+type UpdateOpStateByIdExecutor interface {
```
AnthonyMBonafide: Would it be better to create one exported executor interface for all the updating functionality, since they all accept 0 arguments and return 1 error? 

We are doing this in the `device_profile` get operator:
https://github.com/edgexfoundry/edgex-go/blob/8627af8819795f333cf7a38c6bcd9bb746df233b/internal/core/metadata/operators/device_profile/get.go#L16-L19

In this case we have two types of interfaces as some of the getters return one `device_profile` while others return a slice of `device_profile` so we created 2 interfaces which cover all of the 7 getters.

Here is the interface for the executors which return more a slice:
https://github.com/edgexfoundry/edgex-go/blob/8627af8819795f333cf7a38c6bcd9bb746df233b/internal/core/metadata/operators/device_profile/get.go#L11-L14

internal/core/metadata/operators/device_service/update.go
```diff
@@ -0,0 +1,154 @@
...
+	Execute() error
+}
+
+type deviceServiceOpStateUpdateById struct {
+	id string
+	os contract.OperatingState
+	db DeviceServiceUpdater
+}
+
+func NewUpdateOpStateByIdExecutor(id string, os contract.OperatingState, db DeviceServiceUpdater) UpdateOpStateByIdExecutor {
```
AnthonyMBonafide: Can we add Go doc comments to the exported functions?

internal/core/metadata/rest_deviceservice.go
```diff
@@ -639,39 +614,22 @@ func restUpdateServiceAdminStateByName(w http.ResponseWriter, r *http.Request) {
...
+	w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
 	w.WriteHeader(http.StatusOK)
 	w.Write([]byte("true"))
 }
 
-// Update the admin state for the device service
-func updateServiceAdminState(ds models.DeviceService, as models.AdminState, w http.ResponseWriter) error {
-	ds.AdminState = as
-	if err := dbClient.UpdateDeviceService(ds); err != nil {
-		http.Error(w, err.Error(), http.StatusServiceUnavailable)
```
AnthonyMBonafide: I do not see this `503(Service Unavailable)` response code being used after these changes. Am I missing something or would these changes result in API changes?

For the record, I  think a `503` is not the correct response in this situation as it is meant to tell the client that the server is unavailable not that the entity does not exist.

brandonforster: This changes the API, and I agree with your points.

The RAML doesn't describe a 503, so my best guess is that when we changed all the 503s to 500, we missed this one. We'll be rectifying that now and aligning with the RAML.

PR #1672, opened 2019-08-15
Refactor code so as to have the Executor fetch metrics for services.

akramtexas:
Fix [#1486](https://github.com/edgexfoundry/edgex-go/issues/1486)
- This PR takes the work done in [PR-1609](https://github.com/edgexfoundry/edgex-go/pull/1609) and extends that work by incorporating the remaining feedback received therein.
- To test locally, please use the following (`SMA`) API:
> http://localhost:48090/api/v1/metrics/edgex-core-command
- If you want to test via the 'executor' metrics mechanism, then select (by uncommenting) the following setting in the SMA's `configuration.toml`:
`# ExecutorPath = '<YOUR-=COORDINATES>/go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-executor/sys-mgmt-executor'`
- Following that, based on which of the following two settings is selected (by uncommenting) at a time, respectively...
1. `# MetricsMechanism = 'direct-service'`
2. `# MetricsMechanism = 'executor'`

...you should see a response each (resembling one of the following two samples), respectively, and in turn, from your REST client.

**Per 1:**
```
{
    "Metrics": {
        "edgex-core-command": {
            "CpuBusyAvg": 6.728698199362453,
            "Memory": {
                "Alloc": 1520960,
                "Frees": 1322255,
                "LiveObjects": 6199,
                "Mallocs": 1328454,
                "Sys": 72284408,
                "TotalAlloc": 410826888
            }
        }
    }
}
```
**Per 2:**
```
{
    "Metrics": {
        "edgex-core-command": {
            "block_io": "10.2MB / 0B",
            "cpu_perc": "0.00%",
            "mem_perc": "0.25%",
            "mem_usage": "4.949MiB / 1.952GiB",
            "net_io": "5.71MB / 3.18MB",
            "pids": "18"
        }
    }
}
```


akramtexas:
Recheck

anonymouse64:
I'm not going to be able to deeply review this for a while, but looking at the example output from your comment, it seems that the metrics implementation here is still "hands-off", where what your metrics response looks likes in terms of responses is entirely dependent on the chosen metrics executor. Is this by design? I think it's problematic because, for example if I'm writing a client of the SMA and want to get known CPU usage I need to look at the `CpuBusyAvg` key with a `direct-service` executor, and the `cpu_perc` key for a `executor` executor. How would a client introspect from the SMA which executor is being used to generate metrics (apologies if this is clear from the code)?

I would have thought that we would collate known metrics into a common format for common metrics so that clients of the SMA don't need to know what metric backend the SMA is using. I do see how it could be useful for particular metrics implementations to provide additional information however, so perhaps what we need is a common set of metrics (again organized through interfaces such as `GetCPUMetrics` and `GetMemoryMetrics`, etc. for each understood metric) and then provide a catch-all "passthrough" metrics interface for implementations to dump other kinds of metrics not currently supported or recognized. 

Thoughts?

cmd/config-seed/res/properties/go.mod
```diff
@@ -1 +1,3 @@
 module github.com/edgexfoundry/edgex-go/cmd/config-seed/res/properties
+
+go 1.12
```
tsconn23: FYI PR https://github.com/edgexfoundry/edgex-go/pull/1681 has been created to remove this file altogether.

cmd/sys-mgmt-executor/main.go
```diff
@@ -14,27 +14,22 @@
 package main
 
 import (
-	"encoding/json"
 	"flag"
 	"fmt"
+	"github.com/edgexfoundry/edgex-go/internal/system/executor"
```
tsconn23: Organize imports

cmd/sys-mgmt-executor/main.go
```diff
@@ -58,109 +53,36 @@ func main() {
 	flag.Parse()
 
 	// Setup Logging
-	LoggingClient = logger.NewClient(SystemManagementExecutorKey, EnableRemote, LoggingTarget, LogLevel)
+	LoggingClient := logger.NewClient(SystemManagementExecutorKey, EnableRemote, LoggingTarget, LogLevel)
```
tsconn23: Question -- `EnableRemote` is a constant set to `false`. Should this be the case? If the other services in the platform are logging remotely, shouldn't this one as well? Refer to `setLoggingTarget()` function defined elsewhere for what this entails. Currently the executor doesn't have a configuration, but maybe it should.

cmd/sys-mgmt-executor/main.go
```diff
@@ -58,109 +53,36 @@ func main() {
...
-
-		correctStatus, err := checkDockerContainerStatus(service, expectedStatus)
-		if err != nil {
-			return err
-		}
-		if !correctStatus {
-			return fmt.Errorf("docker %s operation failed for service %s", operation, service)
+		} else {
+			LoggingClient.Error(fmt.Sprintf("the service %s is an unknown one for which the request was made to run Docker command", service))
 		}
```
tsconn23: It occurs to me that this app should utilize os.Exit() as the other services do. If exiting based on successful completion, use `os.Exit(0)`. If terminating due to error, use `os.Exit(1)`

internal/system/agent/router.go
```diff
@@ -55,6 +55,26 @@ func LoadRestRoutes() *mux.Router {
...
+	services = strings.Split(list, ",")
+
+	ctx := r.Context()
+	send, err := InvokeMetrics(services, ctx)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		LoggingClient.Error(err.Error())
+		return
+	}
+	w.Header().Add(clients.ContentType, clients.ContentTypeJSON)
```
tsconn23: Setting this header is redundant. `pkg.Encode` will do this for you.

internal/system/agent/services.go
```diff
@@ -34,18 +37,110 @@ const (
...
+				result, _ := processOutput(out)
+				m.Metrics[service] = ProcessResponse(result)
+			}
+		} else if Configuration.MetricsMechanism == "custom" {
+			err := fmt.Errorf("the requested custom executor (e.g. snap) has not been integrated")
+			LoggingClient.Error(err.Error())
+			m.Metrics[service] = fmt.Sprintf(err.Error())
+		} else {
+			err := fmt.Errorf("the requested metrics mechanism is not supported")
+			LoggingClient.Error(err.Error())
```
tsconn23: Should this return an error?

internal/system/agent/services.go
```diff
@@ -34,18 +37,110 @@ const (
...
+	proceed = false
+
+	for _, line := range allLines {
+
+		if strings.Contains(line, "success performing metrics on") {
+			proceed = true
+		}
+		if proceed {
+			all = append(all, line)
+		}
```
tsconn23: `proceed` is superfluous. It's also not reset to `false` between iterations. This check can be collapsed like so
```
if strings.Contains(line, "success performing metrics on") {
    all = append(all, line)
}
```

internal/system/agent/services.go
```diff
@@ -34,18 +37,110 @@ const (
...
+	}
+	// The following brief logic is to handle an artifact of the way the console output is retrieved
+	// after running the command "docker stats".
+	var relevant string
+	if len(all) > 0 {
+		relevant = all[1]
+	} else {
+		return "", errors.New("got an empty string from the stdout for docker stats command")
+	}
```
tsconn23:
```
if len(all) == 0 {
    return "", errors.New("got an empty string from the stdout for docker stats command")
}
return all[1], nil
```

internal/system/executor/docker.go
```diff
@@ -0,0 +1,138 @@
...
+
+func executeACommand(
+	operation string,
+	service string,
+	executor CommandExecutor,
+	operationPrefix string,
+	shouldBeRunning bool) error {
+
+	if result, err := executor(operation, service); err != nil {
+		return errors.New(executorCommandFailedMessage(operationPrefix, string(result), err.Error()))
```
tsconn23: What you really want here are custom Error types rather than custom string building functions wrapped by the standard error type. We have [examples](https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/data/errors/types.go) of this.

akramtexas:
> I'm not going to be able to deeply review this for a while, but looking at the example output from your comment, it seems that the metrics implementation here is still "hands-off", where what your metrics response looks likes in terms of responses is entirely dependent on the chosen metrics executor. Is this by design? I think it's problematic because, for example if I'm writing a client of the SMA and want to get known CPU usage I need to look at the `CpuBusyAvg` key with a `direct-service` executor, and the `cpu_perc` key for a `executor` executor. How would a client introspect from the SMA which executor is being used to generate metrics (apologies if this is clear from the code)?
> 
> I would have thought that we would collate known metrics into a common format for common metrics so that clients of the SMA don't need to know what metric backend the SMA is using. I do see how it could be useful for particular metrics implementations to provide additional information however, so perhaps what we need is a common set of metrics (again organized through interfaces such as `GetCPUMetrics` and `GetMemoryMetrics`, etc. for each understood metric) and then provide a catch-all "passthrough" metrics interface for implementations to dump other kinds of metrics not currently supported or recognized.
> 
> Thoughts?

I'm back at my desk, after being away for a day: I thank you @anonymouse64 and @tsconn23 for your thoughtful review comments. I am going through the comments and will be in touch 👍

cmd/sys-mgmt-executor/main.go
akramtexas: Good observation. So the `sys-mgmt-executor` was conceived as being an eminently lean executable, without any "frills", much more so than the other executable that reside in packages parallel to the one housing it. But given how it (i.e. the `sys-mgmt-executor`) has grown (a bit, at least) in scope, to now include the handling of `metrics` (in addition to its primary duties of starting, stopping, and restarting services), I do need to revisit the rationale to take into account this feedback 👍

michaelestrin: To keep the executor as thin as possible, it is my intention to remove the logging to stdout and replace it with log entries being returned as part of the JSON result.  Thus, instead of mixing log entries with a JSON result (as is currently the case), the executor will only return a JSON result.  If the executor's log entries are important, the caller can log them through the normal mechanism on the executor's behalf.

internal/system/executor/docker.go
michaelestrin: Acknowledged.  Will review and incorporate in my changes.  Expect to see this in the new PR I intend to open.

michaelestrin:
I'm picking up this work from @akramtexas and will be opening a separate PR that extends his work with some additional changes of my own.  Closing this PR now.

Issue #1642, opened 2019-08-09
Make Retry behavior configurable


iain-anderson:
# 🚀 Feature Request

### Relevant Package

 This feature request is for...
All supported services

### Description
 The Retry loops used when obtaining configuration at startup have hardcoded properties. Specifically they will wait 1s between attempts to contact the registry, and will wait for a total of 30000ms. This should be configurable.

### Describe the solution you'd like
 
Allow use of environment variables:
`edgex_registry_retry_count` for the number of attempts to make
`edgex_registry_retry_wait` for the time to wait in between attempts (in seconds? microseconds?)
With default values corresponding to the existing behavior.

michaelestrin:
@scsosna99 #1739 has been merged.  It contains a basic abstraction around the retry timer interval and timeout.  It also moved the configuration file read and environment variable overrides outside of the retry loop.  You might take another pass at implementing this functionality using the new common bootstrap package and its initial core-data implementation.  Reach out if you'd like assistance.

scsosna99:
starting to look at again

PR #1700, opened 2019-08-22
EF-1642 Added three environment variables which, when provided, overr…

scsosna99:
…ide the default retry timeout/wait, and refactored all existing main.go/init.go usages.

There are three retry values: timeout, wait, count.  Logically, timeout = wait * count (or should), so really might not need a separate timeout value; however, by keeping three the existing usage/paradigm is kept intact, but if preferred someone can specify the n attempts instead.  Just a matter of preference and not forcing people to do unnecessary math.

Refactored func retry() to accept a structure containining the necessary values used rather than continuing to add more individual parameters (previous 3, today 5, tomorrow?).  To support this, a new RetryType was created for holding the three retry values, and BootParam was moved from startup to config so that everything tied together cleanly (or so I believe).

Signed-off-by: Scott C Sosna <me@scottsosna.dev>

### DIFF
--- a/cmd/config-seed/main.go
+++ b/cmd/config-seed/main.go
@@ -16,6 +16,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/startup"
 	"os"
 	"sync"
 
@@ -70,7 +71,7 @@ func bootstrap(profile string) {
 	deps := make(chan error, 2)
 	wg := sync.WaitGroup{}
 	wg.Add(1)
-	var params = config2.BootParams{UseProfile: profile, Retry: config2.GetRetryInfo()}
+	var params = startup.BootParams{UseProfile: profile, Retry: config2.NewRetryInfo()}
 	go config.Retry(params, &wg, deps)
 	go func(ch chan error) {
 		for {
--- a/cmd/config-seed/main.go
+++ b/cmd/config-seed/main.go
@@ -19,7 +19,7 @@ import (
 	"os"
 	"sync"
 
-	"github.com/edgexfoundry/edgex-go/internal"
+	config2 "github.com/edgexfoundry/edgex-go/internal/pkg/config"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/usage"
 	"github.com/edgexfoundry/edgex-go/internal/seed/config"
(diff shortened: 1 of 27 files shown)


tsconn23:
@scsosna99 Please revise the commit message to have a distinct title and body. Reference the issue in the body like this

`Fix #1642`

so that the issue is closed when the PR is merged. Also, please look into why the verification checks failed.

scsosna99:
rescan

scsosna99:
Thanks for the comments, I obviously need some more background into how the configuration parameters in config.toml get injected into the running services.

What I tried to follow was how the edgex_registry URL value gets inserted as an environment variable.  Even though the boot retry timeout exists in config.toml, every retry func implementation had the hardcoded value from internal.BootTimeoutDefault, so /api/v1config/ exposes what's in the .toml, I'm not sure it matters since each usage just hard-coded the default.

My first approach was going down the path of putting this in the config settings themselves - though I still didn't understand how it all put together - but these timeouts are used when getting the configuration and how the timeouts/retries/etc work while getting the configuration.  So I thought - perhaps incorrectly - that putting them into the config itself wouldn't work because we need the values before we actually get the config.  Therefore, the values would be needed from the environment/-D values at the time we are attempting to retrieve the configuration.  Obviously correct me, but this is why the implementation does what it does.

It was a conscious decision to put BootParams and RetryInfo into config rather than startup: dependencies.  If environment.go is where gathering environment variables are centralized - which makes sense - then having BootParams and RetryInfo in startup requires a dependency on boot which didn't make logical sense.  It would be possible for environment.go to just provide the methods to get individual values for retries and have startup create the new RetryInfo, but - based on my understanding - it seemed more logical to have the types in config where all other types were rather than distribute them.

First need clarification on whether config.toml allows the retry values to be available at the time the config is being retried or if we've got the horse-before-the-cart scenario.  Once I have a better understanding of that, the rest should be simple to resolve.

Thanks again.

scsosna99:
Totally fine, I was uncomfortable with the duplicate code and was trying to see a way to abstract out into something common but there was just enough differences to make it out of my experience on the project, so I worked within what there was.  Putting it on hold makes sense.

michaelestrin:
https://github.com/edgexfoundry/edgex-go/issues/1739

tsconn23:
Closing due to age and @michaelestrin comments above RE: #1739

PR #1701, opened 2019-08-22
Unit test -- Core-Metadata Get Device Service By Name/ Addressable Name

### LINKED ISSUE #1696: Unit test -- Core-Metadata Get Device Service By Name/ Addressable Name
https://github.com/edgexfoundry/edgex-go/projects/11#card-23264658

https://github.com/edgexfoundry/edgex-go/projects/11#card-23264689

brandonforster:
Fix #1696

While I was working, I also refactored Addressable's delete code to take advantage of the pattern I used in this PR with the GetByAddressable Executor. Instead of having one factory method that takes both an ID and a name, I have two factory methods to reduce the chance that a consumer does the wrong thing. Technically out of scope, but I wanted to catch it while it was on my mind.

### DIFF
--- a/internal/core/metadata/operators/device_service/get.go
+++ b/internal/core/metadata/operators/device_service/get.go
@@ -53,29 +53,52 @@ func (op deviceServiceLoadAll) Execute() (services []contract.DeviceService, err
 	return
 }
 
-type deviceServiceLoadByAddressableId struct {
-	id string
-	db DeviceServiceLoader
+type deviceServiceLoadByAddressable struct {
+	id   string
+	name string
+	db   DeviceServiceLoader
 }
 
-// NewDeviceServiceLoadByAddressableId creates a new Executor that retrieves all DeviceService associated with a given Addressable ID.
-func NewDeviceServiceLoadByAddressableId(id string, db DeviceServiceLoader) DeviceServiceGetAllExecutor {
-	return deviceServiceLoadByAddressableId{id: id, db: db}
+// NewDeviceServiceLoadByAddressableByName creates a new Executor that retrieves all DeviceService associated with a given Addressable name.
+func NewDeviceServiceLoadByAddressableByName(name string, db DeviceServiceLoader) DeviceServiceGetAllExecutor {
+	return deviceServiceLoadByAddressable{name: name, db: db}
 }
 
-// Execute performs an operation that retrieves all DeviceService associated with a given Addressable ID.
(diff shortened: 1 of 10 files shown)


internal/core/metadata/operators/device_service/get.go
```diff
@@ -53,29 +53,52 @@ func (op deviceServiceLoadAll) Execute() (services []contract.DeviceService, err
...
+// NewDeviceServiceLoadByAddressableName creates a new Executor that retrieves all DeviceService associated with a given Addressable name.
+func NewDeviceServiceLoadByAddressableName(name string, db DeviceServiceLoader) DeviceServiceGetAllExecutor {
+	return deviceServiceLoadByAddressable{name: name, db: db}
 }
 
-// NewDeviceServiceLoadByAddressableId creates a new Executor that retrieves all DeviceService associated with a given Addressable ID.
-func NewDeviceServiceLoadByAddressableId(id string, db DeviceServiceLoader) DeviceServiceGetAllExecutor {
-	return deviceServiceLoadByAddressableId{id: id, db: db}
+// NewDeviceServiceLoadByAddressableID creates a new Executor that retrieves all DeviceService associated with a given Addressable ID.
+func NewDeviceServiceLoadByAddressableID(id string, db DeviceServiceLoader) DeviceServiceGetAllExecutor {
```
tsconn23: If the `DeviceServiceGetAllExecutor` interface is going to be multi-purposed due to its conformant signature, would recommend renaming it since we aren't "getting all" here. Perhaps something like `DeviceServiceListExecutor`.

internal/core/metadata/operators/device_service/get.go
```diff
@@ -53,29 +53,52 @@ func (op deviceServiceLoadAll) Execute() (services []contract.DeviceService, err
...
-	id string
-	db DeviceServiceLoader
+type deviceServiceLoadByAddressable struct {
+	id   string
+	name string
+	db   DeviceServiceLoader
+}
+
+// NewDeviceServiceLoadByAddressableName creates a new Executor that retrieves all DeviceService associated with a given Addressable name.
+func NewDeviceServiceLoadByAddressableName(name string, db DeviceServiceLoader) DeviceServiceGetAllExecutor {
```
tsconn23: If the `DeviceServiceGetAllExecutor` interface is going to be multi-purposed due to its conformant signature, would recommend renaming it since we aren't "getting all" here. Perhaps something like `DeviceServiceListExecutor`.

PR #1707, opened 2019-08-27
Refactoring of the existing security-secrets-setup code

jim-wang-intel:
Refactor the existing code to accept new changes for pki-init integration and better testable code

  - Abstract os.Exit() call so it can be tested without losing code coverage
  - Abstract pki-init option so that it can accept the pki-init options later
  - Change the local logging client to use global one so that it is more consistent with security-proxy and the logger can be used through out all options in this service
  - Refactor the execution of -configFile into function so that it can be integrated later with better generation option
  - Add unit test for main.go for increasing test coverage from the origin
  - Add unit test for x509.go for increasing test coverage from the origin

Related to: Issue #1635 


cmd/security-secrets-setup/main.go
```diff
@@ -16,80 +17,98 @@ package main
...
+	opts := option.PkiInitOption{}
+
+	return option.NewPkiInitOption(opts)
+}
+
+func (dispatcher *pkiInitOptionDispatcher) run() (statusCode int, err error) {
+	optsExecutor, statusCode, err := setupPkiInitOption()
 	if err != nil {
-		log.Fatalf("ERROR: %s: %s", msg, err) // fatalf() =  Prinf() followed by a call to os.Exit(1)
+		return
```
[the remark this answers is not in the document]
jim-wang-intel: @mkbhanda - that is a good catch, should return error with status code

.gitignore
```diff
@@ -26,6 +26,7 @@ cmd/sys-mgmt-agent/sys-mgmt-agent
 cmd/support-scheduler/support-scheduler
 cmd/sys-mgmt-executor/sys-mgmt-executor
 docs/_build/
+cmd/security-secrets-setup/config/*
```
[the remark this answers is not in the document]
jim-wang-intel: ok

jim-wang-intel:
> 
> 
> pkiOption_test.go file may need "gofmt -w -l -s" . Other than that it looks good to me.

@tingyuz Tingyu, I ran that, and gofmt tool reports clean to me on that. so we are fine.

jim-wang-intel:
@anonymouse64 Ian, any feedback? I am expecting to merge this onto master pretty soon.

cmd/security-secrets-setup/main.go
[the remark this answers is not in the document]
jim-wang-intel: yea, this `run()` returns error and error is logged on line 74 above

anonymouse64:
>@anonymouse64 Ian, any feedback? I am expecting to merge this onto master pretty soon.

I will try to review this today

cmd/security-secrets-setup/main_test.go
```diff
@@ -0,0 +1,138 @@
...
+	origArgs := os.Args
+	defer tearDown(t, origArgs)
+	assert := assert.New(t)
+
+	runWithNoOption()
+	assert.Equal(1, (exitInstance.(*testExitCode)).getStatusCode())
+}
+
+func TestMainWithConfigFileOption(t *testing.T) {
+	tearDown := setupTest(t)
```
anonymouse64: this kind of test usually reads much cleaner as:

```go
defer mockExitInstance(newArgs(), newTestExit(), newTestDispatcher())
```

where `mockExitInstance` looks like this:

```go
func mockExitInstance(mockedExiter exit) func() {
    // save for restoring
    prevExitInstance = exitInstance

    // mock the new one
    exitInstance = mockedExiter

    return func() {
        exitInstance = prevExitInstance

        // you can also do cleanup here too
        os.RemoveAll("./config")
    }
}
```

repeating this pattern for all the things you need to mock

cmd/security-secrets-setup/main.go
```diff
@@ -16,80 +17,98 @@ package main
...
+	os.Exit(statusCode)
+}
+
+func newOptionDispatcher() optionDispatcher {
+	return &pkiInitOptionDispatcher{}
 }
 
-// TODO: ELIMINATE THIS ----------------------------------------------------------
-func fatalIfErr(err error, msg string) {
+func setupPkiInitOption() (executor option.OptionsExecutor, status int, err error) {
```
anonymouse64: this function doesn't do much and is only used inside the `run()` method so I think you should just roll it into `run()`. Where you use it in the tests can be similarly rolled inside, running a single function with a default value struct doesn't really warrant sharing code like this

cmd/security-secrets-setup/main_test.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func runWithNoOption() {
+	// case 1: no option given
+	os.Args = []string{"cmd"}
+	printCommandLineStrings(os.Args)
+	main()
+}
+
+func runWithConfigFileOption() {
```
anonymouse64: why not just roll this into the `TestMainWithConfigFileOption` directly? it's not used anywhere else and would make the tests you write easier to follow without having to jump around to all the helper functions

cmd/security-secrets-setup/main_test.go
```diff
@@ -0,0 +1,138 @@
...
+	os.Args = []string{"cmd", "-config", "./res/pkisetup-vault.json"}
+	printCommandLineStrings(os.Args)
+	main()
+}
+
+func printCommandLineStrings(strs []string) {
```
anonymouse64: it seems like this entire function could just be `fmt.Println("command line strings:",strings.Join(strs," "))`

cmd/security-secrets-setup/main.go
```diff
@@ -16,80 +17,98 @@ package main
...
 	"github.com/edgexfoundry/edgex-go/internal/security/setup"
-	"github.com/edgexfoundry/edgex-go/internal/security/setup/certificates"
-	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
+	"github.com/edgexfoundry/edgex-go/internal/security/setup/option"
 )
 
-func main() {
-	start := time.Now()
-	var configFile string
+type exit interface {
```
anonymouse64: I dislike the name of this interface, it's not clear what it does and isn't idiomatic go naming, I would like this better:

```go
type exiter interface {
    Exit()
}
```

cmd/security-secrets-setup/main.go
```diff
@@ -16,80 +17,98 @@ package main
...
+	callExit(int)
+}
+
+type exitCode struct{}
```
anonymouse64: instead of defining a whole new type and interface for testing this kind of thing, what I've seen done in snapd is:

```go
var (
    ExitImpl = os.Exit
)

func main() {
    exitImpl(0)
}
```

then in tests you do this:

```go
func makeTestExit(exitedStatus *int) func(status int) {
    return func(status int){ *exitedStatus = status }
}

func TestTheThings() {
    exitCode := 0
    defer mockExit(makeTestExit(&exitCode))
    runMainWithErrorThings()
    if exitCode == 1 {
        // test pass
    } else {
        // test fail
    }
}

func mockExit(mockedExit func(status int)) func() {
    oldExit := pkg.ExitImpl
    pkg.ExitImpl = mockedExit
    return func() {
        pkg.ExitImpl = oldExit
    }
}
```

cmd/security-secrets-setup/main_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+var hasDispatchError bool
+
+func TestMainWithNoOption(t *testing.T) {
+	tearDown := setupTest(t)
+	origArgs := os.Args
+	defer tearDown(t, origArgs)
+	assert := assert.New(t)
+
+	runWithNoOption()
```
anonymouse64: ditto, I think you should just roll this function into here

internal/pkg/config/x509_test.go
```diff
@@ -15,30 +16,52 @@
...
+			if tt.expectError {
+				assert.NotNil(err, "expected error, none returned")
+				assert.Empty(x509Config)
 			}
 		})
 	}
 }
+
+func TestPkiCAOutDir(t *testing.T) {
+	x509Config, readErr := NewX509Config(testJSONFileName)
```
anonymouse64: I prefer to always just have `err` as the error variable unless you have a specific need to juggle around multiple errors simultaneously and independently

```suggestion
	x509Config, err := NewX509Config(testJSONFileName)
```

internal/security/setup/option/exitCode.go
```diff
@@ -0,0 +1,33 @@
...
+// or implied. See the License for the specific language governing permissions and limitations under
+// the License.
+//
+// SPDX-License-Identifier: Apache-2.0'
+//
+
+package option
+
+// ExitCode is the code used for exit status
+type exitCode int
```
anonymouse64: why have a new type for the exit code and not just define constants for the various exit codes we use for specific situations? seems like an unnecessarily abstraction

internal/pkg/config/x509_test.go
```diff
@@ -15,30 +16,52 @@
...
+	assert := assert.New(t)
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			_, err := NewX509Config(tt.file)
-			if err != nil && !tt.expectError {
-				t.Errorf("unexpected error %v", err)
-				return
+			x509Config, err := NewX509Config(tt.file)
+
+			if !tt.expectError {
```
anonymouse64: this would read better as 

```go
			if tt.expectError {
				assert.NotNil(err, "expected error, none returned")
				assert.Empty(x509Config)
                        } else {
				assert.Nil(err, "unexpected error %v", err)
				assert.NotEmpty(x509Config)
                        }
```

internal/pkg/config/x509_test.go
```diff
@@ -15,30 +16,52 @@
...
 
 import (
+	"path/filepath"
+	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
 
+var testJSONFileName = "./res/pkisetup-test.json"
```
anonymouse64: can we move this to testdata instead of res?

internal/security/setup/option/pkiOption.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// ProcessOptions processes all the input options from the caller
+func (pkiInitOpt *PkiInitOption) ProcessOptions() (int, error) {
+	statusCode, err := pkiInitOpt.executor.executeOptions()
+
+	return statusCode.intValue(), err
+}
+
+func (pkiInitOpt *PkiInitOption) executeOptions(executors ...func(*PkiInitOption) (exitCode, error)) (exitCode, error) {
```
anonymouse64: what's the use case for having multiple executors?

cmd/security-secrets-setup/main.go
```diff
@@ -16,80 +17,98 @@ package main
...
+	flag.Parse()
 
-	// Read the Json config file and unmarshall content into struct type X509Config
-	x509config, err := config.NewX509Config(configFile)
-	if err != nil {
-		lc.Error(err.Error())
+	if len(os.Args) < 2 {
+		fmt.Println("Please specify option for " + option.SecuritySecretsSetup)
+		flag.PrintDefaults()
+		exitInstance.callExit(1)
```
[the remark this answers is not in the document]
jim-wang-intel: @anonymouse64 Ian, I am not against the exit code 0 for --help or -h option.  So it is a matter how we treat no option is provided at all. Currently we treat no option provided as exit code 1 instead of 0 but I am open to that.  @bnevis-i  what do you think?

internal/security/setup/option/constant.go
```diff
@@ -0,0 +1,22 @@
...
+// the License.
+//
+// SPDX-License-Identifier: Apache-2.0'
+//
+
+package option
+
+const (
+	// SecuritySecretsSetup is the name of this module
+	SecuritySecretsSetup = "security-secrets-setup"
```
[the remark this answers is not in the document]
jim-wang-intel: just to reduce the hard-coded same string in multiple places even only in main.

cmd/security-secrets-setup/main.go
```diff
@@ -16,80 +17,98 @@ package main
...
 		return
 	}
 
-	seed, err := setup.NewCertificateSeed(x509config, setup.NewDirectoryHandler(lc))
-	if err != nil {
-		lc.Error(err.Error())
+	if err := setup.Init(); err != nil {
+		exitInstance.callExit(1)
```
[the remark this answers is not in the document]
jim-wang-intel: Sounds ok for me to put some comments in code for that- since the original code already does that logging inside the call, so I'd rather leave it there for less changes.

cmd/security-secrets-setup/main_test.go
jim-wang-intel: Well, my original real purpose to do this is just for prepare Initialization/reset and tearDown for unit tests and not really for mocking purpose per se. But your idea to do this way is interesting for mocking things for other testing purpose as well.  The real mocking is actually achieved via combination of interface and testing struct like I do here.

cmd/security-secrets-setup/main.go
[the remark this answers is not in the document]
bnevis-i: I have seen Unix utilities that do it both ways.  Personally I think if the command didn't do its intended job it should return a non-zero exit status, so that thinks like exec.Command will flag the error.  But since there is no generally accepted standard, if Ian says zero status when displaying usage is the EdgeX convention, I can go along with that.

cmd/security-secrets-setup/main.go
jim-wang-intel: Well, this way is more unit-testable. So someone can just unit-test this setup to make usre it does what it needs to do and not worry about the driver part (which is `run()`).  As a matter of fact, in the unit tests, `run()` is overridden and used for test cases.

cmd/security-secrets-setup/main.go
bnevis-i: First class functions!

cmd/security-secrets-setup/main_test.go
jim-wang-intel: I am fine with that, but it is just code readability concern: with this func name it seems to read better than all codes inside this.

cmd/security-secrets-setup/main.go
jim-wang-intel: It does sound better to me, will change it.

internal/pkg/config/x509_test.go
jim-wang-intel: fine with me

internal/pkg/config/x509_test.go
jim-wang-intel: good point

cmd/security-secrets-setup/main.go
jim-wang-intel: @bnevis-i  can you elaborate it more? What do you mean by "first class" functions?

cmd/security-secrets-setup/main_test.go
jim-wang-intel: same concern- code readability

internal/security/setup/option/exitCode.go
jim-wang-intel: It is less verbose in the constant name without losing the context- if only defined in constant, then one would end up with 
```
const(
   normalExitCode = 0
   noOptionSelectedExitCode = 1
   exitWithErrorExitCode = 2
)
```

internal/pkg/config/x509_test.go
jim-wang-intel: ok

internal/security/setup/option/pkiOption.go
jim-wang-intel: for testing purpose

cmd/security-secrets-setup/main_test.go
jim-wang-intel: ok

cmd/security-secrets-setup/main.go
jim-wang-intel: After some thinking, IMHO, I like interface approach better: the reasons are
1. Interface is clear to convey the idea what type of actions you are trying to do
2. Interface is also easier to put all collections of actions to do in one contract for the caller; though in this case we only have one but with that, it is easier to override the behavior in test and in other places if you find to do so.
3. IMHO, interface is more object-oriented than just using the variables

Moreover, like I said previously, the main purpose of setup and teardown is not to do the mocking but rather to initialize and clean up the things tests needed.

cmd/security-secrets-setup/main.go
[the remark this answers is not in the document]
jim-wang-intel: ok, i am ok to go with 0 status code then

jim-wang-intel:
Thanks for your valuable comments and feedback, @anonymouse64 Ian, I put some replies to your comments and have addressed your PR feedback.  Please check it. Thanks! -Jim

cmd/security-secrets-setup/main.go
anonymouse64: this specific function is not unit tested though, `run()` is what you are unit testing, this function doesn't do anything except call one other function

internal/security/setup/init.go
```diff
@@ -15,27 +15,44 @@
...
 	"github.com/edgexfoundry/go-mod-core-contracts/models"
 )
 
 // Global variables
 var Configuration *ConfigurationStruct
+var LoggingClient logger.LoggingClient
 
-func Init() {
+func Init() error {
 	// Unfortunately I have to do this because of utilization of the LoggingClient in config.LoadFromFile below.
```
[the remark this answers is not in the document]
jim-wang-intel: That is original old comment there.. I agree with you that Trevor should do an issue for that.

cmd/security-secrets-setup/main.go
jim-wang-intel: Adding unit test to it

anonymouse64: sorry the point I was making is that you don't need to unit test this function because it isn't doing much of anything, it's like defining this function:

```go
func doAThing() {
    return doAnotherthing(struct{})
}
```

there's nothing gained by this function and so it doesn't need a test and the function doesn't need to exist is all I'm saying, apologies if that was unclear

jim-wang-intel: Ian, there is something being doing in line 107 `option.NewPkiInitOption(opts)`, this function is doing something.  As a matter of fact, this one serves as the purpose of regression-testing later on when `-generate`, `-cache`, and `-import` option are going to be added.  With this as separate function, then it can be regression tested by the unit test to see if anything break when options are added later.

internal/security/setup/option/optionHelpers.go
```diff
@@ -0,0 +1,65 @@
...
+
+	config "github.com/edgexfoundry/edgex-go/internal/pkg/config"
+	"github.com/edgexfoundry/edgex-go/internal/security/setup"
+	"github.com/edgexfoundry/edgex-go/internal/security/setup/certificates"
+)
+
+// GenTLSAssets generates the TLS assets based on the JSON configuration file
+func GenTLSAssets(jsonConfig string) error {
+	// Read the Json config file and unmarshall content into struct type X509Config
+	x509Config, readErr := config.NewX509Config(jsonConfig)
```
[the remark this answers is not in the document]
jim-wang-intel: good catch i'll change it

jim-wang-intel:
@anonymouse64  Ian, I have rebased and squashed to the single commit. Thanks for great feedbacks. This one commit should be ready to merge onto master.

jim-wang-intel:
Hi @mkbhanda Malini, if changes are good for you, please approve it. Thanks and have a nice weekend! 
-Jim W.

.gitignore
```diff
@@ -19,11 +19,12 @@ cmd/core-metadata/core-metadata
 cmd/export-client/export-client
 cmd/export-distro/export-distro
 cmd/security-secrets-setup/security-secrets-setup
+cmd/security-secrets-setup/config/*
```
mkbhanda: nit: a little out of order

PR #1715, opened 2019-08-28
Fixed timeout not respected bug in all services; commonized server logic

### LINKED ISSUE #1702: Timeout isn't respected in any EdgeX Service
# 🐞 Bug Report

### Affected Services

 The issue is located in: All Services 

### Is this a regression?

 Not that i'm aware of.

Using core-data as an example: 
1) Set 
```toml
[Service]
Timeout = 1
``` 
in configuration.toml
2) Set a sleep (larger than that timeout set in your configuration) in any http handler (i.e. pingHandler) for core-data: 
`time.Sleep(time.Millisecond * time.Duration(5000))`
3) Make the rest request. (i.e. /api/v1/ping) should give you a "Response Timed out" response, but instead you will get pong after however long you put a sleep in the handler for. 

The timeout is ignored 
```
	http.TimeoutHandler(nil, time.Millisecond*time.Duration(data.Configuration.Service.Timeout), "Request timed out")
``` 
in main.go since the handler returned from the timeoutHandler is simply not used, and nil is passed to the TimeoutHandler so it doesn't have any routes to build on. It's basically a no-op. 

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍  Your Environment
**Deployment Environment:**
Native, Windows 10
**EdgeX Version:**
Master Branch

**Anything else relevant?**
the timeoutHandler should be moved into `startHttpServer()`. Something like:
```golang
r := data.LoadRestRoutes()
tr := http.TimeoutHandler(r, time.Millisecond*time.Duration(data.Configuration.Service.Timeout), "Request timed out")
errChan <- http.ListenAndServe(":"+strconv.Itoa(port), context.ClearHandler(tr))
```

brandonforster:
Fix #1702

This PR addresses the bug, and at the same time attempts to refactor the code to make it testable and not repeated. It has been an ongoing pain point that the `main` functions are largely identical, this PR tries to help address that by centralizing the HTTP server logic.

This PR also renames`support-logging`'s router to put it in line with the rest of the services.


anonymouse64:
(I'm not going to have time to do a full review, just happened to see the missing doc-comment and did a drive-by comment, so don't wait for me on this)

cmd/core-data/main.go
```diff
@@ -17,22 +17,20 @@ package main
...
 	"time"
 
-	"github.com/edgexfoundry/edgex-go"
 	"github.com/edgexfoundry/edgex-go/internal"
 	"github.com/edgexfoundry/edgex-go/internal/core/data"
-	"github.com/edgexfoundry/edgex-go/internal/pkg/correlation"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/startup"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/usage"
+
+	"github.com/edgexfoundry/edgex-go"
```
tsconn23: Should come before `github.com/edgexfoundry/edgex-go/internal`. Please review all as it looks like this is all of the main.go files

brandonforster: Is this considered being part of the internal package? I'll move it, I just didn't know.

Issue #1706, opened 2019-08-27
Sending api /api/v1/device/name/{devicename}/command/{commandname}  when the device admin status is locked, got return code 500


cherrycl:
# 🐞 Bug Report

### Affected Services

 The issue is located in: Core-Command

### Is this a regression?

 Yes, the previous version in which this bug was not present was: ....

### Description and Minimal Reproduction
1. Import a device and set the admin state to "Locked".
2. Sending API /api/v1/device/name/{devicename}/command/{commandname} with GET method.

According to the [api doc](https://github.com/edgexfoundry/edgex-go/blob/a224da4f06dcbe362e752b2917792d43d17afe6e/api/raml/core-command.raml#L170), the return code should be 423.
## 🔥 Exception or Error
<pre><code>
Return Code : 500Internal Server Error
Response Body : The device is in admin locked state

</code></pre>

## 🌍  Your Environment
**Deployment Environment:** Ubuntu 18.04

**EdgeX Version:** Master

**Anything else relevant?**

PR #1720, opened 2019-08-30
Check device status, immediately after the device is read.

difince:
Fix https://github.com/edgexfoundry/edgex-go/issues/1706

When user want to execute a command GET/PUT, the device is loaded first. Validate the device.AdminStatus != Locked state in order to proceed with the logic.


brandonforster:
Please address the test failure.

12:39:11 --- FAIL: TestExecuteWithLockedDevice (0.03s)
12:39:11     types_test.go:141: No error should be present for happy path

internal/core/command/types.go
```diff
@@ -15,6 +15,7 @@ package command
 
 import (
 	"bytes"
+	"fmt"
```
brandonforster: Please sort imports

difince: Imports sorted, the test TestExecuteWithLockedDevice - removed.
I have removed this test,  because device.AdminStatus validation is done earlier, immediately after the device is Loaded. So, the test does not return anymore Error because of the device state.

internal/core/command/types_test.go
```diff
@@ -124,20 +124,3 @@ func TestExecuteHttpReadResponseError(t *testing.T) {
 		t.Errorf("The response code should be the default value for failing requests")
 	}
 }
-
-func TestExecuteWithLockedDevice(t *testing.T) {
```
brandonforster: Please don't remove the test. Change the test so that it now reflects current behavior.

We certainly don't want to be dropping test coverage %.

internal/core/command/types.go
```diff
@@ -39,12 +49,6 @@ type serviceCommand struct {
 
 // Execute sends the command to the command service
 func (sc serviceCommand) Execute() (string, int, error) {
-	if sc.AdminState == contract.Locked {
-		LoggingClient.Error(sc.Name + " is in admin locked state")
-
-		return "", DefaultErrorCode, ErrDeviceLocked{}
-	}
-
 	LoggingClient.Info("Issuing" + sc.Request.Method + " command to: " + sc.Request.URL.String())
```
tsconn23: Minor thing -- this should probably be `Debug` instead of `Info`. Otherwise the log is going to fill up.

internal/core/command/types_test.go
tsconn23: A couple things here

1.) Using `go tool cover` comparing coverage on types.go before and after this PR, the coverage percentage is the same %66.7. So removing this test doesn't affect coverage.
2.) @difince is referring to the removal of this test when she says
>I have removed this test, because device.AdminStatus validation is done earlier, immediately after the device is Loaded. So, the test does not return anymore Error because of the device state.
Which has been added in `command/device.go` as part of this PR. We do not currently have a unit test file for this. If Diana would like to add one, then bonus points. But we will need to come back through here refactoring toward the operator pattern anyway.

tsconn23:
@cherrycl Can you please verify?

internal/core/command/types.go
```diff
@@ -15,19 +15,29 @@ package command
...
 // ErrDeviceLocked error indicating the locked state of a device
-type ErrDeviceLocked struct{}
+type ErrDeviceLocked struct {
+	id   string
+	name string
+}
 
-func (ErrDeviceLocked) Error() string {
-	return "The device is in admin locked state"
+func (e ErrDeviceLocked) Error() string {
```
mkbhanda: If at any time only id or name is provided, we could just have Error() defined as 
func (e ErrDeviceLocked) Error() string {
         return fmt.Sprintf("device '%s%s' is in admin locked state", e.id, e.name)
}

cherrycl:
> @cherrycl Can you please verify?

@tsconn23  Felix helped to verify this PR and it works fine.

internal/core/command/types.go
tsconn23: @difince Need to define errors consistently. See [here](https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/metadata/errors/types.go). Specifically, we `New()` them and then `Error()` returns the error message. In this case, the `Error()` implementation can be streamlined as @mkbhanda says.
```
type ErrDeviceLocked struct {
    device string
}

func NewErrDeviceLocked(device string) error {
    return ErrDeviceLocked{device: device}
}

func (e ErrDeviceLocked) Error() string {
    return fmt.Sprintf("device '%s' is in admin locked state", e.id)
}
```
If you were to declare this in an `errors` package similar to how it's done in core-metadata, then the `device` property on the type would not be externally accessible.

internal/core/command/types_test.go
difince: I have added tests for commandByDeviceID, but still no operator pattern. (even though only negative)
We will need to come back for the operator pattern and I would suggest renaming the functions.They are confusing at least for me. I hardly got if a function will get the commands from the database, or the function will execute get command towards real device. Even the function - 'commandByDeviceID' - actually tries to execute a command GET/PUT by both deviceId and commandId. Whenever we could make changes in the APIs, we could stay with just the commandId - no need deviceId to be provided.

internal/core/command/types.go
difince: I have added such struct initiated by New() and now we are able to send in the error msg whatever we want from type string to identify the device - id,  name and etc

Issue #1721, opened 2019-08-30
Apply Operator Pattern - Notification Add


nkstanchev:
This issue addresses the transition of the Notification POST endpoint to the operator pattern and the addition of the accompanying unit tests.

https://github.com/edgexfoundry/edgex-go/projects/11#card-24425059

PR #1722, opened 2019-08-30
Apply Operator Pattern - Notification Add

nkstanchev:
Addresses #1721.


internal/support/notifications/operators/notification/add_test.go
```diff
@@ -0,0 +1,110 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package notification
+
+import (
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
```
tsconn23: Organize imports

internal/support/notifications/rest_notification.go
```diff
@@ -45,24 +45,38 @@ func notificationHandler(w http.ResponseWriter, r *http.Request) {
...
+		case errors.ErrNotificationInUse:
+			http.Error(w, err.Error(), http.StatusConflict)
+		default:
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+		}
 		LoggingClient.Error(err.Error())
 		return
 	}
 
 	if n.Severity == models.NotificationsSeverity(models.Critical) {
```
tsconn23: I think I would advocate including this logic in the AddExecutor. The executor should encapsulate all of the relevant logic for the handler, and this is the only place where we check the severity and send right away. Also note that this is the only place the Distributor is used, so you may be able to just make that functionality a private function on the `notificationAdd` type. This would eliminate the newly added global var, interface and `New` constructor.

nkstanchev: Agree, much more elegant solution. I will edit this commit

internal/support/notifications/normal_distribution.go
```diff
@@ -16,10 +16,13 @@
 package notifications
 
 import (
+	"github.com/edgexfoundry/edgex-go/internal/support/notifications/interfaces"
 	"github.com/edgexfoundry/go-mod-core-contracts/models"
 )
 
-func distributeAndMark(n models.Notification) error {
+type NormalDistributor struct{}
```
AnthonyMBonafide: Could we possibly remove this type, since it does not encapsulate any data and it's only method can be expressed as an exported helper function? I am thinking that we could make this `NormalDistributor` type an executor by doing the following:

1. Move this to the `notification/operator` directory
1. Update the name to match the other executors, maybe something like `DistributionExecutor`
1. Update the name of the `distributeAndMark` method to `Execute` the signature looks good as is.
1. Update the executor to use a "loader" or "updater" type rather than the `dbClient` directly.
1. Create an exported constructor

All the above steps are so that our operators are consistent.

I think this approach also will allow us to eliminate the need for the `distributor` global variable in the init.go file : 
```
var distributor interfaces.Distributor
```

and instead we can just create the executor when we need it and call the `Execute` method like we do for other operators in our REST handler functions. Overall, I do not think we need a global variable here.

internal/support/notifications/operators/notification/add.go
```diff
@@ -0,0 +1,51 @@
...
+
+package notification
+
+import (
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
+	"github.com/edgexfoundry/edgex-go/internal/support/notifications/errors"
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+)
+
+type AddExecutor interface {
```
AnthonyMBonafide: Can we add Go doc comment here?

nkstanchev: Sure

internal/support/notifications/normal_distribution.go
nkstanchev: Agree. I created the global var out of need in order to mock and test notification add executor. 

Moving this to its own executor sounds good. However, when I started implementing this I realised that code in normal_distribution ultimately has calls into sending_service.go and escalation_service.go. I could possibly put all of those functions in one operator, however, it seems to me like they should be in several and, of course, ease of testing needs to be considered also. I'd like to hear your opinion on this

tsconn23:
@nkstanchev Please resolve conflicts and any outstanding comments so we can merge this.

Issue #1725, opened 2019-09-02
Apply Operator Pattern - IntervalAction GetAll


nkstanchev:
This issue addresses the transition of the Scheduler Intervalaction endpoint for GET to the operator pattern and the addition of the accompanying unit tests.

https://github.com/edgexfoundry/edgex-go/projects/11#card-24596170

PR #1728, opened 2019-09-02
Apply Operator Pattern - IntervalAction GetAll

nkstanchev:
Addresses #1725.


internal/support/scheduler/operators/intervalaction/db.go
```diff
@@ -0,0 +1,67 @@
...
+type IntervalActionLoader interface {
+	IntervalActions() ([]contract.IntervalAction, error)
+	IntervalActionsWithLimit(limit int) ([]contract.IntervalAction, error)
+	IntervalActionByName(name string) (contract.IntervalAction, error)
+	IntervalActionById(id string) (contract.IntervalAction, error)
+	interval.IntervalLoader
+}
+
+// IntervalDeleter deletes interval.
+type IntervalActionDeleter interface {
```
tsconn23: It looks like there are a lot of interface declarations in here that are unused, not related to this PR. Please confirm. If so, please limit changes to just the targeted use case.

internal/support/scheduler/rest_intervalaction_test.go
```diff
@@ -0,0 +1,85 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package scheduler
+
+import (
+	"errors"
+	"github.com/stretchr/testify/mock"
```
tsconn23: Organize imports

internal/support/scheduler/operators/intervalaction/get.go
```diff
@@ -0,0 +1,60 @@
...
+ * the License.
+ *******************************************************************************/
+
+package intervalaction
+
+import (
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+)
+
+type CollectionExecutor interface {
```
AnthonyMBonafide: Can we rename this to `IntervalActionsExecutor` to be consistent with the other operators?

AnthonyMBonafide: Can we add Go doc comments to the exported functions, methods and types?

internal/support/scheduler/operators/intervalaction/get.go
```diff
@@ -0,0 +1,60 @@
...
+
+func NewAllExecutor(db IntervalActionLoader, limit int) CollectionExecutor {
+	return intervalActionLoadAll{
+		database: db,
+		limit:    limit,
+	}
+}
+
+func getIntervalActions(db IntervalActionLoader, limit int) ([]contract.IntervalAction, error) {
+	var err error
```
AnthonyMBonafide: Can we simplify this function by removing these variables and simply returning the results of either `db.IntervalActions()` or `db.IntervalActionsWithLimit(limit)`? The surrounding code does not seem to be adding any value. If this is the case and we can reduce this function, my next question would be: does this logic need to exist in a function as it would only be a simple if statement?

internal/support/scheduler/operators/intervalaction/get.go
```diff
@@ -0,0 +1,60 @@
...
+	var intervalActions []contract.IntervalAction
+
+	if limit <= 0 {
+		intervalActions, err = db.IntervalActions()
```
AnthonyMBonafide: I think we are missing some additional logic here. We need to also ensure the number of results do not exceed the `Configuration.Service.MaxResultCount` value. With that being said I think we should always pass a limit to the `db.IntervalActionsWithLimit(limit)`. If my assumption is correct then to add to my previous comment regarding the simplification and removal of the function is further supported as this function would only call `db.IntervalActionsWithLimit(limit)` with either the specified `limit` or `Configuration.Service.MaxResultCount`.

internal/support/scheduler/operators/intervalaction/get.go
```diff
@@ -0,0 +1,60 @@
...
+type CollectionExecutor interface {
+	Execute() ([]contract.IntervalAction, error)
+}
+
+type intervalActionLoadAll struct {
+	database IntervalActionLoader
+	limit    int
+}
+
+func (op intervalActionLoadAll) Execute() ([]contract.IntervalAction, error) {
```
AnthonyMBonafide: Adding to my other comment regarding honoring the `Configuration.Service.MaxResultCount` we should first check that the specified `limit` does not exceed the `Configuration.Service.MaxResultCount` before executing any database operations and simply return an `ErrLimitExceeded` error.

internal/support/scheduler/rest_intervalaction.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/edgexfoundry/edgex-go/internal/support/scheduler/operators/intervalaction"
+)
+
+func restGetIntervalAction(w http.ResponseWriter, r *http.Request) {
+	if r.Body != nil {
+		defer r.Body.Close()
+	}
+	op := intervalaction.NewAllExecutor(dbClient, Configuration.Service.MaxResultCount)
+	intervalActions, err := op.Execute()
+	if err != nil {
```
AnthonyMBonafide: I think we need to add some error type checks so that we are returning the correct responses. The RAML looks like we need to account for `413` which is the `ErrLimitExceeded` error.

https://github.com/edgexfoundry/edgex-go/blob/ba8d5362ba3742b919d0ac002fe8d7221c5310a2/api/raml/support-scheduler.raml#L192-L204

internal/support/scheduler/operators/intervalaction/get.go
nkstanchev: Sure

nkstanchev: Sure

internal/support/scheduler/operators/intervalaction/get.go
nkstanchev: Agree

internal/support/scheduler/operators/intervalaction/get.go
nkstanchev: That's true. However, I think this should be done in the http handler as doing the check directly here will introduce a cyclic dependency. It is possible to pass it as part of the struct and check inside. Up to you for final implementation.

internal/support/scheduler/operators/intervalaction/get.go
nkstanchev: That's true. However, I think this should be done in the http handler as doing the check directly here will introduce a cyclic dependency. It is possible to pass it as part of the struct and check inside. Up to you for final implementation.

internal/support/scheduler/rest_intervalaction.go
nkstanchev: Will depend of location of config max result check

internal/support/scheduler/operators/intervalaction/get.go
AnthonyMBonafide: I think passing in a copy of the configuration to the operator will not create a cycle in dependencies similar to what we are doing here:

https://github.com/edgexfoundry/edgex-go/blob/57817b912cbc23d07f161b86d5b5397305a543ba/internal/core/data/operators/value_descriptor/get.go#L33-L38 

and using that to check the `MaxResultCount` like:

https://github.com/edgexfoundry/edgex-go/blob/57817b912cbc23d07f161b86d5b5397305a543ba/internal/core/data/operators/value_descriptor/get.go#L47-L49

rather than referencing the global configuration variable.

nkstanchev: This endpoint doesn't have a limit GET parameter so by using db.IntervalActionsWithLimit(limit), limit would actually be Configuration.Service.MaxResultCount that is passed from the http handler to the operator's Execute. In that case, how will we even get to a point where we have LimitExceededException as we are the ones specifying how many records to fetch?

I think we should either do db.IntervalActions() and do the check or just do db.IntervalActionsWithLimit(limit) where a check wouldn't be required. The former is going to be slower though.

AnthonyMBonafide: Good point! I agree, this can go either way. Even though I like the idea of using a limit for performance reasons, I think it may be better to have the `intervalActionLoadAll` have a field for the `config.ServiceInfo` and load all the entities with ` db.IntervalActions()`. That way we can check the number of entries against the `MaxResultCount` and decide to either return the results or an `ErrLimitExceeded`. Doing it the other way, which will perform better with large amount of database entries, the user will lose information since they would only get a subset of the data they are trying to obtain, and they will have no indication that the data they have been given is only a subset. The also makes it more difficult to suggest since a `GET` to an endpoint with no limit, range, or filter criteria should return all the results or an error, I do not think there is any decent middle ground there from an API usage point-of-view. I hope this rant makes sense.

tsconn23:
recheck

internal/support/scheduler/rest_intervalaction.go
```diff
@@ -0,0 +1,44 @@
...
+		defer r.Body.Close()
+	}
+	op := intervalaction.NewAllExecutor(dbClient, Configuration.Service)
+	intervalActions, err := op.Execute()
+
+	if err != nil {
+		switch err.(type) {
+		case errors.ErrLimitExceeded:
+			http.Error(w, "Exceeded max limit", http.StatusRequestEntityTooLarge)
+			LoggingClient.Error("Exceeded max limit")
```
AnthonyMBonafide: Can we simply log the `err.Error()` and let the error struct handle the creation of a message, like on line 38? This would make these changes more consistent with the rest of the REST handlers. If this is the case we can move the  `LoggingClient.Error(err.Error())` logic on line 38 to right after the `if err != nil` check.

internal/support/scheduler/operators/intervalaction/get.go
tsconn23: I am going to mark this resolved as I do not see a `limit` parameter in the current logic flow. Also, simply because a limit is specified at 1,000 there may only be 2 records returned, which would not exceed a MaxResultCount of 100.

internal/support/scheduler/operators/intervalaction/get.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// This method gets interval actions from the database.
+func (op intervalActionLoadAll) Execute() ([]contract.IntervalAction, error) {
+	intervalActions, err := op.database.IntervalActions()
+
+	if len(intervalActions) > op.config.MaxResultCount {
+		return intervalActions, errors.NewErrLimitExceeded(len(intervalActions))
+	}
+	if err != nil {
```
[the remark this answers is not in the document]
tsconn23: It appears this is now the order of the error checking. I agree about the `nil` return value, but returning an empty array won't hurt anything.

internal/support/scheduler/rest_intervalaction.go
tsconn23: 413 is returned on line 36 below. Config MaxResult check is inside the operator.

PR #1733, opened 2019-09-04
Edinburgh patch to fix issue #1731 (error handling in core-data POST event/)

### LINKED ISSUE #1731: Error Not Being Checked When Event is Added
In core-data's `eventHandler` we are apparently not checking the error being returned from the `addNewEvent` function call

https://github.com/edgexfoundry/edgex-go/blob/bbadd2666cf18a2b4fea26103b0e8785b0966c47/internal/core/data/router.go#L230

This results in a 200 OK being returned in all cases.

Steps to duplicate
1.) In the core-data/res/configuration.toml set [Writable]MetaDataCheck = true
2.) POST to `http://localhost:48080/api/v1/event` with the following event payload
```
{  
   "created":1471806386919,
   "origin":1471806386919,
   "device":"livingroomthermostatdevice3",
   "readings":[  
      {  
         "origin":1471806386919,
         "name":"luminousIntensity1",
         "value":"40"
      }
   ]
}
```
3.) You will receive a 200 with an empty response body. The response body for a 200 should ALWAYS contain an event ID. The error from metadata telling us the device is invalid is being ignored. 

It appears that the previous logic returning appropriate status codes based on the returned error needs to be restored.

Please verify whether or not this is the case in the Edinburgh release. If so, we may need to consider another patch release.

jdharms:
Fixes #1731 for Edinburgh release.


jdharms:
This is my first time creating a PR for a patch to a released version, so please let me know if I missed anything.

anonymouse64:
Will this trigger us to tag 1.0.2 and release?

tsconn23:
> Will this trigger us to tag 1.0.2 and release a new version for Edinburgh?

No, The TSC voted to not do that this morning. We're just getting it ready in case something else comes up.

anonymouse64:
@tsconn23 if that's the case can we drop the version bump from this PR?

tsconn23:
> @tsconn23 if that's the case can we drop the version bump from this PR?

If we don't then the merge job will overwrite the previously uploaded 1.0.1 images in Nexus.
https://jenkins.edgexfoundry.org/view/edgex-go/job/edgex-go-edinburgh-merge-go/configure

if we change it now then if we do decide to do a 1.0.2 release, subsequent PRs will be included i that patch release.

anonymouse64:
hmm okay, I didn't realize that every merge job will overwrite new images in nexus with the same version, that seems fragile, but oh well. I assume jenkins pipelines will solve this issue for good when everything is transitioned to that

mkbhanda:
😊 No re-writing history. Need to bump up!

Issue #1734, opened 2019-09-04
Error Handling Refactor


AnthonyMBonafide:
# 🚀 Feature Request

### Relevant Package
Core services and possibly mod-core-contracts

### Description
The way we are creating, handling, and logging errors causes a lot of duplicate code and does not provide a clear consistent pattern to be used across the board. For example, the translation of error to an HTTP response is duplicated in many REST handlers across core service:

https://github.com/edgexfoundry/edgex-go/blob/087ead41cf58c57bbed4836ce723319706201d7e/internal/core/metadata/rest_deviceprofile.go#L108-L119

As for logging, there are some situations where we are logging multiple times for the same error. For example, in the `restGetAllDeviceProfiles` REST handler a `ErrLimitExceeded` is logged within the operator:
https://github.com/edgexfoundry/edgex-go/blob/8627af8819795f333cf7a38c6bcd9bb746df233b/internal/core/metadata/operators/device_profile/get.go#L37

and the error is bubbled up to the REST handler and again logged with less context:
https://github.com/edgexfoundry/edgex-go/blob/087ead41cf58c57bbed4836ce723319706201d7e/internal/core/metadata/rest_deviceprofile.go#L39

### Describe the solution you'd like
A pattern, paradigm, and/or strategy which will standardize the common functionality used when dealing with errors so that we can eliminate duplicate code and improve logging. Possibly do this in a non-invasive fashion to minimize the work that needs to be done to adopt these new ideas.

If possible I think it would be nice if we can also address the issue of inconsistent response bodies when an error is encountered. In an effort to harden our APIs I think we should always respond to a request with a proper content type and response body which adheres to that content type. Also, we should come up with an error structure so that clients can unmarshal the error response and use pro grammatically. 

### Describe alternatives you've considered
A proposal has been made and presented in core working group meetings on [August 22nd](https://zoom.us/recording/share/b823jU0h-MM0HOTFMCfgCccyZCORgNdILX3jAaNZdF6wIumekTziMw) , and [August 29th](https://zoom.us/recording/share/N7VEYfqTC8bSvO7c12TdT4VTdbf4ojao4uY4-T85x4CwIumekTziMw). 

You can find the code presented in those meetings:
edgex-go: https://github.com/AnthonyMBonafide/edgex-go/tree/proposal/issue_1671-common_error_type_and_handling

[...]

PR #1736, opened 2019-09-04
Standardize error usage

### LINKED ISSUE #1671: Standardize Error Types to NOT Use Pointers
We have a number of custom error types in EdgeX such as those found here

- [core-data](https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/data/errors/types.go)
- [core-metadata](https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/metadata/errors/types.go)

Older exception types are defined in a manner where their `New()` functions [return a pointer](https://github.com/edgexfoundry/edgex-go/blob/7ef8bff0a329065968754ab0ab9affabe0155e66/internal/core/data/errors/types.go#L104)

Over time, new exception types have been added wherein the `New()` function [does not](https://github.com/edgexfoundry/edgex-go/blob/7ef8bff0a329065968754ab0ab9affabe0155e66/internal/core/metadata/errors/types.go#L119) return a pointer.

We need to ensure universal implementation on this and right now, I'm of the opinion that these error types do not need to be returned from their `New()` functions as pointers. They're small, don't have any pointers as members and we want them to be immutable.

Along with reconciling these definitions, switch statements used for type checking should no longer expect pointers. Example of current code:

```
switch err.(type) {
		case *types.ErrDuplicateName:
		case *types.ErrItemNotFound:
			http.Error(w, err.Error(), http.StatusNotFound)
		default:
		}
```
The pointer dereference will need to be removed from all such statements.

AnthonyMBonafide:
Fix #1671

Update constructors, error type checks, and tests to use values rather
than references(pointers) for errors.

Update tests to conform to the changes in the production code.

### DIFF
--- a/internal/core/data/operators/reading/get_test.go
+++ b/internal/core/data/operators/reading/get_test.go
@@ -17,14 +17,14 @@
 package reading
 
 import (
-	"errors"
+	goErrors "errors"
 	"reflect"
 	"testing"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 
-	errors2 "github.com/edgexfoundry/edgex-go/internal/core/data/errors"
+	"github.com/edgexfoundry/edgex-go/internal/core/data/errors"
 	"github.com/edgexfoundry/edgex-go/internal/core/data/operators/reading/mocks"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
 )
@@ -36,7 +36,7 @@ var TestErrorReadingName = "TestReadingError"
 var TestLoaderLimit = 5
 
 // TestError error used to simulate non EdgeX type errors.
-var TestError = errors.New("testing error")
+var TestError = goErrors.New("testing error")
 
 var TestReading = contract.Reading{
 	Name:        "TestReadingName",
@@ -120,7 +120,7 @@ func TestGetReadingsExecutor(t *testing.T) {
 			config.ServiceInfo{MaxResultCount: 1},
 			nil,
 			true,
-			&errors2.ErrLimitExceeded{},
+			errors.ErrLimitExceeded{},
 		},
 	}
 	for _, test := range tests {
--- a/internal/core/data/operators/value_descriptor/get_test.go
+++ b/internal/core/data/operators/value_descriptor/get_test.go
@@ -17,20 +17,20 @@
 package value_descriptor
 
 import (
+	goErrors "errors"
 	"reflect"
 	"testing"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
-	"github.com/pkg/errors"
 
-	errors2 "github.com/edgexfoundry/edgex-go/internal/core/data/errors"
+	"github.com/edgexfoundry/edgex-go/internal/core/data/errors"
 	"github.com/edgexfoundry/edgex-go/internal/core/data/operators/value_descriptor/mocks"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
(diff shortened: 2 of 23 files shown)


internal/core/data/operators/value_descriptor/get_test.go
```diff
@@ -95,7 +95,7 @@ func TestGetValueDescriptorsByNames(t *testing.T) {
 			config.ServiceInfo{MaxResultCount: 1},
 			nil,
 			true,
-			&errors2.ErrLimitExceeded{},
```
brandonforster: While we're in the neighborhood, should we update this from `errors2`?

internal/core/metadata/rest_deviceprofile.go
```diff
@@ -110,7 +110,7 @@ func restUpdateDeviceProfile(w http.ResponseWriter, r *http.Request) {
 				http.Error(w, err.Error(), err.(types.ErrServiceClient).StatusCode)
 			case errors.ErrDeviceProfileNotFound:
 				http.Error(w, err.Error(), http.StatusNotFound)
-			case *errors2.ErrValueDescriptorsInUse:
+			case errors2.ErrValueDescriptorsInUse:
```
brandonforster: Another `errors2`.

internal/core/data/operators/value_descriptor/get_test.go
```diff
@@ -167,7 +167,7 @@ func TestGetAllValueDescriptors(t *testing.T) {
...
+			errors2.ErrLimitExceeded{},
```
brandonforster: Another `errors2`.

internal/core/data/operators/reading/get_test.go
```diff
@@ -120,7 +120,7 @@ func TestGetReadingsExecutor(t *testing.T) {
 			config.ServiceInfo{MaxResultCount: 1},
 			nil,
 			true,
-			&errors2.ErrLimitExceeded{},
+			errors2.ErrLimitExceeded{},
```
brandonforster: Another `errors2`.

AnthonyMBonafide:
I verified these changes with the black-box tests. Everything seems to be good. However, there were some test failures, but they also existed in the `master` branch. After closer inspection of the test failures I was seeing, none of them were due to these changes. I did not determine why some were failing, I chalked it up to something being wrong with my setup as I had to re-work a few things as most of them were related to state. If I notice anything I will create an issue.

AnthonyMBonafide:
The SonarCloud quality gate is failing due to duplicate code. It is highlighting the code that I changed. The code in question is not new just a change to avoid using pointers. The duplicate code issue is expected to be handled in the Error handling re-work issue #1734 .

internal/core/data/operators/value_descriptor/get_test.go
AnthonyMBonafide: Good point, I think I will follow what you have done recently and alias the `builtin/errors` with `goErrors` and remove the alias from our errors.

tsconn23:
> The SonarCloud quality gate is failing due to duplicate code. It is highlighting the code that I changed. The code in question is not new just a change to avoid using pointers. The duplicate code issue is expected to be handled in the Error handling re-work issue #1734 .

Agreed, this code is pre-existing and SonarQube is treating it as new code based on the inclusion in this PR. What's also interesting is that it's penalizing duplicates not just in a single file but in files across applications (like router.go in both export-client and export-distro). It could be instructive to look at this report and refactor where it makes sense, but that's out of scope for what we're doing here and can be handled via other issues/PRs IMHO.

We also need to take a look at how to get around this `errors` aliasing. Perhaps the package for our custom errors should be unique -- like `internal/pkg/types/errors.go`. That way `types` doesn't conflict with `errors`. And when it conflicts with `types` from `go-mod-core-contracts`, I'd be more comfortable with a descriptive `contractTypes` alias.

PR #1741, opened 2019-09-06
Fix Redis' ReadingsByDevice to get latest readings

### LINKED ISSUE #1643: Query event and reading API with limit always return first $n records when database use redis.
# 🐞 Bug Report

### Affected Services

 The issue is located in: core-data

### Is this a regression? No.

1. Startup EdgeX with device-virtual which database use **redis**
2. Invoke the core-data API to get the events/reading limited to 5 records
http://localhost:48080/api/v1/event/device/Random-Integer-Device/5
http://localhost:48080/api/v1/reading/device/Random-Integer-Device/5

## 🔥 Exception or Error
<pre><code>

Always return first 5 records.
</code></pre>
Expected result:
Should be the same as mongo to return last 5 records.

## 🌍  Your Environment
**Deployment Environment:** ubuntu 18.04

**EdgeX Version:** edinburgh 1.0.0

**Anything else relevant?**

AnthonyMBonafide:
Fix #1643

Update the Redis client to return the latest readings when a limit is
applied to the query. This make the Redis client more consistent with
the desired Mongo client results.

### DIFF
--- a/internal/pkg/db/redis/data.go
+++ b/internal/pkg/db/redis/data.go
@@ -702,7 +702,7 @@ func (c *Client) ReadingsByDevice(id string, limit int) (readings []contract.Rea
 	conn := c.Pool.Get()
 	defer conn.Close()
 
-	objects, err := getObjectsByRange(conn, db.ReadingsCollection+":device:"+id, 0, limit-1)
+	objects, err := getObjectsByRevRange(conn, db.ReadingsCollection+":device:"+id, 0, limit-1)
 	if err != nil {
 		if err != redis.ErrNil {
 			return readings, err
--- a/internal/pkg/db/redis/queries.go
+++ b/internal/pkg/db/redis/queries.go
@@ -17,9 +17,10 @@ import (
 	"encoding/json"
 	"strconv"
 
+	"github.com/gomodule/redigo/redis"
+
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db/redis/models"
-	"github.com/gomodule/redigo/redis"
 )
 
 func getObjectById(conn redis.Conn, id string, unmarshal unmarshalFunc, out interface{}) error {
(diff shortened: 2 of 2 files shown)


AnthonyMBonafide:
I verified these changes manually and also ensured the black-box tests still pass.

AnthonyMBonafide:
SonarCube is failing due to duplicate code. None of the duplicate code it is complaining about is introduced in the PR.

AnthonyMBonafide:
@andresrinivasan, Thank you so much for that suggestion! I like these changes a lot better now.

andresrinivasan:
My OCD would like to refactor the code in `queries.go`

``` go
// getObjectsByRange retrieves the entries for keys enumerated in a sorted set. The entries are retrieved in the sorted set order.
func getObjectsByRange(conn redis.Conn, key string, start, end int) (objects [][]byte, err error) {
	return getObjectsBySomeRange(conn, "RANGE", key, start, end)
}

// getObjectsByRevRange retrieves the entries for keys enumerated in a sorted set. The entries are retrieved in the reverse sorted set order.
func getObjectsByRevRange(conn redis.Conn, key string, start int, end int) (objects [][]byte, err error) {
	return getObjectsBySomeRange(conn, "REVRANGE", key, start, end)
}

//getObjectsBySomeRange retrieves the entries for keys enumerated in a sorted set using the RANGE or REVRANGE Redis command.
func getObjectsBySomeRange(conn redis.Conn, command string, key string, start int, end int) (objects [][]byte, err error) {
	ids, err := redis.Values(conn.Do(command, key, start, end))
	if err != nil && err != redis.ErrNil {
		return nil, err
	}

	if len(ids) > 0 {
		objects, err = redis.ByteSlices(conn.Do("MGET", ids...))
		if err != nil {
			return nil, err
		}
	}

	return objects, nil
```

AnthonyMBonafide:
@andresrinivasan , I do not see this functionality being verified in either of the unit nor black-box tests. @cherrycl, did the black-box tests put up this inconsistency or did you spot this using some other verification technique?

cherrycl:
@AnthonyMBonafide Blackbox testing repository didn't contain any Redis related tests. I found this issue when I collected the response time of the transaction by [performance summary collection](https://github.com/edgexfoundry-holding/performance-summary-collection) repository.

PR #1747, opened 2019-09-09
Add [generate] subcommand from merging pki-init option generate into security-secrets-setup

jim-wang-intel:
To add generate subcommand into security-secrets-setup:

 - TLS assets are generated from "scratch" ("working") dir and currently
   default to `/edgex/security-secretes-setup/scratch` which is under the path
   given by `XDG_RUNTIME_DIR` environment variable

 - TLS assets are arranged into several directories based on service names

 - TLS assets are deployed into "deployed" dir and is currently default to `/run/edgex/secrets` tmpfs directory

 - TLS assets include `CA`, `Vault`, and `Kong` (reverse proxy)

 - Sentinel files are generated upon completion of deploy for services

Note: 
  The legacy option `-config <configuration_file_dir>` or `-c <configuration_file_dir>` is still available under subcommand `legacy` until phase out in the next phase or release.    Example of usage.:  `./security-secrets-setup legacy -config ./res/pkisetup-vault.json`

Related to: 
- Issue #1635
- Issue #1558 

### DIFF
--- a/cmd/security-secrets-setup/main.go
+++ b/cmd/security-secrets-setup/main.go
@@ -128,9 +128,7 @@ func setupPkiInitOption(subcommand string) (executor option.OptionsExecutor, sta
 	case "generate":
 		generateOpt = true
 	default:
-		flag.Usage()
-		exitInstance.exit(0)
-		return
+		return nil, 1, fmt.Errorf("unsupported subcommand %s", subcommand)
 	}
 
 	opts := option.PkiInitOption{
--- a/cmd/security-secrets-setup/main.go
+++ b/cmd/security-secrets-setup/main.go
@@ -19,7 +19,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"strings"
 	"time"
 
 	"github.com/edgexfoundry/edgex-go/internal"
@@ -43,33 +42,28 @@ type pkiInitOptionDispatcher struct{}
 var exitInstance = newExit()
 var dispatcherInstance = newOptionDispatcher()
 var configFile string
-
-func init() {
-	// define and register command line flags:
-	flag.StringVar(&configFile, "config", "", "specify JSON configuration file: /path/to/file.json")
-	flag.StringVar(&configFile, "c", "", "specify JSON configuration file: /path/to/file.json")
-
-	flag.Usage = usage.HelpCallbackSecuritySetup
-}
+var subcommandList = []string{"legacy", "generate", "cache", "import"}
 
 func main() {
 	start := time.Now()
 
-	if len(os.Args) < 2 {
-		fmt.Println("Please specify subcommand or options for " + option.SecuritySecretsSetup)
+	// define and register command line subcommands:
+	legacyCmd := flag.NewFlagSet("legacy", flag.ExitOnError)
+	legacyCmd.StringVar(&configFile, "config", "", "specify JSON configuration file: /path/to/file.json")
+	legacyCmd.StringVar(&configFile, "c", "", "specify JSON configuration file: /path/to/file.json")
+
+	flag.Usage = usage.HelpCallbackSecuritySetup
+
+	flag.Parse()
+
+	if len(os.Args) < 2 || flag.NArg() < 1 {
+		fmt.Println("Please specify subcommand for " + option.SecuritySecretsSetup)
 		flag.Usage()
 		exitInstance.exit(0)
 		return
 	}
 
-	// Before we call Golang's flag.Parse(), we want to make sure that
-	// the subcommand is extracted ans saved if subcommand is used
-	// retrieve subcommand and delete it from os.Args[]
-	// as the Golang flag.Parse() method always parses from
-	// the frist arguments of os.Args[]
-	subcommand := retrieveSubcommand()
-
-	flag.Parse()
+	subcommand := flag.Args()[0]
 
 	if err := setup.Init(); err != nil {
 		// the error returned from Init has already been logged inside the call
@@ -78,24 +72,39 @@ func main() {
 		return
 	}
 
-	if configFile == "" {
-		// run with other options for pki-init
-		statusCode, err := dispatcherInstance.run(subcommand)
+	if checkIfMultipleSubcommands() {
+		setup.LoggingClient.Error("cannot use multiple subcommands, use one at a time")
+		exitInstance.exit(1)
+		return
+	}
+
+	setup.LoggingClient.Debug(fmt.Sprintf("subcommand <%s>", subcommand))
+
+	var exitStatusCode int
+	var err error
+	// legacy mode of operation
+	if "legacy" == subcommand {
+		err = legacyCmd.Parse(flag.Args()[1:])
+		if err != nil {
+			setup.LoggingClient.Error(err.Error())
+			exitInstance.exit(2)
+			return
+		}
+		if err = option.GenTLSAssets(configFile); err != nil {
+			setup.LoggingClient.Error(err.Error())
+			exitInstance.exit(2)
+			return
+		}
+	} else {
+		// pki-init mode of operations
+		exitStatusCode, err = dispatcherInstance.run(subcommand)
 		if err != nil {
 			setup.LoggingClient.Error(err.Error())
 		}
-
-		exitInstance.exit(statusCode)
-		return
-	}
-
-	if err := option.GenTLSAssets(configFile); err != nil {
-		setup.LoggingClient.Error(err.Error())
-		exitInstance.exit(2)
-		return
 	}
 
 	setup.LoggingClient.Info(option.SecuritySecretsSetup+" complete", internal.LogDurationKey, time.Since(start).String())
+	exitInstance.exit(exitStatusCode)
 }
 
 func newExit() exiter {
@@ -110,16 +119,16 @@ func newOptionDispatcher() optionDispatcher {
 	return &pkiInitOptionDispatcher{}
 }
 
-// retrieveSubcommand parses the subcommand out of os.Arg if any
-// otherwise returns empty string ("")
-func retrieveSubcommand() (subcommand string) {
-	// commandline options always starts with -
-	// so it is treated as a subcommand if that is not the case
-	if !strings.HasPrefix(os.Args[1], "-") {
-		subcommand = os.Args[1]
-		os.Args = append(os.Args[:1], os.Args[2:]...)
+func checkIfMultipleSubcommands() bool {
+	numSubCmds := 0
+	for _, arg := range flag.Args() {
+		for _, subcommand := range subcommandList {
+			if arg == subcommand {
+				numSubCmds++
+			}
+		}
 	}
-	return subcommand
+	return numSubCmds > 1
 }
 
 func setupPkiInitOption(subcommand string) (executor option.OptionsExecutor, status int, err error) {
--- a/cmd/security-secrets-setup/main.go
+++ b/cmd/security-secrets-setup/main.go
@@ -19,6 +19,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/edgexfoundry/edgex-go/internal"
@@ -34,7 +35,7 @@ type exiter interface {
 type exitCode struct{}
 
 type optionDispatcher interface {
-	run() (int, error)
+	run(command string) (int, error)
 }
 
 type pkiInitOptionDispatcher struct{}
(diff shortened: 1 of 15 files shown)


jim-wang-intel:
recheck

jim-wang-intel:
@anonymouse64 Ian, @tingyuz, @bnevis-i , @mkbhanda Please review

anonymouse64:
recheck

jim-wang-intel:
> 
> 
> Abstain (already reviewed this code internally 10x times)

LOL, Bryon, you can come back tomorrow and gives an approval when you are ready.   I'll give you a nudge, though. :-)

tsconn23:
recheck

internal/pkg/config/x509_test.go
```diff
@@ -64,3 +64,31 @@ func TestPkiCAOutDir(t *testing.T) {
 	assert.True(strings.Contains(pkiDir,
 		filepath.Join(filepath.Base(x509Config.WorkingDir), x509Config.PKISetupDir, x509Config.RootCA.CAName)))
 }
+
+func TestGetCAPemFileName(t *testing.T) {
+	// ignore the file reading error since it will be caught by other test already
+	x509Config, _ := NewX509Config(testJSONFileName)
+	pemFileName := x509Config.GetCAPemFileName()
+	assert.Equal(t, x509Config.RootCA.CAName+"."+certFileExt, pemFileName)
```
anonymouse64: this isn't a great test since it's duplicating the exact same logic of the function it's testing so if there's some problem with this logic we won't catch it since it's doing the exact same thing that the function is doing to compare the results.
A better test would be:

```go
x509Config, _ := NewX509Config(testJSONFileName)
// code this in the test so if the file changes or the function changes we will catch it in the tests
expected := "the-ca-name.cert"
assert.Equal(t,x509Config.GetCAPemFileName(),expected)
```

internal/pkg/config/x509_test.go
```diff
@@ -64,3 +64,31 @@ func TestPkiCAOutDir(t *testing.T) {
...
+}
+
+func TestGetCAPrivateKeyFileName(t *testing.T) {
+	// ignore the file reading error since it will be caught by other test already
+	x509Config, _ := NewX509Config(testJSONFileName)
+	prvKeyFileName := x509Config.GetCAPrivateKeyFileName()
+	assert.Equal(t, x509Config.RootCA.CAName+"."+skFileExt, prvKeyFileName)
```
anonymouse64: ditto for this test

internal/pkg/config/x509_test.go
```diff
@@ -64,3 +64,31 @@ func TestPkiCAOutDir(t *testing.T) {
...
+}
+
+func TestGetTLSPemFileName(t *testing.T) {
+	// ignore the file reading error since it will be caught by other test already
+	x509Config, _ := NewX509Config(testJSONFileName)
+	pemFileName := x509Config.GetTLSPemFileName()
+	assert.Equal(t, x509Config.TLSServer.TLSHost+"."+certFileExt, pemFileName)
```
anonymouse64: ditto

internal/pkg/config/x509_test.go
```diff
@@ -64,3 +64,31 @@ func TestPkiCAOutDir(t *testing.T) {
...
+}
+
+func TestGetTLSPrivateKeyFileName(t *testing.T) {
+	// ignore the file reading error since it will be caught by other test already
+	x509Config, _ := NewX509Config(testJSONFileName)
+	prvKeyFileName := x509Config.GetTLSPrivateKeyFileName()
+	assert.Equal(t, x509Config.TLSServer.TLSHost+"."+skFileExt, prvKeyFileName)
```
anonymouse64: ditto

internal/security/setup/option/generate.go
```diff
@@ -0,0 +1,188 @@
...
+	return pkiInitOption == nil || !pkiInitOption.GenerateOpt
+}
+
+func generatePkis() (exitCode, error) {
+	baseWorkingDir, err := os.Getwd()
+	if err != nil {
+		return exitWithError, err
+	}
+
+	resourceDirPath := filepath.Join(baseWorkingDir, resourceDirName)
```
anonymouse64: this should be configurable from the command line, this is hard-coding the utility to require that in the current working directory has `res/pki-setup-vault.json`, etc. in it which is wrong, it should be configurable where to find these things, either through a configuration.toml file or through a command line option

jim-wang-intel:
> 
> 
> Okay so I've identified a couple larger issues surrounding command line options and how we handle such things.
> 
>     * When we define `--config`, we never do the new behavior with the subcommands
Yes, that is intended behavior, --config is legacy option to generate TLS assets, and it should be replaced by `generate` subcommand.  So I agree with you that we should error out if both specified and better yet, I like your suggestion below to convert that legacy option to a subcommand like `legacy` or `pkisetup` to be more consistent with rest of pattern subcommands.

> 
>     * The new subcommands ignore the legacy option
> 
Same as bove

>     * there's no verification that takes place if you specify multiple subcommands
> 
> 

agree, will add the checking logic


> this leaves the current implementation with the following concrete bugs IMO:
> 
>     * there is no way to configure where certain things like the json files for the new way, because it's always hard-coded to be `./res/.../pki-setup-vault.json`, probably this should be a subcommand specific option, but it could also be a global option since I think all subcommands will need to know where to find these json files, or at least be able to specify a different directory than `$CWD` with a command line option
> 

This is an enhancement effort in toml file in separate story not the current focus of this PR.  will create an issue in git for this.


>     * specifying both `--config` and a subcommand like `generate` should error out since `--config` is mutually exclusive with `generate` at least as written now
> 
>     * if we want to keep the old legacy behavior, I think at this point it should be put under a specific `legacy` or maybe even `pkisetup` subcommand to make it clear it is mutually exclusive with the new behavior, it could still take a `--config` option and maybe even if we really really wanted to be conservative, we could make that subcommand the default subcommand when no subcommand is specified and print a warning message about defaulting to legacy behavior, however I think that's all unnecessary
> 
I like your suggestion here for converting the legacy option into a subcommand.

>     * trying to specify multiple subcommands should error out somehow to explain that they are mutually exclusive
> 
> 
> Lastly, I think some of the tests are not super useful since they're doing the same thing the function they are supposed to be testing is doing which feels wrong to me. See my comments for what I think is a better way to do it.

internal/security/setup/option/generate.go
jim-wang-intel: will be through a configuration .toml file...in this phase, we only focus on the functionality. will address this issue in another PR.

internal/pkg/config/x509_test.go
jim-wang-intel: the second argument of `assert.Equal()` is actually the expected. so I am not following you in the example you gave above.

anonymouse64: I'm saying construct the expected string without using the X509Config struct

jim-wang-intel: ok, i changed all that to hard-coded in test.

internal/security/setup/option/generate.go
jim-wang-intel: #1767  is filed for this issue

jim-wang-intel:
Hi @anonymouse64 Ian, thanks for your valuable PR feedback and I think I have addressed all your comments. Please re-review. Thanks, Jim

tingyuz:
recheck

jim-wang-intel:
recheck

cmd/security-secrets-setup/main.go
```diff
@@ -34,59 +34,77 @@ type exiter interface {
...
-	if err := option.GenTLSAssets(configFile); err != nil {
-		setup.LoggingClient.Error(err.Error())
-		exitInstance.exit(2)
-		return
+	setup.LoggingClient.Debug(fmt.Sprintf("subcommand <%s>", subcommand))
+
+	var exitStatusCode int
+	var err error
+	// legacy mode of operation
+	if "legacy" == subcommand {
```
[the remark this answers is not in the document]
anonymouse64: See https://play.golang.org/p/_3mHcB4chMj for an example of how I think this should be done in an extensible way that has easy error checking

jim-wang-intel:
@anonymouse64 Ian, please give a green light :-)

cmd/security-secrets-setup/main.go
```diff
@@ -69,24 +75,52 @@ func main() {
...
+
+	switch subcmdName {
+	case "legacy":
+		if err = option.GenTLSAssets(configFile); err != nil {
+			setup.LoggingClient.Error(err.Error())
+			exitInstance.exit(2)
+			return
+		}
+	case "generate", "cache", "import":
+		// no options expected
```
anonymouse64:
```suggestion
		// no arguments expected
```

cmd/security-secrets-setup/main.go
```diff
@@ -69,24 +75,52 @@ func main() {
 		return
 	}
 
-	if configFile == "" {
-		// run with other options for pki-init
-		statusCode, err := dispatcherInstance.run()
-		if err != nil {
-			setup.LoggingClient.Error(err.Error())
-		}
+	if checkIfMultipleSubcommands() {
```
anonymouse64: we don't need this anymore because this case is handled by two different cases below:

- the first subcommand is a known subcommand in which we know it doesn't accept args and fails in the `switch`
- the first subcommand is an unknown subcommand which is handle in the `if !found` ...

cmd/security-secrets-setup/main.go
```diff
@@ -69,24 +75,52 @@ func main() {
...
 		exitInstance.exit(2)
 		return
 	}
 
+	var exitStatusCode int
+	var err error
+
+	switch subcmdName {
+	case "legacy":
+		if err = option.GenTLSAssets(configFile); err != nil {
```
anonymouse64: we should also confirm that there are no additional arguments to the legacy command here like we do below for generate cache, import, etc.

cmd/security-secrets-setup/main.go
```diff
@@ -69,24 +75,52 @@ func main() {
...
+			setup.LoggingClient.Error(err.Error())
+			exitInstance.exit(2)
+			return
+		}
+	case "generate", "cache", "import":
```
anonymouse64: since we don't support `cache` and `import` yet can we drop these in this PR?

cmd/security-secrets-setup/main.go
jim-wang-intel: i thought that is checked or handled by subcommand's parse api itself already like invalid flag provided error message will be given in flag's parse as well

cmd/security-secrets-setup/main.go
jim-wang-intel: ok

cmd/security-secrets-setup/main.go
jim-wang-intel: ok

cmd/security-secrets-setup/main.go
jim-wang-intel: ok

cmd/security-secrets-setup/main.go
anonymouse64: no it is not, what will happen is that the flagset will just have those arguments defined, there's no way of telling the stdlib flag library "this flagset doesn't take any arguments"

jim-wang-intel: ok, but this "legacy" subcommand does accept argument -config flag, though, see the leagcy subcommand flagset definition in the code.

anonymouse64: the argument `-config` does not show up in `flagset.Args()` because it is parsed as the option, because we are using a flag set and defining the option in the flagset

jim-wang-intel: so here is the output of my execution without that checking additional argument logic.  It looks  like the subcomand parse(os.args[1:]) already does the checking for you, doesn't it?  
```
./security-secrets-setup legacy -config ./res/pkisetup-vault.json -extra invalid
level=ERROR ts=2019-09-19T17:25:26.178019174Z app=edgex-security-secrets-setup source=logger.go:105 msg="logTarget cannot be blank, using stdout only"
flag provided but not defined: -extra
Usage of legacy:
  -c string
    	specify JSON configuration file: /path/to/file.json
  -config string
    	specify JSON configuration file: /path/to/file.json
```

jim-wang-intel: ok, I think I got what you meant now. I also added some extra tests for these cases.

jim-wang-intel:
@tingyuz Would you please double check the duplicated code on security-secretstores side.  I just pushed my latest codes and feel innocent about the quality gate failed for the changes I have :-/

jim-wang-intel:
@anonymouse64 Ian, aside from the sonar gate issue, please take a look again. Thanks for your feedback.

bnevis-i:
Note failing "SonarCloud Code Analysis" on duplicated code in `init.go`  Also note that branch is out-of-date.

PR #1748, opened 2019-09-09
Fix issue #1750: add header when sending request to post kong cert into vault; change …

tingyuz:
…the kong services/routes names to lower cases

1. Header information is required inside postCert(), or the post action to KONG will be failed. 
2. When initializing services/routes for KONG, the names need to be lower case, or the initialization will be failed.

### DIFF
--- a/internal/security/proxy/consumer_test.go
+++ b/internal/security/proxy/consumer_test.go
@@ -169,26 +169,3 @@ func TestCreateJWTToken(t *testing.T) {
 		t.Errorf(err.Error())
 	}
 }
-
-/*func TestCreateOAuth2Token(t *testing.T) {
-	t.Skip()
-	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.WriteHeader(http.StatusOK)
-		w.Write([]byte(`{"token_type": "oauth2", "access_token": "test", "expires_in": 1442426001000}`))
-		if r.Method != "POST" {
-			t.Errorf("expected POST request, got %s instead", r.Method)
-		}
-
-		if r.URL.EscapedPath() != "/consumers/testuser/oauth2" {
-			t.Errorf("expected request to /consumers/testuser/oauth2, got %s instead", r.URL.EscapedPath())
-		}
-	}))
-	defer ts.Close()
-
-	co := Consumer{"testuser", &testConsumerRequestor{ts.URL}, &testConsumerConfig{ts.URL}}
-	_, err := co.createOAuth2Token()
-	if err != nil {
-		t.Errorf("failed to creat OAuth2 token for consumer")
-		t.Errorf(err.Error())
-	}
-}*/
--- a/internal/security/proxy/requestor.go
+++ b/internal/security/proxy/requestor.go
@@ -23,9 +23,9 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal"
 )
 
-func NewRequestor(skipVerify bool, timeout int) internal.HttpCaller {
+func NewRequestor(skipVerify bool, timeoutInSecond int) internal.HttpCaller {
 	tr := &http.Transport{
 		TLSClientConfig: &tls.Config{InsecureSkipVerify: skipVerify},
 	}
-	return &http.Client{Timeout: time.Duration(timeout) * time.Second, Transport: tr}
+	return &http.Client{Timeout: time.Duration(timeoutInSecond) * time.Second, Transport: tr}
 }
--- a/internal/security/proxy/service.go
+++ b/internal/security/proxy/service.go
@@ -104,8 +104,12 @@ func (s *Service) Init(cert CertificateLoader) error {
 
 	for _, name := range services {
 		params := Configuration.Clients[name]
+		if params.Host == "" {
+			LoggingClient.Warn("missing config information for service: " + name)
+			continue
+		}
 		serviceParams := &KongService{
-			Name:     name,
+			Name:     strings.ToLower(name),
 			Host:     params.Host,
 			Port:     params.Port,
 			Protocol: params.Protocol,
@@ -117,10 +121,10 @@ func (s *Service) Init(cert CertificateLoader) error {
 		}
 
 		routeParams := &KongRoute{
-			Paths: []string{"/" + name},
-			Name:  name,
+			Paths: []string{"/" + strings.ToLower(name)},
+			Name:  strings.ToLower(name),
 		}
-		err = s.initKongRoutes(routeParams, name)
+		err = s.initKongRoutes(routeParams, strings.ToLower(name))
 		if err != nil {
 			return err
 		}
@@ -160,6 +164,7 @@ func (s *Service) postCert(cert CertificateLoader) error {
 	}
 	tokens := []string{Configuration.KongURL.GetProxyBaseURL(), CertificatesPath}
 	req, err := http.NewRequest(http.MethodPost, strings.Join(tokens, "/"), strings.NewReader(string(data)))
+	req.Header.Add(clients.ContentType, clients.ContentTypeJSON)
 	if err != nil {
 		LoggingClient.Error("failed to create upload cert request -- %s", err.Error())
 		return err


tingyuz:
> This PR should also delete https://github.com/edgexfoundry/edgex-go/blob/c810a60f0fd49918467451174b412dd73294d2e5/internal/interface.go



> This PR should also delete https://github.com/edgexfoundry/edgex-go/blob/c810a60f0fd49918467451174b412dd73294d2e5/internal/interface.go

More details about the discussion to delete this file? I see it still exists in the master branch.

internal/security/proxy/consumer_test.go
```diff
@@ -171,24 +171,24 @@ func TestCreateJWTToken(t *testing.T) {
...
+	 }))
+	 defer ts.Close()
+
+	 co := Consumer{"testuser", &testConsumerRequestor{ts.URL}, &testConsumerConfig{ts.URL}}
+	 _, err := co.createOAuth2Token()
+	 if err != nil {
+		 t.Errorf("failed to creat OAuth2 token for consumer")
+		 t.Errorf(err.Error())
+	 }
+ }*/
```
jim-wang-intel: Can we clean up this commented out dead code? You have that code available in the version-control repository anyway.

internal/security/proxy/certs.go
```diff
@@ -22,21 +22,19 @@ import (
...
 }
 
 type certificate struct {
-	client    internal.HttpCaller
+	client    Requestor
 	certPath  string
 	tokenPath string
 }
 
-func NewCertificateLoader(r internal.HttpCaller, certPath string, tokenPath string) CertificateLoader {
```
anonymouse64: @tingyuz it looks like this is replacing usage of `internal.HttpCaller` with `Requestor`, if that's the case we should do that everywhere, as it is we still have duplication between the HttpCaller and Requestor interfaces - we should only have one of those in the codebase

internal/security/proxy/requestor.go
```diff
@@ -19,11 +19,13 @@ import (
...
-
-	"github.com/edgexfoundry/edgex-go/internal"
 )
 
-func NewRequestor(skipVerify bool, timeout int) internal.HttpCaller {
+type Requestor interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+func NewRequestor(skipVerify bool, timeout int) Requestor {
```
jim-wang-intel: `timeoutInSecond` (instead of just `timeout`) seems better to read at first glance of this func before looking into the details of code below to figure it out.

internal/security/proxy/service.go
```diff
@@ -105,7 +104,7 @@ func (s *Service) Init(cert CertificateLoader) error {
 	for _, name := range services {
 		params := Configuration.Clients[name]
```
jim-wang-intel: What happens if `name`  key not exists in that map?  Should we handle that corner case?

internal/security/proxy/consumer_test.go
tingyuz: will do.

internal/security/proxy/requestor.go
tingyuz: agree and will change the name of the parameter.

internal/security/proxy/certs.go
tingyuz: I just did a review and found that other components such as internal\core\command is still referring it so it may not be a good idea to delete this file. I will prefer to leave this decision to the author of last component that refers to it.

anonymouse64: I would argue that the burden for consolidating the usages falls on you and/or @tsconn23 because y'all's PR is the one that introduced the duplicated code and I specifically asked that it be resolved in the PR but had to agree for the issue to be solved later so here we are

mkbhanda:
There are fewer calls to Requestor than HttpCaller .. keep HttpCaller?

internal/security/proxy/service.go
tingyuz: add handle for this corner case.

tingyuz:
> There are fewer calls to Requestor than HttpCaller .. keep HttpCaller?
I prefer to keep HttpCaller for this PR as it is being used by internal\core\command. If we remove HttpCaller, then we need to refactor \core\command, which has nothing to do with this PR. I would suggest to have another PR later to remove HttpCaller as it is being used by \core\command only.

tingyuz:
@tingyuz @anonymouse64 @jim-wang-intel  please review the changes I made per the suggestion. Thanks!

tingyuz:
recheck

tingyuz:
> Tingyu, would you please create an issue against core-command to get rid of that lone call to HttpCaller.
> Much thanks for addressing our comments.

Create issue #1749 for the request.

tingyuz:
recheck.

internal/security/proxy/requestor.go
```diff
@@ -19,13 +19,15 @@ import (
 	"crypto/tls"
 	"net/http"
 	"time"
-
-	"github.com/edgexfoundry/edgex-go/internal"
 )
 
-func NewRequestor(skipVerify bool, timeout int) internal.HttpCaller {
+type Requestor interface {
```
anonymouse64: I apologize if I didn't make this clear enough before, but this bit of code is the part I was calling out before. What I would like to see for this PR is to not introduce this new type, and instead everywhere we use `Requestor` to instead use `internal.HttpCaller`.

anonymouse64:
@mkbhanda I don't think it's appropriate for this PR to add duplicated code and then file a followup issue asking another WG to remove their interface when this PR is the one introducing duplicated code.

internal/security/proxy/requestor.go
tingyuz: OK I think we understand it now. I feel more comfortable removing Requestor and keep internal.HttpCaller for consistence. will update the project.

internal/security/proxy/consumer_test.go
anonymouse64: Are you planning on removing this code in this PR? I still see there's a large amount of commented out code being added here. I think keeping a git branch on your fork of the repository is a better place to keep this code rather than in the master branch.

internal/security/proxy/requestor.go
anonymouse64: Thank you

internal/security/proxy/service.go
```diff
@@ -104,8 +103,12 @@ func (s *Service) Init(cert CertificateLoader) error {
 
 	for _, name := range services {
 		params := Configuration.Clients[name]
+		if params.Host == "" {
+			LoggingClient.Info("missing params info for name: " + name)
```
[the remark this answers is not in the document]
tingyuz: changed to warning level.

tingyuz:
> (sorry for the duplicated review, for some reason GitHub lost these other comments)
> 
> Also, please include a description in the PR of what is being changed and why. It's not clear from the code where we are adding an additional HTTP Header when talking to Kong, and I don't see why we would need to do that or what happens currently if we don't add that header.

Added information in the description section.

tingyuz:
> (sorry for the duplicated review, for some reason GitHub lost these other comments)
> 
> Also, please include a description in the PR of what is being changed and why. It's not clear from the code where we are adding an additional HTTP Header when talking to Kong, and I don't see why we would need to do that or what happens currently if we don't add that header.

Information added in the description section.

internal/security/proxy/certs.go
tsconn23: @anonymouse64 @tingyuz This WAS removed. Why it's back, I have no idea.
https://github.com/edgexfoundry/edgex-go/pull/1660

tsconn23:
@tingyuz There is no issue for this PR.

tingyuz:
> @tingyuz There is no issue for this PR.

Created and added related issue number in the title of the PR.

internal/security/proxy/certs.go
anonymouse64: @tsconn23 my concern was resolved - see later comments on the PR, I was objecting to the duplicated Requestor and HttpCaller

Issue #1739, opened 2019-09-06
Common Bootstrap Package


michaelestrin:
Spike a flexible common package to be used by each core service to replace the existing copy/paste implementations of init.go and incorporate the existing internal/package/startup package.

AnthonyMBonafide:
I am pretty confident that the changes you are looking into will fix some, if not all, of the code duplication reported by [SonarCloud](https://sonarcloud.io/component_measures?id=edgexfoundry_edgex-go&metric=duplicated_lines&view=list) for the [init.go](https://sonarcloud.io/component_measures?id=edgexfoundry_edgex-go&metric=duplicated_lines&selected=edgexfoundry_edgex-go%3Ainternal%2Fcore%2Fcommand%2Finit.go&view=list) and [main.go](https://sonarcloud.io/component_measures?id=edgexfoundry_edgex-go&metric=duplicated_lines&selected=edgexfoundry_edgex-go%3Acmd%2Fcore-data%2Fmain.go&view=list) files. I just want to make sure we have some trace-ability.

PR #1751, opened 2019-09-11
Common Bootstrap Package

michaelestrin:
Extensible common bootstrap package to be used by each core
service to replace the existing copy/paste implementations of init.go,
incorporate the existing internal/package/startup package, and provide a
unified approach to concurrency and related go routine clean up.

Implementation includes core-data service changes to leverage the new
package.

Fixes https://github.com/edgexfoundry/edgex-go/issues/1739

### DIFF
--- a/internal/pkg/bootstrap/bootstrap.go
+++ b/internal/pkg/bootstrap/bootstrap.go
@@ -1,0 +1,110 @@
+/*******************************************************************************
+ * Copyright 2019 Dell Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package bootstrap
+
+import (
+	"context"
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/configuration"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/interfaces"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/logging"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/startup"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
+
+	"github.com/edgexfoundry/go-mod-registry/registry"
+)
+
+// fatalError logs an error and exits the application.  It's intended to be used only within the bootstrap prior to
+// any go routines being spawned.
+func fatalError(err error, loggingClient logger.LoggingClient) {
+	loggingClient.Error(err.Error())
+	os.Exit(1)
+}
+
+// translateInterruptToCancel spawns a go routine to translate the receipt of a SIGTERM signal to a call to cancel
+// the context used by the bootstrap implementation.
+func translateInterruptToCancel(wg *sync.WaitGroup, cancel context.CancelFunc) {
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		signalStream := make(chan os.Signal)
+		defer func() {
+			signal.Stop(signalStream)
+			close(signalStream)
+		}()
+		signal.Notify(signalStream, os.Interrupt, syscall.SIGTERM)
+		<-signalStream
+		cancel()
+	}()
+}
+
+// Run bootstraps an application.  It loads configuration and calls the provided list of handlers.  Any long-running
+// process should be spawned as a go routine in a handler.  Handlers are expected to return immediately.  Once all of
+// the handlers are called this function will wait for any go routines spawned inside the handlers to exit before
+// returning to the caller.  It is intended that the caller stop executing on the return of this function.
+func Run(
+	configDir, profileDir, configFileName string,
+	useRegistry bool,
+	serviceKey string,
+	config interfaces.Configuration,
+	startupTimer startup.Timer,
+	handlers []interfaces.BootstrapHandler) {
+
+	loggingClient := logging.FactoryToStdout(serviceKey)
+	var err error
+	var registryClient registry.Client
+	var wg sync.WaitGroup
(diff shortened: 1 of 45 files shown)


michaelestrin:
Rebased

michaelestrin:
Rebased.

michaelestrin:
Rebased.

michaelestrin:
Rebased.

michaelestrin:
recheck

michaelestrin:
Rebased.

michaelestrin:
Rebased.  Updated with minor changes to ensure black box test results are equivalent to master.

michaelestrin:
@tsconn23 Ready for review.  Ask is to review/merge.  Will create follow-on issues to integrate go-mod-secrets (with @AnthonyMBonafide), to convert the rest of the core services, and to remove the deprecated code.

internal/pkg/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,107 @@
+/*******************************************************************************
```
tsconn23: This file name is an example of "stutter". How can it be named differently?

internal/pkg/bootstrap/logging/factory.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/interfaces"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
+	"github.com/edgexfoundry/go-mod-core-contracts/models"
+)
+
+// FactoryToStdout returns a logger.LoggingClient that outputs to stdout.
+func FactoryToStdout() logger.LoggingClient {
+	return logger.NewClient(clients.CoreDataServiceKey, false, "", models.DebugLog)
```
tsconn23: I think the service key needs to be passed in as a parameter here rather than hard-coded. This function is called from the generic bootstrap/Run() function and will thus be utilized by any service.

internal/pkg/bootstrap/logging/factory.go
```diff
@@ -0,0 +1,40 @@
...
+// FactoryFromConfiguration returns a logger.LoggingClient based on configuration settings.
+func FactoryFromConfiguration(config interfaces.Configuration) logger.LoggingClient {
+	var target string
+	bootstrapConfig := config.GetBootstrap()
+	if bootstrapConfig.Logging.EnableRemote {
+		target = bootstrapConfig.Clients["Logging"].Url() + clients.ApiLoggingRoute
+	} else {
+		target = bootstrapConfig.Logging.File
+	}
+	return logger.NewClient(clients.CoreDataServiceKey, bootstrapConfig.Logging.EnableRemote, target, config.GetLogLevel())
```
tsconn23: Same applies here w/r/t hard-coded service key

internal/pkg/bootstrap/bootstrap.go
michaelestrin: I wouldn't name it differently.  

The overall purpose of the package is to provide a bootstrap mechanism.  The public implementation is wholly contained with the bootstrap.go file.  That the implementation file shares the package name isn't an issue for me as the file isn't referenced by name anywhere.

I purposefully named the function entry point something other than "Bootstrap()" -- e.g. "Run()" -- so that we didn't end up with stutter in the code (i.e. the code reads bootstrap.Run() and not bootstrap.Bootstrap()).

internal/pkg/bootstrap/logging/factory.go
michaelestrin: You're right.  I missed converting this reference.

tsconn23:
Also notice the following -- All services are running via `make run`. I do a Ctrl-C to shut everything down and a panic occurs. I am on Mac OS.
```
level=INFO ts=2019-09-23T19:16:30.053107Z app=edgex-export-distro source=registrations.go:345 msg="received term signal"
level=WARN ts=2019-09-23T19:16:30.05312Z app=edgex-export-distro source=main.go:76 msg="terminating: interrupt"
make: *** [run] Interrupt: 2

panic: send on closed channel

goroutine 34 [running]:
os/signal.process(0x49e5340, 0xc0000c4a38)
	/usr/local/go/src/os/signal/signal.go:227 +0x189
os/signal.loop()
	/usr/local/go/src/os/signal/signal_unix.go:23 +0x3e
created by os/signal.init.0
	/usr/local/go/src/os/signal/signal_unix.go:29 +0x41
```
Unfortunately that's the extent of the trace.

internal/pkg/bootstrap/bootstrap.go
tsconn23: I see your point from a usage perspective. From a project perspective though we've tried to be consistent in also not having package/package.go.

internal/pkg/bootstrap/logging/factory.go
michaelestrin: You're right. I missed converting this reference.

michaelestrin:
Pushed new version with corrections for hardcoded service keys.  Good catch @tsconn23 !

internal/pkg/bootstrap/bootstrap.go
michaelestrin: No disrespect intended -- that seems like an arbitrary rule to me.  The Go blog defines stutter as:

> Avoid stutter. Since client code uses the package name as a prefix when referring to the package contents, the names for those contents need not repeat the package name. The HTTP server provided by the http package is called Server, not HTTPServer. Client code refers to this type as http.Server, so there is no ambiguity.

By that definition, this isn't stutter.  

I believe bootstrap.go is the appropriate name. 

I don't have an alternative naming suggestion.  Tell me what you'd like me to rename it as and I'll do so.

michaelestrin:
> I do a Ctrl-C to shut everything down and a panic occurs. I am on Mac OS.

We had a side conversation on this.  You confirmed this is reproducible.  You also confirmed this does not occur on the master branch.  Further, you confirmed that running and shutting down core-data by itself does not cause a panic.  

I develop on Windows and have not experienced a panic during shutdown.  I'm guessing there is a race condition somewhere that I'm missing and will investigate.

michaelestrin:
>I do a Ctrl-C to shut everything down and a panic occurs. I am on Mac OS.

I updated translateInterruptToCancel() to stop signal notifications to signalStream immediately before closing signalStream.  Given the error message -- `panic: send on closed channel` -- this seems the most likely cause of the problem (a signal is being received after signalStream is closed but before the application has exited).

I am unable to reproduce this issue locally and ask for your assistance in determining if this change corrects the issue for you. @tsconn23

tsconn23:
@michaelestrin 
>I am unable to reproduce this issue locally and ask for your assistance in determining if this change corrects the issue for you. @tsconn23

This fixed it

PR #1754, opened 2019-09-11
Sets snap version from VERSION file

knitzsche:
Uses adopt-info with new 'version' part.
Appends date and git info to version as previously.


knitzsche:
Closing this since snap build fails on issue 1755 (which has a PR).
I'll make a new PR based on the above PR but with set version changes for edinburgh track.

PR #1756, opened 2019-09-11
#1735 Populate global security configuration in registry

### LINKED ISSUE #1735: Config-Seed: Populate Global Security Configuration in Registry
Per discussions documented in the Core Working Group 15-Aug-2019, the mechanism by which security features will be enabled will affect security operations for the whole EdgeX platform. In other words, security is not a service by service setting. To that end, we need to have a global configuration area for security in Consul that services are aware of.

The current proposal is to have the following section written by the config-seed

`/edgex/security`

For this particular issue, the above section will contain one key `SECRET_STORE` the value of which will be either true/false. This value will be set by providing the following environment variable to the config-seed -- `edgex_security_secret_store`. This follows the convention already present in the config-seed of overriding values in Consul according to an environment variable that maps to a path.

Separate issues will follow for integrating each service with this global security configuration, as well as usage of environment variables should Consul not be in use.

jdharms:
Fixes #1735 

Adds mechanism for config-seed to read security configuration
environment variables and write them to the registry.

To test, run config-seed command with consul running like this:

`edgex_security_SecretStore="false" edgex_security_nested_Key="value" ./config-seed`

You should end up with two values in the consul KV storage (by default located at http://localhost:8500/ui/dc1/kv/edgex/security/1.0/ if running the consul dev agent locally)

Note: I've had some discussions with @tsconn23 and @michaelestrin and we think it would be good if the "global" security configuration settings were stored in a config file for Geneva, likely in a config-seed/res subdirectory.  This will provide some documentation of what settings are expected/available and sensible defaults.  This change will be backwards compatible to the solution presented in this PR, because environment variables will override the config file values.  An issue to track this work will be created when this is merged in.

### DIFF
--- a/cmd/config-seed/Attribution.txt
+++ b/cmd/config-seed/Attribution.txt
@@ -63,6 +63,9 @@ https://github.com/cenkalti/backoff/blob/master/LICENSE
 hashicorp/consul 1.1.0 (Mozilla Public License 2.0) - https://github.com/hashicorp/consul
 https://github.com/hashicorp/consul/blob/master/LICENSE
 
+Nested consul packages are not separately licensed:
+github.com/hashicorp/consul/api
+
 hashicorp/go-cleanhttp (Mozilla Public License 2.0) - https://github.com/hashicorp/go-cleanhttp
 https://github.com/hashicorp/go-cleanhttp/blob/master/LICENSE
 
--- a/cmd/config-seed/Attribution.txt
+++ b/cmd/config-seed/Attribution.txt
@@ -60,11 +60,8 @@ https://github.com/mitchellh/reflectwalk/blob/master/LICENSE
 cenkalti/backoff (MIT) https://github.com/cenkalti/backoff
 https://github.com/cenkalti/backoff/blob/master/LICENSE
 
-hashicorp/consul 1.1.0 (Mozilla Public License 2.0) - https://github.com/hashicorp/consul
+hashicorp/consul/api 1.1.0 (Mozilla Public License 2.0) - https://github.com/hashicorp/consul/api
 https://github.com/hashicorp/consul/blob/master/LICENSE
-
-Nested consul packages are not separately licensed:
-github.com/hashicorp/consul/api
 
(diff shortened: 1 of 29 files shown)


jdharms:
recheck

jdharms:
@lenny-intel @michaelestrin Now that go-mod-registry 0.1.12 has been tagged this PR is ready for review.

cmd/config-seed/Attribution.txt
```diff
@@ -63,6 +63,9 @@ https://github.com/cenkalti/backoff/blob/master/LICENSE
 hashicorp/consul 1.1.0 (Mozilla Public License 2.0) - https://github.com/hashicorp/consul
 https://github.com/hashicorp/consul/blob/master/LICENSE
 
+Nested consul packages are not separately licensed:
+github.com/hashicorp/consul/api
```
anonymouse64: please specify what license this package is under, not just the github link. See other attributions in the file for an example

jdharms: All of that information is already there.  The attribution checker just complains that the nested package consul/api isn't covered.  Please look at the lines above all of my attribution.txt changes, thanks

anonymouse64: ah could you file a bug for me to fix the attribution checker for that then? 

Also, does the checker pass if you delete the above line and just have the LICENSE for the nested api package? I seem to remember us switching to just using the api package so we should be able to delete the root level package from the checker.

jdharms: @anonymouse64 I think I've made the changes you suggested--can you check to make sure I understood you properly?

jdharms: I will file an issue right away

anonymouse64:
Attribution file looks good to me thanks

jdharms:
recheck

tsconn23:
@jdharms Can you please provide a sample command line utilizing the environment variable in the PR description for ease of testing?

jdharms:
@tsconn23 sure, done

Issue #1763, opened 2019-09-13
SMA GenClients usage not thread-safe


michaelestrin:
GenClients is an unprotected map updated in multiple locations (as a direct result of an HTTP request) and referenced elsewhere in the SMA.  Access to it should be protected by a mutex.

Issue #1742, opened 2019-09-06
SMA Configuration - Put the Actual Paths in the Config File


akramtexas:
Currently there are two fields in the SMA configuration.toml that have a default value which isn't usable. These are as follows:

```
ExecutorPath = '/path/to/the/file'
MetricsMechanism = 'custom/executor/direct-service'
```

Attempting to run the SMA without modifying the configuration to point to the actual executor or metrics mechanism results in an error. These should just point to the out-of-the-box locations so the user doesn't have to waste time troubleshooting something that isn't actually a problem.

Changes need to be made in both the `./res/configuration.toml` and `./res/docker/configuration.toml`

DEFAULT
```
ExecutorPath = '../sys-mgmt-executor/sys-mgmt-executor'
MetricsMechanism = 'direct-service'
```

DOCKER
```
ExecutorPath = '/sys-mgmt-executor'
MetricsMechanism = 'executor'
```

It may be necessary to utilize capabilities from Go's `path/filepath` package in order to make these paths interpretable by Windows.

PR #1769, opened 2019-09-16
SMA GenClients usage not thread-safe

michaelestrin:
GenClients is an unprotected map updated in multiple locations (as a
direct result of an HTTP request) and referenced elsewhere in the SMA.
Access to it should be protected by a mutex.

https://github.com/edgexfoundry/edgex-go/issues/1763

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,12 +6,11 @@ require (
 	github.com/OneOfOne/xxhash v1.2.5
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/eclipse/paho.mqtt.golang v1.1.1
-	github.com/edgexfoundry/go-mod-core-contracts v0.1.19
+	github.com/edgexfoundry/go-mod-core-contracts v0.1.25
 	github.com/edgexfoundry/go-mod-messaging v0.1.9
 	github.com/edgexfoundry/go-mod-registry v0.1.12
 	github.com/globalsign/mgo v0.0.0-20181015135952-eeefdecb41b8
 	github.com/go-kit/kit v0.8.0
-	github.com/go-stack/stack v1.8.0 // indirect
 	github.com/gomodule/redigo v2.0.0+incompatible
 	github.com/google/uuid v1.1.0
 	github.com/gorilla/mux v1.7.0
--- a/internal/pkg/startup/endpoint.go
+++ b/internal/pkg/startup/endpoint.go
@@ -20,7 +20,6 @@ import (
 	"time"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
-	registryTypes "github.com/edgexfoundry/go-mod-registry/pkg/types"
 	"github.com/edgexfoundry/go-mod-registry/registry"
 )
 
@@ -29,24 +28,32 @@ type Endpoint struct {
 }
 
 func (e Endpoint) Monitor(params types.EndpointParams, ch chan string) {
-	var endpoint registryTypes.ServiceEndpoint
-	var err error
 	for {
-
-		if e.RegistryClient != nil {
-			endpoint, err = (*e.RegistryClient).GetServiceEndpoint(params.ServiceKey)
-			if err != nil {
-				err = fmt.Errorf("unable to get Service endpoint for %s: %s", params.ServiceKey, err.Error())
-			}
-		} else {
-			err = fmt.Errorf("unable to get Service endpoint for %s: Registry client is nil", params.ServiceKey)
-		}
-
+		url, err := e.buildURL(params)
 		if err != nil {
 			fmt.Fprintln(os.Stdout, err.Error())
 		}
-		url := fmt.Sprintf("http://%s:%v%s", endpoint.Host, endpoint.Port, params.Path)
 		ch <- url
 		time.Sleep(time.Millisecond * time.Duration(params.Interval))
+	}
+}
+
+func (e Endpoint) Fetch(params types.EndpointParams) string {
+	url, err := e.buildURL(params)
+	if err != nil {
+		fmt.Fprintln(os.Stdout, err.Error())
+	}
+	return url
+}
+
+func (e Endpoint) buildURL(params types.EndpointParams) (string, error) {
+	if e.RegistryClient != nil {
+		endpoint, err := (*e.RegistryClient).GetServiceEndpoint(params.ServiceKey)
+		if err != nil {
+			return "", fmt.Errorf("unable to get Service endpoint for %s: %s", params.ServiceKey, err.Error())
+		}
+		return fmt.Sprintf("http://%s:%v%s", endpoint.Host, endpoint.Port, params.Path), nil
+	} else {
+		return "", fmt.Errorf("unable to get Service endpoint for %s: Registry client is nil", params.ServiceKey)
 	}
 }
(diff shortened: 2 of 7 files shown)


anonymouse64:
I don't have time to review this unfortunately

michaelestrin:
Rebased.

michaelestrin:
Rebased.

tsconn23:
@michaelestrin There may be a bug in here. I was testing this PR locally without Consul using the following GET request
```
http://localhost:48090/api/v1/config/edgex-core-data
```
In the master branch, I can see the following entry in the log as a result
```
level=INFO ts=2019-09-17T21:47:09.931Z app=edgex-sys-mgmt-agent source=services.go:111 msg="service edgex-core-data is known to SMA as being in the ready-made list of clients"
```
Using this PR, I see the following
```
level=INFO ts=2019-09-17T21:42:28.193011Z app=edgex-sys-mgmt-agent source=services.go:55 msg="service edgex-core-data not known to SMA as being in the ready-made list of clients"
```
What I would expect in this scenario is that the edgex-core-data client defined in the configuration.toml would be utilized to fetch the metrics. I should note that in both cases, a panic occurs after the message is logged.

FYI, and related to [this issue](https://github.com/edgexfoundry/edgex-go/issues/1742), I have changed the following two entries in my configuration.toml to facilitate local testing.
```
ExecutorPath = '../sys-mgmt-executor/sys-mgmt-executor'
MetricsMechanism = 'direct-service'
```

michaelestrin:
@tsconn23 Appreciate the review and detailed feedback.

It didn't panic for me because I added the explicitly named edgex-core-data service to configuration.toml as follows:

```
  [Clients.edgex-core-data]
  Protocol = 'http'
  Host = 'localhost'
  Port = 48080
```

It does panic when I remove that configuration.toml entry.  _Thank you for catching this!_

Apparently, getConfig() (edgexfoundry/edgex-go/internal/system/agent/services.go:35) assumes the registry flag is always set (similar to getHealth and https://github.com/edgexfoundry/edgex-go/issues/1770).

In lieu of creating a separate issue, I added a check to ensure registryClient is not nil before attempting to dereference it and to return an error message instead of causing a panic.

tsconn23:
@michaelestrin Retesting. For the same scenario above using this PR I receive the following response.
```
GET http://localhost:48090/api/v1/config/edgex-core-data

{"configuration":{"edgex-core-data":"registry is required to obtain configuration."}}
```
It should make the call to core-data using the client information in the configuration.toml and return the service's config.

akramtexas:
> @tsconn23 Appreciate the review and detailed feedback.
> 
> It didn't panic for me because I added the explicitly named edgex-core-data service to configuration.toml as follows:
> 
> ```
>   [Clients.edgex-core-data]
>   Protocol = 'http'
>   Host = 'localhost'
>   Port = 48080
> ```
> 
> It does panic when I remove that configuration.toml entry. _Thank you for catching this!_
> 
> Apparently, getConfig() (edgexfoundry/edgex-go/internal/system/agent/services.go:35) assumes the registry flag is always set (similar to getHealth and #1770).
> 
> In lieu of creating a separate issue, I added a check to ensure registryClient is not nil before attempting to dereference it and to return an error message instead of causing a panic.

Yes, while doing separate testing (with Dockerized services, I can confirm the (implicit) assumption has been made that the `registry flag` is always set for getHealth(). Thus, I got this response to a GET call:
```
{
    "device-virtual": "registry is required to obtain service health status.",
    "edgex-core-command": "registry is required to obtain service health status.",
    "edgex-core-data": "registry is required to obtain service health status.",
    "edgex-core-metadata": "registry is required to obtain service health status.",
    "edgex-export-client": "registry is required to obtain service health status.",
    "edgex-export-distro": "registry is required to obtain service health status.",
    "edgex-support-logging": "registry is required to obtain service health status.",
    "edgex-support-notifications": "registry is required to obtain service health status.",
    "edgex-support-scheduler": "registry is required to obtain service health status."
}
```

tsconn23:
@akramtexas @michaelestrin 
>I can confirm the (implicit) assumption has been made that the registry flag is always set for getHealth().

It doesn't work this way in the Edinburgh release, so this is a breaking change and has to be addressed. To confirm, I just got latest on the Edinburgh branch and re-ran the testing scenario above. Running all of the services outside of containers, i.e. natively, I am able to obtain a populated configuration response.
Screen Shot 2019-09-18 at 3 06 16 PM

michaelestrin:
The implementation I inherited did not include calls to the individual services to obtain configuration information when the registry flag isn't set.  I'll need to go back and compare master against edinburgh to determine what changed and restore the missing implementation.  @tsconn23 

As for the health check, the error message being returned is a very recent change (https://github.com/edgexfoundry/edgex-go/issues/1770) intended to prevent a panic.  The implementation I inherited did not include calls to the individual services to obtain health information.  @akramtexas 

I'll revisit this tomorrow. :)

michaelestrin:
I agree there is a backwards-breaking change in master whereby the SMA -- when started without the --registry flag -- should communicate with a service directly to retrieve and return it's configuration information but does not.  I do not believe correcting that change -- which exists in master prior to this PR's change set -- is within the scope of this PR's issue (which is to correct the unsafe access of GenClients).  I have created a [separate issue](https://github.com/edgexfoundry/edgex-go/issues/1780) to address the backwards-breaking change in master.  

I do not agree that there is a backwards-breaking change in master relative to the health check.  Review of the [getHealth() implementation](https://github.com/edgexfoundry/edgex-go/blob/b4a4fa47c55b1800914bcdfcd3ef6ebcd814ee62/internal/system/agent/services.go#L257) in Edinburgh confirms the code assumes the registry is available.  While this was [recently fixed](https://github.com/edgexfoundry/edgex-go/issues/1770) to avoid a panic when the --registry flag was not set, the correction is probably not the ultimately desired behavior.  It's not clear to me what behavior we might want to substitute and this topic should be brought up in the next system management working group meeting.  Regardless, I believe correcting this is beyond the scope of this PR's issue.

**My ask is that this PR be approved as-is.**

@tsconn23 @jpwhitemn

michaelestrin:
Rebased.

tsconn23:
recheck

jpwhitemn:
Catching up on this... so @tsconn23 you found another bug but should we consider the original change request resolved for now, accept this PR, submit a new issue for the new bug, and work to fix your new found bug?

michaelestrin:
I'm able to duplicate the issue @tsconn23 reported above. Investigating...

michaelestrin:
... @tsconn23's issue is a race in go-mod-core-contract's GeneralClient implementation.  Correcting it is outside the scope of this issue.  I've created a new issue in go-mod-core-contract -- https://github.com/edgexfoundry/go-mod-core-contracts/issues/150

michaelestrin:
Rebased.

michaelestrin:
Rebased.

michaelestrin:
Integrated changes to internal/pkg/startup/endpoint.go provided by @tsconn23 and updated go.mod version reference for go-mod-core-contracts to coordinate with his changes to address https://github.com/edgexfoundry/go-mod-core-contracts/issues/150.

michaelestrin:
recheck

michaelestrin:
recheck

Issue #1771, opened 2019-09-16
SMA Health Check should not be limited to "known" services


michaelestrin:
getHealth() in internal/system/agent/services.go relies on the IsKnownServiceKey() function to determine whether it's possible to query the registry to determine a service's health.  

IsKnownServiceKey() should be eliminated.  The registry should be queried for the health of any specified service.

michaelestrin:
Also dependent on https://github.com/edgexfoundry/edgex-go/issues/1763 merge.

anonymouse64:
Related https://github.com/edgexfoundry/edgex-go/issues/1662

Issue #1770, opened 2019-09-16
SMA /health faults if --registry not set


michaelestrin:
getHealth() in internal/system/agent/services.go relies on registryClient being initialized.  registryClient is only initialized when the service is started with the --registry flag.

PR #1772, opened 2019-09-16
SMA /health faults if --registry not set

michaelestrin:
Updated to return error message when --registry flag not set during
service start up.  Expect this code to be further refactored in
follow-on https://github.com/edgexfoundry/edgex-go/issues/1771.

https://github.com/edgexfoundry/edgex-go/issues/1770


michaelestrin:
Rebased.

PR #1779, opened 2019-09-18
Removing dependency for a Device/Service List and use Configuration

### LINKED ISSUE #1662: Look at Driving "Default Services List" via Configuration
From https://github.com/edgexfoundry/edgex-go/pull/1627#discussion_r311156531

Both the SMA and the security-proxy-setup service enumerate a set of "default services", services which are expected to be up and running in any EdgeX installation. Currently that list of services is defined and exposed [here](https://github.com/edgexfoundry/edgex-go/blob/3c8972b0ec460de0184d9656babf814c76f3b501/internal/pkg/config/types.go#L22).

In the case of the security-proxy-setup service, however, there are two additional services called out.
```
services["edgex-support-rulesengine"] = "Rulesengine"
services["edgex-device-virtual"] = "Virtualdevice"
```
In particular the virtual device service may not be running in a Production installation of EdgeX. It has been proposed that we review this mechanism to determine if it can be driven via configuration.

xmlviking:
Fix #1662

- Modified the security-proxy-setup configuration.toml file(s) docker and local
- Removed the use of the go-mod-core-contracts use of the ServiceList "statically defined"
- Removed the use of the ClientInfo struct and included the new ServiceInfo struct "now featured in the toml config"
- Modified the security/proxy/service.go init func simplifying the KongRoute(s) creational process
- Updated any required helper func(s) as required for the change from ClientInfo to ServiceInfo

### DIFF
--- a/cmd/security-proxy-setup/res/configuration.toml
+++ b/cmd/security-proxy-setup/res/configuration.toml
@@ -91,12 +91,12 @@ SNIS = ["edgex-kong"]
   Host = "localhost"
   Port = 48085
 
-  [Clients.Rulesengine]
+  [Clients.RulesEngine]
   Protocol = "http"
   Host = "localhost"
   Port = 48075
 	
-  [Clients.Virtualdevice]
+  [Clients.VirtualDevice]
   Protocol = "http"
   Host = "localhost"
   Port = 49990
--- a/cmd/security-proxy-setup/res/docker/configuration.toml
+++ b/cmd/security-proxy-setup/res/docker/configuration.toml
@@ -91,12 +91,12 @@ SNIS = ["edgex-kong"]
   Host = "edgex-support-scheduler"
   Port = 48085
 
-  [Clients.Rulesengine]
+  [Clients.RulesEngine]
   Protocol = "http"
   Host = "edgex-support-rulesengine"
   Port = 48075
 
-  [Clients.Virtualdevice]
+  [Clients.VirtualDevice]
   Protocol = "http"
   Host = "edgex-device-virtual"
   Port = 49990
--- a/internal/security/proxy/service.go
+++ b/internal/security/proxy/service.go
@@ -26,7 +26,6 @@ import (
 	"strings"
 
 	"github.com/edgexfoundry/edgex-go/internal"
-	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 )
 
@@ -96,23 +95,12 @@ func (s *Service) Init(cert CertificateLoader) error {
 		return err
 	}
 
-	services := config.ListDefaultServices()
-	//TODO: There are specific other services being routed through Kong. This should probably be driven through configuration
-	//      For now, I'm putting this inline
-	services["edgex-support-rulesengine"] = "Rulesengine"
-	services["edgex-device-virtual"] = "Virtualdevice"
-
-	for _, name := range services {
-		params := Configuration.Clients[name]
-		if params.Host == "" {
-			LoggingClient.Warn("missing config information for service: " + name)
-			continue
-		}
+	for clientName, client := range Configuration.Clients {
 		serviceParams := &KongService{
-			Name:     strings.ToLower(name),
-			Host:     params.Host,
-			Port:     params.Port,
-			Protocol: params.Protocol,
+			Name:     strings.ToLower(clientName),
+			Host:     client.Host,
+			Port:     client.Port,
+			Protocol: client.Protocol,
 		}
 
 		err := s.initKongService(serviceParams)
@@ -121,10 +109,10 @@ func (s *Service) Init(cert CertificateLoader) error {
 		}
 
 		routeParams := &KongRoute{
-			Paths: []string{"/" + strings.ToLower(name)},
-			Name:  strings.ToLower(name),
+			Paths: []string{"/" + strings.ToLower(clientName)},
+			Name:  strings.ToLower(clientName),
 		}
-		err = s.initKongRoutes(routeParams, strings.ToLower(name))
+		err = s.initKongRoutes(routeParams, strings.ToLower(clientName))
 		if err != nil {
 			return err
 		}
--- a/cmd/security-proxy-setup/main.go
+++ b/cmd/security-proxy-setup/main.go
@@ -66,6 +66,7 @@ func main() {
 	startup.Bootstrap(params, proxy.Retry, logBeforeInit)
 
 	req := proxy.NewRequestor(insecureSkipVerify, proxy.Configuration.Writable.RequestTimeout)
+
 	s := proxy.NewService(req)
 
 	err := s.CheckProxyServiceStatus()
@@ -80,7 +81,7 @@ func main() {
 			os.Exit(1)
 		}
 		cert := proxy.NewCertificateLoader(req, proxy.Configuration.SecretService.CertPath, proxy.Configuration.SecretService.TokenPath)
-		err = s.Init(cert)
+		err = s.Init(cert) // Where the Service init is called
 		if err != nil {
 			proxy.LoggingClient.Error(err.Error())
 			os.Exit(1)


cmd/security-proxy-setup/res/docker/configuration.toml
```diff
@@ -66,37 +66,63 @@ SNIS = ["edgex-kong"]
   Host = "edgex-core-command"
   Port = 48082
 
-  [Clients.Notifications]
+[Services]
+  [Services.Coredata]
+  Name = "Coredata"
   Protocol = "http"
-  Host = "edgex-support-notifications"
+  Host = "localhost"
```
anonymouse64: this seems wrong for the docker configuration.toml file, shouldn't it be the name of the docker container to use as the hostname?

internal/security/proxy/service.go
```diff
@@ -96,23 +95,13 @@ func (s *Service) Init(cert CertificateLoader) error {
...
-	services["edgex-support-rulesengine"] = "Rulesengine"
-	services["edgex-device-virtual"] = "Virtualdevice"
-
-	for _, name := range services {
-		params := Configuration.Clients[name]
-		if params.Host == "" {
-			LoggingClient.Warn("missing config information for service: " + name)
-			continue
-		}
+	// EFC issue_1662
```
anonymouse64: what is the purpose of this comment? if this is just about the fact that we added this function for this issue, I don't think that information belongs in the code, I think a better place is in the git commit message

cmd/security-proxy-setup/res/docker/configuration.toml
tsconn23: I called this out separately via Slack. He's on it.

xmlviking: @anonymouse64 , @tsconn23 - Ty for the callout...fixed and updated the PR.

internal/security/proxy/service.go
xmlviking: @anonymouse64 - removed this as well.  I had this in there for my own analysis and debugging, testing.  Removed :)

internal/security/proxy/config.go
```diff
@@ -45,6 +45,30 @@ type KongUrlInfo struct {
...
+	Name string
+	// Host is the hostname or IP address of a service.
+	Host string
+	// Port defines the port on which to access a given service
+	Port int
+	// Protocol indicates the protocol to use when accessing a given service
+	Protocol string
+}
+
+func (cfg *ConfigurationStruct) GetServices() map[string]ServiceInfo {
```
anonymouse64: also since these functions are exported, they should have a short doc comment. even though they are in the internal package, having doc comments is still a really useful thing to have

internal/security/proxy/config.go
```diff
@@ -45,6 +45,30 @@ type KongUrlInfo struct {
...
+	return cfg.Services
+}
+
+func (s ServiceInfo) Url() string {
```
anonymouse64: ditto about doc comment

internal/security/proxy/config.go
xmlviking: Added comment :)

internal/security/proxy/config.go
xmlviking: Added comment on the Url() func

xmlviking:
Awaiting Tinguz approval before merge :)

tingyuz:
Let me run the test to make sure the feature is working as expected then will approve.

internal/security/proxy/service.go
```diff
@@ -96,23 +95,12 @@ func (s *Service) Init(cert CertificateLoader) error {
...
-			LoggingClient.Warn("missing config information for service: " + name)
-			continue
-		}
+	for _, svc := range Configuration.GetServices() {
 		serviceParams := &KongService{
-			Name:     strings.ToLower(name),
-			Host:     params.Host,
-			Port:     params.Port,
-			Protocol: params.Protocol,
+			Name:     svc.Name,
```
tingyuz: The name here needs to be lower case as well so use strings.ToLower(svc.Name) instead.

cmd/security-proxy-setup/res/configuration.toml
```diff
@@ -50,53 +50,63 @@ TokenPath = "res/resp-init.json"
 CACertPath = "res/EdgeXFoundryCA/EdgeXFoundryCA.pem"
 SNIS = ["edgex-kong"]
 
-[Clients]
-  [Clients.Coredata]
+[Services]
```
michaelestrin: Why change from Clients to Services?

michaelestrin:
> Finally, I think it would be good to get @michaelestrin to look at this PR and comment, since this is solving what is essentially a very similar problem to what the SMA has from #1771 but there only the registry is queried and there's no additional configuration in the configuration.toml file whereas this implementation has the requirement to be enumerated in the configuration.toml and not go to consul. 

The changes proposed in #1763 (upon which #1771 is built) remove the population of GenClients using the result from config.ListDefaultServices() when the --registry flag is set.  This follows the assumption that when the --registry flag is set, the registry is the source of truth for configuration.  It also assumes there is little to no benefit to front-load the population of GenClients since the implementation handles population on-the-fly when a requested service does not yet have a corresponding GenClients' entry.

When the --registry flag is not set, GenClients is populated from the information contained in the configuration.toml file.

These changes appear to be specific to the security-proxy-setup service.  It's not clear to me when that service executes relative to the other system services, whether or not the registry would ever be up and running, and whether or not the registry would contain configuration specific to the security-proxy-setup service.  If someone could provide clarity,  I would be able to comment further.

anonymouse64:
> whether or not the registry would ever be up 

yes consul needs to be up by the time security-proxy-setup starts because security-proxy-setup specifically will only be started after Kong is ready in consul and security-secrets-setup is ready in consul (or at least this will be enforced for Fuji)

> whether or not the registry would contain configuration specific to the security-proxy-setup service

yes it could contain configuration.toml information

michaelestrin:
Based on clarification from @anonymouse64 above, I concur the implementation here for retrieving and populating configuration should follow the pattern established elsewhere.  That is, configuration should be retrieved from the registry (if the --registry flag is set) or from the service's configuration.toml (if the --registry flag is not set).

This implies config-seed will populate the registry with the necessary configuration for the security-proxy-setup service.  If it doesn't currently, I would consider those changes in-scope for the parent issue (i.e. they should be in a PR -- this one or another -- tied to the parent issue).

@xmlviking

cmd/security-proxy-setup/res/configuration.toml
```diff
@@ -50,53 +50,63 @@ TokenPath = "res/resp-init.json"
...
+  [Services.Coredata]
+  Name = "Coredata"
```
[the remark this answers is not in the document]
xmlviking: Its used for a unique identification purpose.  It's used in security api as well.

cmd/security-proxy-setup/res/configuration.toml
xmlviking: Due to the comment mentioned earlier.  For identification purposes.  ClientInfo does not provide the capability to offer the unique struct instance name.  Rather then modifying the ClientInfo which is uses in numerous places I chose to use what the existing Security services API uses which is a custom type.   This type "struct" is nearly identical in shape to the KongType.

xmlviking:
> Based on clarification from @anonymouse64 above, I concur the implementation here for retrieving and populating configuration should follow the pattern established elsewhere. That is, configuration should be retrieved from the registry (if the --registry flag is set) or from the service's configuration.toml (if the --registry flag is not set).
> 
> This implies config-seed will populate the registry with the necessary configuration for the security-proxy-setup service. If it doesn't currently, I would consider those changes in-scope for the parent issue (i.e. they should be in a PR -- this one or another -- tied to the parent issue).
> 
> @xmlviking

The timing on this will not work as I understand the security and initialization process works at this time. The service registry will not be populated prior to this running.  There will be nothing for it to use.   This is an ask based on the current state of the codebase.

cmd/security-proxy-setup/res/configuration.toml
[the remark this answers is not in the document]
michaelestrin: Why not use the name embedded in the encompassing toml chapter following the convention established by prior implementations?

cmd/security-proxy-setup/res/configuration.toml
michaelestrin: I don't consider this resolved until I understand your answer to my question.  What "comment mentioned earlier" are you referring to?

michaelestrin:
> The timing on this will not work as I understand the security and initialization process works at this time. The service registry will not be populated prior to this running. There will be nothing for it to use. This is an ask based on the current state of the codebase.

I'm confused.  This information appears to conflict with @anonymouse64's above.  Are you saying config-seed won't be executed prior to the security-proxy-setup service?

cmd/security-proxy-setup/res/configuration.toml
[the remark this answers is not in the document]
michaelestrin: @xmlviking please do not mark comments resolved.  Allow the commenter to mark them resolved once their questions or concerns are satisfactorily addressed.  I responded to your comment here and returned shortly afterwards to find you had marked this thread as resolved.

cmd/security-proxy-setup/res/configuration.toml
xmlviking: @michaelestrin sorry I was referring to the previous comment where i mentioned the use of the ServiceInfo abstraction struct previously in the security api's.  https://github.com/edgexfoundry/security-api-gateway/blob/master/internal/pkg/edgexproxy/service.go .  In the security api there is the use of the KongService struct.  I took the similar approach and used the same pattern used in that API.   The underlying symptom here is there is no logical name for any of the given afore mentioned clientInfo toml configuration entries.  The pattern used in the Security API is the inclusion of the logical name.  We need the name of the service so we can use that to bind not only client but service call.  Rather then including a seperate list of unique names in yet another configuration struct, the pattern reuses the "name" for both those entries.

michaelestrin: >there is no logical name for any of the given afore mentioned clientInfo toml configuration entries.

There is -- the service's name is embedded in the encompassing toml chapter name.  If we depart from that convention, we should do so consistently throughout all of the services.

I feel like I'm missing something.

anonymouse64:
I always assumed that config-seed would be taught to import configuration for the security services but perhaps this was just in my head...

Regardless, it seems like a decision is needed as to whether or not config-seed will import the configuration.toml's into consul for the security services. The only reason I think we couldn't use config-seed before the security services is if config-seed needs to access secrets itself. I think it doesn't need to, but I might be wrong. 
If config-seed doesn't need secrets, then we can order config-seed to run before all of the security services and then security-proxy-setup can talk to Consul for this information. Note however that security-proxy-setup _cannot_ use the health mechanism in consul for getting service config because potentially none of the services in edgex-go other than config-seed and the other security services would have run before security-proxy-setup is done, however accessing the config for other services should be fine even if the service is offline.

@tingyuz and @tsconn23 thoughts on config-seed importing configuration.tomls for the security services into consul and whether or not config-seed needs access to secrets itself?

tsconn23:
>The only reason I think we couldn't use config-seed before the security services is if config-seed needs to access secrets itself. I think it doesn't need to, but I might be wrong.

This is correct. The secrets will effectively override the default username/password info contained in the configuration.toml. So it makes sense to me that config-seed would run as early as possible in the pipeline.

anonymouse64:
Thanks for clarification @tsconn23! 
I don't think this PR is updated to look through the registry for configuration, it seems to still only enumerate the stuff from configuration.toml... Perhaps it hasn't been pushed up here yet?

xmlviking:
> Thanks for clarification @tsconn23!
> I don't think this PR is updated to look through the registry for configuration, it seems to still only enumerate the stuff from configuration.toml... Perhaps it hasn't been pushed up here yet?

Ian,

I'm working though the necessary changes which include the use of useRegistry flag during the init process for the security-proxy-setup service.  This change is a bit more invasive then the simply changes to the PR work previously.  I will get the PR updated accordingly.

michaelestrin:
The corresponding issue grew to include obtaining configuration from the registry. 

Instead of implementing registry support from scratch, it may make more sense to undertake https://github.com/edgexfoundry/edgex-go/issues/1805 first.  Completion thereof would provide standardized registry support.

@xmlviking

cmd/security-proxy-setup/res/configuration.toml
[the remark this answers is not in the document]
xmlviking: I have updated configuration.tom for both docker and local configuration to use the previous pattern for ClientInfo.

internal/security/proxy/service.go
xmlviking: Tingyuz - updated the Name entry with the inclusion of the following:
              Name:     strings.ToLower(clientName),
I have modified the variables to more accurately reflect the usage of the Clients and Name attribute.

michaelestrin:
> The corresponding issue grew to include obtaining configuration from the registry.
> 
> Instead of implementing registry support from scratch, it may make more sense to undertake #1805 first. Completion thereof would provide standardized registry support.

In conversation with @tsconn23, @tingyuz, and @AnthonyMBonafide yesterday, agreed to put #1805 on hold until after October Fuji freeze.  This issue should be worked independently of #1805.

internal/security/proxy/service.go
```diff
@@ -96,23 +95,12 @@ func (s *Service) Init(cert CertificateLoader) error {
...
-	services["edgex-support-rulesengine"] = "Rulesengine"
-	services["edgex-device-virtual"] = "Virtualdevice"
-
-	for _, name := range services {
-		params := Configuration.Clients[name]
-		if params.Host == "" {
-			LoggingClient.Warn("missing config information for service: " + name)
-			continue
-		}
+	for clientName, client := range Configuration.Clients {
```
michaelestrin: @anonymouse64 -- are you OK with this implementation?

It walks the clients defined in configuration and establishes a route for each of them in Kong. Prior implementation used a standard list of services -- a warning was logged if one of the standard services was not defined in configuration but execution was allowed to proceed.

Updated implementation drops the standard service list and corresponding warning. Seems more flexible to me in that we can define -- via configuration -- additional services outside the standard set (if that was desirable).

internal/security/proxy/service.go
xmlviking: @tingyuz am I ok to resolve this conflict or do you still need further review on this change?

xmlviking:
--recheck

JPWKU:
recheck

PR #1794, opened 2019-09-20
Add [cache] subcommand from merging pki-init option cache into security-secrets-setup

jim-wang-intel:
- Cache directory is defined in environment variable $**PKI_CACHE**, and the default is `/etc/edgex/pki` if not defined
- If cache directory is empty, then TLS assets will be generated and then cached
- TLS assets will be deployed from cache directory into deploy directory which defaults to `run/edgex/secrets` tmpfs directory
- CA private key is always securely erased before cached

Example to run: ``` ./security-secrets-setup -confdir ./res cache ```

Related to: 
- Issue #1635
- Issue #1558 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,10 +18,10 @@ cmd/core-data/core-data
 cmd/core-metadata/core-metadata
 cmd/export-client/export-client
 cmd/export-distro/export-distro
-cmd/security-secrets-setup/security-secrets-setup
-cmd/security-secrets-setup/config/*
-cmd/security-secretstore-setup/security-secretstore-setup
 cmd/security-proxy-setup/security-proxy-setup
+cmd/security-secrets-setup/config/*
+cmd/security-secrets-setup/security-secrets-setup
+cmd/security-secretstore-setup/security-secretstore-setup
 cmd/support-logging/support-logging
 cmd/support-notifications/support-notifications
 cmd/support-scheduler/support-scheduler
--- a/internal/security/setup/option/cache.go
+++ b/internal/security/setup/option/cache.go
@@ -1,0 +1,94 @@
+//
+// Copyright (c) 2019 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License. You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distributed under the License
+// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+// or implied. See the License for the specific language governing permissions and limitations under
+// the License.
+//
+// SPDX-License-Identifier: Apache-2.0'
+//
+
+package option
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/edgexfoundry/edgex-go/internal/security/setup"
+)
+
+// Cache option....
+func Cache() func(*PkiInitOption) (exitCode, error) {
+	return func(pkiInitOpton *PkiInitOption) (exitCode, error) {
+
+		if isCacheNoOp(pkiInitOpton) {
+			return normal, nil
+		}
+
+		return cachePkis()
+	}
+}
+
+func isCacheNoOp(pkiInitOption *PkiInitOption) bool {
+	// nop: if the flag is missing or not on
+	return pkiInitOption == nil || !pkiInitOption.CacheOpt
+}
+
+func cachePkis() (statusCode exitCode, err error) {
+	// generate a new one if pkicache dir is empty
+	pkiCacheDir := getPkiCacheDirEnv()
+	cachedCAPrivateKeyFile := filepath.Join(pkiCacheDir, caServiceName, tlsSecretFileName)
+	if empty, err := isDirEmpty(pkiCacheDir); err != nil {
+		return exitWithError, err
+	} else if empty {
+		setup.LoggingClient.Info(fmt.Sprintf("cache dir %s is empty, generating PKI into it...", pkiCacheDir))
+		// perform generate func
+		statusCode, err = generatePkis()
+		if err != nil {
+			return statusCode, err
+		}
+
+		generatedDirPath := filepath.Join(os.Getenv(envXdgRuntimeDir), pkiInitGeneratedDir)
+
+		// always shreds CA private key before cache
+		caPrivateKeyFile := filepath.Join(generatedDirPath, caServiceName, tlsSecretFileName)
+		if err := secureEraseFile(caPrivateKeyFile); err != nil {
+			return exitWithError, err
+		}
+
+		if err = doCache(generatedDirPath); err != nil {
+			return exitWithError, err
+		}
+	} else {
+		// cache dir is not empty: output error message if CA private key is present
+		// when cache is given
+		if checkIfFileExists(cachedCAPrivateKeyFile) {
+			return exitWithError, errCacheNotChangeAfter
+		}
+		setup.LoggingClient.Info(fmt.Sprintf("cached TLS assets from dir %s is present, using cached PKI", pkiCacheDir))
+	}
+
+	// to deploy
+	// copy stuff into dest dir from pkiCache
+	err = deploy(pkiCacheDir, pkiInitDeployDir)
+	if err != nil {
+		return exitWithError, err
+	}
+
+	return normal, nil
+}
+
+func doCache(fromDir string) error {
+	// destination
+	pkiCacheDir := getPkiCacheDirEnv()
+
+	// to cache
+	return copyDir(fromDir, pkiCacheDir)
+}
--- a/internal/security/setup/option/cache.go
+++ b/internal/security/setup/option/cache.go
@@ -24,7 +24,8 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/security/setup"
 )
 
-// Cache option....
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -18,10 +18,10 @@ cmd/core-data/core-data
 cmd/core-metadata/core-metadata
 cmd/export-client/export-client
 cmd/export-distro/export-distro
-cmd/security-secrets-setup/security-secrets-setup
+cmd/security-proxy-setup/security-proxy-setup
 cmd/security-secrets-setup/config/*
+cmd/security-secrets-setup/security-secrets-setup
 cmd/security-secretstore-setup/security-secretstore-setup
-cmd/security-proxy-setup/security-proxy-setup
 cmd/support-logging/support-logging
```
jim-wang-intel: just try to keep alphabetical ordering here

internal/security/setup/option/constant.go
```diff
@@ -31,6 +32,8 @@ const (
 	envXdgRuntimeDir              = "XDG_RUNTIME_DIR"
 	defaultXdgRuntimeDir          = "/tmp"
 	pkiInitBaseDir                = "/edgex/security-secrets-setup"
+	envPkiCache                   = "PKI_CACHE"
+	defaultPkiCacheDir            = "/etc/edgex/pki"
 	tmpfsRunDir                   = "/run"
 	tlsSecretFileName             = "server.key"
```
tingyuz: Would this be a problem if we decide to run the pki-init native in the future ( E.g, running the binary in the windows machine) ?

bnevis-i: > Would this be a problem if we decide to run the pki-init native in the future

@tingyuz Please elaborate on what you mean by "this".  Path names obviously need to change on Windows. The PKI_CACHE env var will be replaced in a future PR with configuration.toml.  On Windows platforms, enabling BitLocker would help protect TLS assets when the system is powered-off.

tingyuz: Hard-coded path is not flexible. It's good if we have plan to move it into configuration.toml or other places so that it becomes configurable. I think we are good here.

internal/security/setup/option/pkiOption_test.go
```diff
@@ -44,21 +44,37 @@ func TestNewPkiInitOption_GenerateOnly(t *testing.T) {
...
+func TestNewPkiInitOption_CacheAndGenerate(t *testing.T) {
+	assert := assert.New(t)
+
+	options := PkiInitOption{
+		GenerateOpt: true,
+		CacheOpt:    true,
+	}
+	optionsExecutor, status, err := NewPkiInitOption(options)
+	assert.Empty(optionsExecutor)
+	assert.Equal(exitWithError.intValue(), status)
```
anonymouse64: this is another instance where I don't see the value from the `exitCode` type since you just end up comparing it with an int anyways, but this is fine for now

internal/security/setup/option/cache_test.go
```diff
@@ -0,0 +1,209 @@
...
+	_ = createDirectoryIfNotExists(pkiCacheDir)
+
+	origDeployDir := pkiInitDeployDir
+	tempDir, tempDirErr := ioutil.TempDir(curDir, "deploytest")
+	if tempDirErr != nil {
+		t.Fatalf("cannot create temporary scratch directory for the test: %v", tempDirErr)
+	}
+	pkiInitDeployDir = tempDir
+
+	return func(t *testing.T) {
```
anonymouse64: what's the purpose of this function taking `testing.T` as an argument? It doesn't seem to be used currently.

The `testing.T` that would normally be used would be the one that's passed here to the setup function so you could just use `t` here if you needed it, with go creating a closure around the `t` value for this function that's returned.

internal/security/setup/option/cache_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+	deployEmpty, emptyErr := isDirEmpty(pkiInitDeployDir)
+	assert.Nil(emptyErr)
+	assert.False(deployEmpty)
+}
+
+func TestCacheOff(t *testing.T) {
+	vaultJSONPkiSetupExist = true
+	kongJSONPkiSetupExist = true
+	tearDown := setupCacheTest(t)
```
anonymouse64: is there a reason not to just do 
```suggestion
	defer setupCacheTest(t)()
```

or if you do really need to be able to pass in `t` to the function returned:

```suggestion
	defer setupCacheTest(t)(t)
```

internal/security/setup/option/cache_test.go
```diff
@@ -0,0 +1,209 @@
...
+	defer tearDown(t)
+
+	options := PkiInitOption{
+		CacheOpt: true,
+	}
+	cacheOn, _, _ := NewPkiInitOption(options)
+	cacheOn.(*PkiInitOption).executor = testExecutor
```
anonymouse64: this bit feels wrong to me, is there not another way for us to override the executor for the test with our mocked one? I just think it's in bad taste for a test to need to type cast things that were returned from an exported function for it to be used later on. I think a better way would be for the package to export some way to mutate the thing that we want to overwrite like this

anonymouse64: also do we already have other examples of this in the codebase?

jim-wang-intel: since the `NewPkiInitOption()` returned is an interface, the only way i can think to avoid the type cast is to add "setter" of executor into interface for overriding the executor.

internal/security/setup/option/cache_test.go
jim-wang-intel: I think it is more readable if we call that variable name out, ie. `tearDown` instead of inline coding.

internal/security/setup/option/cache_test.go
jim-wang-intel: good catch, can remove `testing.T`.  The only point to keep that is to be more symmetric interface signature with setup(t). but yeah, will remove it from this

internal/security/setup/option/generate_test.go
```diff
@@ -13,6 +13,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0'
 //
+
```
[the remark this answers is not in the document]
bnevis-i: I don't think these are cosmetic.  I think these prevent he license header being counted as godoc for the package.

internal/security/setup/option/pkiOption_test.go
```diff
@@ -44,21 +44,37 @@ func TestNewPkiInitOption_GenerateOnly(t *testing.T) {
...
-	// use reflection to find out how many bool type of options in PkiInitOption struct
-	// and create a slice of mock argument interface instances
-	elm := reflect.ValueOf(&PkiInitOption{}).Elem()
-	for i := 0; i < elm.NumField(); i++ {
-		field := elm.Field(i)
-		switch field.Kind() {
-		case reflect.Bool:
-			ifc = append(ifc, mock.AnythingOfTypeArgument("func(*option.PkiInitOption) (option.exitCode, error)"))
-		}
+func TestNewPkiInitOption_CacheOnly(t *testing.T) {
```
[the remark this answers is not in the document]
jim-wang-intel: To make sure the executor instance inside the object has the correct setting with PkiOption, ie `CacheOpt` is set to true for the executor. It is meant to catch the programming error if `CacheOpt` is really set instead of `GenerateOpt` for the executor instance.

internal/security/setup/option/pkiOption.go
```diff
@@ -25,11 +29,17 @@ type OptionsExecutor interface {
...
 	GenerateOpt bool
+	CacheOpt    bool
 	executor    OptionsExecutor
 }
 
 // NewPkiInitOption constructor to get options built for pki-init
 func NewPkiInitOption(opts PkiInitOption) (ex OptionsExecutor, statusCode int, err error) {
+	// cache option cannot used with -generate
+	if opts.CacheOpt && opts.GenerateOpt {
+		return ex, exitWithError.intValue(), errors.New("Cannot attempt cache option with other modes")
```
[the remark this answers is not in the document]
jim-wang-intel: ok

internal/security/setup/option/generate_test.go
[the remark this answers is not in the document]
jim-wang-intel: Thanks!, Good reminder.

internal/security/setup/option/cache.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/edgexfoundry/edgex-go/internal/security/setup"
+)
+
+// Cache option....
```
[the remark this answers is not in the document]
jim-wang-intel: will do

internal/security/setup/option/generate_test.go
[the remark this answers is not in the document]
anonymouse64: well even if they aren't purely cosmetic, we should still separate them out as they are a logically different change from the rest of this PR

internal/security/setup/option/cache_test.go
anonymouse64: okay actually I think you're right and defining it is better, I take back my suggestion

anonymouse64:
in the future, please try to avoid introducing merge commits into feature branches, we prefer rebases on top of master instead. See https://wiki.edgexfoundry.org/display/FA/Committing+Code+Guidelines for more details

jim-wang-intel:
> 
> 
> in the future, please try to avoid introducing merge commits into feature branches, we prefer rebases on top of master instead. See https://wiki.edgexfoundry.org/display/FA/Committing+Code+Guidelines for more details

ok, keep that in mind and will do it in future, thanks for letting me know!

PR #1812, opened 2019-09-24
Convert sys-mgmt-agent to use common bootstrap package

michaelestrin:
Converts system management agent to use the common bootstrap package.

Intentionally reverts global variables eliminated by https://github.com/edgexfoundry/edgex-go/issues/1486
to be consistent with other core service implementations to simplify
intended post-conversion activity to remove global variables from
all core services.

Fixes https://github.com/edgexfoundry/edgex-go/issues/1810

### DIFF
--- a/cmd/sys-mgmt-agent/main.go
+++ b/cmd/sys-mgmt-agent/main.go
@@ -14,29 +14,59 @@
 package main
 
 import (
+	"context"
 	"flag"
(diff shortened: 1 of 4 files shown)


michaelestrin:
Rebased.

michaelestrin:
Rebased.

Issue #1780, opened 2019-09-19
SMA Get Configuration Endpoint Without Registry Flag No Longer Returns Configuration


michaelestrin:
Call to `GET http://localhost:48090/api/v1/config/edgex-core-data` when the system management agent service is not started with the --registry flag should communicate directly with edgex-core-data to retrieve and return edgex-core-data's configuration.

This works in the Edinburgh branch but the related code was removed from the master branch.

This is a backwards-breaking change that must be addressed prior to the Fuji release.

michaelestrin:
My reporting of this issue was apparently flawed.

Upon inspection and testing of master (post #1769 merge), the implementation is working as I would expect when the service is run without the --registry flag.  

It leverages the `[Clients]` chapter in configuration.toml to obtain a service's address and subsequently attempts to connect to that address to retrieve the service's configuration.

If the named service is not found the `[Clients]` chapter in configuration.toml, the implementation will return a response similar to this:

```
{
  "configuration": {
    "foo": "registry is required to obtain configuration."
  }
}
```

Prior to #1769, the implementation would fault.

michaelestrin:
@tsconn23 -- can you confirm this works as you'd expect when the service is started without the --registry flag (and the corresponding configuration.toml has a valid entry for the service name you're passing in the url that you want to obtain configuration for)?

tsconn23:
Currently in `master` (Fuji) the get configuration no longer panics, which is good.

```
GET http://localhost:48090/api/v1/config/edgex-core-data

{"configuration":{"edgex-core-data":"registry is required to obtain configuration."}}
```

However in the Edinburgh release, the same GET request returns the service configuration. So I'm still of the opinion that we have a bug.
```
GET http://localhost:48090/api/v1/config/edgex-core-data
[...]

PR #1819, opened 2019-09-25
SMA Get Configuration Endpoint Without Registry Flag No Longer Returns Configuration

michaelestrin:
Restores backwards compatibility.

Restores prior behavior of mapping service keys to service names in
configuration to get endpoint data for connecting to external services
to obtain configuration (and health, metrics, etc.).

Fixes: https://github.com/edgexfoundry/edgex-go/issues/1780

### DIFF
--- a/internal/system/agent/init.go
+++ b/internal/system/agent/init.go
@@ -21,6 +21,7 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal"
 	bootstrap "github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/interfaces"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/startup"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/endpoint"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/general"
@@ -37,13 +38,7 @@ var RegistryClient registry.Client
 
 func initializeClients(useRegistry bool) {
 	GenClients = NewGeneralClients()
-	if useRegistry {
-		// if we're using the registry, we'll create new general clients as we need them.
-		return
-	}
-
-	// if the registry is not being used, load clients from configurations; assume configuration key is service name
-	for serviceKey := range Configuration.Clients {
+	for serviceKey, serviceName := range config.ListDefaultServices() {
 		GenClients.Set(
 			serviceKey,
 			general.NewGeneralClient(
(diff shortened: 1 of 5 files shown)


michaelestrin:
Rebased.

michaelestrin:
Rebased.

tsconn23:
@michaelestrin Unfortunately, while the use case targeted by the issue is resolved, I think I've found a bug w/r/t Consul. Steps to reproduce below.

1.) Start Consul
2.) Run config-seed
3.) Start all services with the -r cmd line flag
4.) Make sure all services are health checking OK in Consul (http://localhost:8500/ui/dc1/services)
5.) GET http://localhost:48090/api/v1/config/edgex-core-data

On the first call, the following response is returned. Subsequent calls after the first return correctly.
```
{"configuration":{"edgex-core-data":"Get http://:0//api/v1/config: dial tcp :0: connect: can't assign requested address"}}
```
Up to you if you think this should be addressed via a separate issue/PR. The issue that this PR targets is resolved.

internal/system/agent/init.go
```diff
@@ -16,6 +16,7 @@ package agent
 
 import (
 	"context"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
```
tsconn23: organize imports

michaelestrin: Corrected.

michaelestrin:
Will investigate Consul issue tomorrow.

michaelestrin:
>Unfortunately, while the use case targeted by the issue is resolved, I think I've found a bug w/r/t Consul.

I'm unable to duplicate this issue on my local Windows development machine.  Is this behavior consistent on your end (e.g. you're able to duplicate the issue from current tip of master, the response is as you describe regardless of the time passed since the service's are started, etc.)?

On its surface, this issue seems similar/related to https://github.com/edgexfoundry/go-mod-core-contracts/issues/150.  

>Up to you if you think this should be addressed via a separate issue/PR. The issue that this PR targets is resolved.

I'm open to either option.

@tsconn23

michaelestrin:
Minor clean up.  Removed unused parameter from direct.NewMetrics().  Removed unsafe usage of ConfigurationStruct.Clients.  I doubt any of these changes has any effect but please retest @tsconn23.

tsconn23:
> Minor clean up. Removed unused parameter from direct.NewMetrics(). Removed unsafe usage of ConfigurationStruct.Clients. I doubt any of these changes has any effect but please retest @tsconn23.

Retested. This morning the bug I reported above isn't happening, so I think we can go ahead and merge this PR. If it re-appears we'll treat it as a separate issue.

PR #1830, opened 2019-09-26
Add [import] subcommand from merging pki-init option import into secrity-secrets-setup

jim-wang-intel:
- Very similar subcommand operation to cache subcommand but this time it is deployed into _DeployDir_ destination from Cache dir:  it imports from pre-existing cache dir into deploy dir

  - If  PKI in _CacheDir_ is empty, import outputs an error: import requires pre-populated PKI in _CacheDir_ before deploy into _DeployDir_

Example to run: ``` ./security-secrets-setup -confdir ./res import```

Related to: 
- Issue #1635
- Issue #1558 


tingyuz:
recheck

Issue #1826, opened 2019-09-25
Add minimum supported Go version to go.mod


tsconn23:
Locally, I've upgraded to Go 1.13 to test it out. The project hasn't officially rev'ed to that version yet. One of the things I notice is that when I do a build, an entry indicating `go 1.13` as the minimum supported version is put into the `go.mod`. According to the following article, we should have started seeing this behavior with Go 1.12, but I don't recall that being the case

https://golang.org/doc/go1.12#modules

I wouldn't want this to be checked in as `go 1.13` until the project moves to that version. I've tested compilation of the edgex-go services locally with the version directive set to `go 1.12` and everything works as far as I can tell. Again, this directive is only indicating a minimum supported version.

I think we should do this fairly soon as we could easily get contributions from people who've upgraded to Go 1.13 and who might miss this small edit.

jpwhitemn:
Trevor - based on what I read, it sets this by default to the current release if it is not set (which is what you saw when you rev'ed to 1.13).  -> "It will be set to the current release (go 1.12) if no existing version is present."So I am wondering if we just want to leave it out of go.mod and let the default continue to apply.  That way we don't have to have the lock in to the version.  For example, when we do move to 1.13, I presume (maybe incorrectly) that what we do still would compile and work with 1.12 since we aren't using any feature of 1.13 explicitly.  But if we start updating all the go.mods, doesn't this require any developer to have to upgrade to 1.13 (or change their go.mod file back to 1.12)?
In other words, unless we really have a min supported Go that we work with (perhaps that is 1.0 for us??) why set this and artificially require developers to move up when it really isn't necessary?  Shouldn't we use the "minimum required" to indicate just that - minimum and not what we are currently working with?

    On Wednesday, September 25, 2019, 7:47:22 PM GMT+1, Trevor Conn <notifications@github.com> wrote:  
 
 
Locally, I've upgraded to Go 1.13 to test it out. The project hasn't officially rev'ed to that version yet. One of the things I notice is that when I do a build, an entry indicating go 1.13 as the minimum supported version is put into the go.mod. According to the following article, we should have started seeing this behavior with Go 1.12, but I don't recall that being the case

https://golang.org/doc/go1.12#modules

I wouldn't want this to be checked in as go 1.13 until the project moves to that version. I've tested compilation of the edgex-go services locally with the version directive set to go 1.12 and everything works as far as I can tell. Again, this directive is only indicating a minimum supported version.

[...]

PR #1833, opened 2019-09-26
Update README -- Go version pre-req to 1.12.x

### LINKED ISSUE #1832: Update Go version in README
On the main README the Go pre-requisites are listed as 

```
The current targeted version of the Go language runtime is v1.11.5
```

This should read
```
The current targeted version of the Go language runtime is v1.12.x
```

tsconn23:
Fix #1832


anonymouse64:
As mentioned in https://github.com/edgexfoundry/edgex-go/issues/1826#issuecomment-535529566, I think we should say:

* The oldest supported Go compiler version is the oldest currently supported one from the Go team
* What Go compiler we use to build released artifacts

The statement right now (what we target) I think is just the latter point. It would be nice to see a statement about the first part.

anonymouse64:
are you going to add go 1.12 to the go.mod here too?

tsconn23:
@anonymouse64 
> As mentioned in [#1826 (comment)](https://github.com/edgexfoundry/edgex-go/issues/1826#issuecomment-535529566), I think we should say:
> 
> * The oldest supported Go compiler version is the oldest currently supported one from the Go team
> * What Go compiler we use to build released artifacts
> 
> The statement right now (what we target) I think is just the latter point. It would be nice to see a statement about the first part.

Content changed accordingly

>are you going to add go 1.12 to the go.mod here too?

This has been pulled in via rebase.

Issue #1825, opened 2019-09-25
Integrate go-mod-secret into core services


michaelestrin:
Create generic bootstrap handler (to live in internal/pkg/bootstrap/handlers) to instantiate go-mod-secrets-based Vault provider, make the SecretClient.GetSecrets() method available for use by other bootstrap handlers (via global variable for this initial integration effort), and update core-data's configuration.toml files (Docker and non-Docker) with secret-related configuration information.

michaelestrin:
Corresponding configuration.toml file changes for addition of vault.SecretConfig to various core services ConfigurationStruct structures were missed.  Affected service's configuration.toml files need to be updated in a separate PR associated with this issue.

@AnthonyMBonafide

PR #1834, opened 2019-09-26
Add Bootstrap handler for SecretClient creation

AnthonyMBonafide:
Create new handler responsible for creating a secret client and update
core-data to leverage the new handler.

### DIFF
--- a/cmd/core-data/main.go
+++ b/cmd/core-data/main.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"flag"
+
 	"github.com/edgexfoundry/edgex-go"
 	"github.com/edgexfoundry/edgex-go/internal"
 	"github.com/edgexfoundry/edgex-go/internal/core/data"
@@ -54,9 +55,11 @@ func main() {
 		data.Configuration,
 		startupTimer,
 		[]interfaces.BootstrapHandler{
+			handlers.SecretClientBootstrapHandler,
 			data.NewServiceInit(&httpServer).BootstrapHandler,
 			telemetry.BootstrapHandler,
 			httpServer.Handler,
 			handlers.NewStartMessage(clients.CoreDataServiceKey, edgex.Version).Handler,
-		})
+		},
+	)
 }
--- a/cmd/config-seed/Attribution.txt
+++ b/cmd/config-seed/Attribution.txt
@@ -111,11 +111,14 @@ https://github.com/davecgh/go-spew/blob/master/LICENSE
 edgexfoundry/go-mod-core-contracts (Apache 2.0) https://github.com/edgexfoundry/go-mod-core-contracts
 https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/LICENSE
 
+edgexfoundry/go-mod-messaging (Apache 2.0) https://github.com/edgexfoundry/go-mod-messaging
+https://github.com/edgexfoundry/go-mod-messaging/blob/master/LICENSE
+
 edgexfoundry/go-mod-registry (Apache 2.0) https://github.com/edgexfoundry/go-mod-registry
 https://github.com/edgexfoundry/go-mod-registry/blob/master/LICENSE
 
-edgexfoundry/go-mod-messaging (Apache 2.0) https://github.com/edgexfoundry/go-mod-messaging
-https://github.com/edgexfoundry/go-mod-messaging/blob/master/LICENSE
+edgexfoundry/go-mod-secrets (Apache 2.0) https://github.com/edgexfoundry/go-mod-secrets
+https://github.com/edgexfoundry/go-mod-secrets/blob/master/LICENSE
 
 gorilla/context (BSD-3) https://github.com/gorilla/context
 https://github.com/gorilla/context/blob/master/LICENSE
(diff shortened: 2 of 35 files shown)


cmd/core-data/main.go
```diff
@@ -54,6 +54,7 @@ func main() {
 		data.Configuration,
 		startupTimer,
 		[]interfaces.BootstrapHandler{
+			handlers.SecretClientBootstrapHandler,
```
michaelestrin: Is there a command-line switch (or other mechanism) that would preclude the need to execute this handler?

AnthonyMBonafide: Good point. I believe there is, or going to be, a configuration property which will determine if we are using a secret store.I have to look into this.

AnthonyMBonafide:
@tingyuz, did we ever determine how we are going to enable/disable the use of the secret store? I remember talking about it in the Security working group but I do not recall what we decided on other than it is going to be on by default?

michaelestrin:
#1825

cmd/core-data/main.go
AnthonyMBonafide: @michaelestrin , do you know where the configuration property that would drive this be located?

tsconn23: Security is on by default. It can be turned off by running the config-seed with the appropriate environment variable overrides (see PR https://github.com/edgexfoundry/edgex-go/pull/1756). In order to check this value from its location in Consul, a second RegistryClient will be needed.

If no registry is being used, a service can also be invoked with the following env var to disable security `EDGEX_SECURITY_SECRET_STORE=false`. See notes from 15-Aug Core WG. Same is utilized by the docker-edgex-mongo init process.

https://github.com/edgexfoundry/docker-edgex-mongo/blob/0769d9f0c8416f69ea421b9bf45ba0e6c6735a44/internal/pkg/constants.go#L20

AnthonyMBonafide:
Recheck

AnthonyMBonafide:
The CI jobs passed, but I just rebased with the latest changes in `master`

cmd/core-data/main.go
```diff
@@ -45,6 +47,20 @@ func main() {
...
+		data.NewServiceInit(&httpServer).BootstrapHandler,
+		telemetry.BootstrapHandler,
+		httpServer.Handler,
+		handlers.NewStartMessage(clients.CoreDataServiceKey, edgex.Version).Handler,
+	}
+
+	// Unless the secret store functionality has been explicitly disabled, add the SecretClient bootstrap handler as the
+	// first handler since it will be needed by other bootstrap handlers.
+	if env := os.Getenv(internal.SecretStore); env != "false" {
+		bootstrapHandlers = append([]interfaces.BootstrapHandler{handlers.SecretClientBootstrapHandler}, bootstrapHandlers...)
```
michaelestrin: Prefer `bootstrapHandlers = append(bootstrapHandlers, handlers.SecretClientBootstrapHandler)`

cmd/core-data/main.go
```diff
@@ -45,6 +47,20 @@ func main() {
...
+	}
```
michaelestrin: This is an implementation detail that will be common across all services.  Can you revert your changes here, always run the handlers.SecretClientBootstrapHandler, and let it determine in a common location whether or not to do anything based upon the presence or absence of the environment variable?

internal/constants.go
```diff
@@ -23,6 +23,8 @@ const (
 	LogDurationKey                 = "duration"
 	SecuritySecretsSetupServiceKey = "edgex-security-secrets-setup"
 	SecurityProxySetupServiceKey   = "edgex-security-proxy-setup"
+	// Environment variable used to disable the secret store functionality which is ON/TRUE by default.
+	SecretStore = "EDGEX_SECURITY_SECRET_STORE"
```
michaelestrin: This is an implementation detail that belongs in handlers/secret.go.

cmd/core-data/main.go
AnthonyMBonafide: This will go away based on you other comment suggesting that the `handlers.SecretClientBootstrapHandler` should always be included in the bootstrap handlers provided to the `Run` function, and within the handler we should decided to determine if we should create the client. However, the reason why I am doing:
```
bootstrapHandlers = append([]interfaces.BootstrapHandler{handlers.SecretClientBootstrapHandler}, bootstrapHandlers...)
```

rather than:
```
bootstrapHandlers = append(bootstrapHandlers, handlers.SecretClientBootstrapHandler)
```

is because we want the `handlers.SecretClientBootstrapHandler` to be the first handler in the slice that we are passing to the `bootstrap.Run` so that it executes first. The other implementation would add the `handlers.SecretClientBootstrapHandler` to the end of the slice, thus getting executed last. This is a non-issue anyway as the logic for structuring the slice will go away.

michaelestrin: Doh!  Thank you!

AnthonyMBonafide:
> You didn't add `handlers.SecretClientBootstrapHandler` to calls to bootstrap.Run() in the other services that have already been converted to use the new common bootstrap package:
> 
> * core-metadata
> * export-client
> * export-distro
> * support-logging
> * support-notifications
> * support-scheduler
> * sys-mgmt-agent

If I update these other services in this PR then I would like to merge PR #1850 before completing this work so that we can also add this new handler there as well.

AnthonyMBonafide:
Messed something up with the rebase, fixing now.

AnthonyMBonafide:
This ready for review.

michaelestrin:
Rebased

PR #1848, opened 2019-09-30
Refactor SecretStoreClient to redude code duplication

bnevis-i:
This is a refactoring of SecretStoreClient to move request
processing into a common function. This reduces code duplication
and should increase test coverage slightly as a result.

In the future, additional methods will be added to
SecretStoreClient to enable revokation and regeneration of
the root token and stale per-service service tokens.
This refactoring will eanble those methods to be implemented
in less than 10 lines of code in most cases.

TESTING INSTRUCTIONS:
   "go test" in the secretstoreclient folder

### DIFF
--- a/internal/security/secretstoreclient/vault.go
+++ b/internal/security/secretstoreclient/vault.go
@@ -22,7 +22,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 
@@ -30,38 +29,6 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/security/fileioperformer"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 )
-
-// InitRequest contains a Vault init request regarding the Shamir Secret Sharing (SSS) parameters
-type InitRequest struct {
-	SecretShares    int `json:"secret_shares"`
-	SecretThreshold int `json:"secret_threshold"`
-}
-
-// InitResponse contains a Vault init response
-type InitResponse struct {
-	Keys       []string `json:"keys"`
-	KeysBase64 []string `json:"keys_base64"`
-	RootToken  string   `json:"root_token"`
-}
-
-// UnsealRequest contains a Vault unseal request
-type UnsealRequest struct {
-	Key   string `json:"key"`
-	Reset bool   `json:"reset"`
-}
-
-// UnsealResponse contains a Vault unseal response
-type UnsealResponse struct {
-	Sealed   bool `json:"sealed"`
-	T        int  `json:"t"`
-	N        int  `json:"n"`
-	Progress int  `json:"progress"`
-}
-
-// UpdateACLPolicyRequest contains a ACL policy create/update request
-type UpdateACLPolicyRequest struct {
-	Policy string `json:"policy"`
-}
 
 type vaultClient struct {
 	logger logger.LoggingClient
@@ -80,9 +47,8 @@ func NewSecretStoreClient(logger logger.LoggingClient, r internal.HttpCaller, s 
 }
 
 func (vc *vaultClient) HealthCheck() (statusCode int, err error) {
-	url := vc.buildURL(VaultHealthAPI)
 	empty := struct{}{}
-	resp, err := vc.commonRequest(nil, http.MethodGet, url, &empty)
+	resp, err := vc.commonJSONRequest(nil, http.MethodGet, VaultHealthAPI, &empty)
 	if err != nil {
 		vc.logger.Error(fmt.Sprintf("failed on checking status of secret store: %s", err.Error()))
 		return 0, err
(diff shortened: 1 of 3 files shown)


bnevis-i:
recheck

internal/security/secretstoreclient/vault.go
```diff
@@ -272,3 +161,49 @@ func (vc *vaultClient) commonRequest(token *string, method string, url string, j
...
+// and performs standard processing based the the following:
+// operationDescription contains a message describing the operation in logs
+// expectedStatusCode is the expected status code for success
+// responseObject is the address of a JSON struct, or nil
+// If the response is HTTP 200 status and responseObject is not nil,
+// this function will attempt to decode the object
+// delegate is a callback function called to handle the decoded response
+func (vc *vaultClient) processResponse(resp *http.Response, responseError error,
+	operationDescription string, expectedStatusCode int, responseObject interface{},
+	delegate func() error) (statusCode int, err error) {
```
anonymouse64: this is a bit weird that the `delegate` function doesn't take any arguments even though it's effectively acting like a callback. 

Since the `delegate` function is only ever used one place, could you just drop the `delegate` argument entirely and use the code there instead? AFAICT, the only advantage to using this delegate function is that you have error checking inside this function instead of outside it, but to me that is outweighed by the weirdness of having a callback function that takes no arguments.

internal/security/secretstoreclient/vault.go
```diff
@@ -272,3 +161,49 @@ func (vc *vaultClient) commonRequest(token *string, method string, url string, j
...
+		return 0, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != expectedStatusCode {
+		vc.logger.Error(fmt.Sprintf("request to %s failed with status: %s", operationDescription, resp.Status))
+		return resp.StatusCode, err
+	}
+
+	if responseObject != nil && resp.StatusCode == http.StatusOK {
```
anonymouse64: how does this work if I pass `expectedStatusCode` as http.StatusAccepted? Then this won't decode anything into the responseObject AFAICT?

internal/security/secretstoreclient/vault.go
```diff
@@ -272,3 +161,49 @@ func (vc *vaultClient) commonRequest(token *string, method string, url string, j
...
+		err = json.NewDecoder(resp.Body).Decode(responseObject)
+		if err != nil {
+			vc.logger.Error(fmt.Sprintf("failed to parse response body: %s", err.Error()))
+			return resp.StatusCode, err
+		}
+	}
+
+	err = delegate()
```
anonymouse64: as mentioned above I don't think this is really needed

internal/security/secretstoreclient/vault.go
```diff
@@ -80,9 +47,8 @@ func NewSecretStoreClient(logger logger.LoggingClient, r internal.HttpCaller, s
 }
 
 func (vc *vaultClient) HealthCheck() (statusCode int, err error) {
-	url := vc.buildURL(VaultHealthAPI)
 	empty := struct{}{}
-	resp, err := vc.commonRequest(nil, http.MethodGet, url, &empty)
+	resp, err := vc.commonJSONRequest(nil, http.MethodGet, VaultHealthAPI, &empty)
```
anonymouse64: super minor nitpick, but since this empty struct is only used here, could you roll it into it's own line? just a stylistic thing

internal/security/secretstoreclient/vault.go
```diff
@@ -97,68 +63,33 @@ func (vc *vaultClient) Init(config SecretServiceInfo, vmkWriter io.Writer) (stat
...
-	if err != nil {
-		vc.logger.Error(fmt.Sprintf("failed to create Vault init response %s file, HTTP status: %s", config.TokenFolderPath+"/"+config.TokenFile, err.Error()))
-		return 0, err
-	}
+	vc.logger.Info(fmt.Sprintf("vault init strategy (SSS parameters): shares=%d threshold=%d", initRequest.SecretShares, initRequest.SecretThreshold))
 
-	vc.logger.Info("Vault initialization complete.")
-	return resp.StatusCode, nil
+	resp, err := vc.commonJSONRequest(nil, http.MethodPost, VaultInitAPI, &initRequest)
+	return vc.processResponse(resp, err, "initialize secret store", http.StatusOK, &initResp, func() error {
```
anonymouse64: I think it's more readable code to just do:

```go
	err = vc.processResponse(resp, err, "initialize secret store", http.StatusOK, &initResp)
	if err != nil {
	    return err
	}
	return json.NewEncoder(vmkWriter).Encode(initResp)
}
```

you're only saving 3 lines of code checking the error here and IMHO it's more confusing to use this `processResponse` function since you are _requiring_ it to essentially be a closure over some other variable that's probably already provided to the `processResponse` function, so all in all this ends up feeling quite weird to read through.

internal/security/secretstoreclient/vault.go
bnevis-i: Which API call returns http.StatusAccepted with JSON body on success?

internal/security/secretstoreclient/vault.go
bnevis-i: It removes duplication for not only the error checking, but the error logging, and also the log message that the request was performed successfully.  I could in fact pass `responseObject` into the callback, but since I don't know the type I'd have to pass it as interface{} and then write a type assertion in the callback to cast it back to the proper type. I in fact did this without the type assertion and was surprised the callback didn't fail because of the interface{} type, because I found that was using a closure, and the callback already had access to the parsed json body and it had the right type.

In most cases turns a simple API call into a 3-liner

```
	parameters := RevokeTokenAccessorRequest{Accessor: accessor}
	resp, err := vc.commonJSONRequest(&token, http.MethodPost, RevokeAccessorAPI, parameters)
	return vc.processResponse(resp, err, "revoke token accessor", http.StatusNoContent, nil, noop)
```

or a 6-liner if there is something in the JSONresponse we want to keep
```
func (vc *vaultClient) LookupSelf(token string, tokenMetadata *TokenMetadata) (statusCode int, err error) {
	var response TokenLookupResponse
	resp, err := vc.commonRequest(&token, http.MethodGet, LookupSelfAPI, nil)
	return vc.processResponse(resp, err, "lookup self token", http.StatusOK, &response, func() error {
		*tokenMetadata = response.Data
		return nil
	})
}
```

internal/security/secretstoreclient/vault.go
bnevis-i: It saves 7 lines of repetetive boilerplate code per API call, none of which are unit tested in the success flow, times the 7 method calls that are going to use it.  (5 of which will be added in a future PR to support the ability to support token revokation)

internal/security/secretstoreclient/vault.go
bnevis-i: You mean like `resp, err := vc.commonJSONRequest(nil, http.MethodGet, VaultHealthAPI, &struct{}{})` ?

internal/security/secretstoreclient/vault.go
anonymouse64: I don't see how this will be all used here since they all call the noop handler today?

internal/security/secretstoreclient/vault.go
anonymouse64: yes

internal/security/secretstoreclient/vault.go
bnevis-i: I see what you are saying.  In the delegate flow, it will do the json processing then log the success info line.  In this example, processResponse() would log "successfully made request to intialize secret store" or log an error, then do the postprocessing.

internal/security/secretstoreclient/vault.go
bnevis-i: Addressed in #1855

internal/security/secretstoreclient/vault.go
bnevis-i: Addressed in #1855.  Found one API after grep of Vault source where StatusAccepted was returned with a JSON response.

internal/security/secretstoreclient/vault.go
bnevis-i: Addressed in #1855

internal/security/secretstoreclient/vault.go
bnevis-i: Addressed in #1855

internal/security/secretstoreclient/vault.go
bnevis-i: Addressed in #1855

PR #1849, opened 2019-09-30
Resolve spurious unit test failures for fileioperformer in Windows

bnevis-i:
The fileioperfomer unit tests exercise functionality by attempting
to do file I/O on Unix-style paths, which can cause erroneous
unit test failures on Windows. This change makes these tests
only run on Linux -- other platforms automatically pass.

### DIFF
--- a/internal/security/fileioperformer/methods_test.go
+++ b/internal/security/fileioperformer/methods_test.go
@@ -19,6 +19,7 @@ package fileioperformer
 import (
 	"io"
 	"os"
+	"runtime"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


bnevis-i:
> @bnevis-i would it be better to just "fail" on windows environments with a log message or documentation in the test code saying test not for windows as opposed to silently pass?

@mkbhanda No.  We don't want it to fail.  Testing on Linux is good enough.  The goal is to get coverage in the CI/CD pipeline which runs on Linux.

bnevis-i:
Branch out of date. Rebased.

PR #1852, opened 2019-10-01
Uses adopt-info with new 'version' part.

knitzsche:
Executing snapcraft sets snap version for generated snap.
The version is read from VERSION file in root of source tree.
The date and git info is appended to version as previously.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,8 +1,6 @@
 name: edgexfoundry
 base: core18
-version: "replace-me"
-version-script: |
-  echo $(cat VERSION)-$(date +%Y%m%d)+$(git rev-parse --short HEAD)
+adopt-info: version
 summary: Open-source framework for IoT edge computing
 description: |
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -334,6 +332,11 @@ apps:
       - network-bind
 
 parts:
+  version:
+    plugin: nil
+    source: .
```
anonymouse64: as I mentioned over telegram, let's use this instead:

```suggestion
    source: snap/local/build-helpers
```

this better matches other parts like curl which we don't want to have to rebuild every time we make a small change elsewhere in the git tree.

snap/snapcraft.yaml
```diff
@@ -334,6 +332,11 @@ apps:
...
+    override-pull: |
```
anonymouse64: also as we discussed, it would be great if you could leave a comment explaining how this prevents pulling all of the source for the git tree from being staged into the part dir.

PR #1843, opened 2019-09-27
snapcraft.yaml: update docker-edgex-mongo args, add env var to disable security

anonymouse64:
Currently, integration within docker-edgex-mongo to access/setup vault isn't done, so we can't enable integration with vault yet for edgex-mongo, so explicitly turn it off (since if we don't specify it edgex-mongo defaults to trying to turn it on). This env var currently controls whether to use the `Credentials`  (with the env var set to `false` or the `SecretStore` section (with the env var missing or set to true) within the specified configuration.toml (we want Credentials which is just hard-coding the credentials inside the configuration.toml file).

This is a temporary measure until full integration is done. 

See also https://github.com/edgexfoundry/edgex-go/issues/1840

Also, the configuration options were changed to match the rest of EdgeX services, so update that too. 

See https://github.com/edgexfoundry/docker-edgex-mongo/pull/44 and https://github.com/edgexfoundry/docker-edgex-mongo/pull/40 for additional context on the design decisions for docker-edgex-mongo in this area.

To test, just build the snap and install it in devmode. You should see something like this in the logs for `edgex-mongo` service:

```
$ journalctl --no-hostname -ef -u snap.edgexfoundry.edgex-mongo.service 
-- Logs begin at Thu 2019-07-04 19:15:05 CDT. --
Sep 27 15:26:11 systemd[1]: Starting Service for snap application edgexfoundry.edgex-mongo...
Sep 27 15:26:11 edgexfoundry.edgex-mongo[119509]: level=ERROR ts=2019-09-27T20:26:11.856618158Z app=edgex-mongo source=logger.go:105 msg="logTarget cannot be blank, using stdout only"
Sep 27 15:26:11 edgexfoundry.edgex-mongo[119509]: level=INFO ts=2019-09-27T20:26:11.861931256Z app=edgex-mongo source=main.go:32 msg="starting edgex-mongo process ..."
Sep 27 15:26:11 edgexfoundry.edgex-mongo[119509]: level=INFO ts=2019-09-27T20:26:11.861952015Z app=edgex-mongo source=config.go:24 msg="loading the configuration ignoring the secret store"
Sep 27 15:26:11 edgexfoundry.edgex-mongo[119509]: level=INFO ts=2019-09-27T20:26:11.861974608Z app=edgex-mongo source=init.go:47 msg="config file location: /var/snap/edgexfoundry/x1/config/edgex-mongo/res/configuration.toml"
[...]

PR #1855, opened 2019-10-01
Remove delegate function from secretstoreclient processResponse

bnevis-i:
Removes delegate function from secretstoreclient processResponse utility method

Addresses post-commit feedback for #1848.
Fixes potential bug where /auth/token/tidy JSON response would not be
processed because it returns http.StatusAccepted (tidy not currently called).

### DIFF
--- a/internal/security/secretstoreclient/vault.go
+++ b/internal/security/secretstoreclient/vault.go
@@ -47,8 +47,7 @@ func NewSecretStoreClient(logger logger.LoggingClient, r internal.HttpCaller, s 
 }
 
 func (vc *vaultClient) HealthCheck() (statusCode int, err error) {
-	empty := struct{}{}
-	resp, err := vc.commonJSONRequest(nil, http.MethodGet, VaultHealthAPI, &empty)
+	resp, err := vc.commonJSONRequest(nil, http.MethodGet, VaultHealthAPI, struct{}{})
 	if err != nil {
 		vc.logger.Error(fmt.Sprintf("failed on checking status of secret store: %s", err.Error()))
 		return 0, err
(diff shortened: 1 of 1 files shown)


internal/security/secretstoreclient/vault.go
```diff
@@ -185,25 +184,15 @@ func (vc *vaultClient) processResponse(resp *http.Response, responseError error,
 		return resp.StatusCode, err
 	}
 
-	if responseObject != nil && resp.StatusCode == http.StatusOK {
+	if responseObject != nil &&
+		(resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusAccepted) {
```
anonymouse64: this check still feels duplicative with the one above it, I see that this also checks to see if responseObject is nil, but still feels like there should a better way to do it. but it's not logically wrong so approving now anyways.

bnevis-i: @anonymouse64 You are probably right--it is overly conservative. Given the previous check on expected response code, there is little danger that the logic will try to parse a JSON response on an condition such as a 301 or 404 or 500 error -- it will bail out beforehand.  The caller indicates there is no JSON response by passing nil instead of an object.  I have to rebase this anyway.  I could remove the redundant StatusCode check.

anonymouse64: I'll leave the decision on whether to do that or not up to you

bnevis-i: @anonymouse64 I think I will wait until PR #1843 is merged, check Jenkins to see if additional PRs are pending, then rebase this with the extra status code checks removed and try to run the gauntlet again.

Issue #1837, opened 2019-09-27
Integrate simple dependency injection container into common bootstrap


michaelestrin:
Provides a path to replace a service's global variables with instances managed by the DIC which are injected into controller methods.

Dependent upon completion of #1801, #1802, #1803, #1804, #1807, #1808, #1809, and #1810.

michaelestrin:
Now dependent upon #1858

michaelestrin:
TSC discussed during Geneva pre-wire F2F and voted to allow core to merge this prior to Fuji release.

Issue #1858, opened 2019-10-02
System Management Cyclical References


michaelestrin:
Refactored to remove cyclical references that occur when factory of
metricsImpl is moved from main.go into internal/system/agent.

Extracted from https://github.com/edgexfoundry/edgex-go/issues/1837

michaelestrin:
Rebased

PR #1861, opened 2019-10-02
Integrate simple dependency injection container into common bootstrap

michaelestrin:
Provides a path to replace a service's global variables with instances managed by the DIC which are injected into controller methods.

Fixes: https://github.com/edgexfoundry/edgex-go/issues/1837

### DIFF
--- a/internal/pkg/bootstrap/handlers/secret.go
+++ b/internal/pkg/bootstrap/handlers/secret.go
@@ -19,22 +19,15 @@ import (
 	"os"
 	"sync"
 
-	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
-	"github.com/edgexfoundry/go-mod-registry/registry"
-	"github.com/edgexfoundry/go-mod-secrets/pkg"
-	"github.com/edgexfoundry/go-mod-secrets/pkg/providers/vault"
-
-	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/interfaces"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/container"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/startup"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
+	"github.com/edgexfoundry/go-mod-secrets/pkg"
+
+	"github.com/edgexfoundry/go-mod-secrets/pkg/providers/vault"
 )
 
-// Environment variable used to disable the secret store functionality which is ON/TRUE by default.
-const SecretStore = "EDGEX_SECURITY_SECRET_STORE"
-
-// SecretClient is a global variable used to house the SecretClient to be used system-wide for obtaining secrets.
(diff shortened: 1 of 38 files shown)


michaelestrin:
Hold -- dependency on #1858

michaelestrin:
Rebased

michaelestrin:
Rebased

michaelestrin:
Rebased

michaelestrin:
Discussed in today's Core WG call.  Community will continue the higher-level discussion of the pros and cons of DI and implementation in the project at next week's scheduled face-to-face architect meetings in Chicago.

michaelestrin:
Rebased

michaelestrin:
Rebased with additional changes resulting from https://github.com/edgexfoundry/edgex-go/pull/1834 merge.

michaelestrin:
Rebased.

michaelestrin:
Rebased.

internal/pkg/bootstrap/handlers/secret.go
```diff
@@ -16,42 +16,50 @@ package handlers
 
 import (
 	"context"
+	"github.com/edgexfoundry/go-mod-secrets/pkg"
```
tsconn23: organize imports

internal/pkg/bootstrap/handlers/secret.go
```diff
@@ -16,42 +16,50 @@ package handlers
...
 	context context.Context,
 	startupTimer startup.Timer,
-	config interfaces.Configuration,
-	loggingClient logger.LoggingClient,
-	registryClient registry.Client) bool {
+	dic *di.Container) bool {
 
-	if env := os.Getenv(SecretStore); env != "false" {
+	// check for environment variable that turns security off
+	if env := os.Getenv("EDGEX_SECURITY_SECRET_STORE"); env != "false" {
```
tsconn23: `EDGEX_SECURITY_SECRET_STORE` should be a constant. Note that docker-edgex-mongo also uses this.

https://github.com/edgexfoundry/docker-edgex-mongo/blob/0769d9f0c8416f69ea421b9bf45ba0e6c6735a44/internal/pkg/constants.go#L20

I think that means defining it in `go-mod-secrets` would make the most sense. @AnthonyMBonafide for visibility.

internal/support/notifications/init.go
```diff
@@ -20,18 +20,18 @@ package notifications
 import (
 	"context"
 	"fmt"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/container"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
```
tsconn23: organize imports.

internal/pkg/bootstrap/handlers/secret.go
michaelestrin: Agree but also believe this should be worked separately.  Created https://github.com/edgexfoundry/go-mod-secrets/issues/32

PR #1869, opened 2019-10-03
Fix error handling bug in SecretStoreClient.processResponse

bnevis-i:
The processResponse() internal method takes the response object
and eror object from the previous call to commonRequest().
While processing the error, the code formats an error using
the output error object as a source instead of the input
error object as a source, resulting in a nil access error.
A similar error is made when processing an unexpected response
code: this time a nil error object is always returned.

Unit tests have been added to check regressions.

### DIFF
--- a/internal/security/secretstoreclient/vault.go
+++ b/internal/security/secretstoreclient/vault.go
@@ -171,21 +171,22 @@ func (vc *vaultClient) commonRequest(token *string, method string, path string, 
 // delegate is a callback function called to handle the decoded response
 func (vc *vaultClient) processResponse(resp *http.Response, responseError error,
 	operationDescription string, expectedStatusCode int,
-	responseObject interface{}) (statusCode int, err error) {
+	responseObject interface{}) (int, error) {
 
 	if responseError != nil {
-		vc.logger.Error(fmt.Sprintf("unable to make request to %s failed: %s", operationDescription, err.Error()))
-		return 0, err
+		vc.logger.Error(fmt.Sprintf("unable to make request to %s failed: %s", operationDescription, responseError.Error()))
+		return 0, responseError
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode != expectedStatusCode {
-		vc.logger.Error(fmt.Sprintf("request to %s failed with status: %s", operationDescription, resp.Status))
+		err := fmt.Errorf("request to %s failed with status: %s", operationDescription, resp.Status)
+		vc.logger.Error(err.Error())
 		return resp.StatusCode, err
 	}
 
 	if responseObject != nil {
-		err = json.NewDecoder(resp.Body).Decode(responseObject)
+		err := json.NewDecoder(resp.Body).Decode(responseObject)
 		if err != nil {
 			vc.logger.Error(fmt.Sprintf("failed to parse response body: %s", err.Error()))
 			return resp.StatusCode, err
--- a/internal/security/secretstoreclient/vault_test.go
+++ b/internal/security/secretstoreclient/vault_test.go
@@ -17,6 +17,8 @@ package secretstoreclient
 
 import (
 	"encoding/json"
+	"errors"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"os"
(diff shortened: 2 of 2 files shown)


internal/security/secretstoreclient/vault.go
```diff
@@ -171,21 +172,22 @@ func (vc *vaultClient) commonRequest(token *string, method string, path string,
 // delegate is a callback function called to handle the decoded response
 func (vc *vaultClient) processResponse(resp *http.Response, responseError error,
 	operationDescription string, expectedStatusCode int,
-	responseObject interface{}) (statusCode int, err error) {
+	responseObject interface{}) (int, error) {
 
 	if responseError != nil {
-		vc.logger.Error(fmt.Sprintf("unable to make request to %s failed: %s", operationDescription, err.Error()))
-		return 0, err
+		vc.logger.Error(fmt.Sprintf("unable to make request to %s failed: %s", operationDescription, responseError.Error()))
```
jim-wang-intel: IMO, logging should leave it to the caller if possible. so I would suggest just removing this logging here and just return the error.

internal/security/secretstoreclient/vault.go
```diff
@@ -171,21 +172,22 @@ func (vc *vaultClient) commonRequest(token *string, method string, path string,
...
+		return 0, responseError
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode != expectedStatusCode {
-		vc.logger.Error(fmt.Sprintf("request to %s failed with status: %s", operationDescription, resp.Status))
-		return resp.StatusCode, err
+		msg := fmt.Sprintf("request to %s failed with status: %s", operationDescription, resp.Status)
+		vc.logger.Error(msg)
```
jim-wang-intel: removing logging

internal/security/secretstoreclient/vault.go
```diff
@@ -171,21 +172,22 @@ func (vc *vaultClient) commonRequest(token *string, method string, path string,
...
+		return resp.StatusCode, errors.New(msg)
```
jim-wang-intel: just use `fmt.Errorf(...)` for returning error

internal/security/secretstoreclient/vault.go
bnevis-i: As processResponse() was a refactoring to reduce code duplication in the original code, I am going to leave as-is unless someone seconds it, because the error message adds critical context (i.e. which api call failed) that would be lost of we removed it.

internal/security/secretstoreclient/vault.go
bnevis-i: ditto

internal/security/secretstoreclient/vault.go
bnevis-i: Change to fmt.Errorf() and logging err.Error() instead.

internal/security/secretstoreclient/vault.go
jim-wang-intel: Well, this particular one does not seem to add any value of it because the responseError just passes through and the caller has everything in the message.  As a matter of the fact, the caller should have more knowledge on the which api called failed in this case since the response is passed in from the caller side.

bnevis-i:
Will rebase in the morning when the CI queue is empty.

bnevis-i:
Rebased.

internal/security/secretstoreclient/vault.go
anonymouse64: fwiw I agree with @jim-wang-intel here, but not necessary for fixup now

bnevis-i: @anonymouse64  I can further improve the refactor, but need some feedback.  Right now my hangup is that I have two different request() functions, depending on whether the request body has a json object, no request body at all, or potentially an arbitrary payload.  I can combine the request and response functions into a unified request workflow,  but Vault doesn't have named arguments.  I could:
(1) pass interface{} and do a type assertion on it. to determine what to do.  If it is nil or an io.Reader I can just send it on as the request body else assume it is a struct and send it through the JSON marshaller before sending it, or
(2) make a function that just takes a struct{} and have a member JSONData of type interface{} (if not nil, Marshal it) and a member RequestBodyReader of type io.Reader (pass this to the request instead of JSONdata is nil), or
(3) do the same as 2 above, but define the struct and create a method like Do() as a method on the struct.
I kinda lean towards 2 or 3 because 1 will result in a function with 10 or so positional arguments which is almost unreadable.

anonymouse64: I think this fixup is low priority and can wait until you've finished other higher priority tasks for Fuji, but FWIW I like 2

PR #1882, opened 2019-10-08
support for configurable settings for security-secrets-setup

### LINKED ISSUE #1767: Make security-secrets-setup resource directory of pkisetup-vault and pkisetup-kong JSON file configurable
Currently the generate subcommand of security-secrets-setup hard-codes the resource directory path for both pkisetup-vault.json and pkisetup-kong.json JSON files that are used for generating TLS assets.  

This should be configurable through TOML file for security-secrets-setup to use.

eno-intel:
user can specify paths for WorkDir, CacheDir, DeployDir and
CertConfigDir in the configuration.toml

CertConfigDir specifies the directory for certificate configuration json
files (fix #1767)

setting specified in environment variables override the configurable
settings

### DIFF
--- a/internal/security/setup/option/constant.go
+++ b/internal/security/setup/option/constant.go
@@ -18,7 +18,6 @@ package option
 
 import (
 	"errors"
-	"path/filepath"
 )
 
 const (
@@ -27,14 +26,15 @@ const (
 
 	pkiSetupVaultJSON             = "pkisetup-vault.json"
 	pkiSetupKongJSON              = "pkisetup-kong.json"
+	pkiInitScratchDir             = "scratch"
+	pkiInitGeneratedDir           = "generated"
 	resourceDirName               = "res"
 	configTomlFile                = "configuration.toml"
 	envXdgRuntimeDir              = "XDG_RUNTIME_DIR"
-	defaultXdgRuntimeDir          = "/tmp"
-	pkiInitBaseDir                = "/edgex/security-secrets-setup"
-	envPkiCache                   = "PKI_CACHE"
+	defaultWorkDir                = "/tmp"
+	pkiInitBaseDir                = "edgex/security-secrets-setup"
 	defaultPkiCacheDir            = "/etc/edgex/pki"
-	tmpfsRunDir                   = "/run"
+	defaultPkiDeployDir           = "/run/edgex/secrets"
 	tlsSecretFileName             = "server.key"
 	tlsCertFileName               = "server.crt"
 	caCertFileName                = "ca.pem"
@@ -47,8 +47,5 @@ const (
 )
 
 var (
-	pkiInitScratchDir      = filepath.Join(pkiInitBaseDir, "scratch")
-	pkiInitGeneratedDir    = filepath.Join(pkiInitBaseDir, "generated")
-	pkiInitDeployDir       = filepath.Join(tmpfsRunDir, "edgex", "secrets")
 	errCacheNotChangeAfter = errors.New("PKI cache cannot be changed after it was cached previously")
 )
--- a/internal/security/setup/option/generate.go
+++ b/internal/security/setup/option/generate.go
@@ -36,14 +36,27 @@ func Generate() func(*PkiInitOption) (exitCode, error) {
 		if statusCode, err := generatePkis(); err != nil {
 			return statusCode, err
 		}
-		generatedDirPath := filepath.Join(getXdgRuntimeDir(), pkiInitGeneratedDir)
+
+		workDir, err := getWorkDir()
+		if err != nil {
+			return exitWithError, err
+		}
+
+		generatedDirPath := filepath.Join(workDir, pkiInitGeneratedDir)
+		defer os.RemoveAll(generatedDirPath)
+
 		// Shred the CA private key before deploy
 		caPrivateKeyFile := filepath.Join(generatedDirPath, caServiceName, tlsSecretFileName)
 		if err := secureEraseFile(caPrivateKeyFile); err != nil {
 			return exitWithError, err
 		}
 
-		if err := deploy(generatedDirPath, pkiInitDeployDir); err != nil {
+		deployDir, err := getDeployDir()
+		if err != nil {
+			return exitWithError, err
+		}
+
+		if err := deploy(generatedDirPath, deployDir); err != nil {
 			return exitWithError, err
 		}
 
@@ -57,30 +70,35 @@ func isGenerateNoOp(pkiInitOption *PkiInitOption) bool {
 }
 
 func generatePkis() (exitCode, error) {
-	baseWorkingDir, err := os.Getwd()
+	certConfigDir, err := getCertConfigDir()
 	if err != nil {
 		return exitWithError, err
 	}
 
-	resourceDirPath := filepath.Join(baseWorkingDir, resourceDirName)
-	pkiSetupVaultJSONPath := filepath.Join(resourceDirPath, pkiSetupVaultJSON)
-	pkiSetupKongJSONPath := filepath.Join(resourceDirPath, pkiSetupKongJSON)
+	certConfigDir, err = filepath.Abs(certConfigDir)
+	if err != nil {
+		return exitWithError, err
+	}
+	pkiSetupVaultJSONPath := filepath.Join(certConfigDir, pkiSetupVaultJSON)
+	pkiSetupKongJSONPath := filepath.Join(certConfigDir, pkiSetupKongJSON)
 
-	scratchPath := filepath.Join(getXdgRuntimeDir(), pkiInitScratchDir)
+	workingDir, err := getWorkDir()
+	if err != nil {
+		return exitWithError, err
+	}
+	scratchPath := filepath.Join(workingDir, pkiInitScratchDir)
 
 	setup.LoggingClient.Debug(fmt.Sprint("pkiSetupVaultJSONPath: ", pkiSetupVaultJSONPath,
 		"  pkiSetupKongJSONPath: ", pkiSetupKongJSONPath,
 		"  scratchPath: ", scratchPath,
-		"  resourceDirPath: ", resourceDirPath))
+		"  certConfigDir: ", certConfigDir))
 
 	if !checkIfFileExists(pkiSetupVaultJSONPath) {
-		setup.LoggingClient.Error(fmt.Sprint("Vault JSON file for security-secrets-setup not exists in ", pkiSetupVaultJSONPath))
-		return exitWithError, err
+		return exitWithError, fmt.Errorf("Vault JSON file for security-secrets-setup does not exist in %s", pkiSetupVaultJSONPath)
 	}
 
 	if !checkIfFileExists(pkiSetupKongJSONPath) {
-		setup.LoggingClient.Error(fmt.Sprint("Kong JSON file for security-secrets-setup not exists in ", pkiSetupKongJSONPath))
-		return exitWithError, err
+		return exitWithError, fmt.Errorf("Kong JSON file for security-secrets-setup does not exist in %s", pkiSetupKongJSONPath)
 	}
 
 	// create scratch dir if not exists yet:
@@ -88,9 +106,14 @@ func generatePkis() (exitCode, error) {
 		return exitWithError, err
 	}
 
+	currDir, err := os.Getwd()
+	if err != nil {
+		return exitWithError, err
+	}
+
 	// after done, need to change it back to the original working dir to avoid os.Getwd() error
 	// and delete the scratch dir
-	defer cleanup(baseWorkingDir, scratchPath)
+	defer cleanup(currDir, scratchPath)
 
 	// generate TLS certs on the env. of $XDG_RUNTIME_DIR/edgex/pki-init/scratch
 	if err := os.Chdir(scratchPath); err != nil {
@@ -105,10 +128,10 @@ func generatePkis() (exitCode, error) {
 		return exitWithError, err
 	}
 
-	return rearrangePkiByServices(pkiSetupVaultJSONPath, pkiSetupKongJSONPath)
+	return rearrangePkiByServices(workingDir, pkiSetupVaultJSONPath, pkiSetupKongJSONPath)
 }
 
-func rearrangePkiByServices(pkiSetupVaultJSONPath, pkiSetupKongJSONPath string) (exitCode, error) {
+func rearrangePkiByServices(workingDir, pkiSetupVaultJSONPath, pkiSetupKongJSONPath string) (exitCode, error) {
 	vaultConfig, readErr := config.NewX509Config(pkiSetupVaultJSONPath)
 	if readErr != nil {
 		return exitWithError, readErr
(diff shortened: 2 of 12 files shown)


internal/security/setup/option/optionHelpers.go
```diff
@@ -178,6 +178,25 @@ func checkIfFileExists(fileName string) bool {
...
+
+	fileInfo, statErr := os.Stat(path)
+	if os.IsNotExist(statErr) {
+		return
+	}
+
+	exists = fileInfo.IsDir()
+
+	fileMode := int(fileInfo.Mode())
+	writableMask := int(0200)
```
anonymouse64: personally, I prefer this as `(1 << 7)` (with appropriate casting, etc.) as I always forget `0200` is _octal_...

internal/security/setup/option/optionHelpers.go
```diff
@@ -214,20 +233,64 @@ func markComplete(dirPath string) error {
...
 
-func getXdgRuntimeDir() string {
+func getWorkDir() string {
+	var workDir string
+	var err error
+
 	if xdgRuntimeDir, ok := os.LookupEnv(envXdgRuntimeDir); ok {
-		return xdgRuntimeDir
+		workDir = filepath.Join(xdgRuntimeDir, pkiInitBaseDir)
+	} else if setup.Configuration.SecretsSetup.WorkDir != "" {
```
anonymouse64: eventually this should also check consul/registry for the settings, but that's not for this PR

internal/security/setup/option/optionHelpers.go
```diff
@@ -214,20 +233,64 @@ func markComplete(dirPath string) error {
...
+		workDir, err = filepath.Abs(setup.Configuration.SecretsSetup.WorkDir)
+		if err != nil {
+			// use default work dir if error occurs
+			workDir = filepath.Join(defaultWorkDir, pkiInitBaseDir)
```
anonymouse64: I would rather have this function return an error if there's an issue with the configuration.toml. 
Or at the very least log a warning because I have been bitten too many times by trying to understand why something isn't working when it's just silently ignoring my invalid configuration and using the default.

internal/security/setup/option/optionHelpers.go
```diff
@@ -214,20 +233,64 @@ func markComplete(dirPath string) error {
...
+		}
+	} else {
+		// if not specified in config and $XDG_RUNTIME_DIR is undefined, default to /tmp
```
anonymouse64: this comment is wrong, it's not defaulting to `/tmp` it's defaulting to `/tmp/edgex/security-secrets-setup`

internal/security/setup/option/constant.go
```diff
@@ -27,14 +26,15 @@ const (
...
 	pkiSetupVaultJSON             = "pkisetup-vault.json"
 	pkiSetupKongJSON              = "pkisetup-kong.json"
+	pkiInitScratchDir             = "scratch"
+	pkiInitGeneratedDir           = "generated"
 	resourceDirName               = "res"
 	configTomlFile                = "configuration.toml"
 	envXdgRuntimeDir              = "XDG_RUNTIME_DIR"
-	defaultXdgRuntimeDir          = "/tmp"
+	defaultWorkDir                = "/tmp"
 	pkiInitBaseDir                = "/edgex/security-secrets-setup"
```
anonymouse64: since this is no longer the prefix for all the paths that use this variable, I'd prefer to remove the leading `/` from this variable so it's clear looking at the constant definition that this is not meant to be used as a prefix, instead it needs something else to be used as a prefix (i.e. `/run/edgex/security-secrets-setup` or `/tmp/edgex/security-secrets-setup`)

internal/security/setup/option/optionHelpers.go
```diff
@@ -214,20 +233,64 @@ func markComplete(dirPath string) error {
...
-func getPkiCacheDirEnv() string {
-	pkiCacheDir := os.Getenv(envPkiCache)
-	if pkiCacheDir == "" {
-		return defaultPkiCacheDir
+func getCacheDir() (string, error) {
+	cacheDir := defaultPkiCacheDir
+
+	if setup.Configuration.SecretsSetup.CacheDir != "" {
+		cacheDir = setup.Configuration.SecretsSetup.CacheDir
+		if exist, isWritable := checkDirExistsAndIsWritable(cacheDir); !exist || !isWritable {
```
anonymouse64: I'd prefer we return separate error conditions for this, i.e.

```suggestion
		exist, isWritable := checkDirExistsAndIsWritable(cacheDir)
		if !exist {
		     return "", fmt.Errorf("CacheDir from config file, %s, does not exist", cacheDir)
		 }
		 if !isWritable {
		     return "", fmt.Errorf("CacheDir from config file, %s, is not writable", cacheDir)
		 }
```

internal/security/setup/option/optionHelpers.go
```diff
@@ -214,20 +233,64 @@ func markComplete(dirPath string) error {
...
+
+	return cacheDir, nil
+}
+
+func getDeployDir() (string, error) {
+	deployDir := defaultPkiDeployDir
+
+	if setup.Configuration.SecretsSetup.DeployDir != "" {
+		deployDir = setup.Configuration.SecretsSetup.DeployDir
+		if exist, isWritable := checkDirExistsAndIsWritable(deployDir); !exist || !isWritable {
```
anonymouse64: same here, split up the error conditions for more specific error messages

internal/security/setup/option/optionHelpers.go
```diff
@@ -214,20 +233,64 @@ func markComplete(dirPath string) error {
...
+	}
+
+	return workDir
+}
+
+func getCertConfigDir() (string, error) {
+	var certConfigDir string
+	if setup.Configuration.SecretsSetup.CertConfigDir != "" {
+		certConfigDir = setup.Configuration.SecretsSetup.CertConfigDir
+		if exist, _ := checkDirExistsAndIsWritable(certConfigDir); !exist {
```
anonymouse64: a comment here explaining why we don't care if it's writable would be appreciated

internal/security/setup/option/optionHelpers_test.go
```diff
@@ -0,0 +1,153 @@
...
+	// sets up configuration for SecretsSetupInfo
+	tearDown := setupGenerateTest(t)
+	defer tearDown()
+
+	origEnvVal := os.Getenv(envXdgRuntimeDir)
+	os.Unsetenv(envXdgRuntimeDir)
+
+	// test WorkDir that's configured by toml
+	runTimeDir := getWorkDir()
+	expectedWorkDir, _ := filepath.Abs("./workingtest")
```
anonymouse64: why not check the error here?

internal/security/setup/option/optionHelpers_test.go
```diff
@@ -0,0 +1,153 @@
...
+	assert.Equal(t, expectedWorkDir, runTimeDir)
+
+	// test default WorkDir
+	oldConfig := setup.Configuration
+	setup.Configuration = &setup.ConfigurationStruct{}
+	setup.Configuration.SecretsSetup = setup.SecretsSetupInfo{
```
anonymouse64: this could be rolled into the line above it with:

```go
setup.Configuration = &setup.ConfigurationStruct{
    SecretsSetup: {
        WorkDir: "",
    },
}
```

internal/security/setup/option/optionHelpers_test.go
```diff
@@ -0,0 +1,153 @@
...
+	origEnvVal := os.Getenv(envXdgRuntimeDir)
+	os.Unsetenv(envXdgRuntimeDir)
+
+	// test WorkDir that's configured by toml
+	runTimeDir := getWorkDir()
+	expectedWorkDir, _ := filepath.Abs("./workingtest")
+	assert.Equal(t, expectedWorkDir, runTimeDir)
+
+	// test default WorkDir
+	oldConfig := setup.Configuration
```
anonymouse64: I like having `defer tearDown()` like elsewhere in the codebase here rather than just re-assign later because when you modify the global value, the cleanup of that variable/state is right next to the mocking and you don't have to remember to go looking for the cleanup/re-assignment later down in the test.

internal/security/setup/option/optionHelpers_test.go
anonymouse64: ditto for the other tests in this file

internal/security/setup/option/optionHelpers_test.go
anonymouse64: ditto for the other tests in this file

internal/security/setup/option/generate.go
```diff
@@ -57,40 +64,47 @@ func isGenerateNoOp(pkiInitOption *PkiInitOption) bool {
...
 	}
 
-	resourceDirPath := filepath.Join(baseWorkingDir, resourceDirName)
-	pkiSetupVaultJSONPath := filepath.Join(resourceDirPath, pkiSetupVaultJSON)
-	pkiSetupKongJSONPath := filepath.Join(resourceDirPath, pkiSetupKongJSON)
+	certConfigDir, err = filepath.Abs(certConfigDir)
+	if err != nil {
+		return exitWithError, err
+	}
+	pkiSetupVaultJSONPath := filepath.Join(certConfigDir, pkiSetupVaultJSON)
```
anonymouse64: it would be nice if there was a way to specify the actual filename through configuration.toml, but that's not necessary for this PR, but please file an issue for doing that later on. Probably for Geneva or later I think, it's low priority.

internal/security/setup/option/optionHelpers.go
```diff
@@ -214,20 +233,64 @@ func markComplete(dirPath string) error {
 	return nil
 }
 
-func getXdgRuntimeDir() string {
+func getWorkDir() string {
```
anonymouse64: as mentioned above, I think this function should also return an error

internal/security/setup/option/optionHelpers.go
jim-wang-intel: agree, let's put that into future enhancement for using registry

internal/security/setup/option/optionHelpers.go
jim-wang-intel: yea, let's just return an error if operator makes  a mistake

internal/security/setup/option/generate.go
jim-wang-intel: +1 for future enhancement

internal/security/setup/option/optionHelpers.go
jim-wang-intel: ok for that

internal/security/setup/option/optionHelpers.go
jim-wang-intel: we could define that as a constant

internal/security/setup/option/optionHelpers.go
eno-intel: removing comment. it's pretty clear what is happening

internal/security/setup/option/optionHelpers_test.go
eno-intel: the absolute path would be based from the current directory, so i don't expect it to cause an error

internal/security/setup/option/optionHelpers_test.go
eno-intel: not sure i understand the comment above. 
if you are saying that it's better to do the re-assignment of setup.Configuration in the teardown(), rather than in each of the unit tests, then i agree.
otherwise, could you clarify?

anonymouse64: Yes either in the existing teardown function or in a new teardown function

internal/security/setup/option/optionHelpers_test.go
anonymouse64: "don't expect it to cause an error" - those are famous last words :-D please check the error anyways

internal/security/setup/option/generate.go
eno-intel: created an issue for this one. Don't see a way to label it for Geneva, @anonymouse64 ir @jim-wang-intel, could you label it?
https://github.com/edgexfoundry/edgex-go/issues/1885

anonymouse64: done

internal/security/setup/option/optionHelpers.go
bnevis-i: This utility will likely run before the registry is up.  It will definitely run before the registry is up if we generate a TLS certificate for the registry using this tool.

anonymouse64: My understanding was that we will not be generating a TLS certificate for the registry in the short or medium term, so I'd rather see us be consistent with storing configuration in files and also optionally in consul, then deal with the problem of bootstrapping consul later. 

Is there another reason why this utility would run before consul other than generating a TLS certificate for the registry some day?

bnevis-i: Service-to-service encryption.  If Consul Connect were used to provide that using a Vault back-end as a certificate provider, that would also affect the start order by having Consul depend on Vault rather than Vault depend on Consul, or having two Consul instances.  Again, no design implied or proposed -- just possibility thinking here.

anonymouse64: Hmm, well regardless let's file an issue about this so we don't lose this discussion and can bring the issue to a wider audience than just this PR.

internal/security/setup/option/import_test.go
```diff
@@ -167,41 +169,39 @@ func setupImportTest(t *testing.T) func() {
...
+	if err := createDirectoryIfNotExists(pkiInitDeployDir); err != nil {
+		t.Fatalf("cannot create dir %s for the test: %v", pkiInitDeployDir, err)
 	}
-	pkiInitDeployDir = tempDir
 
 	return func() {
 		// cleanup
 		os.Setenv(envXdgRuntimeDir, origEnvXdgRuntimeDir)
-		os.Setenv(envPkiCache, origEnvPkiCache)
+		setup.Configuration = oldConfig
```
anonymouse64: this looks good, thanks

internal/security/setup/option/import_test.go
```diff
@@ -73,7 +74,8 @@ func TestImportFromPKICache(t *testing.T) {
 
 	exitStatus, err := f(importOn.(*PkiInitOption))
 
-	deployEmpty, emptyErr := isDirEmpty(pkiInitDeployDir)
+	deployDir, _ := getDeployDir()
```
anonymouse64: check error here too

internal/security/setup/option/optionHelpers.go
```diff
@@ -214,20 +233,73 @@ func markComplete(dirPath string) error {
...
+
+	return workDir, nil
+}
+
+func getCertConfigDir() (string, error) {
+	var certConfigDir string
+	if setup.Configuration.SecretsSetup.CertConfigDir != "" {
+		certConfigDir = setup.Configuration.SecretsSetup.CertConfigDir
+
+		// we only read files from CertConfigDir, don't care if it's writable
```
anonymouse64: thanks for this

internal/security/setup/option/generate_test.go
```diff
@@ -170,47 +163,43 @@ func setupGenerateTest(t *testing.T) func(t *testing.T) {
 		t.Fatalf("cannot copy %s for the test: %v", configTomlFile, err)
 	}
 
-	setup.Init()
+	_ = setup.Init()
```
anonymouse64: is there a reason not to check err here too?

anonymouse64: I don't see this error being checked, when in the other file we do check the error from `Init`. Please check this error as well

eno-intel: oh, thought i got it

internal/security/setup/option/optionHelpers.go
eno-intel: issue filed here https://github.com/edgexfoundry/edgex-go/issues/1889

PR #1883, opened 2019-10-08
Scheduler IntervalAction returns (413) expected 200

xmlviking:
Fixes#1845 - Address issue in the scheduler schedulerContext code.
 - Was not handling old format correctly for execution in the scheduler_context
 - Fixes Black Box Tests and work correctly.
 - Will introduce a few more Scheduler Black Box Tests which exercise the Interval Frequency

### DIFF
--- a/internal/support/scheduler/utils.go
+++ b/internal/support/scheduler/utils.go
@@ -14,8 +14,13 @@
 package scheduler
 
 import (
+	"regexp"
 	"strconv"
 	"time"
+)
+
+const (
+	frequencyPattern = `^P(\d+Y)?(\d+M)?(\d+D)?(T(\d+H)?(\d+M)?(\d+S)?)?$`
 )
 
 // Convert millisecond string to Time
@@ -34,7 +39,7 @@ func msToTime(ms string) (time.Time, error) {
 }
 
 // Frequency indicates how often the specific resource needs to be polled.
-// It represents as a duration string.
+// It represents as a duration string. Will not do days you must compute to hours
 //
 // Nanosecond Duration = 1
 // Microsecond = 1000 * Nanosecond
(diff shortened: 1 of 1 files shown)


xmlviking:
recheck

akramtexas:
Recheck

akramtexas:
@xmlviking Please squash commits and see if there's a need to rebase again. I have rebased meanwhile.

internal/support/scheduler/utils.go
```diff
@@ -44,7 +49,19 @@ func msToTime(ms string) (time.Time, error) {
 // Hour = 60 * Minute
 func parseFrequency(durationStr string) (time.Duration, error) {
 
-	// check Frequency
+	// Legacy ISO8601 format P1
+	// ^P(\d+Y)?(\d+M)?(\d+D)?(T(\d+H)?(\d+M)?(\d+S)?)?$`
+	matched, _ := regexp.MatchString(frequencyPattern, durationStr)
```
akramtexas: - Can you please point out which new field(s), added to the interval action, regarding `schedulecontext.go` were updated leading up to the issue that resulted in this PR?

akramtexas:
Recheck

internal/support/scheduler/utils.go
akramtexas: @xmlviking Following our chat, we're good here.

michaelestrin: @akramtexas @xmlviking Can ya'll document your conversation here as part of the issue for everyone else's benefit and posterity's sake?

akramtexas: @michaelestrin Sure, will do 👍

xmlviking: There are no additional fields added to the IntervalAction or Interval which affected this particular issue.  I'm sorry if added to any confusion around that.  The issue as I saw it was related to the inclusion of the parseFrequency() function in the scheduler utils.go file.  It did validate both the "legacy frequency" format as well as the new GoLang style, however it did not actually convert the "legacy" time string format into a proper time.Duration.  When creating new SchedulerContext instances (Interval + 1 to N IntervalActions) the frequency field was not going to work in the legacy format.   This fix ensures that if the frequency field is in either formats "legacy" or new "golang time duration style" it will work.

akramtexas: To add (a few supporting details) to that, the existing `func parseFrequency(durationStr string) (time.Duration, error) { ... }` has been augmented with the following call:
```
timeDuration := parseLegacyFrequency(durationStr)
```
That call (to `func parseLegacyFrequency(durationStr string) time.Duration { ... }`) handles the `legacy` approach of parsing frequency (aka `interval`).

akramtexas:
- I'm in the process of validating that the `black box` tests should now pass.

akramtexas:
> * I'm in the process of validating that the `black box` tests should now pass.

Done. The PR LGTM 👍

PR #1887, opened 2019-10-10
The new boot functionality immediate creates a startup parameter with…

### LINKED ISSUE #1886: StartupTimer expects wait in seconds, getting milliseconds
# 🐞 Bug Report

### Affected Services

 The issue is located in: 
New method of bootstrapping the services which use the newly-created StartupTimer

### Is this a regression?

Yes because previously - before the recent bootstrap changes - the milliseconds was acted upon correctly.  No, because this is new functionality.  :)

The method startup.NewStartupTimer which creates the startup timer takes two parameters: retryIntervalInSeconds and maxWaitInSeconds.  The interval is hard-coded as '1' as there's no constant for it.

However, internal.BootTimeoutDefault is 30000, which is milliseconds and is what was correctly used in the previous implementation.  However, it's also used when creating the startup timer which, in fact, is defining the boot timeout as 30,000 seconds, not 30 seconds as I believe is expected. 

## 🔥 Exception or Error
There's no exception or error that's going to occur; just that it's going to take 8 1/3 hours for the service to timeout if the boot does not successfully complete.
<pre><code>

</code></pre>

## 🌍  Your Environment
**Deployment Environment:**

**EdgeX Version:**

**Anything else relevant?**
The change is simple: create a new constant that is number of seconds and not milliseconds.  For completeness, create a constant for the retry interval

scsosna99:
… the parameters '1' and 'internal.BootTimeoutDefault'

BootTimeoutDefault is from the original implementation, in milliseconds, but startup.NewStartupTimer expects seconds.  To fix without breaking the existing usages of internal.BootTimeoutDefault, created a BootTimeoutSecondsDefault and changed all instances of NewStartupTimer to use that.

I also added a BootRetrySecondsDefault for completeness and add it to the calls creating the new startup timer.

Fix #1886

Signed-off-by: Scott C Sosna <me@scottsosna.dev>

### DIFF
--- a/internal/constants.go
+++ b/internal/constants.go
@@ -15,6 +15,8 @@ package internal
 
 const (
 	BootTimeoutDefault             = 30000
+	BootTimeoutSecondsDefault      = 30
+	BootRetrySecondsDefault        = 1
 	ClientMonitorDefault           = 15000
 	ConfigFileName                 = "configuration.toml"
 	ConfigRegistryStemCore         = "edgex/core/"
--- a/internal/constants.go
+++ b/internal/constants.go
@@ -14,7 +14,7 @@
 package internal
 
 const (
-	BootTimeoutDefault             = 30000
+	BootTimeoutDefault             = BootTimeoutSecondsDefault * 1000
 	BootTimeoutSecondsDefault      = 30
 	BootRetrySecondsDefault        = 1
 	ClientMonitorDefault           = 15000
--- a/cmd/core-command/main.go
+++ b/cmd/core-command/main.go
@@ -31,7 +31,7 @@ import (
 )
 
 func main() {
-	startupTimer := startup.NewStartUpTimer(1, internal.BootTimeoutDefault)
+	startupTimer := startup.NewStartUpTimer(internal.BootRetrySecondsDefault, internal.BootTimeoutSecondsDefault)
 
 	var useRegistry bool
 	var configDir, profileDir string
--- a/cmd/core-data/main.go
+++ b/cmd/core-data/main.go
@@ -31,7 +31,7 @@ import (
 )
 
 func main() {
-	startupTimer := startup.NewStartUpTimer(1, internal.BootTimeoutDefault)
+	startupTimer := startup.NewStartUpTimer(internal.BootRetrySecondsDefault, internal.BootTimeoutSecondsDefault)
 
 	var useRegistry bool
 	var profileDir, configDir string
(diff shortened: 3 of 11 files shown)


internal/constants.go
```diff
@@ -15,6 +15,8 @@ package internal
 
 const (
 	BootTimeoutDefault             = 30000
+	BootTimeoutSecondsDefault      = 30
```
michaelestrin: BootTimeoutSecondsDefault = BootTimeoutDefault/1000

scsosna99: The reason I didn't do that is because very soon BootTimeoutDefault will no longer be used and therefore can just be deleted....probably could be made a separate issue to do the cleanup as dependencies on the 2 or 3 remaining usages.

scsosna99: Little confused by comment about needing to multiply by seconds, looks like it's already there.
`func NewStartUpTimer(retryIntervalInSeconds, maxWaitInSeconds int) Timer {
	return Timer{
		startTime: time.Now(),
		duration:  time.Millisecond * time.Duration(maxWaitInSeconds),
		interval:  time.Millisecond * time.Duration(retryIntervalInSeconds),
	}
}
`

scsosna99: so Internally StartupTimer changes the duration and interval to milliseconds so that comparisons to milliseconds the rest of the way works.

michaelestrin: > The reason I didn't do that is because very soon BootTimeoutDefault will no longer be used and therefore can just be deleted....probably could be made a separate issue to do the cleanup as dependencies on the 2 or 3 remaining usages.

These two constants are the same constant expressed in different units of measure -- one is in seconds and the other in milliseconds.  Explicitly tie them together and leave it to a follow-on PR to set BootTimeoutSecondsDefault to 30 when BootTimeoutDefault has all other references removed from the code.

michaelestrin: > Little confused by comment about needing to multiply by seconds, looks like it's already there.

It isn't.  The code is multiplying by `time.Millisecond` and not by `time.Second` which is why the code actually does timeout after 30 seconds (and not the 30,000 seconds as you reported). Even though the variable name denotes unit-of-measure as seconds, we're passing in milliseconds (and multiplying that value by time.Millisecond to get the common duration in nanoseconds.

Run master core data as described above without any changes and verify that it times out after 30 seconds.  Then run it again with the changes in this PR to verify it times out immediately.

michaelestrin:
LGTM pending successful build.

PR #1892, opened 2019-10-11
Add handling for the POST of invalid data to /api/v1/operation.

### LINKED ISSUE #1856: empty POST's to /api/v1/operation return `null`
# 🐞 Bug Report

### Affected Services

 The issue is located in: the system management agent

### Is this a regression?

yes

* empty POST's to /api/v1/operation return `null`:

```
$ curl --verbose -X POST -d "{}" localhost:48090/api/v1/operation; echo
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 48090 (#0)
> POST /api/v1/operation HTTP/1.1
> Host: localhost:48090
> User-Agent: curl/7.64.0
> Accept: */*
> Content-Length: 2
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 2 out of 2 bytes
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Tue, 01 Oct 2019 18:42:28 GMT
< Content-Length: 5
< 
null
* Connection #0 to host localhost left intact
```

## 🌍  Your Environment
**Deployment Environment:**
Ubuntu 19.04
**EdgeX Version:**
master branch

**Anything else relevant?**

### LINKED ISSUE #1866: POST of invalid data to /api/v1/operation works and returns `null`
# 🐞 Bug Report

### Affected Services

 The issue is located in: the system management agent

### Is this a regression?

yes

* POST of invalid data to /api/v1/operation works and returns `null`

```
$ curl --verbose -X POST -d '{"fake":"key"}' localhost:48090/api/v1/operation; echo
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 48090 (#0)
> POST /api/v1/operation HTTP/1.1
> Host: localhost:48090
> User-Agent: curl/7.64.0
> Accept: */*
> Content-Length: 14
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 14 out of 14 bytes
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Tue, 01 Oct 2019 18:45:21 GMT
< Content-Length: 5
< 
null
* Connection #0 to host localhost left intact

```

## 🌍  Your Environment
**Deployment Environment:**
Ubuntu 19.04
**EdgeX Version:**
master branch

**Anything else relevant?**

Extracted from https://github.com/edgexfoundry/edgex-go/issues/1856

akramtexas:
Fix #1866 
Fix #1856 
- As part of DEV testing, I supplied, in turn, the following incorrect and empty body each in with the request:
1. {"fake":"key"}
2. {}
- As expected, got the expected error message back with the response:
```
"incorrect or malformed body was passed in with request"
```

### DIFF
--- a/internal/system/agent/executor/operation.go
+++ b/internal/system/agent/executor/operation.go
@@ -18,7 +18,6 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/system"
 	"github.com/edgexfoundry/edgex-go/internal/system/agent/interfaces"
 	"github.com/edgexfoundry/edgex-go/internal/system/agent/response"
-
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 )
 
--- a/internal/system/agent/router.go
+++ b/internal/system/agent/router.go
@@ -26,6 +26,7 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
 	"github.com/edgexfoundry/edgex-go/internal/system/agent/container"
 	"github.com/edgexfoundry/edgex-go/internal/system/agent/interfaces"
+
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	"github.com/edgexfoundry/go-mod-registry/registry"
 
@@ -110,6 +111,13 @@ func operationHandler(
 	if err = o.UnmarshalJSON(b); err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		loggingClient.Error("error during decoding: %s", err.Error())
+		return
+	}
+
+	if len(o.Services) == 0 || len(o.Action) == 0 {
+		const errorMessage = "incorrect or malformed body was passed in with the request"
+		http.Error(w, errorMessage, http.StatusBadRequest)
+		loggingClient.Error(errorMessage)
 		return
 	}
 


internal/system/agent/executor/operation.go
```diff
@@ -15,6 +15,7 @@
 package executor
 
 import (
+	"fmt"
```
michaelestrin: Empty line after this entry to separate github imports.

internal/system/agent/executor/operation.go
```diff
@@ -45,6 +46,10 @@ func NewOperations(
 // operationViaExecutor delegates a start/stop/restart operation request to the configuration-defined executor.
 func (e operations) Do(services []string, operation string, parameters []string) interface{} {
 	var result []interface{}
+	if len(services) == 0 || operation == "" {
+		return fmt.Sprintf("incorrect or malformed body was passed in with the request")
+	}
+
```
michaelestrin: Is this the same behavior as Edinburgh?  

To ensure backwards contract compatibility, we should implement the same behavior and return the same error message as Edinburgh.

internal/system/agent/executor/operation_test.go
```diff
@@ -32,7 +32,7 @@ func TestOperationDoWithNoServices(t *testing.T) {
 
 	result := sut.Do([]string{}, "operationDoesNotMatter", []string{})
 
-	var emptyResult []interface{}
+	emptyResult := "incorrect or malformed body was passed in with the request"
```
michaelestrin: This string is defined in two separate places; consolidate it to a constant (if it remains after the other requested changes are implemented).

internal/system/agent/executor/operation.go
akramtexas: This is actually _not_ the same behavior as Edinburgh. There was no such error-handling being done in Edinburgh.

internal/system/agent/executor/operation_test.go
akramtexas: So I had removed the definition `var emptyResult []interface{}` and replaced it by the definition emptyResult := "incorrect or malformed body was passed in with the request"`

michaelestrin: This string is defined in two separate places (inside both operation.go and operation_test.go); consolidate it to a constant (if it remains after the other requested changes are implemented).

internal/system/agent/executor/operation.go
michaelestrin: What is the behavior of Edinburgh when invalid data is posted -- what does Edinburgh implementation return as the response body?  We want to duplicate that behavior to ensure backwards compatibility.

akramtexas:
- The behavior of Edinburgh when invalid data is posted is the following:
```
Akrams-MacBook-Pro:edgex-go akramahmad$ curl --verbose -X POST -d '{"fake":"key"}' localhost:48090/api/v1/operation; echo
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 48090 (#0)
> POST /api/v1/operation HTTP/1.1
> Host: localhost:48090
> User-Agent: curl/7.54.0
> Accept: */*
> Content-Length: 14
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 14 out of 14 bytes
* Empty reply from server
* Connection #0 to host localhost left intact
curl: (52) Empty reply from server
```
- The Edinburgh implementation does not return anything as the response body. Instead, the runtime panics:
```
2019/10/16 13:42:04 http: panic serving 172.31.0.1:34648: runtime error: invalid memory address or nil pointer dereference,
goroutine 671 [running]:,
net/http.(*conn).serve.func1(0xc0000928c0),
  /usr/local/go/src/net/http/server.go:1746 +0xd0,
panic(0x9c3960, 0xf0ef70),
  /usr/local/go/src/runtime/panic.go:513 +0x1b9,
github.com/edgexfoundry/edgex-go/internal/system/agent.operationHandler(0xb4bc00, 0xc00057c000, 0xc000142800),
  /go/src/github.com/edgexfoundry/edgex-go/internal/system/agent/router.go:71 +0x7e0,
net/http.HandlerFunc.ServeHTTP(0xaabdd0, 0xb4bc00, 0xc00057c000, 0xc000142800),
  /usr/local/go/src/net/http/server.go:1964 +0x44,
github.com/edgexfoundry/edgex-go/internal/pkg/correlation.OnRequestBegin.func1(0xb4bc00, 0xc00057c000, 0xc000142800),
  /go/src/github.com/edgexfoundry/edgex-go/internal/pkg/correlation/middleware.go:45 +0xa3,
net/http.HandlerFunc.ServeHTTP(0xc00038ef60, 0xb4bc00, 0xc00057c000, 0xc000142800),
  /usr/local/go/src/net/http/server.go:1964 +0x44,
github.com/edgexfoundry/edgex-go/internal/pkg/correlation.OnResponseComplete.func1(0xb4bc00, 0xc00057c000, 0xc000142800),
  /go/src/github.com/edgexfoundry/edgex-go/internal/pkg/correlation/middleware.go:31 +0x9e,
net/http.HandlerFunc.ServeHTTP(0xc00038ef80, 0xb4bc00, 0xc00057c000, 0xc000142800),
  /usr/local/go/src/net/http/server.go:1964 +0x44,
github.com/edgexfoundry/edgex-go/internal/pkg/correlation.ManageHeader.func1(0xb4bc00, 0xc00057c000, 0xc000142700),
  /go/src/github.com/edgexfoundry/edgex-go/internal/pkg/correlation/middleware.go:24 +0x1f0,
net/http.HandlerFunc.ServeHTTP(0xc00038efa0, 0xb4bc00, 0xc00057c000, 0xc000142700),
  /usr/local/go/src/net/http/server.go:1964 +0x44,
github.com/gorilla/mux.(*Router).ServeHTTP(0xc00018c3c0, 0xb4bc00, 0xc00057c000, 0xc00010e300),
  /go/pkg/mod/github.com/gorilla/mux@v1.7.0/mux.go:212 +0xd0,
github.com/gorilla/context.ClearHandler.func1(0xb4bc00, 0xc00057c000, 0xc00010e300),
  /go/pkg/mod/github.com/gorilla/context@v1.1.1/context.go:141 +0x8b,
net/http.HandlerFunc.ServeHTTP(0xc00015b500, 0xb4bc00, 0xc00057c000, 0xc00010e300),
  /usr/local/go/src/net/http/server.go:1964 +0x44,
net/http.serverHandler.ServeHTTP(0xc000163450, 0xb4bc00, 0xc00057c000, 0xc00010e300),
  /usr/local/go/src/net/http/server.go:2741 +0xab,
net/http.(*conn).serve(0xc0000928c0, 0xb4c440, 0xc000089500),
  /usr/local/go/src/net/http/server.go:1847 +0x646,
created by net/http.(*Server).Serve,
  /usr/local/go/src/net/http/server.go:2851 +0x2f5,
```

akramtexas:
> Spin a bug specific to Edinburgh covering the above behavior, reference this issue, and assign it to yourself to correct.
> 
> Does the code in Edinburgh indicate an error return that isn't happening because of the panic? If so, we should correct the code leading to the panic and return that value. If not, does the RAML have any clues what we should return for this use case?
> 
> Edinburgh needs to be fixed to return a proper response for this use case. Whatever response we fix it up to return should be the same response you implement to address #1892.
> 
> @akramtexas

I am following up, starting with the creation of a bug specific to Edinburgh, as follows: [Issue-1932](https://github.com/edgexfoundry/edgex-go/issues/1932)

akramtexas:
> > Spin a bug specific to Edinburgh covering the above behavior, reference this issue, and assign it to yourself to correct.
> > Does the code in Edinburgh indicate an error return that isn't happening because of the panic? If so, we should correct the code leading to the panic and return that value. If not, does the RAML have any clues what we should return for this use case?
> > Edinburgh needs to be fixed to return a proper response for this use case. Whatever response we fix it up to return should be the same response you implement to address #1892.
> > @akramtexas
> 
> I am following up, starting with the creation of a bug specific to Edinburgh, as follows: [Issue-1932](https://github.com/edgexfoundry/edgex-go/issues/1932)

- With the bug-fix applied to `Edinburgh`, the same response will now be implemented here to `master`.

akramtexas:
This PR will also address Issue #1856 since we handle the case when the body supplied with POST is blank as well as when the POST body has invalid data.

PR #1893, opened 2019-10-11
add --confdir switch to security-secrets-setup

eno-intel:


eno-intel:
@jim-wang-intel

eno-intel:
brief history for this PR: the confdir flag was removed as part of the common bootstrap refactor. This is adding that option back for security-secrets-setup.

Issue #1894, opened 2019-10-11
Config settings from Registry no longer properly copied into Services' configuration struct


lenny-intel:
# 🐞 Bug Report
New bootstrap code broke copying configuration from registry into service's configuration struct. No changes made in Consul will make it into service's configuration struct.

### Affected Services
All core services using new bootstrap 

The issue is located in each services implementation of `UpdateFromRaw` since they all try to reassign the pointer to the config that the method was called on. This only changes it for the current scope of the function and doesn't work as intended. 

`c = configuration` in the code below doesn't work

```
func (c *ConfigurationStruct) UpdateFromRaw(rawConfig interface{}) bool {
	configuration, ok := rawConfig.(*ConfigurationStruct)
	if ok {
		// Check that information was successfully read from Registry
		if configuration.Service.Port == 0 {
			return false
		}
		c = configuration
	}
	return ok
}
```

This need to be changed to copy the values of each top level config field into the current config pointer.

### Is this a regression?

Yes, this worked in Edinburgh

### Description and Minimal Reproduction

- Make a change to Support Notifications `Service.Port` in Consul
- Restart Support Notifications
- Support Notifications will use the Port defined in the local configuration.toml not the value from Registry

## 🌍  Your Environment
**Deployment Environment:**
Linux
**EdgeX Version:**
Fuji WIP

michaelestrin:
Great catch @lenny-intel!

lenny-intel:
@michaelestrin , the benefit of having internal use cases using our WIP... :-)

PR #1897, opened 2019-10-12
Config settings from Registry no longer properly copied into Services' configuration struct

michaelestrin:
Fixes: https://github.com/edgexfoundry/edgex-go/issues/1894


michaelestrin:
Rebased

Issue #1868, opened 2019-10-03
Extract dbClient implementations into single common bootstrap handler


michaelestrin:
Extract dbClient creation from individual service-specific bootstrap handlers into a common bootstrap handler.

Will affect core-command, core-data, core-metadata, export-client, support-logging, support-notifications, and support-scheduler.

Dependent on https://github.com/edgexfoundry/edgex-go/issues/1837

PR #1902, opened 2019-10-14
Extract dbClient implementations into single common bootstrap handler

michaelestrin:
Extracted dbClient creation from individual service-specific bootstrap
handlers into a common bootstrap handler.  Affected core-command,
core-data, core-metadata, export-client, support-logging,
support-notifications, and support-scheduler.

Fixes: https://github.com/edgexfoundry/edgex-go/issues/1868

### DIFF
--- a/cmd/core-command/main.go
+++ b/cmd/core-command/main.go
@@ -22,6 +22,7 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/core/command"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/handlers"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/handlers/database"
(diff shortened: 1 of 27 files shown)


michaelestrin:
Rebased

michaelestrin:
recheck

AnthonyMBonafide:
I bit too late :)

michaelestrin:
> What did catch my attention was the separate check for the core service and subsequent conditional to call a different initialization method for the DB. Could this be expressed by embedding the Core DB struct in a DB struct and accomplishing a method override?

Byproduct of the existing implementation I was converting from.  Not sure what you've got in mind.  Thought my approach encapsulated the two different possibilities using a different factory function to create each.

> The next thing that caught my attention was objects like `LoggingClient` are still global. Is the plan to iterate and inject every object?

Yes,the plan is to eliminate the global variables that currently reside in each core service's init.go.  See the system management agent for an example (where this has already been done).




> Finally, why pass di.Container as a pointer to `BootstrapHandler` when its being treated as a read object? The compiler is really good at escape analysis.

Because it's not always used as a read object.  It's passed as a reference to allow a bootstrap handler to add its own service definitions.

PR #1914, opened 2019-10-15
Update secret BootstrapHandler to read token file

### LINKED ISSUE #1876: Update Secret Bootstrap to Read Vault Token from Common File
Update bootstrap implementation to replace the SecretConfig.Authentication.AuthToken value read from configuration.toml file with value read from file (see #1864 for the referential fully-qualified file name where the vault token will be mapped into each core service's container; root_token is the json field within the file that contains the vault token).

Depends on completion of #1825 and #1864.

AnthonyMBonafide:
Fix #1876

Update secret.BootstrapHandler to read the Vault token from a file
specified in the configuration.

Create a new SecretStoreInfo type which encapsulates the SecretConfig
and adds a configuration property for tokenfile information.

Update service specific configurations to use the newly created
SecretStoreInfo rather than SecretConfig so the tokenfile information
can be captured in the struct.

### DIFF
--- a/cmd/core-command/res/docker/configuration.toml
+++ b/cmd/core-command/res/docker/configuration.toml
@@ -49,9 +49,9 @@ Path = '/v1/secrets/edgex/core-command/'
 Protocol = 'http'
 RootCaCertPath = '/vault/config/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem'
 ServerName = 'edgex-vault'
+TokenFile = '/vault/config/assets/resp-init.json'
   [SecretStore.Authentication]
   AuthType = 'X-Vault-Token'
-  AuthToken= 'edgex'
 
 [Startup]
 Duration = 30
--- a/internal/pkg/bootstrap/handlers/secret/secret.go
+++ b/internal/pkg/bootstrap/handlers/secret/secret.go
@@ -21,7 +21,11 @@ import (
 
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/container"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/startup"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
+	"github.com/edgexfoundry/edgex-go/internal/security/authtokenloader"
+	"github.com/edgexfoundry/edgex-go/internal/security/fileioperformer"
+
 	"github.com/edgexfoundry/go-mod-secrets/pkg"
 
 	"github.com/edgexfoundry/go-mod-secrets/pkg/providers/vault"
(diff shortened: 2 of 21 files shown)


AnthonyMBonafide:
I manually tested these changes by using the the test file that we are using for unit tests and confirmed the data is being loaded correctly:

 https://github.com/edgexfoundry/edgex-go/blob/ea1f15c5e52ec41aeee6b49e69e44486ee9ba6fa/internal/security/secretstore/testdata/test-resp-init.json#L10-L9

cmd/core-command/res/docker/configuration.toml
```diff
@@ -49,6 +49,6 @@ Path = '/v1/secrets/edgex/core-command/'
 Protocol = 'http'
 RootCaCertPath = '/vault/config/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem'
 ServerName = 'edgex-vault'
+TokenFile = '/vault/config/assets/resp-init.json'
```
AnthonyMBonafide: As per my conversation with @tingyuz, the Vault token file will be available to the core-service Docker containers at this location. The PR related to providing this file is https://github.com/edgexfoundry/developer-scripts/pull/139

internal/pkg/bootstrap/handlers/secret/secret.go
```diff
@@ -63,3 +72,34 @@ func BootstrapHandler(
 
 	return err == nil
 }
+
+// getSecretConfig creates a SecretConfig based on the SecretStoreInfo configuration properties.
+// If a tokenfile is present it will override the Authentication.AuthToken value.
+func getSecretConfig(secretStoreInfo config.SecretStoreInfo) (vault.SecretConfig, error) {
```
AnthonyMBonafide: I am not completely a fan of doing this logic at this level. I think that it would be best if this lived in `go-mod-secrets` and the `SecretConfig` strruct containing the `tokenfile` field. That way consumers of the `go-mod-secrets` module can simply pass the configuration properties to the constructor without having to perform this logic. The reason this cannot be done now is that the logic which parses the file lives in `edgex-go` and moving the logic to `go-mod-secrets` would create a cyclic dependency. If this seems reasonable I will create a followup issue to migrate the `AuthTokenLoader` functionality to `go-mod-secrets`. I would like to not tackle this effort here as we are approaching a code freeze for the Fuji release.

AnthonyMBonafide:
Recheck

michaelestrin:
Please rebase and address conflicts.  @AnthonyMBonafide

internal/pkg/bootstrap/handlers/secret/secret.go
michaelestrin: > I am not completely a fan of doing this logic at this level.

Agreed.  Create a follow up issue for post-Fuji resolution to move this code into go-mod-secrets.

PR #1915, opened 2019-10-15
Fixes #1814 (nil panic while logging missing res/configuration.toml)

### LINKED ISSUE #1814: panic running security-secretstore-setup and security-proxy-setup
# 🐞 Bug Report

Running security-secretstore-setup with res/configuration.toml missing causes a panic due to undefined LoggingClient.

See:

```
./security-secretstore-setup --init=true
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x8a5ae4]

goroutine 17 [running]:
github.com/edgexfoundry/edgex-go/internal/pkg/config.LoadFromFile(0x0, 0x0, 0x0, 0x0, 0x8e1480, 0xc0000b3d40, 0x5d8a5505, 0xc0223daddb)
        /home/ijohnson/git/edgex/edgex-go/internal/pkg/config/loader.go:52 +0x214
github.com/edgexfoundry/edgex-go/internal/security/secretstore.initializeConfiguration(0xbf5ab321623dad00, 0x0, 0x0, 0x0, 0x0, 0xe47d20, 0xe47d01, 0x1)
        /home/ijohnson/git/edgex/edgex-go/internal/security/secretstore/init.go:68 +0x78
github.com/edgexfoundry/edgex-go/internal/security/secretstore.Retry(0xc000140100, 0x0, 0x0, 0x0, 0x0, 0x7530, 0xc00013c050, 0xc0001401e0)
        /home/ijohnson/git/edgex/edgex-go/internal/security/secretstore/init.go:37 +0x131
created by github.com/edgexfoundry/edgex-go/internal/pkg/startup.Bootstrap
        /home/ijohnson/git/edgex/edgex-go/internal/pkg/startup/bootstrap.go:37 +0xe4
```

This is because the LoggingClient in LoadFromFile is nil because we haven't been able to load the configuration file at this point in time yet.

### Affected Services

The issue is located in: security-secretstore-setup

### Is this a regression?

Maybe...

This was from master @ abc9af1de59ab5c4130efe3914aaf2c6688b4e14

bnevis-i:
security-secretsstore-setup is not migrating to common bootstrap
in this release. The config module can panic() while calling
deprecated methods to load the configuration while attempting
to log that configuration.toml cannot be found.

This patch sets the global variable before attempting to call
the deprecated configuration loading functions and then forces
the utility to exit if configuration cannot be loaded.
This is a temporary workaround until the utility can be
migrated to use the common bootstrap.

### DIFF
--- a/cmd/security-secretstore-setup/main.go
+++ b/cmd/security-secretstore-setup/main.go
@@ -69,6 +69,10 @@ func main() {
 		BootTimeout: internal.BootTimeoutDefault,
 	}
 	startup.Bootstrap(params, secretstore.Retry, logBeforeInit)
+	if secretstore.Configuration == nil {
+		// secretstore.LoggingClient wasn't initialized either
+		os.Exit(1)
+	}
 
 	//step 1: boot up secretstore general steps same as other EdgeX microservice
 
--- a/internal/constants.go
+++ b/internal/constants.go
@@ -14,17 +14,18 @@
 package internal
 
 const (
-	BootTimeoutDefault             = BootTimeoutSecondsDefault * 1000
-	BootTimeoutSecondsDefault      = 30
-	BootRetrySecondsDefault        = 1
-	ClientMonitorDefault           = 15000
-	ConfigFileName                 = "configuration.toml"
-	ConfigRegistryStemCore         = "edgex/core/"
-	ConfigRegistryStemSecurity     = "edgex/security/"
-	ConfigMajorVersion             = "1.0/"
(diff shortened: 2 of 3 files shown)


bnevis-i:
Testing instructions: from the original issue, all that is needed is to run security-secretstore-setup --init=true with res/configuration.toml missing.  The bug is fixed if the tool exits cleanly instead of crashing with a panic().

tingyuz:
recheck

PR #1916, opened 2019-10-15
security-secretstore-setup Credential Creation

### LINKED ISSUE #1676: Implementation of credential creation for Microservices in security-secretstore-setup
Some micro services in EdgeX needs credentials to access mongodb. Security-secretstore-setup is responsible of creating such credentials and store them into Vault so that other micro services are able to retrieve them and use them to access mongodb.

jdharms:
Fixes #1676 

This PR implements the functionality to create database/mongodb credentials and store them in Vault during security-secretstore-setup initialization.

There are a few subtleties around this process.

1. The databases to have credentials created are specified in a new section of the security-secretstore-setup config file (for now).  I know that @tingyuz and @michaelestrin have had discussions on other ways to implement this, but they're out of scope for Fuji.

2. For some databases (core-data, for example), a specific service needs access to the credentials as well as the mongo service.  For others (admin, authorization, config, local), only the mongo service needs access.   This separation is handled by having a "service" key in the config section of each database to specify if the credential should be added to a service-specific path (e.g., `secret/edgex/coredata/mongodb`).

3. The gokey library used to create credentials uses the vault token and the service name to generate durable (subsequent requests to generate passwords for the same token/service name will generate the same password), secure passwords.  As a result of this, the password for each service is generated, then vault is checked to see if it already contains credentials at the path, and if not they are stored at the path.  For the credentials that need to be stored in two paths, if for some reason they are found in just one (as a result of an interrupted initialization, for example), re-generating and storing in just the missing one will work fine.

## To test:
1. Inside `cmd/security-secretstore-setup/res/configuration.toml`, edit the `[SecretStore]` section for the following fields:
```
server = "localhost"
scheme = "http"
cafilepath = "/Users/jdharms/workspace/edgex-go/cmd/security-secretstore-setup/EdgeXFoundryCA.pem"
certfilepath = "/Users/jdharms/workspace/edgex-go/cmd/security-secretstore-setup/edgex-kong.pem"
keyfilepath = "/Users/jdharms/workspace/edgex-go/cmd/security-secretstore-setup/edgex-kong.priv.key"
tokenfolderpath = "/Users/jdharms/workspace/edgex-go/cmd/security-secretstore-setup"
```
2. Ensure the paths in the previous step point to files on your file system and not mine.  The contents of these files are not important right now; I just used the filenames.
3. Run vault with flag `-dev-kv-v1` to ensure the right version of the secrets engine (alternatively bring up the v1 kv some other way)
```
vault server -dev -dev-kv-v1
```
4. `VAULT_ADDR='http://127.0.0.1:8200' vault operator seal` will seal the local development server.
4. Create a resp-init.json file if it doesn't exist inside the tokenfolderpath above.  The contents should be:
```
{
  "keys": [
    "test-keys"
  ],
  "keys_base64": [
    "UsI8HbpFCt9zqVePbo4CvsijhlNaC7ckKiDnNiBt3F4="
  ],
  "root_token": "s.MRPe8MVhaSePWcfDdHQsMImy"
}
```
Where the `root_token` value is the root token for your dev mode vault instance and `keys_base64` is the Unseal Key.
5. Make build and executing the security-secretstore-setup with --init=true binary should succeed.
6. Check for existence of mongodb credentials in vault:
```
$ vault kv get secret/edgex/mongo/coredata
$ vault kv get secret/edgex/coredata/mongodb # should be the same
$ vault kv get secret/edgex/admin/mongodb # should not exist
$ vault kv get secret/edgex/mongo/admin # should exist

### DIFF
--- a/cmd/security-secretstore-setup/main.go
+++ b/cmd/security-secretstore-setup/main.go
@@ -167,14 +167,7 @@ func main() {
 	gk := secretstore.NewGokeyGenerator(absPath)
 	cred := secretstore.NewCred(req, absPath, gk)
 	for dbname, info := range secretstore.Configuration.Databases {
-		// derive paths from database config info
 		service := info.Service
-		servicePath := ""
-		if len(service) != 0 {
-			servicePath = fmt.Sprintf("/v1/secret/edgex/%s/mongodb", service)
-		}
-		mongoPath := fmt.Sprintf("/v1/secret/edgex/mongo/%s", dbname)
-
 		// generate credentials
 		password, err := cred.GeneratePassword(dbname)
 		if err != nil {
@@ -187,7 +180,8 @@ func main() {
 		}
 
 		// add credentials to service path if specified and they're not already there
-		if len(servicePath) != 0 {
+		if len(service) != 0 {
+			servicePath := fmt.Sprintf("/v1/secret/edgex/%s/mongodb", service)
 			existing, err := cred.AlreadyInStore(servicePath)
 			if err != nil {
 				secretstore.LoggingClient.Error(err.Error())
@@ -204,6 +198,7 @@ func main() {
 			}
 		}
 
+		mongoPath := fmt.Sprintf("/v1/secret/edgex/mongo/%s", dbname)
 		// add credentials to mongo path if they're not already there
 		existing, err := cred.AlreadyInStore(mongoPath)
 		if err != nil {
@@ -214,6 +209,7 @@ func main() {
 			err = cred.UploadToStore(&pair, mongoPath)
 			if err != nil {
 				secretstore.LoggingClient.Error(fmt.Sprintf("failed to upload credential pair for db %s on path %s", dbname, mongoPath))
+				os.Exit(1)
 			}
 		} else {
 			secretstore.LoggingClient.Info(fmt.Sprintf("credentials for %s already present at path %s", dbname, mongoPath))
--- a/cmd/security-secretstore-setup/main.go
+++ b/cmd/security-secretstore-setup/main.go
@@ -78,6 +78,10 @@ func main() {
 
 	//step 2: initialize the communications
 	req := secretstore.NewRequester(insecureSkipVerify)
+	if req == nil {
+		secretstore.LoggingClient.Error("unable to create http requester")
+		os.Exit(1)
+	}
 	vaultScheme := secretstore.Configuration.SecretService.Scheme
 	vaultHost := fmt.Sprintf("%s:%v", secretstore.Configuration.SecretService.Server, secretstore.Configuration.SecretService.Port)
 	intervalDuration := time.Duration(vaultInterval) * time.Second
@@ -92,7 +96,7 @@ func main() {
 		func() {
 			tokenFile, err := os.Open(absPath)
 			if err != nil {
-				secretstore.LoggingClient.Error(fmt.Sprintf("unable to open token file at %s%s", path, filename))
+				secretstore.LoggingClient.Error(fmt.Sprintf("unable to open token file at %s", absPath))
 				os.Exit(1)
 			}
 			defer tokenFile.Close()
@@ -160,7 +164,8 @@ func main() {
 	//TODO: create vault access token for different roles
 
 	// credential creation
-	cred := secretstore.NewCred(req, absPath)
+	gk := secretstore.NewGokeyGenerator(absPath)
+	cred := secretstore.NewCred(req, absPath, gk)
 	for dbname, info := range secretstore.Configuration.Databases {
 		// derive paths from database config info
 		service := info.Service
@@ -171,7 +176,7 @@ func main() {
 		mongoPath := fmt.Sprintf("/v1/secret/edgex/mongo/%s", dbname)
 
 		// generate credentials
-		password, err := cred.Generate(dbname)
+		password, err := cred.GeneratePassword(dbname)
 		if err != nil {
 			secretstore.LoggingClient.Error(fmt.Sprintf("failed to generate credential pair for service %s", service))
 			os.Exit(1)
--- a/cmd/security-secretstore-setup/main.go
+++ b/cmd/security-secretstore-setup/main.go
@@ -165,6 +165,7 @@ func main() {
 
 	// credential creation
 	gk := secretstore.NewGokeyGenerator(absPath)
+	secretstore.LoggingClient.Warn("WARNING: The gokey generator is a reference implementation for credential generation and the underlying libraries not been reviewed for cryptographic security. The user is encouraged to perform their own security investigation before deployment.")
 	cred := secretstore.NewCred(req, absPath, gk)
 	for dbname, info := range secretstore.Configuration.Databases {
 		service := info.Service
--- a/cmd/security-secretstore-setup/main.go
+++ b/cmd/security-secretstore-setup/main.go
@@ -93,6 +93,7 @@ func main() {
 			tokenFile, err := os.Open(absPath)
 			if err != nil {
 				secretstore.LoggingClient.Error(fmt.Sprintf("unable to open token file at %s%s", path, filename))
+				os.Exit(1)
 			}
 			defer tokenFile.Close()
 			sCode, _ := vc.HealthCheck()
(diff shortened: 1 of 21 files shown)

### CODE REVIEW

**Comment by tingyuz** [comment]
recheck

**Comment by jdharms** [comment]
recheck

**Comment by tingyuz** [comment]
> I can't find this anywhere in the WG meeting notes, but I thought during one of the security WG meetings we had agreed to have the generation of passwords (i.e. secrets) implemented using an external executor, like what the SMA does, and provide a reference implementation of how that works (possibly using this gokey library, etc.), but making it very clear we are not guaranteeing the security or reliability of that executor.
> 
> The reason we agreed to do this I thought was that no one (to my knowledge and to the working group's knowledge when I brought this up) had done a security review of the password generation code referenced by @tingyuz in the original implementation and that rather than try to do a security review of any such library we should just punt on the issue and provide a reference implementation that we explicitly described as not being secure and actual deployments should perform their own security review of the password generation executor and optionally provide their own implementation that meets their standards.
> 
> I however recognize that we are under a deadline to get security-secretstore-setup as close as possible as it was to Edinburgh before Fuji is released, but I still _strongly_ think that we should either have a clearly insecure solution (i.e. just hard-code the passwords here) that is obvious to folks not to be relied on for security _or_ we do the configurable executor implementation so that folks at least have a choice.
> 
> @jdharms I don't see that kind of implementation here, have you explored that at all?

@anonymouse64 since the interface of CredentialGenerator is defined to create the password, and coding with gokey is just a reference implementation to the interface, I think we are OK here. 

I would propose to add log warning information before creating the passwords, something like "this is for reference implementation and the user is encouraged to perform security investigation on password generation  mechanism independently".

**Comment by anonymouse64** [comment]
> since the interface of CredentialGenerator is defined to create the password, and coding with gokey is just a reference implementation to the interface

I think this in insufficient, how does a given deployment decide to do use a non-reference implementation of the interface? They would have to fork and rebuild edgex-go to change the implementation. If we were using go plugins, then we could do that and users could build their own plugin but as is, I don't think we should be requiring them to fork the code and rewrite parts of the codebase to implement the interface with their own version. 

Providing them with the option to modify the configuration.toml file and write a small program (or even designing the executor to "just work" with something like the openssl CLI tool with appropriate options) that is separate from edgex-go is the only way to provide this right now with our current set of community agreed upon tools.

**Comment by anonymouse64** [comment]
also is this PR also part of #1891 ?

**Comment by tingyuz** [comment]
> also is this PR also part of #1891 ?

Yes it is correct. I have talked with @difince regarding this PR so that she can work on her side ( mongoDB client) to make sure we are on the same page.

**Comment by tingyuz** [comment]
> > since the interface of CredentialGenerator is defined to create the password, and coding with gokey is just a reference implementation to the interface
> 
> I think this in insufficient, how does a given deployment decide to do use a non-reference implementation of the interface? They would have to fork and rebuild edgex-go to change the implementation. If we were using go plugins, then we could do that and users could build their own plugin but as is, I don't think we should be requiring them to fork the code and rewrite parts of the codebase to implement the interface with their own version.
> 
> Providing them with the option to modify the configuration.toml file and write a small program (or even designing the executor to "just work" with something like the openssl CLI tool with appropriate options) that is separate from edgex-go is the only way to provide this right now with our current set of community agreed upon tools.

I think we have different understanding on the decision we made regarding this issue during past security WG meeting. On Sept 25th WG meeting we agreed the solution as below ( copied from meeting minutes):

•    3rd party package to generate credentials for Microservice
o    No No for writing own module for credential creation
o    Provide interface and reference implementation for future extension
    1Password/spg : https://github.com/1Password/spg
    Gokey : https://github.com/tingyuz/gokey

My understanding is this PR meeting requirements above - not writing own module for credential create, providing interface and reference implementation. Having executor and defining the module in the configuration.toml file sounds interesting but it is not what we asked in the meeting. I would prefer to have such idea captured in issue list and it may be addressed in the next release.

**Comment by anonymouse64** [comment]
@tingyuz regardless of what's in the meeting minutes, how would a user go about using a non-reference implementation here?

**Comment by mkbhanda** [comment]
@anonymouse64, it is about what you "expect", in this case we expect secrets in some location in some format. We do  provide the user a script/executable that writes those secrets (aka a reference implementation). Users could create their own script replacing goKey with their favourite key generator. It is not a "fork" but a custom implementation.

**Comment by tingyuz** [comment]
> @tingyuz regardless of what's in the meeting minutes, how would a user go about using a non-reference implementation here?

The user needs to implement the generate interface with his own method. It is not perfect and needs coding. However if we have executor and define the binary in configuration.toml file, he still needs to code for password generation part and compile to a binary, then update the configuration file and rebuild the image. IMO it almost takes same efforts if he decides to go with his own method.

**Comment by bnevis-i** [comment]
I would like to weight in on this as well.  First, I don't think this issue is worthy of blocking Fuji release, but I do believe that security-secrets-setup is on the monolithic side.  I would like to see @anonymouse64 submit a github issue for his feature request.

> However if we have executor and define the binary in configuration.toml file, he still needs to code for password generation part and compile to a binary, then update the configuration file and rebuild the image. IMO it almost takes same efforts if he decides to go with his own method.

In many cases, it is possible to inject an executable into the container's file system, and it will be possible to override configuration.toml entries with environment variables.  So I disagree with the part about adding a hook requiring the user to build a new container image in order to use it.

**Comment by tingyuz** [comment]
> I would like to weight in on this as well. First, I don't think this issue is worthy of blocking Fuji release, but I do believe that security-secrets-setup is on the monolithic side. I would like to see @anonymouse64 submit a github issue for his feature request.
> 
> > However if we have executor and define the binary in configuration.toml file, he still needs to code for password generation part and compile to a binary, then update the configuration file and rebuild the image. IMO it almost takes same efforts if he decides to go with his own method.
> 
> In many cases, it is possible to inject an executable into the container's file system, and it will be possible to override configuration.toml entries with environment variables. So I disagree with the part about adding a hook requiring the user to build a new container image in order to use it.

I will put this into today's security WG meeting for a panel discussion and we will make decision there.

**Comment by anonymouse64** [comment]
Thank you @tingyuz for putting this on the agenda...

I agree with @bnevis-i that I don't think it's a significant amount of additional work to change the configuration.toml and add a new script to the container. Additionally, it's specifically harder to do this for docker than it is to do for other deployment methods, i.e. snaps 

Additionally, I'd still like to hear from @jdharms as the original author of this PR if this feature was explored at all.

I believe there was a misunderstanding in the minutes that were taken as I never suggested an interface like this I believe I specifically mentioned having an external executor using the SMA as an example but what was mentioned doesn't matter so much. 

The main issue I see with this approach is that for users of edgex with the security services turned on, there's an (many times invalid) assumption of things being secure especially when things "just work". By making this just work with something that hasn't been vetted (i.e. gokey has not been vetted and there don't seem to be any CVE's, etc.) we are giving the impression to these users that this was vetted when in fact it was not. The message that @tingyuz suggested is a step in the right direction but I don't think it's sufficient, especially since we aren't providing users with a way to change out the implementation with something that would be vetted by their own deployment admins or system integrators, etc. So their only option is to go with something that's unvetted.

**Comment by jdharms** [comment]
@anonymouse64 :
> Additionally, I'd still like to hear from @jdharms as the original author of this PR if this feature was explored at all.


If I understand you correctly you are asking if I explored having an external binary generate the passwords? I did not. I also thought this PR met the requirements that Tingyu pasted from the 9/25 security wg minutes.  Agree with @bnevis-i above that requested enhancements should be noted in an issue (less ephemeral than a PR review).

File: internal/security/secretstore/password.go [hunk 1]
```diff
@@ -0,0 +1,216 @@
...
+
+type UserPasswordPair struct {
+	User     string `json:"user,omitempty"`
+	Password string `json:"password,omitempty"`
+}
+
+type Cred struct {
+	client    internal.HttpCaller
+	tokenPath string
+	g         CredentialGenerator
```
jim-wang-intel: Can we spell out this field name in struct like `generator`?  I am a fan of self-document code so full name is more descriptive and clear to the developer.

cmd/security-secretstore-setup/main.go
```diff
@@ -154,8 +159,62 @@ func main() {
...
+		// add credentials to mongo path if they're not already there
+		existing, err := cred.AlreadyInStore(mongoPath)
+		if err != nil {
+			secretstore.LoggingClient.Error(err.Error())
+			os.Exit(1)
+		}
+		if !existing {
+			err = cred.UploadToStore(&pair, mongoPath)
+			if err != nil {
+				secretstore.LoggingClient.Error(fmt.Sprintf("failed to upload credential pair for db %s on path %s", dbname, mongoPath))
```
jim-wang-intel: Do we also need an `os.Exit(1)` after error here?

jdharms: Good catch, thank you

jdharms:
Rebased and resolved merge conflicts and addressed @jim-wang-intel 's code review feedback.

cmd/security-secretstore-setup/main.go
```diff
@@ -154,8 +155,61 @@ func main() {
...
+	// credential creation
+	cred := secretstore.NewCred(req, absPath)
+	for dbname, info := range secretstore.Configuration.Databases {
+		// derive paths from database config info
+		service := info.Service
+		servicePath := ""
+		if len(service) != 0 {
+			servicePath = fmt.Sprintf("/v1/secret/edgex/%s/mongodb", service)
+		}
+		mongoPath := fmt.Sprintf("/v1/secret/edgex/mongo/%s", dbname)
```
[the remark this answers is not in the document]
tingyuz: @jim-wang-intel I think it is needed here. The first path ( /v1/secret/edgex/%s/mongodb) is for micoservices of edgex (which Anthony is working on), and the second path ( /v1/secret/edgex/mongo/%s) is for MongoDB client ( which Diana is working on), so both are required IMO.

cmd/security-secretstore-setup/main.go
```diff
@@ -158,8 +163,59 @@ func main() {
 	//Step 4:
 	//TODO: create vault access token for different roles
 
-	//step 5 :
-	//TODO: implement credential creation
+	// credential creation
+	gk := secretstore.NewGokeyGenerator(absPath)
+	secretstore.LoggingClient.Warn("WARNING: The gokey generator is a reference implementation for credential generation and the underlying libraries not been reviewed for cryptographic security. The user is encouraged to perform their own security investigation before deployment.")
```
anonymouse64: this is fine, thanks

internal/security/secretstore/password.go
```diff
@@ -0,0 +1,216 @@
...
+		return nil, e
+	}
+
+	return &cred.Pair, nil
+}
+
+func (cr *Cred) credPathURL(path string) (string, error) {
+	baseURL, err := url.Parse(Configuration.SecretService.GetSecretSvcBaseURL())
+	if err != nil {
+		e := fmt.Errorf("error parsing secret-service url.  check server and port properties")
```
anonymouse64: it would be nice if the error provided here was specific to the actual issue. in general I dislike using tools that tell me something's wrong, especially when the tool has information about what specifically is wrong. i.e. something like:
```go
e := fmt.Sprintf("error parsing secret-service url: %s",err.Error())
```

internal/security/secretstore/password.go
```diff
@@ -0,0 +1,216 @@
...
+	return cr.generator.Generate(service)
+}
+
+func (cr *Cred) UploadToStore(pair *UserPasswordPair, path string) error {
+	t, err := GetAccessToken(cr.tokenPath)
+	if err != nil {
+		return err
+	}
+
+	LoggingClient.Info("trying to upload the credential pair into secret store")
```
anonymouse64: this feels like a debug level message

anonymouse64:
also did we ever get an issue filed for geneva to refactor this into an executor style implementation?

bnevis-i:
> also did we ever get an issue filed for geneva to refactor this into an executor style implementation?

No, but I volunteer take care of that right now for you.

bnevis-i:
> also did we ever get an issue filed for geneva to refactor this into an executor style implementation?

@anonymouse64 Filed #1946. Feel free to edit description directly to include anything I missed.

internal/security/secretstore/password.go
anonymouse64: sorry if I was unclear, but I didn't mean for @tsconn23 to address this, rather @jdharms address this in a followup

Issue #1917, opened 2019-10-15
Update security-secrets-setup package name under internal to be consistent with other core services


eno-intel:
Currently, the package for security-secrets-setup is under internal/security/setup. This is misleading, as we have 3 security services that use "setup" in their name (security-proxy-setup, security-secrets-setup and security-secretstore-setup) and could cause confusion in the future. 

security-proxy-setup and security-secretstore-setup have their packages under internal/security/proxy and internal/security/secretstore, respectively.

It seems appropriate to change security-secrets-setup package to be under internal/security/secrets to be consistent with structure of the other core services package's naming.

jim-wang-intel:
@tingyuz , @bnevis-i , @anonymouse64 Ian, can we have this in Fuji?  @eno-intel has already implemented the changes on his fork and can create an PR fairly quickly.

anonymouse64:
if this is just moving code around that should be a simple PR I'd say open it and if it gets hairy/complicated then we'll close it

bnevis-i:
I think that's a good observation and am 60/40 in-favor given that it is localized to just this one utility.

anonymouse64:
Fixed with #1920

PR #1920, opened 2019-10-15
Update security-secrets-setup package name under internal to be consi…

eno-intel:
…stent with other core services #1917


eno-intel:
@jim-wang-intel @anonymouse64 @tingyuz @bnevis-i

PR #1936, opened 2019-10-16
Remove BSON from Service RAML Docs

### LINKED ISSUE #1918: Audit RAML docs for BSON references
I notice that our RAML docs still make reference to {Id} route parameters as being of a BSON datatype.

https://github.com/edgexfoundry/edgex-go/blob/72ebc617e5544205d3e97b6458656f707fb7934c/api/raml/core-metadata.raml#L72

In other places, BSON IDs are used in the route examples

https://github.com/edgexfoundry/edgex-go/blob/72ebc617e5544205d3e97b6458656f707fb7934c/api/raml/core-metadata.raml#L239

We should clean these up and make them UUIDs

tsconn23:
Fix #1918

### DIFF
--- a/api/raml/core-command.raml
+++ b/api/raml/core-command.raml
@@ -45,7 +45,7 @@ schemas:
                 description: The service's metrics as JSON document
 /v1/device/{id}/command/{commandid}:
     displayName: Issue command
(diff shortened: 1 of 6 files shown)


tsconn23:
recheck

PR #1945, opened 2019-10-17
Update to obtain DB credentials from secret store

### LINKED ISSUE #1341: retrieving credentials ( database passwords, account credentials etc.) from secret service.
Currently the database passwords and other sensitive information are saved in the configuration file or consul. With new secret service added into the edgex architecture, it is needed to keep these credentials in the secret service and retrieve them from there to enhance the security. This applies to all the core microservices that keep sensitive information in the configuration file/consul currently.

AnthonyMBonafide:
Fix #1341

Update database.BootstrapHandler and support-logging's init.go to use
the SecretClient to obtain database credentials when security features
have been enabled.

Create helper functions for determining if security is enabled and
another for obtaining database credentials.

### DIFF
--- a/internal/pkg/bootstrap/container/credentials.go
+++ b/internal/pkg/bootstrap/container/credentials.go
@@ -1,5 +1,5 @@
-/*******************************************************************************
- * Copyright 2019 Dell Inc.
+/********************************************************************************
+ *  Copyright 2019 Dell Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -15,15 +15,15 @@
 package container
 
 import (
-	"github.com/edgexfoundry/go-mod-secrets/pkg"
-
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/interfaces"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
 )
 
-// SecretClientName contains the name of the registry.Client implementation in the DIC.
-var SecretClientName = di.TypeInstanceToName((*pkg.SecretClient)(nil))
+// CredentialsProviderName contains the name of the interfaces.CredentialsProvider implementation in the DIC.
+var CredentialsProviderName = di.TypeInstanceToName((*interfaces.CredentialsProvider)(nil))
 
-// SecretClientFrom helper function queries the DIC and returns the pkg.SecretClient implementation.
-func SecretClientFrom(get di.Get) pkg.SecretClient {
-	return get(SecretClientName).(pkg.SecretClient)
+// CredentialsProviderFrom helper function queries the DIC and returns the interfaces.CredentialsProviderName
+// implementation.
+func CredentialsProviderFrom(get di.Get) interfaces.CredentialsProvider {
+	return get(CredentialsProviderName).(interfaces.CredentialsProvider)
 }
--- a/internal/pkg/bootstrap/handlers/database/database.go
+++ b/internal/pkg/bootstrap/handlers/database/database.go
@@ -23,6 +23,7 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/container"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/interfaces"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/startup"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 	dbInterfaces "github.com/edgexfoundry/edgex-go/internal/pkg/db/interfaces"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db/mongo"
@@ -63,7 +64,7 @@ func NewDatabaseForCoreData(httpServer httpServer, database interfaces.Database)
 }
 
 // Return the dbClient interface
-func (d Database) newDBClient(loggingClient logger.LoggingClient) (dbInterfaces.DBClient, error) {
+func (d Database) newDBClient(loggingClient logger.LoggingClient, credentials config.Credentials) (dbInterfaces.DBClient, error) {
 	databaseInfo := d.database.GetDatabaseInfo()["Primary"]
 	switch databaseInfo.Type {
 	case db.MongoDB:
@@ -73,8 +74,8 @@ func (d Database) newDBClient(loggingClient logger.LoggingClient) (dbInterfaces.
 				Port:         databaseInfo.Port,
 				Timeout:      databaseInfo.Timeout,
 				DatabaseName: databaseInfo.Name,
-				Username:     databaseInfo.Username,
-				Password:     databaseInfo.Password,
+				Username:     credentials.Username,
+				Password:     credentials.Password,
 			})
 	case db.RedisDB:
 		if d.isCoreData {
@@ -100,11 +101,23 @@ func (d Database) BootstrapHandler(
 
 	loggingClient := container.LoggingClientFrom(dic.Get)
 
+	// get database credentials.
+	var credentials config.Credentials
+	for startupTimer.HasNotElapsed() {
+		var err error
+		credentials, err = container.CredentialsProviderFrom(dic.Get).GetDatabaseCredentials(d.database.GetDatabaseInfo()["Primary"])
+		if err == nil {
+			break
+		}
+		loggingClient.Warn(fmt.Sprintf("couldn't retrieve database credentials: %v", err.Error()))
+		startupTimer.SleepForInterval()
+	}
+
 	// initialize database.
 	var dbClient dbInterfaces.DBClient
 	for startupTimer.HasNotElapsed() {
 		var err error
-		dbClient, err = d.newDBClient(loggingClient)
+		dbClient, err = d.newDBClient(loggingClient, credentials)
 		if err == nil {
 			break
 		}
--- a/internal/pkg/bootstrap/handlers/secret/secret.go
+++ b/internal/pkg/bootstrap/handlers/secret/secret.go
@@ -16,6 +16,7 @@ package secret
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"sync"
 
@@ -31,42 +32,41 @@ import (
 	"github.com/edgexfoundry/go-mod-secrets/pkg/providers/vault"
 )
 
-// SecretClientBootstrapHandler creates a secretClient to be used for obtaining secrets from a secret store manager.
+type SecretProvider struct {
+	secretClient pkg.SecretClient
+}
+
+func NewSecret() *SecretProvider {
+	return &SecretProvider{}
+}
+
+// SecretClientBootstrapHandler creates a secretClient to be used for obtaining secrets from a SecretProvider store manager.
 // NOTE: This BootstrapHandler is responsible for creating a utility that will most likely be used by other
 // BootstrapHandlers to obtain sensitive data, such as database credentials. This BootstrapHandler should be processed
 // before other BootstrapHandlers, possibly even first since it has not other dependencies.
-func BootstrapHandler(
+func (s *SecretProvider) BootstrapHandler(
 	wg *sync.WaitGroup,
 	context context.Context,
 	startupTimer startup.Timer,
 	dic *di.Container) bool {
+	loggingClient := container.LoggingClientFrom(dic.Get)
 
-	// check for environment variable that turns security off
-	if env := os.Getenv("EDGEX_SECURITY_SECRET_STORE"); env == "false" {
-		dic.Update(di.ServiceConstructorMap{
-			container.SecretClientName: func(get di.Get) interface{} {
-				return nil
-			},
-		})
-		return true
-	}
-
-	// attempt to create a new secret client
+	// attempt to create a new SecretProvider client
 	configuration := container.ConfigurationFrom(dic.Get)
-	scc, err := getSecretConfig(configuration.GetBootstrap().SecretStore)
+	secretConfig, err := s.getSecretConfig(configuration.GetBootstrap().SecretStore)
 	if err != nil {
+		loggingClient.Error(fmt.Sprintf("unable to parse secret store configuration: %s", err.Error()))
 		return false
 	}
 
-	secretClient, err := vault.NewSecretClient(scc)
-
-	var result *pkg.SecretClient
-	if err == nil {
-		result = &secretClient
+	s.secretClient, err = vault.NewSecretClient(secretConfig)
+	if err != nil {
+		loggingClient.Error(fmt.Sprintf("unable to create SecretClient: %s", err.Error()))
 	}
+
 	dic.Update(di.ServiceConstructorMap{
-		container.SecretClientName: func(get di.Get) interface{} {
-			return result
+		container.CredentialsProviderName: func(get di.Get) interface{} {
+			return s
 		},
 	})
 
@@ -75,7 +75,7 @@ func BootstrapHandler(
 
 // getSecretConfig creates a SecretConfig based on the SecretStoreInfo configuration properties.
 // If a tokenfile is present it will override the Authentication.AuthToken value.
-func getSecretConfig(secretStoreInfo config.SecretStoreInfo) (vault.SecretConfig, error) {
+func (s *SecretProvider) getSecretConfig(secretStoreInfo config.SecretStoreInfo) (vault.SecretConfig, error) {
 	secretConfig := vault.SecretConfig{
 		Host:           secretStoreInfo.Host,
 		Port:           secretStoreInfo.Port,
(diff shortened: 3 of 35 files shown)


internal/pkg/bootstrap/handlers/database/database.go
```diff
@@ -63,18 +63,23 @@ func NewDatabaseForCoreData(httpServer httpServer, database interfaces.Database)
 }
 
 // Return the dbClient interface
-func (d Database) newDBClient(loggingClient logger.LoggingClient) (dbInterfaces.DBClient, error) {
+func (d Database) newDBClient(loggingClient logger.LoggingClient, dic *di.Container) (dbInterfaces.DBClient, error) {
```
michaelestrin: Don't pass container around; pull the required dependencies out of it and pass those explicitly into this method as parameters.

internal/pkg/config/environment.go
```diff
@@ -37,3 +37,9 @@ func OverrideFromEnvironment(registry RegistryInfo) RegistryInfo {
 	}
 	return registry
 }
+
+// IsSecurityEnabled determines if security has been enabled.
+func IsSecurityEnabled() bool {
+	env := os.Getenv("EDGEX_SECURITY_SECRET_STORE")
+	return env != "false"
+}
```
michaelestrin: I think this should be moved to the bootstrap secret package.

internal/pkg/db/db.go
```diff
@@ -76,3 +80,28 @@ type Configuration struct {
 func MakeTimestamp() int64 {
 	return time.Now().UnixNano() / int64(time.Millisecond)
 }
+
+// GetDatabaseCredentials retrieves the credentials to be used to connect to the underlying database.
+// If security is enabled then the credentials will be retrieved from the secret store.
+func GetDatabaseCredentials(dic *di.Container, configuration config.DatabaseInfo) (config.Credentials, error) {
```
michaelestrin: Assumes "primary"; I wonder if it would be better to pass in a single database struct (as opposed to the entire map).

internal/support/logging/init.go
```diff
@@ -39,13 +39,13 @@ func NewServiceInit(server server) ServiceInit {
 	}
 }
 
-func getPersistence() (persistence, error) {
+func getPersistence(dic *di.Container) (persistence, error) {
```
michaelestrin: Don't pass the dic around.  It's a god object intended to bridge the gap after bootstrap/init between routing and controller functions/methods.  Pull out required dependencies in the caller and pass those dependencies explicitly to getPersistence.

internal/support/logging/init.go
```diff
@@ -39,13 +39,13 @@ func NewServiceInit(server server) ServiceInit {
...
 	switch Configuration.Writable.Persistence {
 	case PersistenceFile:
 		return &fileLog{filename: Configuration.Logging.File}, nil
 	case PersistenceDB:
 		// TODO: Integrate db layer with internal/pkg/db/ types so we can support other databases
-		ms, err := connectToMongo()
+		ms, err := connectToMongo(dic)
```
michaelestrin: See above.

internal/support/logging/init.go
```diff
@@ -73,7 +73,7 @@ func (s ServiceInit) BootstrapHandler(
 	// initialize database.
 	for startupTimer.HasNotElapsed() {
 		var err error
-		dbClient, err = getPersistence()
+		dbClient, err = getPersistence(dic)
```
michaelestrin: See above.

internal/support/logging/mongo.go
```diff
@@ -22,13 +24,18 @@ type mongoLog struct {
 	session *mgo.Session // Mongo database session
 }
 
-func connectToMongo() (*mgo.Session, error) {
+func connectToMongo(dic *di.Container) (*mgo.Session, error) {
```
michaelestrin: Don't pass the dic around. It's a god object intended to bridge the gap after bootstrap/init between routing and controller functions/methods. Pull out required dependencies in the caller and pass those dependencies explicitly to connectToMongo.

internal/pkg/db/db.go
AnthonyMBonafide: I think we are limited in this regard since the `DatabaseInfo` is a type alias for a `map` where the map's key is a `string` and the value is an anonymous struct. This limits us since we cannot use that anonymous struct in the function signature. Do you see anything wrong with giving an identity to that anonymous struct, or see another way to achieve this?

https://github.com/edgexfoundry/edgex-go/blob/7d0b6b305c2451dd3125ab97d00e56bcb3189c1a/internal/pkg/config/types.go#L115-L122

michaelestrin: I can't conceive of any reason not to turn the anonymous struct into a named type.

internal/pkg/config/environment.go
```diff
@@ -37,3 +37,9 @@ func OverrideFromEnvironment(registry RegistryInfo) RegistryInfo {
 	}
 	return registry
 }
+
+// IsSecurityEnabled determines if security has been enabled.
+func IsSecurityEnabled() bool {
+	env := os.Getenv("EDGEX_SECURITY_SECRET_STORE")
+	return env != "false"
```
jim-wang-intel: IMO, from purely usage point of view, there is some issue here: `os.Getenv` cannot distinguish from `empty vaule` and `non-existing`. See [line 98-100](https://golang.org/src/os/env.go) from Golang's comments.    So in this case, `"false"` string and empty string `""` should be treated the same.  Suggest changing to use `os.LookupEnv("EDGEX_SECURITY_SECRET_STORE")` to cover both cases.

AnthonyMBonafide: This logic takes that into consideration. We are assuming security is enabled unless the environment value is present AND `false`. Therefore if the value is empty, non-existent, or blank we will get a non `false` value which we will ignore and treat security as being enabled, which is intended.

AnthonyMBonafide: Using `os.LookupEnv("EDGEX_SECURITY_SECRET_STORE")` would yield the same results just a little more verbose, as we would then inspect the `bool` value first, and if not present we would return false. If it were present we would have to verify that it is `false`. Essentially the same thing with less code.

jim-wang-intel: Got cha.  I was not sure that was the assumption and thanks for your explanation.

internal/pkg/bootstrap/handlers/database/database.go
```diff
@@ -104,7 +105,8 @@ func (d Database) BootstrapHandler(
 	var dbClient dbInterfaces.DBClient
 	for startupTimer.HasNotElapsed() {
 		var err error
-		dbClient, err = d.newDBClient(loggingClient)
+		credentials, err := db.GetDatabaseCredentials(dic, d.database.GetDatabaseInfo()["Primary"])
```
michaelestrin: You're still passing the container around.  Pull out the required dependencies and pass them explicitly.  Results in the called function/method being infinitely more testable.

AnthonyMBonafide: I was trying to not duplicate code, but it seems like we are going to have to here since we will need to check if security is enabled before getting the `SecretClient` out of the container. At that point the `GetDatabaseCredentials` helper function is useless. I will just remove that helper function.

difince: Missing check for err != nil

internal/pkg/bootstrap/handlers/database/database.go
```diff
@@ -104,7 +105,14 @@ func (d Database) BootstrapHandler(
 	var dbClient dbInterfaces.DBClient
 	for startupTimer.HasNotElapsed() {
 		var err error
-		dbClient, err = d.newDBClient(loggingClient)
+		credentialProvider := container.CredentialsFrom(dic.Get)
+		dbCredentials, err := credentialProvider.GetDatabaseCredentials(d.database.GetDatabaseInfo()["Primary"])
+		if err != nil {
+			loggingClient.Error(fmt.Sprintf("unable to get database credentials: %s", err.Error()))
+			return false
```
michaelestrin: Is this truly a fatal error that should cause the bootstrapping process to be aborted and the service to be stopped?  Will a retry potentially succeed?

internal/pkg/bootstrap/container/credentials.go
```diff
@@ -0,0 +1,28 @@
...
+
+package container
+
+import (
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/interfaces"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
+)
+
+// CredentialsName contains the name of the registry.Client implementation in the DIC.
+var CredentialsName = di.TypeInstanceToName((*interfaces.CredentialsProvider)(nil))
```
michaelestrin: Nitpick - CredentialsName -> CredentialsProviderName

internal/pkg/bootstrap/container/credentials.go
```diff
@@ -0,0 +1,28 @@
...
+
+// CredentialsFrom helper function queries the DIC and returns the secret.CredentialsName implementation.
+func CredentialsFrom(get di.Get) interfaces.CredentialsProvider {
```
michaelestrin: Nitpick - CredentialsFrom -> CredentialsProviderFrom

internal/pkg/bootstrap/handlers/secret/secret.go
```diff
@@ -31,51 +32,51 @@ import (
...
-	}
 	dic.Update(di.ServiceConstructorMap{
-		container.SecretClientName: func(get di.Get) interface{} {
-			return result
+		container.CredentialsName: func(get di.Get) interface{} {
+			return s
 		},
 	})
 
+	if err != nil {
```
michaelestrin: This should be moved up to be immediately under `s.secretClient, err = vault.NewSecretClient(secretConfig)`

internal/pkg/bootstrap/handlers/secret/secret.go
```diff
@@ -103,3 +104,9 @@ func getSecretConfig(secretStoreInfo config.SecretStoreInfo) (vault.SecretConfig
 
 	return secretConfig, nil
 }
+
+// IsSecurityEnabled determines if security has been enabled.
+func (s *SecretProvider) IsSecurityEnabled() bool {
```
michaelestrin: Should this method be public?

internal/support/logging/init.go
```diff
@@ -73,7 +74,8 @@ func (s ServiceInit) BootstrapHandler(
 	// initialize database.
 	for startupTimer.HasNotElapsed() {
 		var err error
-		dbClient, err = getPersistence()
+		credentials, err := container.CredentialsFrom(dic.Get).GetDatabaseCredentials(Configuration.Databases["Primary"])
+		dbClient, err = getPersistence(credentials)
```
michaelestrin: You're not dealing with err result from the first line before overwriting it in the second.

internal/pkg/bootstrap/handlers/database/database.go
AnthonyMBonafide: I do not think we have enough information to determine if the type of error can be resolved with a retry, so I think we can still retry since we are protected by the time limit. Overall, I think it would be best if the errors had some more information where we make a better decision.

Here are the errors that will eventually bubble up: https://github.com/edgexfoundry/go-mod-secrets/blob/master/pkg/types.go

internal/pkg/bootstrap/handlers/secret/secret.go
AnthonyMBonafide: No, this does not need to be exported. Good catch.

internal/pkg/bootstrap/handlers/database/database.go
AnthonyMBonafide: Also, I separated the retrieval of credentials into its own retry-loop, so that if we fail to connect to the database we do not throw away the credentials that we already obtained.

internal/pkg/bootstrap/container/credentials.go
```diff
@@ -0,0 +1,28 @@
...
+ *******************************************************************************/
+
+package container
+
+import (
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/interfaces"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
+)
+
+// CredentialsProviderName contains the name of the registry.Client implementation in the DIC.
```
michaelestrin: `registry.Client` should be `interfaces.CredentialsProvider`

internal/pkg/bootstrap/container/credentials.go
```diff
@@ -0,0 +1,28 @@
...
+var CredentialsProviderName = di.TypeInstanceToName((*interfaces.CredentialsProvider)(nil))
+
+// CredentialsProviderFrom helper function queries the DIC and returns the secret.CredentialsProviderName implementation.
```
michaelestrin: `secret.CredentialsProviderName` should be `interfaces.CredentialsProviderName`

internal/pkg/bootstrap/handlers/database/database.go
```diff
@@ -100,11 +101,22 @@ func (d Database) BootstrapHandler(
...
 
+	// get database credentials.
+	var credentials config.Credentials
+	for startupTimer.HasNotElapsed() {
+		var err error
+		credentials, err = container.CredentialsProviderFrom(dic.Get).GetDatabaseCredentials(d.database.GetDatabaseInfo()["Primary"])
+		if err == nil {
+			break
+		}
+		startupTimer.SleepForInterval()
```
michaelestrin: You should log the error when err != nil -- similar to what's being done below when d.NewDBClient() returns an err.

internal/support/logging/init.go
```diff
@@ -70,10 +71,21 @@ func (s ServiceInit) BootstrapHandler(
...
 
+	// get database credentials.
+	var credentials config.Credentials
+	for startupTimer.HasNotElapsed() {
+		var err error
+		credentials, err = container.CredentialsProviderFrom(dic.Get).GetDatabaseCredentials(Configuration.Databases["Primary"])
+		if err == nil {
+			break
+		}
+		startupTimer.SleepForInterval()
```
michaelestrin: You should log the error when err != nil -- similar to what's being done below when getPersistence() returns an err.

AnthonyMBonafide:
> There is an issue here. In the configuration file we define path as:
> /v1/secret/edgex/core-data/. However the credential needs to be retrieved at
> /v1/secret/edgex/core-data/mongodb. Here is an example of request/response with curl:
> curl -k --header "X-Vault-Token: s.JCgXTjP7fmUnt5YAvPy73XCG" https://127.0.0.1:8200/v1/secret/edgex/coredata/mongodb
> 
> {"request_id":"33d28afc-e423-6996-d294-4ef764b2a646","lease_id":"","renewable":false,"lease_duration":604800,"data":{"password":"w94M{\L0"Vqt/maB","username":"core"},"wrap_info":null,"warnings":null,"auth":null}

We are retrieving the secrets at `/v1/secret/edgex/core-data/mongodb`. We are providing the `/v1/secret/edgex/core-data/` portion in the `path` configuration property:

https://github.com/edgexfoundry/edgex-go/blob/6815423dab2a6ecf1be83452d991cbd123005ee5/cmd/core-data/res/configuration.toml#L61

and then appending the `mongodb` portion when we request the `SecretClient` to get the secrets:

https://github.com/edgexfoundry/edgex-go/blob/68a9646b7fb56197e3f3021ca32ca4eab73f93d7/internal/pkg/bootstrap/handlers/secret/credentials.go#L33

where `database.Type` is `mongodb.

The final result will actually be the desired `/v1/secret/edgex/core-data/mongodb`, so I do not think we need to do anything here.

mkbhanda:
I recall the "path" discussion in the secret client. @tingyuz would it help to add some documentation stating differences between pure curl request and using the secret client.

tingyuz:
> > There is an issue here. In the configuration file we define path as:
> > /v1/secret/edgex/core-data/. However the credential needs to be retrieved at
> > /v1/secret/edgex/core-data/mongodb. Here is an example of request/response with curl:
> > curl -k --header "X-Vault-Token: s.JCgXTjP7fmUnt5YAvPy73XCG" https://127.0.0.1:8200/v1/secret/edgex/coredata/mongodb
> > {"request_id":"33d28afc-e423-6996-d294-4ef764b2a646","lease_id":"","renewable":false,"lease_duration":604800,"data":{"password":"w94M{\L0"Vqt/maB","username":"core"},"wrap_info":null,"warnings":null,"auth":null}
> 
> We are retrieving the secrets at `/v1/secret/edgex/core-data/mongodb`. We are providing the `/v1/secret/edgex/core-data/` portion in the `path` configuration property:
> 
> https://github.com/edgexfoundry/edgex-go/blob/6815423dab2a6ecf1be83452d991cbd123005ee5/cmd/core-data/res/configuration.toml#L61
> 
> and then appending the `mongodb` portion when we request the `SecretClient` to get the secrets:
> 
> https://github.com/edgexfoundry/edgex-go/blob/68a9646b7fb56197e3f3021ca32ca4eab73f93d7/internal/pkg/bootstrap/handlers/secret/credentials.go#L33
> 
> where `database.Type` is `mongodb.
> 
> The final result will actually be the desired `/v1/secret/edgex/core-data/mongodb`, so I do not think we need to do anything here.

Thats great! I will go ahead and send me approval soon.

AnthonyMBonafide:
@tingyuz , has asked me to remove the hyphens in the `path` configuration. I will remove these and update this branch.

AnthonyMBonafide:
Rebased and updated with removal of the hyphen in the `path` configuration property.

michaelestrin:
Currently security versions of the blackbox test Jenkins builds are broken.  I believe this PR is needed to resolve the issue.  Can we expedite the review and merge of this PR?

PR #1953, opened 2019-10-17
Add git rev to snap version

knitzsche:
version part override-pull cds to root so git rev provides the required data to set the version

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -347,7 +347,7 @@ parts:
       cd $SNAPCRAFT_PROJECT_DIR
       PROJECT_VERSION=$(cat VERSION)
       GIT_REVISION=$(git rev-parse --short HEAD)
-      snapcraftctl set-version ${PROJECT_VERSION}-$(date +%Y%m%d)+${GIT_REVISION}}
+      snapcraftctl set-version ${PROJECT_VERSION}-$(date +%Y%m%d)+${GIT_REVISION}
   curl:
     plugin: nil
     # the default source for a part that doesn't specify one is ".", which
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -340,9 +340,14 @@ apps:
 parts:
   version:
     plugin: nil
-    source: .
+    # as with curl plugin, the source dir is unrelated to this part and is used
+    # since it changes rarely and therefore will not trigger a new pull
+    source: snap/local/build-helpers
     override-pull: |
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -340,9 +340,14 @@ apps:
...
-    source: .
+    # as with curl plugin, the source dir is unrelated to this part and is used
+    # since it changes rarely and therefore will not trigger a new pull
+    source: snap/local/build-helpers
     override-pull: |
-      snapcraftctl set-version $(cat $SNAPCRAFT_PROJECT_DIR/VERSION)-$(date +%Y%m%d)+$(git rev-parse --short HEAD)
+      cd $SNAPCRAFT_PROJECT_DIR
+      PROJECT_VERSION=$(cat VERSION)
+      GIT_REVISION=$(git rev-parse --short HEAD)
+      snapcraftctl set-version ${PROJECT_VERSION}-$(date +%Y%m%d)+${GIT_REVISION}}
```
anonymouse64: same as the master branch, need to drop the last `}`:

```suggestion
      snapcraftctl set-version ${PROJECT_VERSION}-$(date +%Y%m%d)+${GIT_REVISION}
```

anonymouse64:
Looks like this is working:

```
15:24:07 Snapping 'edgexfoundry' ...
15:25:51 
15:25:51 Snapped edgexfoundry_1.0.2-20191018+0247237b_amd64.snap
```

Will merge after the arm64 test is done

anonymouse64:
@jamesrgregg out of curiosity why was this PR added to the security WG project?

jamesrgregg:
it was moved to the icebox on the project tracker so it could be dispositioned between either core or security wgs
it was tagged as edinburgh so it seemed like it was late and by moving it to the icebox column.. it was a means to start the dialog 
it's good that this spawned some action on your part because it seems this issue was addressed.

anonymouse64:
I guess my question was more directed towards putting it on the _security_ board instead of the _core_ board. This doesn't have anything to do with the security WG work.

Note that this PR is just changing packaging of the snap, which was recently broken (hence this PR to fix it), so if there was a devops/packaging board that would be more appropriate for this I think (this should also be true of any other PR tagged "snap").

jamesrgregg:
hey @anonymouse64 use the tags or create new ones so we can make sure that the issues can be organized on the kanban board so as to not create confusion.
thanks for taking action on this one

anonymouse64:
Sorry I saw the board had a "lock" on it, so I didn't realize I have perms to change things there, it appears I do have perms to move things though.

Thanks!

jamesrgregg:
the tracker for security is a top level org tracker found here - https://github.com/orgs/edgexfoundry/projects/28

PR #1959, opened 2019-10-18
snap: use postgres in the snap for kong's DB

### LINKED ISSUE #1222: Switch to using postgres in the snap
The snapd team has [started work](https://forum.snapcraft.io/t/phase-1-of-opt-in-per-snap-users-groups-aka-the-daemon-user/10624) on implementing support for snap daemons to privilege drop to another user which would allow us to bundle postgres in the snap to use with Kong instead of Cassandra as we currently do. This is preferable because postgres is lighter weight, requiring less CPU and RAM than Cassandra as well as a smaller disk foot print, making the snap smaller in size.

However, it's not currently clear if this work will be completed before Edinburgh freeze, so this may happen after Edinburgh, but it would be quite nice for the snap to be able to switch to this before hand, so if we can switch we should.

anonymouse64:
This removes cassandra from the snap and adds postgres in it's place. 

In addition to being generally faster, this also:
- decreases the size of the snap from 390M to 335M,. :tada: 
- decreases RAM usage from ~75% to ~35% (this is in a single core, 1 GB RAM Ubuntu 19.10 Eoan Ermine KVM) :tada: 

To test this, build + install the snap in devmode, then check:

- [ ] that the postgres services starts up and stays running:
```
$ snap services edgexfoundry.postgres 
Service                Startup  Current  Notes
edgexfoundry.postgres  enabled  active   -
```
- [ ] that kong-daemon starts up and stays running:
```
$ snap services edgexfoundry.kong-daemon
Service                   Startup  Current  Notes
edgexfoundry.kong-daemon  enabled  active   -
```
- [ ] that the kong db is present in postgres using psql:
```
$ sudo edgexfoundry.psql -d kong
psql (10.10 (Ubuntu 10.10-0ubuntu0.18.04.1))
Type "help" for help.

kong=# \l
                                  List of databases
   Name    |    Owner    | Encoding | Collate |  Ctype  |      Access privileges      
-----------+-------------+----------+---------+---------+-----------------------------
 kong      | snap_daemon | UTF8     | C.UTF-8 | C.UTF-8 | 
 postgres  | snap_daemon | UTF8     | C.UTF-8 | C.UTF-8 | 
 template0 | snap_daemon | UTF8     | C.UTF-8 | C.UTF-8 | =c/snap_daemon             +
           |             |          |         |         | snap_daemon=CTc/snap_daemon
 template1 | snap_daemon | UTF8     | C.UTF-8 | C.UTF-8 | =c/snap_daemon             +
           |             |          |         |         | snap_daemon=CTc/snap_daemon
(4 rows)

kong=# \c kong
You are now connected to database "kong" as user "snap_daemon".
kong=# \dt
                   List of relations
 Schema |             Name              | Type  | Owner 
--------+-------------------------------+-------+-------
 public | acls                          | table | kong
 public | apis                          | table | kong
 public | basicauth_credentials         | table | kong
 public | certificates                  | table | kong
 public | cluster_ca                    | table | kong
 public | cluster_events                | table | kong
 public | consumers                     | table | kong
 public | hmacauth_credentials          | table | kong
 public | jwt_secrets                   | table | kong
 public | keyauth_credentials           | table | kong
 public | locks                         | table | kong
 public | oauth2_authorization_codes    | table | kong
 public | oauth2_credentials            | table | kong
 public | oauth2_tokens                 | table | kong
 public | plugins                       | table | kong
 public | ratelimiting_metrics          | table | kong
 public | response_ratelimiting_metrics | table | kong
 public | routes                        | table | kong
 public | schema_meta                   | table | kong
 public | services                      | table | kong
 public | snis                          | table | kong
 public | targets                       | table | kong
 public | ttls                          | table | kong
 public | upstreams                     | table | kong
(24 rows)

kong=# 
```
- [ ] that turning on/off the security-api-gateway turns on/off postgres
```
$ snap set edgexfoundry security-api-gateway=off
$ snap services edgexfoundry.postgres 
Service                Startup   Current   Notes
edgexfoundry.postgres  disabled  inactive  -
$ snap services edgexfoundry.kong-daemon
Service                   Startup   Current   Notes
edgexfoundry.kong-daemon  disabled  inactive  -
$ snap set edgexfoundry security-api-gateway=on
$ snap services edgexfoundry.postgres 
Service                Startup  Current  Notes
edgexfoundry.postgres  enabled  active   -
$ snap services edgexfoundry.kong-daemon
Service                   Startup  Current  Notes
edgexfoundry.kong-daemon  enabled  active   -
```
Note that technically the same test should be performed for security-secret-store, but that actually doesn't work properly due to https://github.com/edgexfoundry/edgex-go/issues/1963 which I just found, however that seems to be an existing issue and is not related to my PR here, so let's fix that later. There's also the possibility that the security config for Fuji may change considerably, so it might not even be worth fixing that now.

Fixes #1222


tonyespy:
When I install the snap, I'm getting the following error:

`error: exec: "/snap/edgexfoundry/x1/bin/psql-wrapper.sh": stat: /snap/edgexfoundry/x1/bin/psql-wrapper.sh: no such file or directory`

snap/snapcraft.yaml
```diff
@@ -359,6 +436,56 @@ parts:
...
       - curl
+
+  # snapcraft preload is necessary to make postgres just use a different
+  # lockfile location in /dev/shm
+  # NOTE: if this ever breaks in really scary ways when compiling with lots of
+  # warnings, see the comment on the kong part, tldr probably some other C/C++
+  # part in the build broke snapcraft-preload by running before
+  # snapcraft-preload
+  snapcraft-preload:
+    source: https://github.com/sergiusens/snapcraft-preload.git
```
tonyespy: As there've been no releases of snapcraft-preload, should we pin this to a specific commit?  The README for this code doesn't fully explain how it works, especially with respect to the environment variable SNAPCRAFT_PRELOAD_REDIRECT_ONLY_SHM.

snap/snapcraft.yaml
```diff
@@ -359,6 +436,56 @@ parts:
...
+          - g++-multilib-arm-linux-gnueabihf
+          - gcc-multilib-arm-linux-gnueabihf
+      - else:
+          - gcc-multilib
+          - g++-multilib
+    stage-packages:
+      - to amd64:
+          - lib32stdc++6
+  gosu:
+    source: https://github.com/tianon/gosu.git
```
tonyespy: Can we pin this to the latest release (1.11)?

snap/hooks/install
```diff
@@ -135,11 +134,43 @@ if [ ! -f "$SNAP_DATA/config/security-api-gateway/kong.conf" ]; then
...
+gosu snap_daemon "$SNAP/usr/lib/postgresql/10/bin/initdb" -D "$SNAP_DATA/postgresql/10/main"
+
+# ensure the sockets dir exists and is properly owned
+mkdir -p "$SNAP_COMMON/sockets"
+chown -R snap_daemon:snap_daemon "$SNAP_COMMON/sockets" 
+
+# start postgres up and wait a bit for it so we can create the database and user
+# for kong
+snapctl start "$SNAP_NAME.postgres"
+sleep 2
```
tonyespy: Is there a cleaner approach to this? Problems related to sleep commands are notoriously hard to track down if/when they go wrong...

snap/snapcraft.yaml
anonymouse64: sure I'll use https://github.com/sergiusens/snapcraft-preload/commit/d654bbe8a5add5a4bea14d96342b656990c5c818 since that's the most recent commit and also the one that adds `SNAPCRAFT_PRELOAD_REDIRECT_ONLY_SHM`, but also agreed there's not a lot of documentation for snapcraft-preload.

snap/hooks/install
anonymouse64: :thinking: well the problem is that we need to run the postgres commands below to bootstrap postgres for kong, which talk to the postgres daemon, so I guess we could loop trying to run the commands until they succeed instead with the sleep there instead. Let me try that out

snap/snapcraft.yaml
anonymouse64: done

snap/snapcraft.yaml
anonymouse64: done

anonymouse64:
> One question about your choice to include gosu. Did you also investigate whether chroot or setpriv could be used instead? Both to me would be preferable to gosu if functionally equivalent, as their both source-able from the archive.

chroot doesn't work without modification to it's usage of `setgroups()`, i.e. 

```
$ snap install test-postgres-system-usernames --edge
$ sudo snap run --shell test-postgres-system-usernames.psql-any
# chroot --userspec=snap_daemon:snap_daemon /
chroot: failed to set supplemental groups: Operation not permitted
```
This also triggers a Seccomp denial:

```
kernel: audit: type=1326 audit(1571694863.553:20137): auid=1000 uid=0 gid=0 ses=3 pid=82701 comm="chroot" exe="/usr/sbin/chroot" sig=0 arch=c000003e syscall=116 compat=0 ip=0x7f74f523ce65 code=0x50000
```

Where syscall 116 is `setgroups`.
I didn't originally investigate using setpriv but I just did and setpriv fails the same way, producing an identical denial:

```
audit: type=1326 audit(1571696348.142:34856): auid=1000 uid=584788 gid=584788 ses=2 pid=109529 comm="setpriv" exe="/snap/postgres/x1/usr/bin/setpriv" sig=0 arch=c000003e syscall=116 compat=0 ip=0x7f672b263e65 code=0x50000
```

See also jdstrand's doc on this: https://snapcraft.io/docs/system-usernames. TLDR, `chroot` and other stdlib functions _could_ be made to do this but `setgroups()` usage needs to be patched to use the "non-portable, Linux-specific, invocation _size 0_ and _NULL list_ to clear the list completely". 

Due to this, I went with something that doesn't use the stdlib `setgroups()` at all, i.e. static Go.

snap/hooks/install
anonymouse64: yes looping works fine, I went with this approach

snap/hooks/install
```diff
@@ -135,11 +134,60 @@ if [ ! -f "$SNAP_DATA/config/security-api-gateway/kong.conf" ]; then
...
+        echo "failed to create kong user in postgres database after $iter_num iterations"
+        exit 1
+    fi
+done
+iter_num=0
+until gosu snap_daemon "$SNAP/bin/perl5lib-launch.sh" "$SNAP/usr/bin/createdb" kong; do
+    sleep 1
+    iter_num=$(( iter_num + 1 ))
+    if [ $iter_num -gt $MAX_POSTGRES_INIT_ITERATIONS ]; then
+        echo "failed to create kong user in postgres database after $iter_num iterations"
```
tonyespy: shouldn't this say "db" instead of "user"?

anonymouse64:
> Also re: the gosu usage. I read jdstrand's documentation, but it mentions nothing about sudo, but instead focuses on the system calls an application would use to drop privs. Is the issue with postgres that it doesn't have this code builtin already? In the most most daemons I've worked with that do this, just require the user name to be configured, and the daemon drops privs itself after starting as root.

1. Postgres the daemon doesn't drop by itself
2. Sudo does not and will not ever work inside a snap by design
3. The other issue with dropping privileges is that we need to drop privileges because snapd doesn't drop privileges for us and the other tools like psql, createuser, etc. don't drop privileges by themselves.

anonymouse64:
> Note - I still find the use of ld_preload and the afore-mentioned env var SNAPCRAFT_PRELOAD_REDIRECT_ONLY_SHM black magic, which might warrant a better explanation via comments.

Okay I'll add another comment about why we need it in the snapcraft.yaml

snap/hooks/install
anonymouse64: Yes... Sorry will fix this

anonymouse64:
Note this PR is blocked waiting for #1975 to land and unbreak the snap builds

anonymouse64:
recheck

Issue #1851, opened 2019-10-01
SMA should validate parameters passed to the service executors


anonymouse64:
Continuing from https://github.com/edgexfoundry/edgex-go/issues/770#issuecomment-536821382 ...

The SMA should validate parameters passed to it via HTTP before passing those parameters to the executors. 

> By definition, parameters are passed through the agent to an executor as-is without verification.

Sorry I haven't been attending the SMA meetings regularly recently, but where was this decided? 

> I'd propose we leave it to the executor to validate and ignore incoming parameters based on the requirements of its specific implementation.

I disagree with this choice specifically because this is exactly the kind of avenue for privilege escalation against the docker socket that I think could be our first critical remote code execution CVE we get hit with after discovering some set of docker options to `start`/`stop`/`restart` allows executing arbitrary containers. 

I also want to apologize for approving the original PR without having thought more deeply about this issue, if folks don't think we can fix this issue specifically in time, I'd like us to revert that PR since I think it's just too big of a security risk to allow arbitrary options to be passed over HTTP to the docker command.

michaelestrin:
> The SMA should validate parameters passed to it via HTTP before passing those parameters to the executors.

I disagree for the reasons stated on #770 and below.

> > By definition, parameters are passed through the agent to an executor as-is without verification.
> 
> Sorry I haven't been attending the SMA meetings regularly recently, but where was this decided?

It's implied.  Unless there's a standard set of optional parameters that each executor implementation is obligated to support, the optional parameters are executor implementation-specific.   This was discussed in passing during today's SMA WG call and consensus seemed to reinforce this and the approach of executor-validation of these parameters.

> > I'd propose we leave it to the executor to validate and ignore incoming parameters based on the requirements of its specific implementation.
> 
> I disagree with this choice specifically because this is exactly the kind of avenue for privilege escalation against the docker socket that I think could be our first critical remote code execution CVE we get hit with after discovering some set of docker options to `start`/`stop`/`restart` allows executing arbitrary containers.

Please explain how modifying the Docker executor implementation to validate and ignore incoming parameters will lead to security risk.  I don't understand.

> I'd like us to revert that PR since I think it's just too big of a security risk to allow arbitrary options to be passed over HTTP to the docker command.

Let's fix the problem -- after we agree on a solution -- instead of reverting the PR.  

[...]

PR #1960, opened 2019-10-18
Revert 1847

anonymouse64:
This reverts commit 82e87b26c3ad337de0a3e78cb0efff640788bae8.

See https://github.com/edgexfoundry/edgex-go/issues/1851#issuecomment-543806893

Note that while there is an equivalent PR reverting the go-mod-core-contracts change (see https://github.com/edgexfoundry/go-mod-core-contracts/pull/178), this PR does not depend on that PR since without that revert, we just end up ignoring the Parameters field here. Regardless, we should still remove the Parameters field there too until there's consensus on how to proceed for this feature.

This and https://github.com/edgexfoundry/go-mod-core-contracts/pull/178 will close https://github.com/edgexfoundry/edgex-go/issues/1851

### DIFF
--- a/internal/system/agent/executor/executor-stub.go
+++ b/internal/system/agent/executor/executor-stub.go
@@ -40,7 +40,7 @@ func NewStub(results []stubCall) Stub {
 
 // CommandExecutor provides the common callout to the configuration-defined executor.  This is a stub implementation of
 // the CommandExecutor interface.
-func (m *Stub) CommandExecutor(executorPath, serviceName, operation string, parameters []string) (string, error) {
+func (m *Stub) CommandExecutor(executorPath, serviceName, operation string) (string, error) {
 	m.Called++
 	m.capturedArgs = append(m.capturedArgs, []string{executorPath, serviceName, operation})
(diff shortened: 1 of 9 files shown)


michaelestrin:
https://github.com/edgexfoundry/edgex-go/blob/master/internal/system/executor/README.md will also need to be updated to be consistent with these changes.

anonymouse64:
@michaelestrin since that's just a docs change can we do that in a follow-up PR to ensure this is merged ASAP before the freeze next week?

michaelestrin:
Seems to me you'd want the documentation to match the implementation.  Since both reside in the same repository, it would make sense to me to change both at the same time.

anonymouse64:
> Seems to me you'd want the documentation to match the implementation. Since both reside in the same repository, it would make sense to me to change both at the same time.

Okay, I have updated the docs for the executor too. @michaelestrin and/or @akramtexas can you review the docs change since y'all committed to these docs recently?

michaelestrin:
LGTM pending successful builds.

PR #1962, opened 2019-10-18
Correct usage of MarshalJSON methods in test code

jdharms:
Corrects two unit tests explicitly using DeviceProfile.MarshalJSON()
method instead of using json.Marshal(dp)

MarshalJSON() is the method that needs to be implemented for a type to satisfy the "Marshaler" interface in the json standard library package.  Passing the Marshaler to json.Marshal() insulates us from changes around whether or not the MarshalJSON() method gets overridden in the future.

This change is required for edgexfoundry/go-mod-core-contracts#143 to be resolved, but won't break anything in the meantime.


jdharms:
Blackbox tests *shouldn't* be impacted because these are only changes to unit tests.  That being said, I welcome the sanity checking :)

Rebased and pushed.

tsconn23:
Closing as this change is included, along with bump in go-mod-core-contract version number, in #1967

PR #1969, opened 2019-10-20
Fix security-secretstore-setup secret store initialization bug.

### LINKED ISSUE #1968: Security-secretstore-setup cannot init secret store due to opening of resp-init.json readonly
# 🐞 Bug Report

### Affected Services
Affects security-secretstore-setup.

When initializing the secret store for the first time, it is necessary to have read/write access to resp-init.json.  The current implementation, ported over from security-secret-store repo with changes, contains a bug wherein the resp-init.json cannot be written to, only read.  This causes secret store initialization to fail.

## 🔥 Exception or Error
`security-secretstore-setup` contains the following logic:

<pre><code>
tokenFile, err := os.Open(absPath)
...
defer tokenFile.Close()
...
_, err := vc.Init(secretstore.Configuration.SecretService, tokenFile)
if err == nil {
	_, err = vc.Unseal(secretstore.Configuration.SecretService, tokenFile)
...
</code></pre>

The `os.Open` method opens the file in read-only mode, which causes `vc.Init()` to fail to write the file.  Moreover, the file pointer must be reset to the start of the file prior to `vc.Unseal()` or the unsealing code will run into an EOF situation when trying to unseal the previously init-ed secret store.

## 🌍  Your Environment
**Deployment Environment:**
Ubuntu 16.04

**EdgeX Version:**
Fuji pre-release

**Anything else relevant?**
No

cc: @jdharms

bnevis-i:
Security-secretstore-setup opens resp-init.json file in
read-only mode prior to checking if secret store
is initialized.  If the secret store must be initialized,
the code is unable to write the init response to disk.

This fix opens resp-init.json in read/write mode initially,
and during the init+unseal flow, resets the file pointer
prior to unsealing.

Fixes #1968

### DIFF
--- a/cmd/security-secretstore-setup/main.go
+++ b/cmd/security-secretstore-setup/main.go
@@ -94,7 +94,7 @@ func main() {
 	for shouldContinue := true; shouldContinue; {
 		// Anonymous function used to prevent file handles from accumulating
 		func() {
-			tokenFile, err := os.Open(absPath)
+			tokenFile, err := os.OpenFile(absPath, os.O_CREATE|os.O_RDWR, 0600)
 			if err != nil {
 				secretstore.LoggingClient.Error(fmt.Sprintf("unable to open token file at %s", absPath))
 				os.Exit(1)
@@ -113,6 +113,7 @@ func main() {
 				secretstore.LoggingClient.Info(fmt.Sprintf("vault is not initialized (status code: %d). Starting initialisation and unseal phases", sCode))
 				_, err := vc.Init(secretstore.Configuration.SecretService, tokenFile)
 				if err == nil {
+					tokenFile.Seek(0, 0) // Read starting at beginning
 					_, err = vc.Unseal(secretstore.Configuration.SecretService, tokenFile)
 					if err == nil {
 						shouldContinue = false


bnevis-i:
# Testing instructions

Check out the security-secret-store Git repository to another directory and run the following:

```
docker-compose down ; docker volume prune ; docker-compose up -d vault
```

The `docker volume prune` ensures that the test case starts with an empty Vault

This will bring up Vault and its dependencies but not initialize the secret store.

Then then compile and run security-secretstore-setup

```
make -C ../.. cmd/security-secretstore-setup/security-secretstore-setup 
./security-secretstore-setup --confdir testdata --init=true --insecureSkipVerify=true
```

There `testdata/configuration.toml` contains below or equivalent:
```
[Writable]
LogLevel = 'DEBUG'

[Logging]
EnableRemote = false
File = './logs/security-secretstore-setup.log'

[SecretService]
scheme = "https"
server = "localhost"
port = 8200
certpath = "v1/secret/edgex/pki/tls/edgex-kong"
cafilepath = "EdgeXFoundryCA.pem"
certfilepath = "edgex-kong.pem"
keyfilepath = "edgex-kong.priv.key"
tokenfolderpath = "."
tokenfile = "resp-init.json"
vaultsecretshares = 5
vaultsecretthreshold = 3
```

bnevis-i:
recheck

tingyuz:
recheck

PR #1890, opened 2019-10-11
Apply errorConcept pattern to core-command REST Handler

brandonforster:
Fixes #1860.

This PR contains a large refactor for core command. This PR's functionality should be verified as well as ensuring there are no regressions from the previous code.

PR #1972, opened 2019-10-21
Apply errorConcept pattern to core-command REST Handler

### LINKED ISSUE #1860: Apply error-handling pattern to core-command and core-data REST Handlers
# 🚀 Feature Request

### Relevant Package

- core-data
- core-command

### Description

In core-metadata, a pattern has been put in place in effort to create common way of handling HTTP errors. This approach uses a set of structs referred to as ErrorConcepts. These structs encapsulate all of the properties needed for type-checking, setting the error on the responseWriter, and logging the error. 

We need to apply this same pattern of error-handling to the core-data and core-command services.

The recommended solution is to go through core-data and core-command updating all error-handling to use the same pattern used in core-metadata REST handlers.

This effort may include abstracting out common errors in both core-command and core-data and reusing ErrorConcepts where applicable.

None

brandonforster:
Fixes #1860, supercedes #1890.

This PR contains a large refactor for core command. This PR's functionality should be verified as well as ensuring there are no regressions from the previous code.

### DIFF
--- a/internal/core/command/device.go
+++ b/internal/core/command/device.go
@@ -15,43 +15,34 @@ package command
 
 import (
 	"context"
+	goErrors "errors"
 	"fmt"
 	"net/http"
 
-	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
-	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
-
 	"github.com/edgexfoundry/edgex-go/internal/core/command/errors"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
+
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 )
 
-func commandByDeviceID(deviceID string, commandID string, body string, queryParams string, isPutCommand bool, ctx context.Context) (string, int) {
+func commandByDeviceID(deviceID string, commandID string, body string, queryParams string, isPutCommand bool,
+	ctx context.Context) (string, error) {
 	d, err := mdc.Device(deviceID, ctx)
 	if err != nil {
 		LoggingClient.Error(err.Error())
-
-		chk, ok := err.(types.ErrServiceClient)
-		if ok {
-			return err.Error(), chk.StatusCode
-		} else {
-			return err.Error(), http.StatusInternalServerError
-		}
+		return "", err
 	}
 
 	if d.AdminState == contract.Locked {
 		LoggingClient.Error(d.Name + " is in admin locked state")
-		return errors.NewErrDeviceLocked(d.Name).Error(), http.StatusLocked
+		return "", errors.NewErrDeviceLocked(d.Name)
 	}
 
-	//once command service have its own persistence layer this call will be changed.
+	// once command service have its own persistence layer this call will be changed.
 	commands, err := dbClient.GetCommandsByDeviceId(d.Id)
 	if err != nil {
 		LoggingClient.Error(err.Error())
-		if err == db.ErrNotFound {
-			return err.Error(), http.StatusNotFound
-		} else {
-			return err.Error(), http.StatusInternalServerError
-		}
+		return "", err
 	}
 
 	var c contract.Command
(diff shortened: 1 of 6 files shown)


brandonforster:
@michaelestrin I'm also seeing those four `device-error` failures in master. It seems like it wants the device to be missing in the database, but it's hanging up on the device being locked.

I believe I have addressed the `device` failures, however.

jpwhitemn:
@brandonforster - still a Fuji release item?  Label Geneva?

jpwhitemn:
@brandonforster - still a Fuji release item?  Label Geneva?

brandonforster:
@jpwhitemn This is ready to go, but it's my understanding that it's low priority and review resources are needed elsewhere to manage high priority Fuji deliverables.

michaelestrin:
@jbonafide623 please do an initial review today and comment your findings.  Do not approve.

michaelestrin:
@brandonforster please rebase.

internal/core/command/device.go
```diff
@@ -15,43 +15,30 @@ package command
...
-		} else {
-			return err.Error(), http.StatusInternalServerError
-		}
+		return "", err
 	}
 
 	if d.AdminState == contract.Locked {
-		LoggingClient.Error(d.Name + " is in admin locked state")
-		return errors.NewErrDeviceLocked(d.Name).Error(), http.StatusLocked
+		return "", errors.NewErrDeviceLocked(d.Name)
```
jbonafide623: Is there a way that we can move instances like this to the general error-checking approach? The goal of applying the general Error Handling is to provide a single place for error-checking. This will help facilitate that common error-handling approach we are aiming for in v2.

My first thought is that these utility methods (i.e. commandByDeviceId) are not too flexible in the sense that they are only re-used for testing sake. With that, more complexity has been introduced making it difficult to apply the common error-handling approach we've applied elsewhere.  Do you see the opportunity of refactoring this in such a way that we can cover scenarios like this which are not handled by the common handler? Possibly gutting this file of utility methods as a whole and find a way to reuse code while also applying common error handling.

internal/core/command/device.go
```diff
@@ -63,45 +50,33 @@ func commandByDeviceID(deviceID string, commandID string, body string, queryPara
 	}
 
 	if c.String() == (contract.Command{}).String() {
-		errMsg := fmt.Sprintf("Command with id '%v' does not belong to device with id '%v'.", commandID, deviceID)
-		LoggingClient.Error(errMsg)
-		return errMsg, http.StatusNotFound
+		return "", errors.NewErrCommandNotAssociatedWithDevice(commandID, deviceID)
```
jbonafide623: See previous comment

internal/core/command/device.go
```diff
@@ -63,45 +50,33 @@ func commandByDeviceID(deviceID string, commandID string, body string, queryPara
...
-		chk, ok := err.(types.ErrServiceClient)
-		if ok {
-			return err.Error(), chk.StatusCode
-		} else {
-			return err.Error(), http.StatusInternalServerError
-		}
+		return "", err
 	}
 
 	if d.AdminState == contract.Locked {
```
jbonafide623: See previous comment on errors which are not handled by common-error-handler

internal/core/command/device.go
```diff
@@ -111,90 +86,61 @@ func commandByDevice(device contract.Device, command contract.Command, body stri
...
+		return "", err
 	}
 
 	responseBody, responseCode, err := ex.Execute()
 	if err != nil {
-		LoggingClient.Error(err.Error())
-		return err.Error(), http.StatusInternalServerError
+		return "", err
+	}
+	if responseCode != 200 {
```
jbonafide623: See previous comment on errors which are not handled by common-error-handler

brandonforster:
@jbonafide623 @michaelestrin Responding to all your comments in one block, I agree there's some more work that can be done here to do this right and not just have it be done, especially now that it's a Geneva item. I'm going to close the issue I was working under and create a new issue to capture this refactor effort, and we can possibly explore breaking that out into subtasks with this branch forming a basis for that work, or if we want to pivot this branch can be informing that work.

PR #1974, opened 2019-10-21
Add .git directory to .dockerignore file

### LINKED ISSUE #1973: .dockerignore file should include .git/ directory
# 🚀 Feature Request

### Description
 A clear and concise description of the problem or missing capability...
Currently, the `.git/` directory is sent in the docker build context for each image built.  This currently represents 230MB of transfer that needs to happen, as well as bloat in potentially cached images on developer systems.  Adding .git to the .dockerignore file will speed up docker build times for developers and in CI/CD

I think we should add .git and log files to the .dockerignore file.

jdharms:
Fixes #1973

There's no reason to have the .git/ directory in our docker build context,
having it present slows down builds and bloats cached images.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,4 @@
 docs
 *.snap
+.git
+**/logs


anonymouse64:
Does go when used with modules care about what git tag version of the code you are building? If so, then we don't want this because go won't realize it's building a git repository. If it doesn't then this looks fine to me

jdharms:
@anonymouse64 the module directive in the go.mod file lets go know what package it is building.  No .git directory necessary.

PR #1977, opened 2019-10-22
Ensure token path exists before creating token file

### LINKED ISSUE #1976: security-secretstore-setup should ensure the path exists before creating token file
# 🐞 Bug Report

### Affected Services

 The issue is located in: security-secretstore-setup

security-secretstore-setup can run into an issue creating the token file if the token path doesn't exist.  We should ensure this path exists before continuing.

jdharms:
Fixes #1976

Creates token path if it doesn't exist on the filesystem.

### DIFF
--- a/cmd/security-secretstore-setup/main.go
+++ b/cmd/security-secretstore-setup/main.go
@@ -91,6 +91,14 @@ func main() {
 	path := secretstore.Configuration.SecretService.TokenFolderPath
 	filename := secretstore.Configuration.SecretService.TokenFile
 	absPath := filepath.Join(path, filename)
+
+	// Ensure token path exists
+	err := os.MkdirAll(path, 0700)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


jdharms:
@bnevis-i I agree, thanks.  Applying a hold label.  Feel free to take it off if you decide later it is necessary.

tingyuz:
After conversation with @jim-wang-intel I don't think we need this PR, but need to update docker-compose file to attach volume of /vault/config into each container. 

As Jim showed me, we did create /vault/config/asset folder in the dockerfile of security-secrets-setup, 
https://github.com/edgexfoundry/edgex-go/blob/master/cmd/security-secrets-setup/Dockerfile

And all we need to do is to attach /vault/config with our customized Vault container like others:

volumes:
      - vault-config:/vault/config

jim-wang-intel:
The missing volume mapping vault-config is addressed in [this PR of developer-scripts](https://github.com/edgexfoundry/developer-scripts/pull/153)

tingyuz:
Close this PR as it is not the root cause of the issue.

PR #1979, opened 2019-10-23
log resp-init.json file creation error

tingyuz:
Add detailed log entry when error of creating resp-init.json file happens.


jdharms:
LGTM

PR #1982, opened 2019-10-23
Fix marshaling of secretstore.UserPasswordPair

### LINKED ISSUE #1981: use "username" as the field name when creating credential pairs in secretstore
Currently when creating credential pairs the security-secretstore-setup is using "user" as the field name as shown below (in password.go):

type UserPasswordPair struct {
    User     string `json:"user,omitempty"`
    Password string `json:"password,omitempty"`
}

However both MongoDB client as well as secretclient of microservice are searching "username" field to retrieve the user name. It is suggested to use "username" as well in the UserPasswordPair struct so that the search will be successful.

jdharms:
Fixes #1981

UserPasswordPair should be marshaled into json as
{username, password} instead of {user, password}.

### DIFF
--- a/internal/security/secretstore/password.go
+++ b/internal/security/secretstore/password.go
@@ -65,7 +65,7 @@ type CredCollect struct {
 }
 
 type UserPasswordPair struct {
-	User     string `json:"user,omitempty"`
+	User     string `json:"username,omitempty"`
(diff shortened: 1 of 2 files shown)


bnevis-i:
#1981

PR #1983, opened 2019-10-23
Fix security based TOML configurations

AnthonyMBonafide:
This addresses part of issue 282

Update TOML configuration files to use the correct path and protocol
when obtaining secrets from Vault.

Update export-distro and sys-mgmt-agent to not use
secret.BootstrapHandler as they do not need to obtain secrets even when
security is enabled.

Update security-secretstore-setup to include core-command as part of the
secret generation process by updating the associated TOML.


michaelestrin:
https://github.com/edgexfoundry/blackbox-testing/issues/282

Issue #1984, opened 2019-10-24
Edgex-proxy failed to retrieve CA certificate when insecureSkipVerify set to false


jim-wang-intel:
# 🐞 Bug Report

### Affected Services

  
 The issue is located in:  
  **edgex-security-proxy-setup** 

One can see the x509 certificate error.  Issue exists when the switch or flag `insecureSkipVefiry` set to **false**.  No issue is observed when that flag is set to true.

### Is this a regression?

 

No, when the flag of  ```insecureSkipVerify``` is false, it appears to have this issue already. 

### Description and Minimal Reproduction
The TLS transport layer requires CA certificate to be present when the communication between proxy and the vault starts in order to retrieve the certificate for Kong.  Because the current mechanism on the requestor is missing the loading of CA certificate into http TLS layer, the `x509: certificate signed by unknown authority` error will surface when the `insecureSkipVerify` flag is off or false. 

To reproduce, 
 1. one can set the insecureSkipVerify=false on the edgex-security-proxy-setup service in the docker-compose file indicating here:
https://github.com/edgexfoundry/developer-scripts/blob/7d7f885b77450c78d4f2c1640c88d406175141f2/releases/nightly-build/compose-files/docker-compose-nexus.yml#L204

Change "true" to "false like
```yaml
command: ["--init=true", "--insecureSkipVerify=false"]
```
2. Save the file
3. Run docker-compose up 
4. In the docker log of the `security-proxy-setup`, one can see the error:

```
level=ERROR ts=2019-10-23T23:33:45.941083418Z app=edgex-security-proxy-setup source=certs.go:96 msg="failed to retrieve certificate on path v1/secret/edgex/pki/tls/edgex-kong with error Get https://edgex-vault:8200/v1/secret/edgex/pki/tls/edgex-kong: x509: certificate signed by unknown authority"
level=ERROR ts=2019-10-23T23:33:45.945461964Z app=edgex-security-proxy-setup source=main.go:86 msg="Get https://edgex-vault:8200/v1/secret/edgex/pki/tls/edgex-kong: x509: certificate signed by unknown authority"
```

## 🔥 Exception or Error
<pre>failed to retrieve certificate on path v1/secret/edgex/pki/tls/edgex-kong with error Get https://edgex-vault:8200/v1/secret/edgex/pki/tls/edgex-kong: x509: certificate signed by unknown authority<code>

</code></pre>

## 🌍  Your Environment
**Deployment Environment:**
[...]

PR #1985, opened 2019-10-24
Fix issue #1984: EdgeX Proxy Setup failed to retrieve CA certificate

jim-wang-intel:
Add the missing code for the requestor to load CA certificate on HTTP TLS transport layer setup

This PR should resolve the issue #1984 

### DIFF
--- a/internal/security/proxy/requestor.go
+++ b/internal/security/proxy/requestor.go
@@ -17,6 +17,8 @@ package proxy
 
 import (
 	"crypto/tls"
+	"crypto/x509"
+	"io/ioutil"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/security/proxy/requestor.go
```diff
@@ -17,15 +17,37 @@ package proxy
...
+			return nil
+		}
+		LoggingClient.Info("successfully loaded rootCA certificate.")
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
+
+		tr = &http.Transport{
+			TLSClientConfig: &tls.Config{
+				RootCAs:            caCertPool,
+				InsecureSkipVerify: skipVerify,
```
anonymouse64: minor nitpick you could just drop this line since in this block we already know that it's false and the default value for a bool in Go is false so we don't gain anything by having this line and IMHO it just opens us up to accidental typos or something modifying this here when we shouldn't be

jim-wang-intel:
rebase

jim-wang-intel:
@anonymouse64 Ian, is this looking good to you now? Thanks!

anonymouse64:
@jim-wang-intel will review after I get back from lunch this afternoon

anonymouse64:
I had to also remove the `--insecureSkipVerify` from the docker-compose command with this patch, but after doing so then I can confirm that now the edgex-proxy container runs fine.

```diff
diff --git a/releases/nightly-build/compose-files/docker-compose-nexus.yml b/releases/nightly-build/compose-files/docker-compose-nexus.yml
index 5bf924f..946e7a2 100644
--- a/releases/nightly-build/compose-files/docker-compose-nexus.yml
+++ b/releases/nightly-build/compose-files/docker-compose-nexus.yml
@@ -226,10 +226,10 @@ services:
         - consul
 
   edgex-proxy:
-    image: nexus3.edgexfoundry.org:10004/docker-edgex-security-proxy-setup-go:1.1.0
+    image: edgexfoundry/docker-edgex-security-proxy-setup-go:1.1.0-dev
     container_name: edgex-proxy
     hostname: edgex-proxy
-    command: ["--init=true", "--insecureSkipVerify=true"]
+    command: ["--init=true"]
     networks:
       edgex-network:
         aliases:
```

Can we please submit a PR against developer-scripts removing this option from the container? Thanks

anonymouse64:
> Can we please submit a PR against developer-scripts removing this option from the container? Thanks

Well I just ended up filing it myself here: https://github.com/edgexfoundry/developer-scripts/pull/159

michaelestrin:
Merged before Fuji branch cut; verified in Fuji branch.

PR #1992, opened 2019-10-24
feat(config-seed): Change Config Seed rules engine properties for App-Service-Configurable

### LINKED ISSUE #1991: Rules Engine properties need to be changed so data is received from App-Service-Configurable
# 🐞 Bug Report

### Affected Services
config-seed

Rules Engine properties need to be change so the default is that it receives messages from new App Service Configurable (rules-engine profile).

Two properties have to change for rules engine to receive the message from App-Service-Configurable.

- export.client=false
- export.zeromq.host=tcp://edgex-app-service-configurable-rules

### Is this a regression?

No. Just a missed updating for Fuji

Once Export Client and Export Distro are removed for default running services, the Rules engine will not received data

## 🔥 Exception or Error

Rules engine will continue to try to register with Export Client

## 🌍  Your Environment
**Deployment Environment:**
Any
**EdgeX Version:**
Fuji

**Anything else relevant?**
No

lenny-intel:
Rules Engine properties changed so the default is that it now receives messages from new App Service Configurable (rules-engine profile).

closes #1991

### DIFF
--- a/cmd/config-seed/res/properties/edgex-support-rulesengine/application.properties
+++ b/cmd/config-seed/res/properties/edgex-support-rulesengine/application.properties
@@ -28,7 +28,7 @@ server.port=48075
 #-----------------Export Service Config----------------------------------------
 #Turn on/off registration of rules engine as export distro client.  
 #Set to false to receive messages directly from core data
-export.client=true
+export.client=false
 expect.serializedjava=false
 export.client.registration.url=http://localhost:48071/api/v1
 #export.client.registration.url=http://edgex-export-client:48071/api/v1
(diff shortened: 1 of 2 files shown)


lenny-intel:
@michaelestrin , the plan is to add app-service-configurable to the compose files once this is merged. We are just changing the default. Export services will still be available if someone wants to switch back to using them. We had this discussion in the applications slack channel and agreed to changing the default. This will also be on the agenda for today's App Services WG meeting.

lenny-intel:
@michaelestrin , I reviewed the BB Tests and is dependent on Rules Engine actually receiving data from Export Distro. I think we are good to merg this one.

PR #2032, opened 2019-10-28
Refactor SecretStoreClient to simplfy request/response handling

### LINKED ISSUE #1923: Refactor SecretStoreClient to combine common~~Request and commonResponse methods
# 🚀 Feature Request

### Relevant Package
Affects secretstoreclient package

### Description
The SecretStoreClient has three common methods for formatting requests and processing responses from the secret store: `commonRequest`, `commonJSONRequest`, and `commonResponse`.

The request methods differ in that `commonJSONrequest` takes a JSON object that is serialized before calling `commonRequest`.  Request that do not require a JSON object call `commonRequest` directly.  The `common~~Request` and `commonResponse` objects are intended to be called together, where the output of the former is passed to the latter.  This creates an awkward situation that was raised in code review that the first lines of `commonResponse` handle the errors from `commonRequest`.

The proposal is to combine the request and response methods into a single function that performs the entire request flow, and define a parameter struct so that the caller can supply named arguments to the function.  Otherwise, the argument list to the combined function would be unwieldy.  The named parameter struct is to take care of the special case of whether the invoked API method requires serialization of a JSON object.

The following behavior is proposed:
- If `params.JSONObject (interface{})` is non-nil, the object is sent through the JSON marshaller and passed  to the `http.NewRequest` method.
- Otherwise, `params.BodyReader (io.Reader)` is passed to the `http.NewRequest` method unchanged.

Other named parameters will be defined for the existing positional arguments.

In code review, it was determined that providing two named arguments, JSONObject and BodyReader with unique types would be cleaner than passing interface{} and using type assertions to dynamically decide what to do with the argument based on its type.

bnevis-i:
Fixes #1923

SecretStoreClient has three common methods for formatting requests
and processing responses from the secret store:
commonRequest, commonJSONRequest, and commonResponse.

The improved implementation has a common doRequest() method
and takes a parameter struct that enables named parameters to
the function.  This resolves an awkward situation that was raised
in code review that the first lines of commonResponse handle
the errors from commonRequest.


internal/security/secretstoreclient/common.go
```diff
@@ -0,0 +1,99 @@
+/*******************************************************************************
```
bnevis-i: The common doRequest() and tests have been refactored into common.go and common_test.go.

internal/security/secretstoreclient/vault.go
```diff
@@ -46,15 +45,25 @@ func NewSecretStoreClient(logger logger.LoggingClient, r internal.HttpCaller, s
 	}
 }
 
-func (vc *vaultClient) HealthCheck() (statusCode int, err error) {
-	resp, err := vc.commonJSONRequest(nil, http.MethodGet, VaultHealthAPI, struct{}{})
-	if err != nil {
-		vc.logger.Error(fmt.Sprintf("failed on checking status of secret store: %s", err.Error()))
+func (vc *vaultClient) HealthCheck() (int, error) {
```
bnevis-i: All of the methods changed to use the common doRequest() method;
return arguments have been de-identifed to reduce risk of introducing unintended bugs.

internal/security/secretstoreclient/vault_test.go
```diff
@@ -49,13 +47,54 @@ func TestHealthCheck(t *testing.T) {
...
 
+	assert.NoError(t, err)
 	if code != http.StatusOK {
 		t.Errorf("incorrect vault health check status.")
 	}
 }
 
+// TestHealthCheckUninit tests that a no error object is returned
+// for uninitialized vault (with 5xx error)
+func TestHealthCheckUninit(t *testing.T) {
```
bnevis-i: Add additional test case to ensure that HealthCheck doesn't return Error object when the status code is in the 5xx range.

internal/security/secretstoreclient/vault_test.go
```diff
@@ -49,13 +47,54 @@ func TestHealthCheck(t *testing.T) {
...
+
+	assert.NoError(t, err)
+	if code != http.StatusNotImplemented {
+		t.Errorf("incorrect vault health check status.")
+	}
+}
+
+// TestHealthCheckSealed tests that a no error object is returned
+// for sealed vault (with 5xx error)
+func TestHealthCheckSealed(t *testing.T) {
```
bnevis-i: Ditto.

internal/security/secretstoreclient/vault_test.go
```diff
@@ -208,42 +247,3 @@ func TestCreateToken(t *testing.T) {
 	assert.Equal(http.StatusOK, code)
 	assert.Equal("f00341c1-fad5-f6e6-13fd-235617f858a1", response["request_id"].(string))
 }
-
```
bnevis-i: Removed tests replaced with equivalent tests in common_test.go

anonymouse64:
Don't have time to review this unfortunately

internal/security/secretstoreclient/common.go
```diff
@@ -0,0 +1,99 @@
...
+	if resp.StatusCode != params.ExpectedStatusCode {
+		err := fmt.Errorf("request to %s failed with status: %s", params.OperationDescription, resp.Status)
+		vc.logger.Error(err.Error())
+		return resp.StatusCode, err
+	}
+
+	if params.ResponseObject != nil {
+		err := json.NewDecoder(resp.Body).Decode(params.ResponseObject)
+		if err != nil {
+			vc.logger.Error(fmt.Sprintf("failed to parse response body: %s", err.Error()))
```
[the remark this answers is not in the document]
bnevis-i: No. resp.Body is a stream.  It is consumed by the decoding. There would be nothing left to print.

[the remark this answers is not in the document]
jim-wang-intel: You can still read it out to the memory as string and include it as logging message if want:

For example: 
```go
bodyBytes, err := ioutil.ReadAll(resp.Body)
    if err != nil {
        log.Fatal(err)
    }
    bodyString := string(bodyBytes)
```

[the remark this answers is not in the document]
bnevis-i: I still don't think it's worth the added complexity and extra buffering, given that we have a guard `resp.StatusCode != params.ExpectedStatusCode` in the lines immediately above (i.e. we should only be attempting to parse JSON for known response codes).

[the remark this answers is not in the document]
jim-wang-intel: ok

bnevis-i:
Rebased for geneva; re-review requested.

PR #2033, opened 2019-10-28
Update VERSION on master to avoid conflicts with fuji

lranjbar:
Signed-off-by: Lisa Rashidi-Ranjbar <lisa.a.rashidi-ranjbar@intel.com>


lranjbar:
recheck

PR #2042, opened 2019-10-30
Remove 'LoggingClient' global usage in core-data

### LINKED ISSUE #2024: Data: refactor to remove LoggingClient global variable

AnthonyMBonafide:
Fix #2024

Update core-data to leverage the dependency injection container(DIC)
during bootstrap and pass a 'LoggingClient' instance down the call stack
to eliminate the need for the 'LoggingClient' global variable.

Update tests to use a DIC when loading REST routes as they will need the
DIC to obtain a reference of the 'LoggingClient'. Also, update any
references to methods/functions which have updated signatures to accept
a 'LoggingClient' as part of changes to the production code.

### DIFF
--- a/cmd/core-data/main.go
+++ b/cmd/core-data/main.go
@@ -17,6 +17,8 @@ package main
 import (
 	"flag"
 
+	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+
 	"github.com/edgexfoundry/edgex-go"
 	"github.com/edgexfoundry/edgex-go/internal"
 	"github.com/edgexfoundry/edgex-go/internal/core/data"
@@ -30,8 +32,6 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/telemetry"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/usage"
-
-	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 )
 
 func main() {
@@ -49,7 +49,8 @@ func main() {
 	flag.Usage = usage.HelpCallback
 	flag.Parse()
 
-	httpServer := httpserver.NewBootstrap(data.LoadRestRoutes())
+	dic := di.NewContainer(di.ServiceConstructorMap{})
+	httpServer := httpserver.NewBootstrap(data.LoadRestRoutes(dic))
 	bootstrap.Run(
 		configDir,
 		profileDir,
@@ -58,7 +59,7 @@ func main() {
 		clients.CoreDataServiceKey,
 		data.Configuration,
 		startupTimer,
-		di.NewContainer(di.ServiceConstructorMap{}),
+		dic,
 		[]interfaces.BootstrapHandler{
 			secret.NewSecret().BootstrapHandler,
 			database.NewDatabaseForCoreData(&httpServer, data.Configuration).BootstrapHandler,
--- a/internal/core/data/device_test.go
+++ b/internal/core/data/device_test.go
@@ -5,28 +5,23 @@ import (
 	"fmt"
 	"math"
 	"net/http"
-	"os"
 	"sync"
 	"testing"
 	"time"
 
 	"github.com/globalsign/mgo/bson"
 	"github.com/google/uuid"
-	"github.com/gorilla/mux"
 	"github.com/stretchr/testify/mock"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata/mocks"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
-	"github.com/edgexfoundry/go-mod-messaging/messaging"
-	msgTypes "github.com/edgexfoundry/go-mod-messaging/pkg/types"
 
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 )
 
 var testEvent contract.Event
-var testRoutes *mux.Router
 
 const (
 	testDeviceName string = "Test Device"
@@ -40,7 +35,7 @@ func TestCheckMaxLimit(t *testing.T) {
 
 	testedLimit := math.MinInt32
 
-	expectedNil := checkMaxLimit(testedLimit)
+	expectedNil := checkMaxLimit(testedLimit, logger.NewMockClient())
 
 	if expectedNil != nil {
 		t.Errorf("Should not exceed limit")
@@ -52,28 +47,11 @@ func TestCheckMaxLimitOverLimit(t *testing.T) {
 
 	testedLimit := math.MaxInt32
 
-	expectedErr := checkMaxLimit(testedLimit)
+	expectedErr := checkMaxLimit(testedLimit, logger.NewMockClient())
 
 	if expectedErr == nil {
 		t.Errorf("Exceeded limit and should throw error")
 	}
-}
-
-func TestMain(m *testing.M) {
-	testRoutes = LoadRestRoutes()
-	LoggingClient = logger.NewMockClient()
-	mdc = newMockDeviceClient()
-	// no need to mock this since it's all in process
-	msgClient, _ = messaging.NewMessageClient(msgTypes.MessageBusConfig{
-		PublishHost: msgTypes.HostInfo{
-			Host:     "*",
-			Protocol: "tcp",
-			Port:     5563,
-		},
-		Type: "zero",
-	})
-	chEvents = make(chan interface{}, 10)
-	os.Exit(m.Run())
 }
 
 // Supporting methods
@@ -162,7 +140,7 @@ func buildReadings() []contract.Reading {
 }
 
 func handleDomainEvents(bitEvents []bool, wait *sync.WaitGroup, t *testing.T) {
-	until := time.Now().Add(500 * time.Millisecond) //Kill this loop after half second.
+	until := time.Now().Add(500 * time.Millisecond) // Kill this loop after half second.
 	for time.Now().Before(until) {
 		select {
 		case evt := <-chEvents:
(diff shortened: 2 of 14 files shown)


AnthonyMBonafide:
I verified the changes with the blackbox tests for `core-data.` The tests passed and the `edgex-core-data` Docker container logs showed the logs are being populated.

cmd/core-data/main.go
```diff
@@ -49,7 +52,17 @@ func main() {
...
-	httpServer := httpserver.NewBootstrap(data.LoadRestRoutes())
+	configuration := &data.ConfigurationStruct{}
+	dic := di.NewContainer(di.ServiceConstructorMap{
+		container.ConfigurationName: func(get di.Get) interface{} {
+			return configuration
+		},
+		bootstrapContainer.ConfigurationInterfaceName: func(get di.Get) interface{} {
+			return get(container.ConfigurationName)
+		},
+	})
```
michaelestrin: You need a new empty container here.  

bootstrap.Run() will populate ConfigurationInterfaceName.  

ConfigurationName isn't used elsewhere within your code and should be dropped.  You'll likely add these as part of #2031.

internal/core/data/device_test.go
```diff
@@ -52,16 +55,20 @@ func TestCheckMaxLimitOverLimit(t *testing.T) {
...
 
 func TestMain(m *testing.M) {
-	testRoutes = LoadRestRoutes()
-	LoggingClient = logger.NewMockClient()
+	dic := di.NewContainer(di.ServiceConstructorMap{
+		container.LoggingClientInterfaceName: func(get di.Get) interface{} {
+			return mockLoggingClient
+		}})
+
+	testRoutes = LoadRestRoutes(dic)
```
michaelestrin: Per our side conversation, I question what purpose this test serves beyond code coverage.  It doesn't appear to be actually testing anything.  If that's the case, it should be removed.

As a more general statement, there's something amiss if the dependency injection container is required for a unit test.

internal/core/data/reading_test.go
```diff
@@ -43,7 +43,7 @@ func TestGetAllReadingsOverLimit(t *testing.T) {
 
 	dbClient = newReadingsMockDB()
 
-	_, err := getAllReadings()
+	_, err := getAllReadings(mockLoggingClient)
```
michaelestrin: Commented here; applies elsewhere.

mockLoggingClient is defined in a different file.  It's usage complicates potential future refactoring of the current monolithic data package into a series of more cohesive subpackages.

In my opinion, all references to mockLoggingClient (in this file and elsewhere) should simply be replaced with its assigned value; i.e. logger.NewMockClient().  This would keep the tests "clean" by instantiating required dependencies within the scope of each test (instead of relying on an existing variable that may be tainted by its use in a test that ran prior).

internal/core/data/router_test.go
```diff
@@ -184,7 +182,10 @@ func TestRestValueDescriptorsUsageHandler(t *testing.T) {
 			dbClient = tt.dbMock
 			Configuration.Service = tt.config
 			rr := httptest.NewRecorder()
-			handler := http.HandlerFunc(restValueDescriptorsUsageHandler)
+			handler := http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
```
michaelestrin: I wonder if this series of tests shouldn't be modified to call restValueDescriptorsUsageHandler() directly.  Beyond goosing code coverage, I don't see any value going through the full HTTP request/response cycle.

internal/core/data/router_test.go
```diff
@@ -69,8 +70,7 @@ var TestValueDescriptor3 = contract.ValueDescriptor{
 func TestRestValueDescriptorsUsageHandler(t *testing.T) {
 	Configuration = &ConfigurationStruct{Service: config.ServiceInfo{MaxResultCount: 10}}
 	defer func() { Configuration = &ConfigurationStruct{} }()
-	LoggingClient = logger.MockLogger{}
-	httpErrorHandler = errorconcept.NewErrorHandler(LoggingClient)
+	httpErrorHandler = errorconcept.NewErrorHandler(logger.NewMockClient())
```
michaelestrin: I think these should be moved inside the test loop.  

This ensures fresh instances for each test (and avoids the potential for cross-test contamination).

I would also ensure that you're using a single logging instance for the test; you're most recent set of changes injects a different logger mock into httpErrorHandler than the logger mock passed to restValueDescriptorsUsageHandler() (and I total understand why you did that given the current structure of the code).

brandonforster:
@michaelestrin I'm not seeing any regressions from master on the blackbox tests, but I am getting failures in both codebases.

internal/core/data/device_test.go
AnthonyMBonafide: These tests are a bit messed up. I removed this test, which doesn't look like it would affect anything, and other tests started to fail due to panics. I will need to look deeper into this and restructure these tests.

AnthonyMBonafide: It turns out that this test was instanciating the global variables:

- mdc (meta-data client)
- msgClient
- chEvents 

These two tests were expecting those variables to be populated, but were not doing the population themselves:

https://github.com/edgexfoundry/edgex-go/blob/dc82d8d40a8e1e3d622a5a3b2b216fa3e9cf0243/internal/core/data/valuedescriptor_test.go#L314

https://github.com/edgexfoundry/edgex-go/blob/dc82d8d40a8e1e3d622a5a3b2b216fa3e9cf0243/internal/core/data/event_test.go#L210

So I updated them to populate the global variables they need to execute the test. These will go away with my future work to eliminate the global variables.

AnthonyMBonafide:
I removed some unnecessary tests that were not testing any functionality, but rather misleading the test code coverage. This will decrease code coverage and is expected.

internal/core/data/event_test.go
```diff
@@ -207,8 +211,20 @@ func newAddEventMockDB(persist bool) *dbMock.DBClient {
...
+	// no need to mock this since it's all in process
+	msgClient, _ = messaging.NewMessageClient(msgTypes.MessageBusConfig{
+		PublishHost: msgTypes.HostInfo{
+			Host:     "*",
+			Protocol: "tcp",
+			Port:     5563,
+		},
+		Type: "zero",
+	})
+
```
AnthonyMBonafide: I needed to populate this global variable as the tests would panic without it. Sadly, the test `TestMain` I removed in `device_test.go` was populating this and causing this test to be dependent on the execution of that test.  This also goes for the `chEvents` global variable being populated for this test.

internal/core/data/valuedescriptor_test.go
```diff
@@ -311,13 +313,15 @@ func TestGetValueDescriptorsByTypeNotFound(t *testing.T) {
 
 func TestGetValueDescriptorsByTypeError(t *testing.T) {
 	reset()
+
 	myMock := &mocks.DBClient{}
 
 	myMock.On("ValueDescriptorsByType", mock.Anything).Return([]models.ValueDescriptor{}, fmt.Errorf("some error"))
 
 	dbClient = myMock
+	mdc = newMockDeviceClient()
```
AnthonyMBonafide: See my other comment regarding the global variable changes needed due to the removal of the `TestMain` test in `device_test.go`

internal/core/data/event_test.go
```diff
@@ -207,8 +211,20 @@ func newAddEventMockDB(persist bool) *dbMock.DBClient {
...
+	// no need to mock this since it's all in process
+	msgClient, _ = messaging.NewMessageClient(msgTypes.MessageBusConfig{
+		PublishHost: msgTypes.HostInfo{
+			Host:     "*",
+			Protocol: "tcp",
+			Port:     5563,
+		},
+		Type: "zero",
+	})
```
AnthonyMBonafide: I needed to populate this global variable as the tests would panic without it. Sadly, the test TestMain I removed in device_test.go was populating this and causing this test to be dependent on the execution of that test. This also goes for the chEvents global variable being populated for this test.

cmd/core-data/main.go
```diff
@@ -49,7 +49,8 @@ func main() {
 	flag.Usage = usage.HelpCallback
 	flag.Parse()
 
-	httpServer := httpserver.NewBootstrap(data.LoadRestRoutes())
+	dic := di.NewContainer(di.ServiceConstructorMap{})
```
michaelestrin: This container should be passed into `bootstrap.Run()` below.

internal/core/data/device_test.go
michaelestrin: @AnthonyMBonafide great illustration of why tests should be self-contained.  The hidden dependency on these global variables being set in different tests would be hard to troubleshoot/find if the tests that depended on them failed because of related changes.

internal/core/data/router_test.go
```diff
@@ -69,8 +70,6 @@ var TestValueDescriptor3 = contract.ValueDescriptor{
 func TestRestValueDescriptorsUsageHandler(t *testing.T) {
 	Configuration = &ConfigurationStruct{Service: config.ServiceInfo{MaxResultCount: 10}}
```
michaelestrin: This should also be moved inside the test loop.  See prior PR comments.

AnthonyMBonafide:
Rebased

internal/core/data/router_test.go
```diff
@@ -67,10 +68,7 @@ var TestValueDescriptor3 = contract.ValueDescriptor{
 }
 
 func TestRestValueDescriptorsUsageHandler(t *testing.T) {
-	Configuration = &ConfigurationStruct{Service: config.ServiceInfo{MaxResultCount: 10}}
 	defer func() { Configuration = &ConfigurationStruct{} }()
```
michaelestrin: (You're going to hate me.) 

Clean up should occur at the same level of allocation/change.  

Move this down below the `Configuration` assignment in the `t.Run()` method.

AnthonyMBonafide: Hahaha, I never can hate someone taking the time to make MY code better. I appreciate it. Plus you are right :)

AnthonyMBonafide: Actually this is my fault. I should have looked closer. My apologises.

PR #2045, opened 2019-10-30
Remove 'LoggingClient' global usage in core-command

### LINKED ISSUE #2013: Command: refactor to remove LoggingClient global variable

brandonforster:
Fix #2013 

Update core-command to leverage the dependency injection container(DIC)
during bootstrap and pass a 'LoggingClient' instance down the call stack
to eliminate the need for the 'LoggingClient' global variable.

### DIFF
--- a/internal/core/command/device.go
+++ b/internal/core/command/device.go
@@ -27,8 +27,14 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 )
 
-func commandByDeviceID(deviceID string, commandID string, body string, queryParams string, isPutCommand bool,
-	ctx context.Context, loggingClient logger.LoggingClient) (string, int) {
+func commandByDeviceID(
+	deviceID string,
+	commandID string,
+	body string,
+	queryParams string,
+	isPutCommand bool,
+	ctx context.Context,
+	loggingClient logger.LoggingClient) (string, int) {
 	d, err := mdc.Device(deviceID, ctx)
 	if err != nil {
 		loggingClient.Error(err.Error())
@@ -74,7 +80,13 @@ func commandByDeviceID(deviceID string, commandID string, body string, queryPara
 	return commandByDevice(d, c, body, queryParams, isPutCommand, ctx, loggingClient)
 }
 
-func commandByNames(dn string, cn string, body string, queryParams string, isPutCommand bool, ctx context.Context,
+func commandByNames(
+	dn string,
+	cn string,
+	body string,
+	queryParams string,
+	isPutCommand bool,
+	ctx context.Context,
 	loggingClient logger.LoggingClient) (string, int) {
 	d, err := mdc.DeviceForName(dn, ctx)
 	if err != nil {
@@ -105,8 +117,14 @@ func commandByNames(dn string, cn string, body string, queryParams string, isPut
 	return commandByDevice(d, command, body, queryParams, isPutCommand, ctx, loggingClient)
 }
 
-func commandByDevice(device contract.Device, command contract.Command, body string, queryParams string,
-	isPutCommand bool, ctx context.Context, loggingClient logger.LoggingClient) (string, int) {
+func commandByDevice(
+	device contract.Device,
+	command contract.Command,
+	body string,
+	queryParams string,
+	isPutCommand bool,
+	ctx context.Context,
+	loggingClient logger.LoggingClient) (string, int) {
 	var ex Executor
 	var err error
 	if isPutCommand {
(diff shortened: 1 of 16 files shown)


internal/core/command/types.go
```diff
@@ -29,14 +30,19 @@ type serviceCommand struct {
 	contract.Device
 	internal.HttpCaller
 	*http.Request
+	loggingClient logger.LoggingClient
```
michaelestrin: Alternate approach to simply adding field to receiver is to create a factory function and replace all existing non-error struct literal definitions with a call to the factory function.  [Boy Scout Rule](http://pragmaticcraftsman.com/2011/03/the-boy-scout-rule/).  Ensures single consistent mechanism for instantiating receiver (which ensures all fields in receiver are always properly initialized).

internal/core/command/types.go
```diff
@@ -29,14 +30,19 @@ type serviceCommand struct {
...
 }
 
 // Execute sends the command to the command service
 func (sc serviceCommand) Execute() (string, int, error) {
-	LoggingClient.Debug("Issuing" + sc.Request.Method + " command to: " + sc.Request.URL.String())
+	if sc.loggingClient != nil {
```
michaelestrin: Commented here; applies elsewhere.

You shouldn't ever need to check for `!= nil`; serviceCommand should always have a properly initialized loggingClient field when Execute() is called.

internal/core/command/types_test.go
```diff
@@ -58,7 +57,6 @@ func (MockBody) Close() error {
 func TestExecute(t *testing.T) {
 	expectedResponseBody := "Sample Response Body"
 	expectedResponseCode := http.StatusOK
-	LoggingClient = logger.MockLogger{}
```
michaelestrin: Commented here; applies elsewhere.

You're removing this.  Isn't a logger.LoggingClient needed for this test?

brandonforster:
@michaelestrin I'm seeing the same blackbox test failures on master, can you confirm? If so, I'll open a separate issue to address those.

internal/core/command/device.go
```diff
@@ -64,17 +67,18 @@ func commandByDeviceID(deviceID string, commandID string, body string, queryPara
...
 		return errMsg, http.StatusNotFound
 	}
 
-	return commandByDevice(d, c, body, queryParams, isPutCommand, ctx)
+	return commandByDevice(d, c, body, queryParams, isPutCommand, ctx, loggingClient)
 }
 
-func commandByNames(dn string, cn string, body string, queryParams string, isPutCommand bool, ctx context.Context) (string, int) {
+func commandByNames(dn string, cn string, body string, queryParams string, isPutCommand bool, ctx context.Context,
+	loggingClient logger.LoggingClient) (string, int) {
```
michaelestrin: Commented here; applies elsewhere.

For readability and consistency (not yet, but I'm pushing towards a readable standard), when folding function/method declarations, place one parameter on each line.  For example:

```
func commandByNames(
    dn string, 
    cn string, 
    body string, 
    queryParams string, 
    isPutCommand bool, 
    ctx context.Context,
    loggingClient logger.LoggingClient) (string, int) {
```

internal/core/command/device_test.go
```diff
@@ -17,10 +31,21 @@ import (
...
+	d200c500 = "d200-c500"
+
+	TestCommandId = "TestCommandID"
+)
 
-	LoggingClient = logger.NewMockClient()
+// commandByDeviceID
+func TestExecuteGETCommandByDeviceIdAndCommandId(t *testing.T) {
 	mdc = newMockDeviceClient()
 	dbClient = newCommandMock()
```
michaelestrin: Commented here; potentially applies elsewhere.

I think these should be moved inside the test loop.

This ensures fresh instances for each test (and avoids the potential for cross-test contamination).

internal/core/command/types.go
```diff
@@ -49,3 +52,13 @@ func (sc serviceCommand) Execute() (string, int, error) {
 	}
 	return buf.String(), resp.StatusCode, nil
 }
+
+func NewServiceCommand(device contract.Device, caller internal.HttpCaller, req *http.Request,
```
michaelestrin: Package public function returning a package private struct.  I think this should be `newServiceCommand`.

internal/core/command/device_test.go
brandonforster: These are setting the global variables that are getting removed anyway in my upcoming work. Refactoring it without removing them seems unnecessary and an entry point for bugs, but if you insist I can do it.

michaelestrin: Makes sense.  I was reviewing within the context of the issue and didn't take into account the follow-on issues related to this.

Issue #1472, opened 2019-06-20
Expose edgexproxy command in the snap


anonymouse64:
We currently run edgexproxy as a one-shot service in the snap, however it can also be used as a user-driven app to configure OAuth tokens and such when provided with specific arguments and a different configuration.toml file. There is currently no way to override the arguments passed to edgexproxy in order to use it like this, so we should add a specific new app (not a daemon) to the snap to allow using edgexproxy in this way.

anonymouse64:
Assigning this to @knitzsche

tonyespy:
I removed the fuji label as PR #2102 landed for Fuji just prior to the release however it didn't include a "Fixes" line.

anonymouse64:
Fixed for geneva/master with https://github.com/edgexfoundry/edgex-go/pull/2136

PR #2050, opened 2019-10-31
Add edgexproxy-cmd as app so it can be run by user

knitzsche:
Untested, as mentioned previously. Might want to change edgexproxy (snapcraft daemon) to  edgexproxy-daemon so that the command introduced here can be named simply edgexproxy: let me know.

Fixes https://github.com/edgexfoundry/edgex-go/issues/1472

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -175,6 +175,12 @@ apps:
     plugs:
       - network
       - network-bind
+  edgexproxy-cmd:
+    adapter: none
+    command: bin/edgexproxy
+    plugs:
+      - network
+      - network-bind
   edgex-mongo:
     adapter: full
     after: [mongod]


snap/snapcraft.yaml
```diff
@@ -175,6 +175,12 @@ apps:
     plugs:
       - network
       - network-bind
+  edgexproxy-cmd:
+    adapter: none
+    command: bin/edgexproxy-wrapper.sh
```
anonymouse64: could you change this to run the edgexproxy command and not the wrapper? The wrapper is currently setup to only run as a daemon so the options will be lost when using the wrapper.

knitzsche: will do

anonymouse64:
This needs to be reworked to use security-proxy-setup instead of edgexproxy. I will re-open, and if I can get a review from Tony we can pull this in for Fuji, if not it can slip to Geneva.

Issue #1840, opened 2019-09-27
Update snap to use new security services from edgex-go


anonymouse64:
This is a holding issue to update the snap for various things that have changed surrounding the security-services. Namely:

- [ ] Change `edgexproxy` from security-api-gateway to use `security-proxy-setup`
- [ ] Change `pkisetup` from `security-secret-store` to use `security-secrets-setup`
- [ ] Change `vault-worker` from `security-secret-store` to use `security-secretstore-setup`
- [ ] Setup edgex-mongo from docker-edgex-mongo to be configurable to use vault for tokens/etc.
- [ ] Setup the rest of the core services to be configurable to use vault for mongodb access, etc.

anonymouse64:
Note that when docker-edgex-mongo is updated, we will need to make the env var `EDGEX_SECURITY_SECRET_STORE` configurable through a snap set call, and we will need to configure (using sed probably) the locations of the vault files in the configuration.toml for docker-edgex-mongo when building and installing the snap, as currently the paths are hard-coded to the vault locations.

anonymouse64:
Note that this may slip to after the freeze date, but definitely will be done before the final release date. I think this is okay since theoretically this is just build/packaging changes (absent any unforeseen bugs in the security services...)

anonymouse64:
PR #2053 fixes this for Fuji, but we need to forward-port that PR to master/geneva now. This can wait til next week or the week after though

anonymouse64:
This is fixed for Geneva with https://github.com/edgexfoundry/edgex-go/pull/2135

PR #2052, opened 2019-10-31
many/configuration.toml: set default security key values, fix proxy SNIs

anonymouse64:
See the commit messages for a more thorough explanation, but basically we should be consistent in setting these values for a few reasons, one of which is that it greatly simplifies setting up the security workflow for the snap as a part of my work on #1840

If we don't do this, then I have to either:
- convert the toml file to json to use jq to set these values, then convert back to toml
- use a really nasty sed expression to replace lines with stuff that is really going above the matching lines so that we can insert these keys into the middle of the file

Neither of which I really want to do.

The regression potential here is pretty low since we're simply adding default values (same as for the docker configuration.toml) to config files.

The other change here is changing the default `SNIS` setting for security-proxy-setup to localhost instead of edgex-kong for the non-docker configuration.toml file.

### DIFF
--- a/bin/edgex-launch.sh
+++ b/bin/edgex-launch.sh
@@ -21,6 +21,9 @@ function cleanup {
 	pkill edgex
 }
 
+# disable secret-store integration
+export EDGEX_SECURITY_SECRET_STORE=false
+
 ###
 # Support logging
 ###
--- a/cmd/core-command/res/configuration.toml
+++ b/cmd/core-command/res/configuration.toml
@@ -48,6 +48,9 @@ Port = 8200
 # Use the core-meta data secrets due to core-command using core-meta-data's database for persistance.
 Path = '/v1/secret/edgex/metadata/'
 Protocol = 'https'
+RootCaCertPath = '/vault/config/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem'
+ServerName = 'localhost'
+TokenFile = '/vault/config/assets/resp-init.json'
   [SecretStore.Authentication]
   AuthType = 'X-Vault-Token'
   AuthToken = 'edgex'
--- a/cmd/core-data/res/configuration.toml
+++ b/cmd/core-data/res/configuration.toml
@@ -60,6 +60,9 @@ Host = 'localhost'
 Port = 8200
 Path = '/v1/secret/edgex/coredata/'
 Protocol = 'https'
+RootCaCertPath = '/vault/config/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem'
+ServerName = 'localhost'
+TokenFile = '/vault/config/assets/resp-init.json'
   [SecretStore.Authentication]
   AuthType = 'X-Vault-Token'
   AuthToken = 'edgex'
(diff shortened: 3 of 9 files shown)


anonymouse64:
If it is possible that one day the other services (i.e. export-distro and sys-mgmt-agent) will access secrets from vault then I can add this section to those config files for consistency.

michaelestrin:
> If it is possible that one day the other services (i.e. export-distro and sys-mgmt-agent) will access secrets from vault then I can add this section to those config files for consistency.

Lets not add configuration that isn't being used for consistency's sake.  It will only add confusion.

anonymouse64:
https://jenkins.edgexfoundry.org/job/edgex-go-fuji-verify-go/3/ seems to be stuck, so I'm going to close and reopen

anonymouse64:
recheck

AnthonyMBonafide:
What is the intent of the non-docker configuration TOML files? Are they meant to used when running EdgeX's core services locally without security? I ask because originally those configuration properties where omitted because they require some additional setup. For example, we would need the security setup process to create self-signed certs, so executing `make run` locally would not work.

anonymouse64:
@AnthonyMBonafide The non-docker files are meant to be used both locally when running with `make run` as you described as well as with the snap. The snap is (going to be) setup to run with security on by default. Note that the specific values of these properties in the files don't much matter to me in the snap, since I'm just replacing them with `sed`, but I need them to at least be present in the file so that they are easily replaced. 

If you're concerned about `make run` not working because now the security services will also need to be running I can add the env var to `edgex-launch.sh` to turn off security services so that nothing tries to use these properties.

AnthonyMBonafide:
@anonymouse64 , good point! that would work. However, I am concerned that the default configurations are getting complicated. The only thing I can think of is to create separate configuration files for SNAP, but that does not really solve the problem as now it would be more complex due to the number of duplicate files. Any thoughts?

anonymouse64:
@AnthonyMBonafide 

> However, I am concerned that the default configurations are getting complicated

Well yes, but I don't think this can be avoided. I don't think adding these 3 properties are significantly adding to the complexity much. If you wanted these to not be complex, then probably should just delete the whole SecretStore section entirely. 

> The only thing I can think of is to create separate configuration files for SNAP

I **_really_** **_really_** **_really_** don't want to have the snap have a completely separate set of config files from the non-docker ones. It gets to be a big maintenance burden and TBH we already suffer a lot from the situation where devs only update the docker config files because that's all they test and we only find out about changes and other issues to the non-docker ones after the issue has already been pushed to users and such.

Let me know if you want me to add the security env var to edgex-launch.sh

AnthonyMBonafide:
@anonymouse64, I agree. Could you please add the environment variable to edgex-launch.sh? Thank you for the your thoughts and explanation :+1:

anonymouse64:
> Could you please add the environment variable to edgex-launch.sh

Done, please review if you have time :smile: thanks!

anonymouse64:
recheck

AnthonyMBonafide:
Recheck

Issue #1963, opened 2019-10-18
Turning off security-secret-store in the snap doesn't also turn off security-api-gateway


anonymouse64:
# 🐞 Bug Report

### Affected Services

 The issue is located in: the snap packaging, specifically the configure hook

### Is this a regression?
Unclear if it ever worked properly, the logic in the configure hook seems to imply no?

### Description and Minimal Reproduction

```
$ snap get edgexfoundry security-api-gateway
on
$ snap get edgexfoundry security-secret-store
on
$ snap set edgexfoundry security-secret-store=off
$ snap get edgexfoundry security-api-gateway
on
$ snap services edgexfoundry.kong-daemon
Service                   Startup  Current  Notes
edgexfoundry.kong-daemon  enabled  active   -
```

When setting security-secret-store to off, we need to set security-api-gateway off as well, as technically there is a linkage between the two, with the gateway accessing it's tokens from the secret-store. 

There is also the inverse problem where if you try to turn on security-api-gateway, this should turn on security-secret-store if that is off, but does not.

## 🌍  Your Environment
**Deployment Environment:** ubuntu 19.04 desktop

**EdgeX Version:** edge channel of the snap, i.e. master

bnevis-i:
@tonyespy  is this being replanned for Hanoi?

tonyespy:
This should've been closed awhile ago as it was fixed in fuji. The steps to repeat actually only work with the edinburgh snap, as the security-api-gateway config option was removed in fuji.

I've verified that disabling security-secret-store disables Kong, and the associated services with the fuji and geneva (built from lastest master) snaps.

PR #2053, opened 2019-10-31
Update the snap to use edgex-go security services

### LINKED ISSUE #1840: Update snap to use new security services from edgex-go

anonymouse64:
This set of changes updates the snap to work with the new security services from edgex-go as well as work with the integrations from edgex-mongo that setup database credentials inside secret store (i.e. Vault)  and finally updates all of the main core services to get their mongo credentials from the secret store.

# Testing

- [x] Test that the snap builds and installs successfully
- [x] Test that turning off security-proxy will disable `kong-daemon` and `postgres` services
- [x] Test that setting `security-secret-store` to off _will_ set the expected env var for the various core services
```
$ snap install --devmode edgexfoundry_1.1.0-20191031+43344467_amd64.snap
edgexfoundry 1.1.0-20191031+43344467 installed
$ snap set edgexfoundry security-secret-store=off
$ snap run --shell edgexfoundry.core-data
To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.

$ env | grep EDGEX
EDGEX_SECURITY_SECRET_STORE=false
$
```
- [x] Test that setting `security-secret-store` to off will disable vault, security-proxy-setup, security-secrets-setup, security-secretstore-setup, kong-daemon, and postgres (subject to issue https://github.com/edgexfoundry/edgex-go/issues/1963)
- [x] Test that the expected certificates are generated inside $SNAP_DATA/pki

```
$ sudo tree /var/snap/edgexfoundry/current/pki
/var/snap/edgexfoundry/current/pki
├── ca
│   └── ca.pem
├── EdgeXFoundryCA
│   ├── EdgeXFoundryCA.pem
│   ├── EdgeXFoundryCA.priv.key
│   ├── localhost.pem
│   └── localhost.priv.key
├── edgex-kong
│   ├── ca.pem
│   ├── server.crt
│   └── server.key
└── edgex-vault
    ├── ca.pem
    ├── server.crt
    └── server.key
```

Specifically, the EdgeXFoundryCA.pem and EdgeXFoundryCA.priv.key are generated which are the certificate authority files, from which we derive the single localhost.pem and localhost.priv.key files which is used by Kong and Vault when serving up HTTPS. The `edgex-vault` folder contains the keys for Vault and same for Kong with `edgex-kong`.

- [x] Test that one can use the certificate from $SNAP_DATA/pki/ca/ca.pem to connect with TLS host verification on with Vault and Kong

copy the CA to a user-readable area:
```
$ sudo cat /var/snap/edgexfoundry/current/pki/ca/ca.pem > ca.pem
```

Kong uses a cert signed by the CA (on port 8443):
```
$ curl --cacert ca.pem https://localhost:8443; echo
{"message":"no Route matched with those values"}
```
Kong uses the specific cert from the edgex-kong folder (note you need `-servername` here to set the Host header and thus have Kong respond with the right cert, it can use different certs for different Host/SNIs settings):
```
$ sudo cat /var/snap/edgexfoundry/current/pki/edgex-kong/server.crt > expected-kong.pem
$ openssl s_client -connect localhost:8443 -servername localhost 2>/dev/null </dev/null |  sed -ne '/-BEGIN CERTIFICATE-/,/-END CERTIFICATE-/p' > kong.pem
$ diff -u kong.pem expected-kong.pem
```

Vault uses a cert signed by the CA (on port 8200):
```
$ curl --cacert ca.pem https://localhost:8200; echo 
<a href="/ui/">Temporary Redirect</a>.


```

Vault uses the specific cert from edgex-vault folder:
```
$ openssl s_client -connect localhost:8200 -servername localhost 2>/dev/null </dev/null |  sed -ne '/-BEGIN CERTIFICATE-/,/-END CERTIFICATE-/p' > vault.pem
$ sudo cat /var/snap/edgexfoundry/current/pki/edgex-vault/server.crt > expected-vault.pem
$ diff -u vault.pem expected-vault.pem
```

Fixes #1840


anonymouse64:
recheck

anonymouse64:
> I found everything disabled except for security-secrets-setup. Looking at the configure hook, it's not modified when security-secretstore-setup is touched. So I think this may be a typo?

Ah sorry, actually this is my problem. `security-secrets-setup` should be disabled when we turn off `security-secret-store`. I am not doing this in the configure hook, I will fix this.

> You also seem to imply that kong-daemon and postgres may not be disabled per issue #1963, however both were disabled for me. Am I missing something?

I mentioned this because it _may_ affect the snap, I'm not sure if it will or not. If you have kong-daemon and postgres disabled then that's fine, I was just saying if they weren't disabled then that is probably bug 1963 coming up. How bug 1963 gets triggered is that we have logic wherein if you turn off a service that has other services dependent on it, then those other services should be turned off as well (and the inverse problem with turning on, so when you turn on a service with dependent services, those dependent services should also be turned on). If I were to guess, you can probably reproduce that bug with this PR by doing the following (in order):

1. `snap set security-proxy=on`
2. `snap set security-secret-store=off`

and you may find that kong, postgres, etc. are still enabled after the second command because of how the logic/loop works in the configure hook. We don't currently have a way of asking snapd "what changed" to cause the configure hook to run, so we have to loop over all the services checking them all, and depending on the order in which we loop over them we might disable the services from a prior key, then a later key will inadvertently re-enable them. I don't have time to fix this issue just yet in the configure hook, hence why I mentioned it might happen.

> Finally, for the last two tests:

Sure, sorry this is on me for not providing more complete testing instructions. I will update the PR description with these.

anonymouse64:
@tonyespy updated testing instructions

anonymouse64:
@tonyespy So after looking into this a bit, I have determined that the behavior around security-proxy-setup is a known bug with it and I think there's nothing specific we should do in the snap around this. We could drop a "completed" file somewhere to prevent it from running multiple times, but I think that security-proxy-setup should handle this more gracefully. 

I filed https://github.com/edgexfoundry/edgex-go/issues/2089 to this effect, but I think we should continue with this PR regardless of whether that issue gets fixed for Fuji or not.

PR #2065, opened 2019-11-04
notifications: refactor to remove dbClient global variable

michaelestrin:
Fixes: #2061 

Signed-off-by: Michael Estrin m.estrin@dell.com

PR #2066, opened 2019-11-04
notifications: refactor to remove Configuration global variable

michaelestrin:
Fixes: #2062

Signed-off-by: Michael Estrin m.estrin@dell.com

### DIFF
--- a/internal/support/notifications/container/config.go
+++ b/internal/support/notifications/container/config.go
@@ -12,10 +12,19 @@
  * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
- *
  *******************************************************************************/
 
-package notifications
+package container
 
-// Global variables
-var Configuration = &ConfigurationStruct{}
+import (
+	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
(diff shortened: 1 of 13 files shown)


michaelestrin:
Builds on https://github.com/edgexfoundry/edgex-go/pull/2065

michaelestrin:
Rebased.

internal/support/notifications/container/config.go
```diff
@@ -1,7 +1,5 @@
 /*******************************************************************************
- * Copyright 2017 Dell Inc.
- * Copyright 2018 Dell Technologies Inc.
- * Copyright (c) 2019 Intel Corporation
```
brandonforster: According to the Apache License, derivative works must have original copyrights preserved. Please restore the originals.

michaelestrin: This change was made by Goland when I moved the file from one directory to another.  It wasn't an intentional change on my part and I missed it when I reviewed the changes prior to creating a PR.

While I don't agree that their remove is a violation of the Apache license, I will manually restore the copyright chain as requested.

michaelestrin:
Restored copyright notices as requested.

internal/support/notifications/container/config.go
brandonforster: Totally understand. I just don't want to do anything that might offend any companies or come ANYWHERE NEAR lawyers getting mad.

PR #2067, opened 2019-11-04
Refactor Metadata-related code to remove LoggingClient global variable.

### LINKED ISSUE #2018: Metadata: refactor to remove LoggingClient global variable

akramtexas:
Fix #2018 
- Update core-metadata to leverage the Dependency Injection Container(DIC) during bootstrap and pass a `LoggingClient` instance down the call stack to eliminate the need for the `LoggingClient` global variable.
- Update tests to use a DIC when loading REST routes as they will need the DIC to obtain a reference of the `LoggingClient`.
- Also, update any references to methods/functions which have updated signatures to accept a `LoggingClient` as part of changes to the production code.

### DIFF
--- a/cmd/core-metadata/main.go
+++ b/cmd/core-metadata/main.go
@@ -48,7 +48,8 @@ func main() {
 	flag.Usage = usage.HelpCallback
 	flag.Parse()
 
-	httpServer := httpserver.NewBootstrap(metadata.LoadRestRoutes())
+	dic := di.NewContainer(di.ServiceConstructorMap{})
+	httpServer := httpserver.NewBootstrap(metadata.LoadRestRoutes(dic))
 	bootstrap.Run(
 		configDir,
 		profileDir,
@@ -57,7 +58,7 @@ func main() {
 		clients.CoreMetaDataServiceKey,
 		metadata.Configuration,
 		startupTimer,
-		di.NewContainer(di.ServiceConstructorMap{}),
+		dic,
 		[]interfaces.BootstrapHandler{
 			secret.NewSecret().BootstrapHandler,
 			database.NewDatabase(&httpServer, metadata.Configuration).BootstrapHandler,
--- a/internal/core/metadata/init.go
+++ b/internal/core/metadata/init.go
@@ -27,7 +27,6 @@ import (
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/coredata"
-	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/notifications"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
 )
@@ -35,7 +34,6 @@ import (
 // Global variables
 var Configuration = &ConfigurationStruct{}
 var dbClient interfaces.DBClient
-var LoggingClient logger.LoggingClient
 var nc notifications.NotificationsClient
 var vdc coredata.ValueDescriptorClient
 var httpErrorHandler errorconcept.ErrorHandler
@@ -43,10 +41,9 @@ var httpErrorHandler errorconcept.ErrorHandler
 // BootstrapHandler fulfills the BootstrapHandler contract and performs initialization needed by the metadata service.
 func BootstrapHandler(wg *sync.WaitGroup, ctx context.Context, startupTimer startup.Timer, dic *di.Container) bool {
 	// update global variables.
-	LoggingClient = container.LoggingClientFrom(dic.Get)
 	dbClient = container.DBClientFrom(dic.Get)
 
-	httpErrorHandler = errorconcept.NewErrorHandler(LoggingClient)
+	httpErrorHandler = errorconcept.NewErrorHandler(container.LoggingClientFrom(dic.Get))
 
 	// initialize clients required by service.
 	registryClient := container.RegistryFrom(dic.Get)
--- a/internal/core/metadata/rest_addressable_test.go
+++ b/internal/core/metadata/rest_addressable_test.go
@@ -20,7 +20,6 @@ import (
 	"errors"
 	"net/http"
 	"net/http/httptest"
-	"os"
 	"strconv"
 	"testing"
 
@@ -32,6 +31,7 @@ import (
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+
 	"github.com/gorilla/mux"
 	"github.com/stretchr/testify/mock"
 )
@@ -52,13 +52,9 @@ var ErrorPathParam = "%zz"
 // ErrorPortPathParam path parameter used to trigger an error in the `restGetAddressableByPort` function where the port variable is expected to be a number.
 var ErrorPortPathParam = "abc"
 
-func TestMain(m *testing.M) {
-	LoggingClient = logger.NewMockClient()
-	httpErrorHandler = errorconcept.NewErrorHandler(LoggingClient)
-	os.Exit(m.Run())
-}
-
 func TestGetAllAddressables(t *testing.T) {
+
+	httpErrorHandler = errorconcept.NewErrorHandler(logger.NewMockClient())
 	Configuration = &ConfigurationStruct{Service: config.ServiceInfo{MaxResultCount: 10}}
 	defer func() { Configuration = &ConfigurationStruct{} }()
 
@@ -98,8 +94,7 @@ func TestGetAllAddressables(t *testing.T) {
 
 			dbClient = tt.dbMock
 			rr := httptest.NewRecorder()
-			handler := http.HandlerFunc(restGetAllAddressables)
-			handler.ServeHTTP(rr, tt.request)
+			restGetAllAddressables(rr, logger.NewMockClient())
 			response := rr.Result()
 			if response.StatusCode != tt.expectedStatus {
 				t.Errorf("status code mismatch -- expected %v got %v", tt.expectedStatus, response.StatusCode)
@@ -159,8 +154,7 @@ func TestAddAddressable(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			dbClient = tt.dbMock
 			rr := httptest.NewRecorder()
-			handler := http.HandlerFunc(restAddAddressable)
-			handler.ServeHTTP(rr, tt.request)
+			restAddAddressable(rr, tt.request, logger.NewMockClient())
 			response := rr.Result()
 			if response.StatusCode != tt.expectedStatus {
 				t.Errorf("status code mismatch -- expected %v got %v", tt.expectedStatus, response.StatusCode)
@@ -233,8 +227,7 @@ func TestUpdateAddressable(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			dbClient = tt.dbMock
 			rr := httptest.NewRecorder()
-			handler := http.HandlerFunc(restUpdateAddressable)
-			handler.ServeHTTP(rr, tt.request)
+			restUpdateAddressable(rr, tt.request, logger.NewMockClient())
 			response := rr.Result()
 			if response.StatusCode != tt.expectedStatus {
 				t.Errorf("status code mismatch -- expected %v got %v", tt.expectedStatus, response.StatusCode)
--- a/internal/core/metadata/rest_device_test.go
+++ b/internal/core/metadata/rest_device_test.go
@@ -22,18 +22,14 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/core/metadata/interfaces"
 	"github.com/edgexfoundry/edgex-go/internal/core/metadata/interfaces/mocks"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
-	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 	"github.com/pkg/errors"
 )
 
 func TestGetAllDevices(t *testing.T) {
 	Configuration = &ConfigurationStruct{Service: config.ServiceInfo{MaxResultCount: 1}}
-	req, err := http.NewRequest(http.MethodGet, clients.ApiBase+"/"+DEVICE, nil)
-	if err != nil {
-		t.Error(err)
-		return
-	}
 
 	tests := []struct {
 		name           string
@@ -48,9 +44,7 @@ func TestGetAllDevices(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			dbClient = tt.dbMock
 			rr := httptest.NewRecorder()
-			handler := http.HandlerFunc(restGetAllDevices)
-
-			handler.ServeHTTP(rr, req)
+			restGetAllDevices(rr, logger.NewMockClient())
 			response := rr.Result()
 			if response.StatusCode != tt.expectedStatus {
 				t.Errorf("status code mismatch -- expected %v got %v", tt.expectedStatus, response.StatusCode)
--- a/internal/core/metadata/rest_deviceprofile.go
+++ b/internal/core/metadata/rest_deviceprofile.go
@@ -16,7 +16,6 @@ package metadata
 
 import (
 	"encoding/json"
-	"gopkg.in/yaml.v2"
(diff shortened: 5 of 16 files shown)


cmd/core-metadata/main.go
```diff
@@ -15,11 +15,15 @@ package main
 
 import (
 	"flag"
+	"github.com/edgexfoundry/edgex-go/internal/system/agent/container"
```
jdharms: These imports are in the wrong location, should be below.

akramtexas: @jdharms I had made a slew of updates to `reorder` imports (across numerous `core-metadata` files) and looks like I missed this one. Thanks for catching it 👍

internal/core/metadata/rest_device_test.go
```diff
@@ -19,17 +19,20 @@ import (
...
 	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
-	"github.com/edgexfoundry/go-mod-core-contracts/clients"
-	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
-	"github.com/pkg/errors"
 )
 
 func TestGetAllDevices(t *testing.T) {
 	Configuration = &ConfigurationStruct{Service: config.ServiceInfo{MaxResultCount: 1}}
-	req, err := http.NewRequest(http.MethodGet, clients.ApiBase+"/"+DEVICE, nil)
+	_, err := http.NewRequest(http.MethodGet, clients.ApiBase+"/"+DEVICE, nil)
```
jdharms: @michaelestrin If we're not using the request anymore--do we still need to call `http.NewRequest` at all?

internal/core/metadata/rest_deviceprofile_test.go
```diff
@@ -109,12 +109,10 @@ func TestGetAllProfiles(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			LoggingClient = logger.MockLogger{}
 			Configuration = &ConfigurationStruct{Service: config.ServiceInfo{MaxResultCount: len(TestDeviceProfiles)}}
 			dbClient = tt.dbMock
 			rr := httptest.NewRecorder()
-			handler := http.HandlerFunc(restGetAllDeviceProfiles)
-			handler.ServeHTTP(rr, tt.request)
```
jdharms: @michaelestrin Similarly, as we're not using `tt.request` anymore, should it be refactored out now?

internal/core/metadata/rest_devicereport.go
```diff
@@ -261,31 +283,39 @@ func restDeleteReportByName(w http.ResponseWriter, r *http.Request) {
...
 
 	return nil
 }
 
 // Notify the associated device services to the device report
-func notifyDeviceReportAssociates(dr models.DeviceReport, action string) error {
+func notifyDeviceReportAssociates(
+	dr models.DeviceReport,
+	action string,
+	loggingClient logger.LoggingClient) error {
```
jdharms: @michaelestrin This is extremely minor, but what do we think about renaming loggingClient to lc now that the scope is no longer global?

internal/core/metadata/utils.go
```diff
@@ -25,20 +26,20 @@ const (
...
-		LoggingClient.Error(err.Error())
+		loggingClient.Error(err.Error())
 	}
 
-	LoggingClient.Info(bodyString)
+	loggingClient.Info(bodyString)
 }
 
-func checkMaxLimit(limit int) error {
+func checkMaxLimit(limit int, loggingClient logger.LoggingClient) error {
```
jdharms: @michaelestrin I'm wondering if it's more appropriate for `checkMaxLimit` to not log its error and push that responsibility to the caller.

internal/core/metadata/rest_devicereport.go
michaelestrin: @jdharms and I discussed off-line earlier today.  Decision reached to leave implementation here (and elsewhere) as is and handle parameter naming conversion in subsequent PR(s).

cmd/core-metadata/main.go
```diff
@@ -48,16 +51,25 @@ func main() {
...
 
-	httpServer := httpserver.NewBootstrap(metadata.LoadRestRoutes())
+	configuration := &metadataConfig.ConfigurationStruct{}
+	dic := di.NewContainer(di.ServiceConstructorMap{
+		container.ConfigurationName: func(get di.Get) interface{} {
+			return configuration
+		},
+		bootstrapContainer.ConfigurationInterfaceName: func(get di.Get) interface{} {
+			return get(container.ConfigurationName)
+		},
```
michaelestrin: At this point, your container should be instantiated with an empty map.  

When you tackle the follow-on issue to eliminate the Configuration global, you'll need to add these.  And when you do, you'll need to spin your own helper to type assert and return metadata's configuration struct (you're using system management agent's helper and returning the system management agent's configuration struct in the above code).

cmd/core-metadata/main.go
```diff
@@ -48,16 +51,25 @@ func main() {
 	flag.Usage = usage.HelpCallback
 	flag.Parse()
 
-	httpServer := httpserver.NewBootstrap(metadata.LoadRestRoutes())
+	configuration := &metadataConfig.ConfigurationStruct{}
```
michaelestrin: Premature.  Remove and reference the existing metadata.Configuration global where configuration is actually referenced/used in the code that follows.

internal/core/metadata/init.go
```diff
@@ -27,26 +27,24 @@ import (
...
 
 // BootstrapHandler fulfills the BootstrapHandler contract and performs initialization needed by the metadata service.
 func BootstrapHandler(wg *sync.WaitGroup, ctx context.Context, startupTimer startup.Timer, dic *di.Container) bool {
 	// update global variables.
-	LoggingClient = container.LoggingClientFrom(dic.Get)
+	loggingClient := container.LoggingClientFrom(dic.Get)
 	dbClient = container.DBClientFrom(dic.Get)
 
-	httpErrorHandler = errorconcept.NewErrorHandler(LoggingClient)
+	httpErrorHandler = errorconcept.NewErrorHandler(loggingClient)
```
michaelestrin: Since this is the only usage of loggingClient, you could eliminate line 44 above (loggingClient assignment) and replace `loggingClient` here with `container.LoggingClientFrom(dic.Get)`

internal/core/metadata/rest_addressable_test.go
```diff
@@ -53,8 +53,7 @@ var ErrorPathParam = "%zz"
 var ErrorPortPathParam = "abc"
 
 func TestMain(m *testing.M) {
-	LoggingClient = logger.NewMockClient()
-	httpErrorHandler = errorconcept.NewErrorHandler(LoggingClient)
+	httpErrorHandler = errorconcept.NewErrorHandler(logger.NewMockClient())
 	os.Exit(m.Run())
 }
```
michaelestrin: `TestMain` doesn't actually test anything and only exists, I think, to goose code coverage stats.  Please remove it.  This is consistent with the equivalent/corresponding work done on other services.

internal/core/metadata/rest_device_test.go
michaelestrin: I concur with @jdharms' assessment.  This (and the follow on `if` construct should be removed as they serve no purpose.  (Nice catch @jdharms!)

internal/core/metadata/rest_deviceprofile.go
```diff
@@ -457,12 +473,16 @@ func restGetYamlProfileById(w http.ResponseWriter, r *http.Request) {
 }
 
 // Notify the associated device services for changes in the device profile
-func notifyProfileAssociates(dp models.DeviceProfile, dl device.DeviceLoader, action string) error {
+func notifyProfileAssociates(
+	dp models.DeviceProfile,
+	dl device.DeviceLoader,
+	action string,
+	loggingClient logger.LoggingClient) error {
```
michaelestrin: Calling out once; may be other instances.  When folding function/method definitions, to assist readability please leave an empty line after the function's/method's definition and before the first line of the function's/method's implementation.

internal/core/metadata/rest_deviceprofile_test.go
michaelestrin: I concur.  Since we're not using it anywhere, `request` should be removed from the anonymous struct assigned to the `tests` variable.

internal/core/metadata/rest_deviceservice_test.go
```diff
@@ -45,7 +47,7 @@ var testError = errors.New("some error")
 
 func TestGetAllDeviceServices(t *testing.T) {
 	Configuration = &ConfigurationStruct{Service: config.ServiceInfo{MaxResultCount: 1}}
-	req, err := http.NewRequest(http.MethodGet, testDeviceServiceURI, nil)
+	_, err := http.NewRequest(http.MethodGet, testDeviceServiceURI, nil)
```
michaelestrin: Same as commented elsewhere.  request is not used.  Code should be removed.

internal/core/metadata/rest_provisionwatcher.go
```diff
@@ -371,15 +397,23 @@ func updateProvisionWatcherFields(from models.ProvisionWatcher, to *models.Provi
...
 		return err
 	}
 
 	// Notify the service
-	if err = notifyAssociates([]models.DeviceService{ds}, pw.Id, action, models.PROVISIONWATCHER); err != nil {
+	if err = notifyAssociates([]models.DeviceService{ds},
+		pw.Id,
+		action,
+		models.PROVISIONWATCHER,
+		loggingClient); err != nil {
```
michaelestrin: Preference would be to do the assignment before the if statement.  This results in a single line if statement and is IMO more readable/parseable.

internal/core/metadata/router.go
```diff
@@ -19,47 +19,61 @@ import (
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 	"github.com/gorilla/mux"
 
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
```
michaelestrin: Group your imports.

internal/core/metadata/utils.go
```diff
@@ -15,6 +15,7 @@ package metadata
 
 import (
 	"github.com/edgexfoundry/edgex-go/internal/core/data/errors"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
```
michaelestrin: Group your imports; separate with system imports first.

internal/core/metadata/utils.go
michaelestrin: Concur.  Further, believe the existing implementation is already logging (via caller's subsequent call to `httpErrorHandler.Handle` when `err != nil`) and thus the call to `loggingClient.Error` below is redundant and can be removed.

internal/core/metadata/rest_addressable_test.go
akramtexas: - It turns out that removal of `TestMain` leads to the following error during `make test`, so I have retained it for now:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x1588c62]

goroutine 23 [running]:
testing.tRunner.func1(0xc00016ce00)
        /usr/local/go/src/testing/testing.go:830 +0x392
panic(0x1671920, 0x1cbbd50)
        /usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/edgexfoundry/edgex-go/internal/core/metadata.restGetAllAddressables(0x184e720, 0xc000129300, 0x185a800, 0x1ceebb0)
        /Users/akramahmad/go/src/github.com/edgexfoundry/edgex-go/internal/core/metadata/rest_addressable.go:40 +0x422
github.com/edgexfoundry/edgex-go/internal/core/metadata.TestGetAllAddressables.func2(0xc00016ce00)
        /Users/akramahmad/go/src/github.com/edgexfoundry/edgex-go/internal/core/metadata/rest_addressable_test.go:92 +0xf2
testing.tRunner(0xc00016ce00, 0xc000133360)
```
where, Line 40:
    httpErrorHandler.HandleOneVariant(w, err, errorconcept.Common.LimitExceeded, errorconcept.Default.InternalServerError)

and, Line 92:
      restGetAllAddressables(rr, logger.NewMockClient())

internal/core/metadata/utils.go
akramtexas: - Done ✅ 
- Please let me know if we're good with following the following convention:
`standard library imports` -> `blank line` -> `third party imports` -> `blank line` -> `edgex imports`

akramtexas:
@michaelestrin PR is now ready again for your review, thanks.

internal/core/metadata/rest_addressable_test.go
brandonforster: To remove `TestMain()`, you need to add this line to every test until you remove the `httpErrorHandler` global:

`httpErrorHandler = errorconcept.NewErrorHandler(logger.NewMockClient())`

michaelestrin: Discussed out-of-band on Slack; request is to remove hidden dependency on TestMain so you can remove it as part of this PR.

internal/core/metadata/rest_provisionwatcher.go
michaelestrin: Not updated.  Expecting this:

```
err = notifyAssociates([]models.DeviceService{ds}, pw.Id, action, models.PROVISIONWATCHER, loggingClient);
if  err != nil {
		return err
}```

File: internal/core/metadata/router.go [hunk 15]
```diff
@@ -17,49 +17,62 @@ import (
 	"net/http"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	"github.com/gorilla/mux"
```
michaelestrin: Empty line before.  Move to after all github.com/edgexfoundry sections.

michaelestrin:
>Please let me know if we're good with following the following convention:
standard library imports -> blank line -> third party imports -> blank line -> edgex imports

The convention I've seen in the code and try to follow is standard library imports, edgex imports, third party imports (all separated with black lines).  Further, edgex imports are sectioned out by module .

internal/core/metadata/rest_addressable_test.go
akramtexas: - Yes, the next step here is for me to figure out the hidden dependencies some tests have on `TestMain` and fulfill those dependencies within each test that has them so I can remove `TestMain` 👍

akramtexas:
> > Please let me know if we're good with following the following convention:
> > standard library imports -> blank line -> third party imports -> blank line -> edgex imports
> 
> The convention I've seen in the code and try to follow is standard library imports, edgex imports, third party imports (all separated with black lines). Further, edgex imports are sectioned out by module .
- Sounds good: I will follow up on implementing the convention you point out above 👍

internal/core/metadata/rest_addressable_test.go
akramtexas: @michaelestrin Once I have addressed this comment as well, I will notify (via `Request Review` option up atop) at that time that this PR is ready again for your review.

internal/core/metadata/rest_addressable_test.go
```diff
@@ -52,13 +52,9 @@ var ErrorPathParam = "%zz"
...
 
-func TestMain(m *testing.M) {
-	LoggingClient = logger.NewMockClient()
-	httpErrorHandler = errorconcept.NewErrorHandler(LoggingClient)
-	os.Exit(m.Run())
-}
-
 func TestGetAllAddressables(t *testing.T) {
+
+	httpErrorHandler = errorconcept.NewErrorHandler(logger.NewMockClient())
```
michaelestrin: Nitpick: you create a logger.NewMockClient() here and elsewhere in this test function.  You should only instantiate and use one through the test function.

michaelestrin:
@akramtexas please review comments to incorporate nitpick changes into follow-on issue's PR.

internal/core/metadata/rest_provisionwatcher.go
```diff
@@ -21,12 +21,17 @@ import (
...
+
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	"github.com/edgexfoundry/go-mod-core-contracts/models"
+
 	"github.com/gorilla/mux"
 )
 
-func restGetProvisionWatchers(w http.ResponseWriter, _ *http.Request) {
+func restGetProvisionWatchers(
```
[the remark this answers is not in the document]
akramtexas: @michaelestrin Good point: And yep, I will incorporate this while working on the follow-on issue's PR 👍

PR #2079, opened 2019-11-05
Command: refactor to remove mdc global variable

### LINKED ISSUE #2015: Command: refactor to remove mdc global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2014

brandonforster:
Fixes #2015 

### DIFF
--- a/internal/core/metadata/containers/device.go
+++ b/internal/core/metadata/containers/device.go
@@ -1,0 +1,29 @@
+/*******************************************************************************
+ * Copyright 2019 Dell Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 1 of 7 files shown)


internal/core/metadata/containers/device.go
```diff
@@ -0,0 +1,29 @@
...
+package container
+
+import (
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata"
+
+	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
+)
+
+// DeviceClientName contains the name of the client implementation in the DIC.
+var DeviceClientName = di.TypeInstanceToName((*metadata.DeviceClient)(nil))
```
michaelestrin: `DeviceClientName` -> `MetadataDeviceClientName`

internal/core/metadata/containers/device.go
```diff
@@ -0,0 +1,29 @@
...
+
+// DeviceClientFrom helper function queries the DIC and returns the client implementation.
+func DeviceClientFrom(get di.Get) metadata.DeviceClient {
```
michaelestrin: `DeviceClientFrom` -> `MetadataDeviceClientFrom`

Issue #2012, opened 2019-10-25
Refactor core-command and apply errorConcepts


brandonforster:
# 🚀 Feature Request

### Relevant Package

- core-command

### Description

In #1860, work was attempted to apply errorConcepts to core-command. It was realized during the course of that work (see comments on #1972) that core-command, as written, does error checking in too many places and has a large body of code with a very different style and flow to the other microservices. To this end, this issue attempts to capture two work items:

- Refactor core-command to bring its structure and flow in line with the patterns established by core-data and core-metadata
- During that refactor, ensure that errors are processed and handled centrally so that the errorConcepts pattern can be used and error handling can be more readily modified in the future.
- This is a good opportunity to write unit tests.

michaelestrin:
Recommend you don't apply the operator pattern here.  Related issues are being created to remove the global variables from each core service which will obviate the need to use the existing operator pattern.

michaelestrin:
During a conversation, @brandonforster decided to make this work dependent upon the completion of https://github.com/edgexfoundry/edgex-go/issues/2017.

PR #2082, opened 2019-11-06
Command: refactor to remove httpErrorHandler global variable

### LINKED ISSUE #2016: Command: refactor to remove httpErrorHandler global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2015

brandonforster:
Fixes #2016.

Must be merged AFTER #2079.

### DIFF
--- a/internal/core/command/init.go
+++ b/internal/core/command/init.go
@@ -19,7 +19,8 @@ import (
 	"context"
 	"sync"
 
-	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/container"
+	container "github.com/edgexfoundry/edgex-go/internal/core/metadata/containers"
+	bootstrapContainer "github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/container"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/startup"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/endpoint"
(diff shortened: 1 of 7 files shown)


brandonforster:
This is trivial because core-command doesn't actually use errorConcepts. I added the global here in prep for the second half of #1860 which has now been superseded by #2012. I expect that #2012 will implement actual direct injection for the httpErrorHandler.

internal/core/command/init.go
```diff
@@ -34,14 +34,11 @@ import (
...
 
-// Global ErrorConcept variables
-var httpErrorHandler errorconcept.ErrorHandler
-
 // BootstrapHandler fulfills the BootstrapHandler contract and performs initialization needed by the command service.
 func BootstrapHandler(wg *sync.WaitGroup, ctx context.Context, startupTimer startup.Timer, dic *di.Container) bool {
 	loggingClient := bootstrapContainer.LoggingClientFrom(dic.Get)
 	registryClient := bootstrapContainer.RegistryFrom(dic.Get)
-	httpErrorHandler = errorconcept.NewErrorHandler(loggingClient)
+	errorconcept.NewErrorHandler(loggingClient)
```
michaelestrin: Is this call actually needed?

brandonforster: Not really, but without it we remove the logging client pull which is then going to get put right back by #2012.

Truthfully, I think we should close this PR without merge, close the issue, and handle this particular variable as a part of #2012. Anything reasonable done here is going to be immediately undone by #2012, and anything silly (like the current status of the PR) probably isn't worth doing.

michaelestrin:
> Truthfully, I think we should close this PR without merge, close the issue, and handle this particular variable as a part of #2012. Anything reasonable done here is going to be immediately undone by #2012, and anything silly (like the current status of the PR) probably isn't worth doing.

I'm aligned with that.  Will close this PR and the corresponding issue.

Issue #2034, opened 2019-10-29
Metadata: Update Description Does Not


michaelestrin:
From Phan Van Hai via Slack core channel:

> I want to change the "description" of the device. I sent a request to update, although it received a successful response but in reality, it was not updated. 
> 
> ![image](https://user-images.githubusercontent.com/464327/67742482-29c5c080-f9ea-11e9-8770-46d2eb8dd28f.png)
> 
> ![update_descriptor](https://user-images.githubusercontent.com/464327/67742547-5d084f80-f9ea-11e9-86af-b41d443ddf1b.png)
> 
> ![get_descriptor](https://user-images.githubusercontent.com/464327/67742551-60034000-f9ea-11e9-990a-145cd16128e0.png)
> 

Reported against Edinburgh (1.0.1).  Verify/validate/fix against Edinburgh, Fuji, and master.

PR #2085, opened 2019-11-06
fix: #2034 fixes bug around named return values (edinburgh)

jdharms:
This is a duplicate of #2083 (which targets the fuji branch)

/pkg/db/mongo/models/device.go's ToContract() method assigned
some values to the named contract.Device in the return value.
This value was overwritten at the end of the method causing these
assignments to be discarded.

This change removes all named return values from the file and
adjusts the methods accordingly.

fixes: #2034


anonymouse64:
Does this really need to be fixed for Edinburgh considering that Edinburgh will no longer be supported in approximately a week?

PR #2086, opened 2019-11-06
Fixes issue #1710 for security proxy setup

### LINKED ISSUE #1710: Develop dependency tracker for security functionality containers to ensure successful launch.
Need dependency tracker to determine each container service involved in security functionality set up is launched and available.
Docker-compose's hook of "launched" is inadequate. Use the Ubuntu snaps scripts as a starting point.

Tasks to complete this:

- [x] ~~In go-mod-registry, add function to allow registering a script as a health check~~ - this is unnecessary anymore, see followup comment
- [x] in docker-edgex-consul, write script to register things that aren't go services like Kong to be registered as services in consul - this is supported via config files with consul
- [x] in edgex-go modify the security-secrets-setup container to wait for security-secrets-setup to be done before running vault (note we will defer splitting security-secrets-setup into its own container separate from vault to later) (also note this is basically already done)
- [x] in docker-edgex-consul, write consul health script for kong-migration - https://github.com/edgexfoundry/docker-edgex-consul/pull/2
- [x] in docker-edgex-consul, write consul health script for kong-database - https://github.com/edgexfoundry/docker-edgex-consul/pull/2
- [x] in docker-edgex-consul, write consul health script for kong - https://github.com/edgexfoundry/docker-edgex-consul/pull/2
- [x] in docker-edgex-volume, register services kong, kong-migration, kong-database as services in consul with local health scripts defined - this is supported via config files with consul - https://github.com/edgexfoundry/docker-edgex-volume/pull/9
- [x] in docker-edgex-volume, register services such as security-secrets-setup, security-proxy-setup, and security-secretstore-setup with local health scripts defined - this is supported via config files with consul
- [x] in docker-edgex-consul, write consul health script for security-secrets-setup and define in consul config file
- [x] in docker-edgex-consul, write consul health script for security-secretstore-setup and define in consul config file
- [x] in docker-edgex-consul, write consul health script for security-proxy-setup and define in consul config file
- [x] In go-mod-secrets, update to wait for the service known as `security-secretstore-setup` in consul is healthy before accessing vault
- [x] update edgex-go to use new version of go-mod-secrets

jim-wang-intel:
Fixes #1710  as part of dependency check for security proxy setup bootstrap starting.
Add for-loop of retries until it succeeds.

### DIFF
--- a/internal/security/proxy/certs.go
+++ b/internal/security/proxy/certs.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/edgexfoundry/edgex-go/internal"
 )
@@ -58,9 +59,15 @@ func (cs certificate) Load() (*CertPair, error) {
 	if err != nil {
 		return nil, err
 	}
-	cp, err := cs.retrieve(t)
-	if err != nil {
-		return nil, err
+	var cp *CertPair
+	for {
+		cp, err = cs.retrieve(t)
+		if err != nil {
+			//return nil, err
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		break
 	}
 	err = cs.validate(cp)
 	if err != nil {


internal/security/proxy/certs.go
```diff
@@ -58,9 +59,15 @@ func (cs certificate) Load() (*CertPair, error) {
 	if err != nil {
 		return nil, err
 	}
-	cp, err := cs.retrieve(t)
-	if err != nil {
-		return nil, err
+	var cp *CertPair
+	for {
```
anonymouse64: I don't typically like infinite loops that don't error out at some point. 
That being said, I think the ideal thing we want is for security-proxy-setup to use go-mod-secrets right? If that's the case this is probably fine for Fuji, but we shouldn't pull in this fix for master.

PR #2087, opened 2019-11-06
Add configuration for SecretClient retries (master)

### LINKED ISSUE #2080: Add retry logic within go-mod-secret to utilize the dependence check offered by customized consul
With the merge of https://github.com/edgexfoundry/developer-scripts/pull/172, the configuration for go-mod-secret within edgex-go needs to be changed as well to include retry logic so that it can utilize the features of dependence check provided across the microservice.

AnthonyMBonafide:
Fix #2080

Add configuration properties needed to leverage the retry logic within
the SecretClient. This helps address the issue of temporal coupling,
when security is enabled, between services, and bootstrapping steps by
giving some grace period for dependencies before unsuccessfully
exiting(shutting down) the service.

Update to use the latest version of `go-mod-secrets` which has the retry
logic.

### DIFF
--- a/cmd/core-command/res/configuration.toml
+++ b/cmd/core-command/res/configuration.toml
@@ -51,6 +51,10 @@ Protocol = 'https'
 RootCaCertPath = '/vault/config/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem'
 ServerName = 'localhost'
 TokenFile = '/vault/config/assets/resp-init.json'
+# Number of attemtps to retry retrieving secrets before failing to start the service.
+AdditionalRetryAttempts = 10
+# Amount of time to wait before attempting another retry in nanoseconds.
+RetryWaitPeriod = 1000000000
   [SecretStore.Authentication]
   AuthType = 'X-Vault-Token'
   AuthToken = 'edgex'
--- a/cmd/core-command/res/docker/configuration.toml
+++ b/cmd/core-command/res/docker/configuration.toml
@@ -51,6 +51,10 @@ Protocol = 'https'
 RootCaCertPath = '/vault/config/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem'
 ServerName = 'edgex-vault'
 TokenFile = '/vault/config/assets/resp-init.json'
+# Number of attemtps to retry retrieving secrets before failing to start the service.
+AdditionalRetryAttempts = 10
+# Amount of time to wait before attempting another retry in nanoseconds.
+RetryWaitPeriod = 1000000000
   [SecretStore.Authentication]
   AuthType = 'X-Vault-Token'
 
--- a/cmd/core-data/res/configuration.toml
+++ b/cmd/core-data/res/configuration.toml
@@ -63,6 +63,10 @@ Protocol = 'https'
 RootCaCertPath = '/vault/config/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem'
 ServerName = 'localhost'
 TokenFile = '/vault/config/assets/resp-init.json'
+# Number of attemtps to retry retrieving secrets before failing to start the service.
+AdditionalRetryAttempts = 10
+# Amount of time to wait before attempting another retry in nanoseconds.
+RetryWaitPeriod = 1000000000
   [SecretStore.Authentication]
   AuthType = 'X-Vault-Token'
   AuthToken = 'edgex'
--- a/cmd/core-data/res/docker/configuration.toml
+++ b/cmd/core-data/res/docker/configuration.toml
@@ -62,6 +62,10 @@ Protocol = 'https'
 RootCaCertPath = '/vault/config/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem'
 ServerName = 'edgex-vault'
 TokenFile = '/vault/config/assets/resp-init.json'
+# Number of attemtps to retry retrieving secrets before failing to start the service.
+AdditionalRetryAttempts = 10
+# Amount of time to wait before attempting another retry in nanoseconds.
+RetryWaitPeriod = 1000000000
   [SecretStore.Authentication]
   AuthType = 'X-Vault-Token'
 
--- a/cmd/core-metadata/res/configuration.toml
+++ b/cmd/core-metadata/res/configuration.toml
@@ -63,6 +63,10 @@ Protocol = 'https'
 RootCaCertPath = '/vault/config/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem'
 ServerName = 'localhost'
 TokenFile = '/vault/config/assets/resp-init.json'
+# Number of attemtps to retry retrieving secrets before failing to start the service.
+AdditionalRetryAttempts = 10
+# Amount of time to wait before attempting another retry in nanoseconds.
+RetryWaitPeriod = 1000000000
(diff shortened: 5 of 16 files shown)


AnthonyMBonafide:
Once we get this approved, I will cherry pick the commit and add it to the Fuji branch

cmd/core-command/res/configuration.toml
```diff
@@ -51,7 +51,11 @@ Protocol = 'https'
 RootCaCertPath = '/vault/config/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem'
 ServerName = 'localhost'
 TokenFile = '/vault/config/assets/resp-init.json'
-  [SecretStore.Authentication]
+# Number of attemtps to retry retrieving secrets before failing to start the service.
+AdditionalRetryAttempts = 10
+# Amount of time to wait before attempting another retry in nanoseconds.
+RetryWaitPeriod =  1000000000
```
anonymouse64: I think this is wrong, we shouldn't be using nanoseconds (i.e. ints) here for the time, we should be using a string that is parsed into a time.Duration.

See how I did this for docker-edgex-mongo here: https://github.com/edgexfoundry/docker-edgex-mongo/pull/59

michaelestrin: (What Ian said.)

cmd/core-command/res/configuration.toml
```diff
@@ -51,7 +51,11 @@ Protocol = 'https'
...
+[SecretStore.Authentication]
```
michaelestrin: Is indentation important here?  Before, chapter heading was indented 2 spaces.  Now it isn't.  Commenting once; occurs multiple times.

anonymouse64: FWIW TOML format doesn't care, but we should try to keep indentation consistent for readability

cmd/core-command/res/configuration.toml
AnthonyMBonafide: I'll go forward with this suggestion. However, I disagree with how this is being achieved. In the `docker-edgex-mongo` PR we are duplicating the `SecretConfig` struct and the one that lives in `docker-edgex-mongo` has diverged from `go-mod-secrets` and the `RetryWaitPeriod` is a `string`:

https://github.com/edgexfoundry/docker-edgex-mongo/blob/4c0243a28f29f77b6a324602919251fb9d36520f/internal/pkg/config.go#L64

where the version that exists in `go-mod-secrets` it is a `time.Duration`:

https://github.com/edgexfoundry/go-mod-secrets/blob/f1caf4c75e0add8e50e323efbf869f084199d470/pkg/providers/vault/config.go#L34

My vision for the core-services, is to be ignorant to the underlying configuration for the `SecretClient`(as much as possible), so that we can swap out different versions of `go-mod-secrets` and the only thing we would have to do is change configuration. Keeping on that philosophy, I think it would be better if either:

1. `go-mod-secrets:SecretConfig` is updated so that `RetryWaitPeriod` is an `int` and we pick a default unit of time.

2. Update the `go-mod-secrets` to perform the `time.Duration` parsing and we can simply pass the configuration to `go-mod-secrets and this will be consistent for all consumers.

Just my thoughts

AnthonyMBonafide: I'll go forward with this suggestion. However, I disagree with how this is being achieved. In the `docker-edgex-mongo` PR we are duplicating the `SecretConfig` struct and the one that lives in `docker-edgex-mongo` has diverged from `go-mod-secrets` and the `RetryWaitPeriod` is a `string`:

https://github.com/edgexfoundry/docker-edgex-mongo/blob/4c0243a28f29f77b6a324602919251fb9d36520f/internal/pkg/config.go#L64

where the version that exists in `go-mod-secrets` it is a `time.Duration`:

https://github.com/edgexfoundry/go-mod-secrets/blob/f1caf4c75e0add8e50e323efbf869f084199d470/pkg/providers/vault/config.go#L34

My vision for the core-services, is to be ignorant to the underlying configuration for the `SecretClient`(as much as possible), so that we can swap out different versions of `go-mod-secrets` and the only thing we would have to do is change configuration. Keeping on that philosophy, I think it would be better if either:

1. `go-mod-secrets:SecretConfig` is updated so that `RetryWaitPeriod` is an `int` and we pick a default unit of time.

2. Update the `go-mod-secrets` to perform the `time.Duration` parsing and we can simply pass the configuration to `go-mod-secrets and this will be consistent for all consumers.

Just my thoughts

anonymouse64: @AnthonyMBonafide 

> so that we can swap out different versions of go-mod-secrets and the only thing we would have to do is change configuration

what's the use case for this? It kinda sounds like you're saying "don't break APIs" in go-mod-secrets (i.e. the different versions of go-mod-secrets being just future versions).

I'm +1 on not breaking APIs, but having slightly inconsistent configuration.toml derived structs and SecretConfig structs passed to go-mod-secrets to me is okay, because really the problem we have is that go-mod-secrets is something that talks "go" (and thus gets cool and correct typing with time.Duration) and these other things like docker-edgex-mongo are things that talk "user provided config files" (and thus only get basic strings and numbers). That's a more general problem though and I'd rather push that complexity to clients like docker-edgex-mongo than have that complexity in go-mod-secrets. But that's a personal preference...

FWIW if you feel strongly about this I'd rather go with 2 and have the parsing happen in go-mod-secrets than have all configuration.tomls provide the wait time in nanoseconds.

anonymouse64:
@AnthonyMBonafide @tsconn23 now that https://github.com/edgexfoundry/docker-edgex-mongo/pull/59 is merged if y'all want to switch to having go-mod-secrets do the time.Duration parsing, I suggest we do that for Geneva. 

1. The only change to do what Anthony wants is in what go-mod-secrets accepts from an API perspective (i.e. accept a string instead of a time.Duration)
1. This change doesn't break compatibility for client services config format
1. We haven't tagged go-mod-secrets with a 1.X release so I think it's fine to break go-mod-secret's API

Thoughts? If we do want to make that change for Fuji I can submit a PR doing so to go-mod-secrets and docker-edgex-mongo as well as edgex-go tonight, but I'm worried that we would have to delay the release to account for everyone traveling from the F2F tomorrow/this weekend.

michaelestrin:
We're attempting to troubleshoot an issue with the arm security build.  We need the basic functionality of this PR (i.e. the additional configuration.toml entries).  While I believe the switch to string representation is desirable, it significantly complicates our short term troubleshooting attempts.  

I've created a separate issue to address the conversion of these configuration options from nanoseconds (int) to seconds (string).  I intend to pull this PR in as-is (without the additional conversion).

michaelestrin:
@AnthonyMBonafide Please create a PR to target Fuji and cross reference here.

michaelestrin:
recheck

AnthonyMBonafide:
Recheck

anonymouse64:
@michaelestrin 

> We're attempting to troubleshoot an issue with the arm security build. We need the basic functionality of this PR (i.e. the additional configuration.toml entries). While I believe the switch to string representation is desirable, it significantly complicates our short term troubleshooting attempts.

if you're trying to debug some security arm issues right now, don't you want to pull in this change for Fuji and not master? It seems like you could just close this PR open the same thing against Fuji, merge that and figure out what to do for master/geneva next week

PR #2088, opened 2019-11-06
Fixes issue #1710 for security proxy setup [master branch]

bnevis-i:
Fixes issue #1710 for security proxy setup [master branch]

Signed-off-by: Jim Wang <yutsung.jim.wang@intel.com>7

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -22,6 +22,9 @@ ALL_SERVICES="$ALL_SERVICES support-logging"
 ALL_SERVICES="$ALL_SERVICES support-scheduler"
 ALL_SERVICES="$ALL_SERVICES support-rulesengine"
 
+# app-services
+ALL_SERVICES="$ALL_SERVICES app-service-configurable"
+
 # export services
 ALL_SERVICES="$ALL_SERVICES export-distro"
 ALL_SERVICES="$ALL_SERVICES export-client"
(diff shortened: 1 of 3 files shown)


bnevis-i:
@michaelestrin   Cherry picked for PR #2087 for master too.

internal/security/proxy/certs.go
```diff
@@ -58,9 +59,15 @@ func (cs certificate) Load() (*CertPair, error) {
 	if err != nil {
 		return nil, err
 	}
-	cp, err := cs.retrieve(t)
-	if err != nil {
-		return nil, err
+	var cp *CertPair
+	for {
```
anonymouse64: as mentioned in https://github.com/edgexfoundry/edgex-go/pull/2086/files#r343421187, I think this is the wrong thing to do for master. Can't we use go-mod-secrets in security-proxy-setup? If so then I think we should fix this the _right_ way for master/geneva. 

If this requires changes to go-mod-secrets I'd rather see those in master/geneva than have this on master.

bnevis-i: > Can't we use go-mod-secrets in security-proxy-setup?

Yes.  But by delaying synchronizing the branches we risk introducing a regression in geneva.

> If so then I think we should fix this the right way for master/geneva.

By vetoing the fix you are essentially signing up to do so.

anonymouse64: > But by delaying synchronizing the branches we risk introducing a regression in geneva.

Well it's not necessarily a regression since it's always been broken for Geneva... it never worked to begin with so by not merging this code we're not making Geneva any worse than it would have been anyways...

> By vetoing the fix you are essentially signing up to do so.

I disagree

Issue #2089, opened 2019-11-07
Running security-proxy-setup multiple times now fails


anonymouse64:
Running security-proxy-setup from fuji/master currently fails if done multiple times:

The first time:
```
level=ERROR ts=2019-11-07T02:10:27.987831776Z app=edgex-security-proxy-setup source=logger.go:73 msg="logTarget cannot be blank, using stdout only"
Creating directory: ./logs
level=INFO ts=2019-11-07T02:10:27.98893277Z app=edgex-security-proxy-setup source=requestor.go:40 msg="successfully loaded rootCA certificate."
level=INFO ts=2019-11-07T02:10:27.997430451Z app=edgex-security-proxy-setup source=service.go:64 msg="the service on http://kong:8001 is up successfully"
level=ERROR ts=2019-11-07T02:10:28.039450024Z app=edgex-security-proxy-setup source=certs.go:117 msg="failed to retrieve certificate on path v1/secret/edgex/pki/tls/edgex-kong with error code 404"
level=DEBUG ts=2019-11-07T02:10:29.084115014Z app=edgex-security-proxy-setup source=service.go:147 msg="trying to upload cert to proxy server"
level=INFO ts=2019-11-07T02:10:29.102275442Z app=edgex-security-proxy-setup source=service.go:169 msg="successfully added certificate to the reverse proxy"
level=INFO ts=2019-11-07T02:10:29.117267804Z app=edgex-security-proxy-setup source=service.go:207 msg="successful to set up proxy service for virtualdevice"
level=INFO ts=2019-11-07T02:10:29.131375224Z app=edgex-security-proxy-setup source=service.go:245 msg="successful to set up route for virtualdevice"
level=INFO ts=2019-11-07T02:10:29.13846105Z app=edgex-security-proxy-setup source=service.go:207 msg="successful to set up proxy service for coredata"
level=INFO ts=2019-11-07T02:10:29.14779438Z app=edgex-security-proxy-setup source=service.go:245 msg="successful to set up route for coredata"
level=INFO ts=2019-11-07T02:10:29.154649346Z app=edgex-security-proxy-setup source=service.go:207 msg="successful to set up proxy service for scheduler"
level=INFO ts=2019-11-07T02:10:29.163932141Z app=edgex-security-proxy-setup source=service.go:245 msg="successful to set up route for scheduler"
[...]

PR #2090, opened 2019-11-07
change to more user-friendly return message if the cert already exists in proxy

tingyuz:
This is to fix isssue #2089 . Update user-friendly log message if the certificate already exists in reverse proxy.

Signed-off-by: Tingyu Zeng <Tingyu.Zeng@rsa.com>


tingyuz:
recheck

internal/security/proxy/service.go
```diff
@@ -173,8 +173,12 @@ func (s *Service) postCert(cert CertificateLoader) error {
 		if err != nil {
 			return err
 		}
-		e := fmt.Sprintf("failed to add certificate with errorcode %d, error %s", resp.StatusCode, string(b))
+		reason := string(b)
+		e := fmt.Sprintf("failed to add certificate with errorcode %d, error %s", resp.StatusCode, reason)
 		LoggingClient.Error(e)
+		if (resp.StatusCode == http.StatusBadRequest) && (strings.Index(reason, "existing certificate") != -1) {
+			e = fmt.Sprintf("certificate already exists on reverse proxy")
```
jim-wang-intel: @tingyuz why the http status code not be `http.StatusConflict` when the certificate already exists?

anonymouse64: FWIW the docs don't seem to say what happens when there's already a cert uploaded: https://docs.konghq.com/1.0.x/admin-api/#create-certificate

tingyuz: Please refer my comments on #2089. The original code even checks http.StatusConflict but kong doesn't follow such standard, so we have to get this done this way.

bnevis-i:
recheck

bnevis-i:
The default clause of the following switch statement continues to have no code coverage:
```go
        switch resp.StatusCode {
        case http.StatusOK, http.StatusCreated, http.StatusConflict:
                LoggingClient.Info("successfully added certificate to the reverse proxy")
                break
        default:
                b, err := ioutil.ReadAll(resp.Body)
                if err != nil {
                        return err, false
                }
                reason := string(b)
                e := fmt.Sprintf("failed to add certificate with errorcode %d, error %s", resp.StatusCode, reason)
                LoggingClient.Error(e)
                if (resp.StatusCode == http.StatusBadRequest) && (strings.Index(reason, "existing certificate") != -1) {
                        e = fmt.Sprintf("certificate already exists on reverse proxy")
                        return nil, true
                }
                return errors.New(e), false
        }
```

internal/security/proxy/service.go
jim-wang-intel: ok

tingyuz:
Add the unit test case for the changes, which is, when the cert already exists on the proxy.

internal/security/proxy/service.go
```diff
@@ -171,13 +175,18 @@ func (s *Service) postCert(cert CertificateLoader) error {
...
 		LoggingClient.Error(e)
-		return errors.New(e)
+		if (resp.StatusCode == http.StatusBadRequest) && (strings.Index(reason, "existing certificate") != -1) {
+			e = fmt.Sprintf("certificate already exists on reverse proxy")
+			return nil, true
+		}
+		return errors.New(e), false
 	}
-	return nil
+	return nil, false
```
jim-wang-intel: should return `true` here, too.
```suggestion
	return nil, true
```

bnevis-i: This boolean means "existed" not "was successful"

jim-wang-intel: ok, then please ignore my comment then.

bnevis-i:
Note that we need to pull this fix into master too.

mkbhanda:
To Jim's comment and your response Bryon, it indicates need for better self-documenting code. 
Go supports named return values, see https://tour.golang.org/basics/7
Worth editing.

bnevis-i:
While we're on the the subject of the bad signature for postCert(), generally the error  is the last item in the return list.  I generally dislike returning information in a bool.  Moreover, for this function, there are multiple possible outcomes:  there was an error, it succeeded replacing the cert, or succeeded uploading the cert for the first time.  It may not be unreasonable that a future outcome might be, certs exists but was not replaced.  As such, I think that postCert() should just return an error, and it should take a pointer argument that returns a information enum that tells how it was handled if the function succeeded.

anonymouse64:
+1

tingyuz:
recheck

Issue #2076, opened 2019-11-05
Delete event func for Redis misses the deletion of pushed and device collection


cloudxxx8:
# 🐞 Bug Report

### Affected Services
The issue is located in:  Core-Data

### Is this a regression?

No, it also happens in Delhi and Edinburgh release

### Description and Minimal Reproduction

When creating the event, there are two additional collections added to reference to the event id
```
	conn.Send("ZADD", db.EventsCollection, 0, id)
	conn.Send("ZADD", db.EventsCollection+":created", e.Created, id)
	conn.Send("ZADD", db.EventsCollection+":pushed", e.Pushed, id)
	conn.Send("ZADD", db.EventsCollection+":device:"+e.Device, e.Created, id)
```
https://github.com/edgexfoundry/edgex-go/blob/master/internal/pkg/db/redis/data.go#L1056-L1059

However, the deleteEvent func only deletes two of them
```
	_ = conn.Send("ZREM", db.EventsCollection, id)
	_ = conn.Send("ZREM", db.EventsCollection+":created", id)
```
https://github.com/edgexfoundry/edgex-go/blob/master/internal/pkg/db/redis/data.go#L1095-L1096

## 🔥 Exception or Error
The data will keep growing until out of memory

## 🌍  Your Environment
**Deployment Environment:**
Ubuntu

**EdgeX Version:**
delhi, edinburgh, and fuji

**Anything else relevant?**

cloudxxx8:
add @TomFlem and @FelixTing for awareness

michaelestrin:
cc: @AnthonyMBonafide

AnthonyMBonafide:
This bug seems valid. I just want to point out that there is some async processing happening when events are deleted. See:

https://github.com/AnthonyMBonafide/edgex-go/blob/f07acc231909531283c898e4100e3cc5d499f8b9/internal/pkg/db/redis/data.go#L254-L267

I want to point this out so we can verify this functionality still works and is valid after we provide a fix for this.

brandonforster:
@michaelestrin master on Delhi does not currently build. Please open other issues to resolve that or remove the Delhi tag from this bugfix.

cloudxxx8:
@brandonforster 
I just noticed that 
```
	if e.Checksum != "" {
		_ = conn.Send("ZADD", db.EventsCollection+":checksum:"+e.Checksum, 0, e.ID)
	}
```
Thus, we should also delete checksum for binary event

[...]

PR #2094, opened 2019-11-07
(fuji) Delete event func for Redis misses the deletion of pushed and device collection

brandonforster:
#2076 in Fuji.

Verification steps:

Start Redis. Start core-data with Redis configured as the DB.

First, flush your Redis database with this redis-cli command:
`redis-cli FLUSHALL`

Run the following cURL command:

```
curl -X POST \
  http://localhost:48080/api/v1/event \
  -H 'Accept: */*' \
  -H 'Accept-Encoding: gzip, deflate' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Content-Length: 230' \
  -H 'Content-Type: application/json' \
  -H 'Host: localhost:48080' \
  -H 'cache-control: no-cache' \
  -d '{  
   "origin":1471806386919,
   "device":"cartachometer",
   "readings":[  
      {  
        "origin":1471806386919,
		"device":"cartachometer",
        "name":"luminousIntensity",
        "value":"40"
      }
   ]
}'
```

Copy the UUID returned by the request. Run this additional cURL command to delete the event:

```
curl -X DELETE \
  http://localhost:48080/api/v1/event/id/{UUID} \
  -H 'Accept: */*' \
  -H 'Accept-Encoding: gzip, deflate' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Content-Length: 0' \
  -H 'Host: localhost:48080' \
  -H 'cache-control: no-cache'
```

Finally, run this redis-cli command to verify that the event is deleted and there is nothing in the database:
`redis-cli --scan --pattern '*'`

There should be no returned output from that command.

### DIFF
--- a/internal/pkg/db/redis/data.go
+++ b/internal/pkg/db/redis/data.go
@@ -109,7 +109,7 @@ func (c *Client) UpdateEvent(e correlation.Event) (err error) {
 
 	id := event.ID
 
-	o, err := eventByID(conn, id)
+	o, err := eventByID(conn, id, false)
 	if err != nil {
 		if err == redis.ErrNil {
 			return db.ErrNotFound
@@ -137,7 +137,7 @@ func (c *Client) EventById(id string) (event contract.Event, err error) {
 	conn := c.Pool.Get()
 	defer conn.Close()
 
-	event, err = eventByID(conn, id)
+	event, err = eventByID(conn, id, false)
 	if err != nil {
 		if err == redis.ErrNil {
 			return event, db.ErrNotFound
@@ -201,11 +201,11 @@ func (c *Client) EventCountByDeviceId(id string) (count int, err error) {
 // Delete an event by ID. Readings are not deleted as this should be handled by the contract layer
 // 404 - Event not found
 // 503 - Unexpected problems
-func (c *Client) DeleteEventById(id string) (err error) {
+func (c *Client) DeleteEventById(id string) error {
 	conn := c.Pool.Get()
 	defer conn.Close()
 
-	err = deleteEvent(conn, id)
+	err := deleteEvent(conn, id)
 	if err != nil {
 		if err == redis.ErrNil {
 			return db.ErrNotFound
@@ -1089,11 +1089,21 @@ func addEvent(conn redis.Conn, e correlation.Event) (id string, err error) {
 }
 
 func deleteEvent(conn redis.Conn, id string) error {
+	o, err := eventByID(conn, id, true)
+	if err != nil {
+		if err == redis.ErrNil {
+			return db.ErrNotFound
+		}
+		return err
+	}
+
 	_ = conn.Send("MULTI")
 	_ = conn.Send("UNLINK", id)
 	_ = conn.Send("UNLINK", db.EventsCollection+":readings:"+id)
 	_ = conn.Send("ZREM", db.EventsCollection, id)
 	_ = conn.Send("ZREM", db.EventsCollection+":created", id)
+	_ = conn.Send("ZREM", db.EventsCollection+":pushed", id)
+	_ = conn.Send("ZREM", db.EventsCollection+":device:"+o.Device, id)
 	res, err := redis.Values(conn.Do("EXEC"))
 	if err != nil {
 		return err
@@ -1108,7 +1118,7 @@ func deleteEvent(conn redis.Conn, id string) error {
 	return nil
 }
 
-func eventByID(conn redis.Conn, id string) (event contract.Event, err error) {
+func eventByID(conn redis.Conn, id string, skipReadings bool) (event contract.Event, err error) {
 	obj, err := redis.Bytes(conn.Do("GET", id))
 	if err == redis.ErrNil {
 		return event, db.ErrNotFound
@@ -1117,7 +1127,7 @@ func eventByID(conn redis.Conn, id string) (event contract.Event, err error) {
 		return event, err
 	}
 
-	event, err = unmarshalEvent(obj)
+	event, err = unmarshalEvent(obj, skipReadings)
 	if err != nil {
 		return event, err
 	}
--- a/internal/pkg/db/redis/data.go
+++ b/internal/pkg/db/redis/data.go
@@ -109,7 +109,7 @@ func (c *Client) UpdateEvent(e correlation.Event) (err error) {
 
 	id := event.ID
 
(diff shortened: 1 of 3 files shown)


internal/pkg/db/redis/event.go
```diff
@@ -60,31 +60,34 @@ func unmarshalEvents(objects [][]byte, events []contract.Event) (err error) {
...
+		Modified: s.Modified,
+		Origin:   s.Origin,
+	}
 
 	conn, err := getConnection()
 	if err != nil {
-		return event, err
+		return contract.Event{}, err
 	}
+
```
jdharms: nitpicky, but remove blank line.  Discourages separation of the connection creation and deferred close in the future.

jdharms:
Thank you for providing test instructions 🎉

internal/pkg/db/redis/queries.go
```diff
@@ -137,6 +139,10 @@ func getObjectsBySomeRange(conn redis.Conn, command string, key string, start in
 		}
 	}
 
+	if len(objects) == 1 && objects[0] == nil {
```
brandonforster: This check was added because without it, the single nil element trips up https://github.com/edgexfoundry/edgex-go/pull/2094/files#diff-e0713026144dfb12cdabe915f0403939R96 and makes it error out.

AnthonyMBonafide: Does this affect the other `getObjectsBy...Range` functions? If so, can we fix them here or create issues so we do not forget to fix it?

brandonforster: It would have to, they're just syntactic sugar for this function.

Truly, this feels like treating a side effect and not the underlying problem, but it seems like an underlying problem with the Redis driver code not returning an empty array when it doesn't find anything.

brandonforster:
> I have statically reviewed this but don't have a runtime environment where I can confirm the test at the moment. This looks correct to me.
> 
> Do we need to define an EdgeX convention for `func foo() (event contract.Event, err error) {` vs `func foo() (contract.Event, error) {`?

I think we ought to, and I firmly believe we should have unnamed return variables unless there's a compelling engineering reason to have them. We have a bug that was just fixed that came about soley because of named return variables. (#2034)

michaelestrin:
> I think we ought to, and I firmly believe we should have unnamed return variables unless there's a compelling engineering reason to have them. We have a bug that was just fixed that came about soley because of named return variables. (#2034)

I believe this mischaracterizes #2034.  The issue there was an incorrect implementation that frankly could have been solved without removing the named return variables (and the resulting correction would have actually been smaller/more concise).

@brandonforster

michaelestrin:
Builds appear stuck.  You may want to close this PR and open an new one (referencing this one to ensure continuity of conversation).  @brandonforster

internal/pkg/db/redis/data.go
```diff
@@ -1089,11 +1089,21 @@ func addEvent(conn redis.Conn, e correlation.Event) (id string, err error) {
...
+		return err
+	}
+
 	_ = conn.Send("MULTI")
 	_ = conn.Send("UNLINK", id)
 	_ = conn.Send("UNLINK", db.EventsCollection+":readings:"+id)
 	_ = conn.Send("ZREM", db.EventsCollection, id)
 	_ = conn.Send("ZREM", db.EventsCollection+":created", id)
+	_ = conn.Send("ZREM", db.EventsCollection+":pushed", id)
+	_ = conn.Send("ZREM", db.EventsCollection+":device:"+o.Device, id)
```
cloudxxx8: please also delete checksum if o.Checksum != ""

brandonforster:
Closing to attempt to fix erroneous snap failure.

internal/pkg/db/redis/data.go
brandonforster: I'm not seeing any `checksum` fields added anywhere, can you provide steps to duplicate so I can verify I'll be removing the correct thing?

brandonforster: I found what you were talking about, addressing.

brandonforster:
rebased

internal/pkg/db/redis/queries.go
andresrinivasan: The redigo driver is a very thin wrapper on top of raw Redis commands. Though according to the [RESP](https://redis.io/topics/protocol#array-reply)spec

> A client library API should return a null object and not an empty Array when Redis replies with a Null Array. This is necessary to distinguish between an empty list and a different condition (for instance the timeout condition of the BLPOP command).

This could be a bug in redigo. I've added a task to my backlog to dig into this.

cloudxxx8: @andresrinivasan we have modified this part in another PR https://github.com/edgexfoundry/edgex-go/pull/2192/files
please help take a look.  thanks.

andresrinivasan: Dug into this a bit more. TL;DR: The driver is right.

From the [MGET documentation](https://redis.io/commands/mget)

> Returns the values of all specified keys. For every key that does not hold a string value or does not exist, the special value nil is returned. Because of this, the operation never fails.

In action this looks like

```
127.0.0.1:6379> set foo bar
OK
127.0.0.1:6379> set baz mumble
OK
127.0.0.1:6379> mget x y z
1) (nil)
2) (nil)
3) (nil)
127.0.0.1:6379> mget foo baz zoo
1) "bar"
2) "mumble"
3) (nil)
```

If we're getting _nil_ results from `MGET`, we're seeing a referential integrity challenge. In other words, the list of ids retrieved from the sorted set don't all exist as keys. I don't think we'll be able to fix the referential integrity challenge soon so I recommend checking all the values returned by `MGET` before trying to use them. 

The correct example is from PR #2192 

```
	var result [][]byte
	if len(ids) > 0 {
		result, err = redis.ByteSlices(conn.Do("MGET", ids...))
		if err != nil {
			return nil, err
		}
	}

	for _, obj := range result {
		if obj != nil {
			objects = append(objects, obj)
		}
	}
```

Vs this incomplete solution as it doesn't check all the results from `MGET`

```
	if len(ids) > 0 {
		objects, err = redis.ByteSlices(conn.Do("MGET", ids...))
		if err != nil {
			return nil, err
		}
	}

	if len(objects) == 1 && objects[0] == nil {
		objects = [][]byte{}
	}
```

PR #2095, opened 2019-11-07
(edinburgh) Delete event func for Redis misses the deletion of pushed and device collection

brandonforster:
#2076 in Edinburgh.

Please see #2094 for verification steps and discussion on design.


brandonforster:
recheck

tsconn23:
@brandonforster Thanks for the diligence in creating this PR. However we won't be releasing any more Edinburgh images. Fix only needed in `fuji` and `master`.

PR #2096, opened 2019-11-07
(master) Delete event func for Redis misses the deletion of pushed and device collection

brandonforster:
#2076 in master.

Please see #2094 for verification steps and discussion on design.

Signed-off-by: Brandon Forster brandonforster@gmail.com


brandonforster:
rebased

PR #2116, opened 2019-11-12
Add app-service-config with rules-engine profile to the snap

### LINKED ISSUE #2115: Add app-service-configurable with rules-engine profile to the snap
For Fuji, we need to add the app-service-configurable with the rules-engine profile to use with support-rulesengine in place of the export services.

This was communicated to me after the F2F by @tonyespy, whom apparently learned of this from @lenny-intel.

anonymouse64:
Add the app-service-configurable service to the snap so that it can be used with the rules-engine profile in place of the export services for Fuji when a user wants to use support-rulesengine.

To test this, build the snap and then install it. Then:

- [x] Check that app-service-configurable is disabled by default
```
$ snap services edgexfoundry.app-service-configurable
Service                                Startup   Current   Notes
edgexfoundry.app-service-configurable  disabled  inactive  -
```

- [x] Turning on support-rulesengine will also turn on the app-service-configurable service:

```
$ snap set edgexfoundry support-rulesengine=on
$ snap services edgexfoundry.app-service-configurable
Service                                Startup  Current  Notes
```

- [x] After app-service-configurable is started it is running with the rules-engine profile (indicated by `app=AppService-rules-engine` in the log output here):
```
edgexfoundry.app-service-configurable  enabled  active   -
$ snap logs edgexfoundry.app-service-configurable 
2019-11-12T16:52:52Z edgexfoundry.app-service-configurable[14568]: level=INFO ts=2019-11-12T16:52:52.087281208Z app=AppService-rules-engine source=sdk.go:662 msg="Listening for changes from registry"
2019-11-12T16:52:52Z edgexfoundry.app-service-configurable[14568]: level=INFO ts=2019-11-12T16:52:52.087296057Z app=AppService-rules-engine source=telemetry.go:79 msg="Starting CPU Usage Average loop"
2019-11-12T16:52:52Z edgexfoundry.app-service-configurable[14568]: level=INFO ts=2019-11-12T16:52:52.091976547Z app=AppService-rules-engine source=main.go:43 msg="Loading Configurable Pipeline..."
2019-11-12T16:52:52Z edgexfoundry.app-service-configurable[14568]: level=INFO ts=2019-11-12T16:52:52.098097641Z app=AppService-rules-engine source=sdk.go:690 msg="Writable configuration has been updated from Registry"
2019-11-12T16:52:52Z edgexfoundry.app-service-configurable[14568]: level=INFO ts=2019-11-12T16:52:52.106220898Z app=AppService-rules-engine source=sdk.go:508 msg="MessageBus trigger selected"
2019-11-12T16:52:52Z edgexfoundry.app-service-configurable[14568]: level=INFO ts=2019-11-12T16:52:52.109691598Z app=AppService-rules-engine source=sdk.go:759 msg="Reloaded Configurable Pipeline from Registry"
2019-11-12T16:52:52Z edgexfoundry.app-service-configurable[14568]: level=INFO ts=2019-11-12T16:52:52.114009216Z app=AppService-rules-engine source=messaging.go:46 msg="Initializing Message Bus Trigger. Subscribing to topic: events on port 5563 , Publish Topic:  on port 5566"
2019-11-12T16:52:52Z edgexfoundry.app-service-configurable[14568]: level=INFO ts=2019-11-12T16:52:52.118777304Z app=AppService-rules-engine source=sdk.go:154 msg="StoreAndForward disabled. Not running retry loop."
2019-11-12T16:52:52Z edgexfoundry.app-service-configurable[14568]: level=INFO ts=2019-11-12T16:52:52.127130702Z app=AppService-rules-engine source=sdk.go:157 msg="AppServiceConfigurable-rules-engine has Started"
2019-11-12T16:52:52Z edgexfoundry.app-service-configurable[14568]: level=INFO ts=2019-11-12T16:52:52.131084595Z app=AppService-rules-engine source=server.go:232 msg="Starting HTTP Server on port :50001"
```

Fixes #2115

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -22,6 +22,9 @@ ALL_SERVICES="$ALL_SERVICES support-logging"
 ALL_SERVICES="$ALL_SERVICES support-scheduler"
 ALL_SERVICES="$ALL_SERVICES support-rulesengine"
 
+# app-services
+ALL_SERVICES="$ALL_SERVICES app-service-configurable"
+
 # export services
 ALL_SERVICES="$ALL_SERVICES export-distro"
 ALL_SERVICES="$ALL_SERVICES export-client"
@@ -110,6 +113,13 @@ for key in $ALL_SERVICES; do
             fi
             handle_svc "vault" "$status"
             handle_svc "security-secretstore-setup" "$status"
+            ;;
+        support-rulesengine)
+            # if we are turning rulesengine on, make sure 
+            # app-service-configurable is on too
+            if [ "$status" = "on" ]; then
+                handle_svc "app-service-configurable" "on"
+            fi
             ;;
         *)
             # default case for all other services just enable/disable the service using
--- a/snap/hooks/install
+++ b/snap/hooks/install
@@ -8,6 +8,7 @@ SNAP_CURRENT=${SNAP/%$SNAP_REVISION/current}
 
 # install all the config files from $SNAP/config/SERVICE/res/configuration.toml 
 # into $SNAP_DATA/config
+# note that app-service-configurable is handled separately
 mkdir -p "$SNAP_DATA/config"
 for service in edgex-mongo security-proxy-setup security-secrets-setup security-secretstore-setup core-command config-seed core-data core-metadata export-client export-distro support-logging support-notifications support-scheduler sys-mgmt-agent device-random device-virtual; do
     if [ ! -f "$SNAP_DATA/config/$service/res/configuration.toml" ]; then
@@ -20,6 +21,26 @@ for service in edgex-mongo security-proxy-setup security-secrets-setup security-
             "$SNAP_DATA/config/$service/res/configuration.toml"
     fi
 done
+
+# handle app-service-configurable's various profiles:
+# 1. ensure all the directories from app-service-configurable exist
+# 2. copy the config files from $SNAP into $SNAP_DATA
+# 3. replace the various env vars that might be in that config file with their
+#    "current" symlink equivalent
+mkdir -p "$SNAP_DATA/config/app-service-configurable"
+cd "$SNAP/config/app-service-configurable"
+find . -maxdepth 2 -type d -exec mkdir -p "$SNAP_DATA/config/app-service-configurable/"{} \;
+find ./res -name "configuration.toml" | \
+    while read -r fname; do
+	if [ ! -f "$SNAP_DATA/config/app-service-configurable/$fname" ]; then
+	    cp "./$fname" "$SNAP_DATA/config/app-service-configurable/$fname"
+
+            sed -i -e "s@\$SNAP_COMMON@$SNAP_COMMON@g" \
+                -e "s@\$SNAP_DATA@$SNAP_DATA_CURRENT@g" \
+                -e "s@\$SNAP@$SNAP_CURRENT@g" \
+                "$SNAP_DATA/config/app-service-configurable/$fname"
+        fi
+    done
 
 # handle device-random device profile
 cp "$SNAP/config/device-random/res/device.random.yaml" "$SNAP_DATA/config/device-random/res/device.random.yaml"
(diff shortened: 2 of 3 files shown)


lenny-intel:
@anonymouse64 , changes look appropriate to me. I don't have my Linux box to test this today. One concern that is see in the log above is the log messages have `app=AppService` when it should be `app=AppService-rules-engine`. It seem it is using the default profile rather than `rules-engine` profile.

 Can you confirm that Consul has the config under `AppService-rules-engine` and that `Writable.Pipeline.ExecutionOrder` is set to just `SetOutputData`?

anonymouse64:
> Can you confirm that Consul has the config under AppService-rules-engine and that Writable.Pipeline.ExecutionOrder is set to just SetOutputData?

Hmm it seems that Consul has the config under `AppService` and `Writable.Pipeline.ExecutionOrder` is set to`FilterByDeviceName, TransformToXML, SetOutputData`.

I see now that the service wrapper also has to be patched to read the `app-service-configurable-profile` config setting not the `profile` setting. Good catch!

anonymouse64:
@lenny-intel fixed that issue, and added another test to the PR description checking specifically for what you mentioned. Thanks for reviewing!

anonymouse64:
> What's the justification for including app-service-configurable-profile as a config item? When I first discussed this with @lenny-intel, the idea was to include app-service-configurable in the edgexfoundry snap solely for the purpose of supporting the rules-engine. If you want to use different profiles, install the standalone app-service-configurable-snap...

Well TBH I don't really understand the justification for including app-service-config in the edgexfoundry snap _at all_ when there's already another snap for it available. If we really didn't want folks to use the export services this badly then we should have removed them...

I guess another part of it is that I dislike having things hard-coded in a snap like this when we can make them configurable. We can make them configurable here, so why not do so? It adds complexity, but if complexity is your concern then I would say we shouldn't include app-service-config in the snap at all since it's already going to be maintained elsewhere.

lenny-intel:
@anonymouse64 , you can think of `app-service-configurable-rule-engine` as distinct application service for feeding the rules engine. Currently it doesn't filter, but someone could add filtering via Consul live updates. The other snap for app-service-configurable is for all other uses.

As for export services, that are not deprecated yet, but want them turned off so folks look at using App Service Configurable or App Functions SDK instead. Maybe they should be removed from the edgexfoundry snap and put into a separate export-services snap.

anonymouse64:
The PR is updated with what we discussed on slack to not expose the config for setting the profile for app-service-configurable in the edgexfoundry snap, but for posterity: 

> The other snap for app-service-configurable is for all other uses.

But it could also serve this use too? That's what I don't get is why we can't use that other snap and we need the user to use this version of app-service-configurable in edgexfoundry. 

> As for export services, that are not deprecated yet, but want them turned off so folks look at using App Service Configurable or App Functions SDK instead

Sure, nudging people towards using things are are going to be supported is fine. Export services have always been off by default in the snap (at least for as long as I can remember, perhaps in California they were on by default) and additionally turning on support-rulesengine never turned the export services on anyways, so it had to be a conscious decision for them to turn on the export services. That doesn't change with this PR. 

> Maybe they should be removed from the edgexfoundry snap and put into a separate export-services snap.

I don't think we should remove functionality that's still supported from the snap and put it into a separate snap that will just be used for 6 months (or however long) until Geneva is ready.

To me, this proposal sounds an awful lot like what I have been trying to argue for which is that if a user wanted to use support-rulesengine in Fuji, they should install the app-service-configurable snap and use that in conjunction with support-rulesengine instead of using the export services. This can be handled with docs and still enables folks already using the export services to continue doing so without having to manually disable the app-service-configurable, while folks wanting to use app-service-configurable with support-rulesengine can just install that other snap and get a well maintained version of the snap that they can use for other purposes as well.

Introducing a duplicated version of the app-service-configurable into the edgexfoundry when it's already maintained separately seems like it will just lead to maintenance issues in my experience. Something will change upstream in app-service-configurable and that snap will be fixed but we will forget to update the version in edgexfoundry until the last minute (or maybe not even at all) because the users of app-service-configurable in edgexfoundry seems like they will be far less numerous than the normal users of app-service-configurable as a standalone snap. 

:man_shrugging:

PR #2118, opened 2019-11-12
Command: refactor to remove configuration global variable

### LINKED ISSUE #2017: Command: refactor to remove Configuration global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2016

brandonforster:
Fixes #2017


internal/core/command/init.go
```diff
@@ -31,9 +31,6 @@ import (
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
 )
 
-// Global variables
-var Configuration = &ConfigurationStruct{}
-
 // Global ErrorConcept variables
 var httpErrorHandler errorconcept.ErrorHandler
```
michaelestrin: https://github.com/edgexfoundry/edgex-go/issues/2016#issuecomment-550421660

Agreed previously to defer this to #2112.  Reconsidered and want you to remove this global as part of this PR (since this is the last PR to remove service-specific globals).  #2112 won't "redo" this work as it'll have to implement a non-global httpErrorHandler.

Will have a minor knock on effect since we won't need a call to `errorconcept.NewErrorHandler` and that's the only usage of `loggingClient`.

internal/core/command/router.go
```diff
@@ -37,7 +38,7 @@ func LoadRestRoutes(dic *di.Container) *mux.Router {
 
 	// Configuration
 	r.HandleFunc(clients.ApiConfigRoute, func(w http.ResponseWriter, r *http.Request) {
-		configHandler(w, bootstrapContainer.LoggingClientFrom(dic.Get))
+		configHandler(w, bootstrapContainer.LoggingClientFrom(dic.Get), *container.ConfigurationFrom(dic.Get))
```
michaelestrin: Commented once; may apply elsewhere.

Don't copy the resulting configuration struct; pass a reference (and have configHandler continue to pass a reference down the call chain as needed).

brandonforster: Removing the deference syntax throws an error about the underlying struct not adhering to the interface.

PR #2122, opened 2019-11-12
Fix update device functionality to correct method

### LINKED ISSUE #2119: Incorrect event type used when updating a device
# 🐞 Bug Report

### Affected Services
Core-metadata

### Is this a regression?

No, this looks like it has been this way for awhile.  However, the operator which encompasses this bug was created in this [commit](https://github.com/edgexfoundry/edgex-go/commit/818be3dfd721e644cbac9b5acdcbef0f0d1fb2ce)

The event type should be `PUT` since the associated operation is an update, but instead it is a `POST`

https://github.com/edgexfoundry/edgex-go/blob/818be3dfd721e644cbac9b5acdcbef0f0d1fb2ce/internal/core/metadata/operators/device/update.go#L143

1. Start core-services
2. Create a device
3. Update the device

See the Slack conversation for more details: https://edgexfoundry.slack.com/archives/CE6902Q6S/p1573569721008100

## 🔥 Exception or Error
An error is logged when the callback is invoked:

https://github.com/edgexfoundry/edgex-go/blob/21750bd69222753f69a7a217e480b07ca6bb6dd6/internal/core/metadata/operators/device/notify.go#L113

## 🌍  Your Environment
**Deployment Environment:**
Linux Fedora

**EdgeX Version:**
Fuji, this is also present in master(Geneva) 

**Anything else relevant?**

AnthonyMBonafide:
Fix #2119

Update the HTTP method from POST to PUT when invoking a callback for
a device service when a device has been updated.

### DIFF
--- a/internal/core/metadata/operators/device/update.go
+++ b/internal/core/metadata/operators/device/update.go
@@ -140,7 +140,7 @@ func (op updateDevice) Execute() (err error) {
 
 	evt.DeviceId = op.device.Id
 	evt.DeviceName = op.device.Name
-	evt.HttpMethod = http.MethodPost
+	evt.HttpMethod = http.MethodPut
 	evt.ServiceId = op.device.Service.Id
 
 	op.events <- evt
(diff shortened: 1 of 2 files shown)


tsconn23:
@AnthonyMBonafide We need this same fix in the `Fuji` branch I believe

PR #2123, opened 2019-11-12
security-proxy-setup: refactor to remove LoggingClient global variable

michaelestrin:
Fixes: https://github.com/edgexfoundry/edgex-go/issues/2120

Dependency on (includes work from) https://github.com/edgexfoundry/edgex-go/pull/2117.  That PR must be reviewed and merged prior to this one.

PR #2124, opened 2019-11-12
security-proxy-setup: refactor to remove Configuration global variable

michaelestrin:
Fixes: security-proxy-setup: refactor to remove Configuration global variable

Dependency on (includes work from) #2123. That PR must be reviewed and merged prior to this one.


michaelestrin:
Rebased.

PR #2126, opened 2019-11-13
Convert security-secretstore-setup to use common bootstrap package

michaelestrin:
Fixes: https://github.com/edgexfoundry/edgex-go/issues/1806

Signed-off-by: Michael Estrin <m.estrin@dell.com>

PR #2129, opened 2019-11-13
security-secretstore-setup: refactor to remove LoggingClient global variable

michaelestrin:
Fixes: https://github.com/edgexfoundry/edgex-go/issues/2127

Dependency on (includes work from) https://github.com/edgexfoundry/edgex-go/pull/2126. That PR must be reviewed and merged prior to this one.


michaelestrin:
Rebased

PR #2133, opened 2019-11-13
design(Refactor Registry): Create a design for refactoring the Registry component

lenny-intel:
Design for refactoring the Registry component to separate out the Service Configuration and Service Registration.

Also create initial `adr` folder and README for location to hold Architecture Decision Records.

Note that this design doesn't address the configuration hierarchy that @jpwhitemn brought up in the PHX F2F. I feel that is out of scope for this effort and should be addressed separately.


adr/Registry Refactoring Design.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Command line option changes
+
+The new `-cp/-configProvider` command line option will be added to each service which will have a value specified using the format `{type}.{protocol}://{host}:{port}` e.g `consul.http://localhost:8500`.  This new command line option will be overridden by the `edgex_configuration_provider` environment variable when it is set. This environment variable's value has the same format as the command line option value. 
+
+If **no value** is provided to the `-cp/-configProvider` option, i.e. just `-cp`, and no environment variable override is specified, the default value of `consul.http://localhost:8500` will be used. 
+
+if `-cp/-configProvider` **not used** and no environment variable override is specified the local configuration file is used, as is it now.
+
```
mkbhanda: what happens if the local configuration file has no cp entry? use the `consul.http://localhost:8500`?

A little concerned if services start-up out of sync, that is one referring to a registry foo and another to bar.
Guess an error message may work  that attempts a connection check to registry and if it fails, emit an misconfiguration error message and promptly exit. 

A la dependency injection, would it make sense to

adr/Registry Refactoring Design.md
```diff
@@ -0,0 +1,144 @@
...
+```go
+type Client interface {
+	HasConfiguration() (bool, error)
+	PutConfigurationToml(configuration *toml.Tree, overwrite bool) error
+	PutConfiguration(configStruct interface{}, overwrite bool) error
+	GetConfiguration(configStruct interface{}) (interface{}, error)
+	WatchForChanges(updateChannel chan<- interface{}, errorChannel chan<- error,
+                    configuration interface{}, waitKey string)
+	IsAlive() bool
```
mkbhanda: IsAlive feels more like registry related?

adr/Registry Refactoring Design.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+##### New Configuration Client
+
+This following is the new `Configuration Client` Interface which contains the  **Service Configuration** specific portion from the above current `Registry Client`.
+
+```go
+type Client interface {
+	HasConfiguration() (bool, error)
+	PutConfigurationToml(configuration *toml.Tree, overwrite bool) error
```
iain-anderson: Do we need this? If we have a toml-backed implementation of the Configuration Client, the configuration in toml.Tree form won't need to be seen outside of that

adr/Registry Refactoring Design.md
```diff
@@ -0,0 +1,144 @@
...
+### Command line option changes
+
+The new `-cp/-configProvider` command line option will be added to each service which will have a value specified using the format `{type}.{protocol}://{host}:{port}` e.g `consul.http://localhost:8500`.  This new command line option will be overridden by the `edgex_configuration_provider` environment variable when it is set. This environment variable's value has the same format as the command line option value. 
+
+If **no value** is provided to the `-cp/-configProvider` option, i.e. just `-cp`, and no environment variable override is specified, the default value of `consul.http://localhost:8500` will be used. 
+
+if `-cp/-configProvider` **not used** and no environment variable override is specified the local configuration file is used, as is it now.
+
+The existing `-r/-registry` command line option will be removed from all services. 
+
```
iain-anderson: Is the -p/-profile option being removed also?

adr/Registry Refactoring Design.md
lenny-intel: IsAlive() just determines if the Provider is alive, i.e. available. Applicable to Config and Registry

adr/Registry Refactoring Design.md
lenny-intel: Nope. Still need it.

adr/Registry Refactoring Design.md
lenny-intel: Yes, it is used by Config Seed since it only deals with the TOML

cloudxxx8: Does this mean update the configurations from TOML file?  If it does, we had better rename it to:
`PutConfigurationByToml` or `PutConfigurationFromToml`

Also, if it is only for Config Seed, it might not be put in this interface.  Config Seed can convert to configStruct on its own and call `PutConfiguration`.

lenny-intel:
@cloudxxx8 , PutConfigurationToml is currently only used by Config-seed, but from a module perspective it should stay in the interface. I do like the name change for clarity. Thx!

PR #2134, opened 2019-11-14
Add app-service-config with rules-engine profile to the snap (master)

anonymouse64:
Duplicate of #2116 for Geneva/master.

Unclear what we want to do long term with this, but since we don't have a new replacement rules engine yet and the export services are supposed to be dropped sometime this cycle, let's just make sure that master is in the same state as Fuji w/respect to using rules-engine.

Also note that this builds from master app-service-configurable rather than the tagged 1.0 release of that repo since we want master to be rolling.

### DIFF
--- a/snap/hooks/install
+++ b/snap/hooks/install
@@ -8,6 +8,7 @@ SNAP_CURRENT=${SNAP/%$SNAP_REVISION/current}
 
 # install all the config files from $SNAP/config/SERVICE/res/configuration.toml 
 # into $SNAP_DATA/config
+# note that app-service-configurable is handled separately
 mkdir -p "$SNAP_DATA/config"
 for service in edgex-mongo security-proxy-setup security-secrets-setup security-secretstore-setup core-command config-seed core-data core-metadata export-client export-distro support-logging support-notifications support-scheduler sys-mgmt-agent device-random device-virtual; do
     if [ ! -f "$SNAP_DATA/config/$service/res/configuration.toml" ]; then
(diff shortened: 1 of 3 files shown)


snap/hooks/install
```diff
@@ -21,6 +22,26 @@ for service in edgex-mongo security-proxy-setup security-secrets-setup security-
...
 
+# handle app-service-configurable's various profiles:
+# 1. ensure all the directories from app-service-configurable exist
+# 2. copy the config files from $SNAP into $SNAP_DATA
+# 3. replace the various env vars that might be in that config file with their
+#    "current" symlink equivalent
+mkdir -p "$SNAP_DATA/config/app-service-configurable"
+cd "$SNAP/config/app-service-configurable"
+find . -maxdepth 2 -type d -exec mkdir -p "$SNAP_DATA/config/app-service-configurable/"{} \;
+find ./res -name "configuration.toml" | \
```
tonyespy: I let this slide in the first iteration due to pressure schedule, however I think for this PR, we really only should be copying the rules-engine profile.

snap/snapcraft.yaml
```diff
@@ -1241,3 +1246,34 @@ parts:
...
+      # install the service binary
+      install -DT "./app-service-configurable" \
+         "$SNAPCRAFT_PART_INSTALL/bin/app-service-configurable"
+
+      # create config dirs
+      find ./res -maxdepth 1 -type d -exec install -d "$SNAPCRAFT_PART_INSTALL/config/app-service-configurable/"{} \;
+
+      # replace relative log paths in configuration.toml files with fully qualified paths
+      # prefixed with $SNAP_COMMON
+      find ./res -name "configuration.toml" | \
```
tonyespy: Same as the comment for install hook...

snap/hooks/install
anonymouse64: ok

snap/snapcraft.yaml
anonymouse64: ok

snap/hooks/install
anonymouse64: Done

snap/snapcraft.yaml
anonymouse64: Done

PR #2140, opened 2019-11-15
Refactor metadata to remove the httpErrorHandler global variable.

### LINKED ISSUE #2020: Metadata: refactor to remove httpErrorHandler global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2019

akramtexas:
Fix #2020

### DIFF
--- a/internal/core/metadata/init.go
+++ b/internal/core/metadata/init.go
@@ -20,10 +20,10 @@ import (
 
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/container"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/startup"
+	errorContainer "github.com/edgexfoundry/edgex-go/internal/pkg/container"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/endpoint"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/errorconcept"
-
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/coredata"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/notifications"
@@ -34,12 +34,9 @@ import (
 var Configuration = &ConfigurationStruct{}
 var nc notifications.NotificationsClient
 var vdc coredata.ValueDescriptorClient
-var httpErrorHandler errorconcept.ErrorHandler
 
 // BootstrapHandler fulfills the BootstrapHandler contract and performs initialization needed by the metadata service.
 func BootstrapHandler(wg *sync.WaitGroup, ctx context.Context, startupTimer startup.Timer, dic *di.Container) bool {
-
-	httpErrorHandler = errorconcept.NewErrorHandler(container.LoggingClientFrom(dic.Get))
 
 	// initialize clients required by service.
 	registryClient := container.RegistryFrom(dic.Get)
@@ -61,6 +58,13 @@ func BootstrapHandler(wg *sync.WaitGroup, ctx context.Context, startupTimer star
 			Interval:    Configuration.Service.ClientMonitor,
 		},
 		endpoint.Endpoint{RegistryClient: &registryClient})
+
+	// add dependencies to container
+	dic.Update(di.ServiceConstructorMap{
+		errorContainer.ErrorHandlerName: func(get di.Get) interface{} {
+			return errorconcept.NewErrorHandler(container.LoggingClientFrom(get))
+		},
+	})
 
 	return true
 }
--- a/internal/core/metadata/rest_addressable_test.go
+++ b/internal/core/metadata/rest_addressable_test.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	"net/http"
 	"net/http/httptest"
 	"strconv"
@@ -29,15 +30,14 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/errorconcept"
 
-	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 
 	"github.com/gorilla/mux"
 	"github.com/stretchr/testify/mock"
 )
 
-// AddressableTestURI this is not really used since we are using the HTTP testing framework and not creating routes, but rather
-// creating a specific handler which will accept all requests. Therefore, the URI is not important.
+// AddressableTestURI this is not really used since we are using the HTTP testing framework and not creating routes,
+// but rather creating a specific handler which will accept all requests. Therefore, the URI is not important.
 var AddressableTestURI = "/addressable"
 var TestAddress = "TestAddress"
 var TestPort = 8080
@@ -46,15 +46,15 @@ var TestTopic = "TestTopic"
 var TestName = "AddressableName"
 var TestId = "123e4567-e89b-12d3-a456-426655440000"
 
-// ErrorPathParam path parameter value which will trigger the 'mux.Vars' function to throw an error due to the '%' not being followed by a valid hexadecimal number.
+// ErrorPathParam path parameter value which will trigger the 'mux.Vars' function to throw an error due to the '%' not
+// being followed by a valid hexadecimal number.
 var ErrorPathParam = "%zz"
 
-// ErrorPortPathParam path parameter used to trigger an error in the `restGetAddressableByPort` function where the port variable is expected to be a number.
+// ErrorPortPathParam path parameter used to trigger an error in the `restGetAddressableByPort` function where the
+// port variable is expected to be a number.
 var ErrorPortPathParam = "abc"
 
 func TestGetAllAddressables(t *testing.T) {
-
-	httpErrorHandler = errorconcept.NewErrorHandler(logger.NewMockClient())
 	Configuration = &ConfigurationStruct{Service: config.ServiceInfo{MaxResultCount: 10}}
 	defer func() { Configuration = &ConfigurationStruct{} }()
 
@@ -93,7 +93,12 @@ func TestGetAllAddressables(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 
 			rr := httptest.NewRecorder()
-			restGetAllAddressables(rr, logger.NewMockClient(), tt.dbMock)
+			var loggerMock = logger.NewMockClient()
+			restGetAllAddressables(
+				rr,
+				loggerMock,
+				tt.dbMock,
+				errorconcept.NewErrorHandler(loggerMock))
 			response := rr.Result()
 			if response.StatusCode != tt.expectedStatus {
 				t.Errorf("status code mismatch -- expected %v got %v", tt.expectedStatus, response.StatusCode)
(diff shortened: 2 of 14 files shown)


internal/core/metadata/init.go
```diff
@@ -62,5 +58,12 @@ func BootstrapHandler(wg *sync.WaitGroup, ctx context.Context, startupTimer star
 		},
 		endpoint.Endpoint{RegistryClient: &registryClient})
 
+	// add dependencies to container
+	dic.Update(di.ServiceConstructorMap{
+		metadataContainer.ErrorHandlerName: func(get di.Get) interface{} {
+			return metadataContainer.ErrorHandlerName
+		},
```
michaelestrin: This map has a key with the service name and a ServiceConstructor-typed value that you're satisfying with a closure.  The name is correct.  The value implementation is not.  Your closure implementation is where the dependency is instantiated; it should provide an instance of errorconcept.Handler (it's currently providing a string).

While the code compiles and runs, it panics when the blackbox testing metadata importer is run:

```
2019/11/15 07:09:39 http: panic serving 127.0.0.1:57743: interface conversion: interface {} is string, not *errorconcept.Handler
```

Look at [this](https://github.com/edgexfoundry/edgex-go/blob/3fb26d3d713f167aaf69d4122db684b602a4c852/internal/system/agent/init.go#L53) for an example implementation.

akramtexas:
Recheck

internal/core/metadata/rest_addressable_test.go
```diff
@@ -93,7 +94,11 @@ func TestGetAllAddressables(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 
 			rr := httptest.NewRecorder()
-			restGetAllAddressables(rr, logger.NewMockClient(), tt.dbMock)
+			restGetAllAddressables(
+				rr,
+				logger.NewMockClient(),
+				tt.dbMock,
+				errorconcept.NewErrorHandler(logger.NewMockClient()))
```
michaelestrin: Commented once; applies elsewhere.

Should `errorconcept.ErrorHandler` be an actual instance or a mock?

internal/core/metadata/rest_addressable_test.go
```diff
@@ -152,7 +157,11 @@ func TestAddAddressable(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			rr := httptest.NewRecorder()
-			restAddAddressable(rr, tt.request, logger.NewMockClient(), tt.dbMock)
+			restAddAddressable(
+				rr,
+				tt.request,
+				logger.NewMockClient(),
+				tt.dbMock, errorconcept.NewErrorHandler(logger.NewMockClient()))
```
michaelestrin: These two parameters should be on separate lines to adhere to the folding pattern.

internal/core/metadata/rest_addressable_test.go
```diff
@@ -674,6 +698,7 @@ func createAddressables(howMany int) []contract.Addressable {
 }
 
 func createMockAddressLoaderStringArg(howMany int, methodName string, arg string) interfaces.DBClient {
+
```
michaelestrin: Commented here; applies elsewhere.

Empty line after a function/method declaration should only be inserted after a multi-line (i.e. folded) function/method declaration.  This is a one line declaration.  Remove the empty line.

internal/core/metadata/rest_deviceprofile.go
```diff
@@ -68,40 +73,40 @@ func restAddDeviceProfile(
...
-	addDeviceProfile(dp, dbClient, w)
+	addDeviceProfile(dp, dbClient, w, errorHandler)
 }
 
 func restUpdateDeviceProfile(
 	w http.ResponseWriter,
 	r *http.Request,
 	loggingClient logger.LoggingClient,
-	dbClient interfaces.DBClient) {
+	dbClient interfaces.DBClient, errorHandler errorconcept.ErrorHandler) {
```
michaelestrin: Commented elsewhere in a separate context.

To follow folding pattern, this line with two parameters should be two lines with one parameter each.

internal/core/metadata/rest_deviceprofile.go
```diff
@@ -456,7 +509,7 @@ func restGetYamlProfileById(
...
```
michaelestrin: Two parameters on one line needs to be two lines with one parameter each.

Applies elsewhere; not commenting additional instances of this.

internal/core/metadata/rest_deviceprofile_test.go
```diff
@@ -692,10 +742,17 @@ func TestUpdateDeviceProfile(t *testing.T) {
...
+				Service:  config.ServiceInfo{MaxResultCount: 1}}
 			vdc = MockValueDescriptorClient{}
 			rr := httptest.NewRecorder()
-			restUpdateDeviceProfile(rr, tt.request, logger.NewMockClient(), tt.dbMock)
+			restUpdateDeviceProfile(
+				rr,
+				tt.request,
+				logger.NewMockClient(),
+				tt.dbMock,
+				errorconcept.NewErrorHandler(logger.NewMockClient()))
```
michaelestrin: Called out earlier if `errorconcept.Handler` should be a mock.

If you didn't change the existing code to replace the `NewErrorHandler` call with a mock, I'd comment that you're using two different logger instances in the same test where you should only be using one.

internal/pkg/errorconcept/handler.go
```diff
@@ -26,7 +26,7 @@ type ErrorConceptType interface {
 	message(err error) string
 }
 
-type handler struct {
+type Handler struct {
```
michaelestrin: Why was this made public?

akramtexas: - That was done to export it. 
- Otherwise, the following code (in `/internal/core/metadata/container/error_handler.go`), which is under `metadata`, won't be able to access it, resulting in these compile-time errors:
```
internal/core/metadata/container/error_handler.go:23:46: cannot refer to unexported name errorconcept.handler
internal/core/metadata/container/error_handler.go:26:36: cannot refer to unexported name errorconcept.handler
internal/core/metadata/container/error_handler.go:27:33: cannot refer to unexported name errorconcept.handler
```

brandonforster: Also running into the issue; commenting to see what the resultant solution will be.

If we don't want to export the default implementation, I supposed we could duplicate it per service?

akramtexas: @brandonforster Yep, let's stay in touch regarding this so we can have an aligned, uniform solution  👍

michaelestrin: Change to make `Handler` public is appropriate.  Since both you and Brandon need what you've got in .../container/error_handler.go, consider a common implementation in internal/pkg/container that both of you use (first merge wins).  

@akramtexas @brandonforster

akramtexas: @michaelestrin Excellent, thanks for the feedback and confirmation 👍

internal/core/metadata/rest_addressable_test.go
AnthonyMBonafide: Good catch! @akramtexas  and I were looking at getting this changed to used a mock and we ran into a few hurdles. The issue is the expected side effect in the `ErrorHandler` where the HTTP status code is being determined and the written to the HTTP response:

https://github.com/edgexfoundry/edgex-go/blob/21750bd69222753f69a7a217e480b07ca6bb6dd6/internal/pkg/errorconcept/handler.go#L44-L49

The issue is the tests are validating the the response has been passed to the `http.ResponseWriter` and providing a simple mock will not perform that operation. This lead to us creating an anonymous function to be executed when the mock was invoked that would pass the response to the `http.ResponseWriter`. This approach is not really desirable as it invalidates the test since we are just writing the "hard-coded" status code value to the `http.ResponseWriter` rather than letting the REST handler process the request and populate the response naturally.

I think these issues are based on that the tests are not structured properly since the addition of the `ErrorHandler`. The tests now are testing two things:

1. The REST handler's functionality
2. The `ErrorHandler` functionality

I think these should be split up into two different test suites. The tests which validate the REST handlers can verify the functionality that is explicitly performed by the REST handler --like ensuring the proper `Content-Type` is written to the response-- and for tests with expected errors we can simply verify that the `ErrorHandler` mock was called with the proper parameters rather than verifying the status code. We could then test the propagation of the status code from the `ErrorHandler` to the `http.ResponseWriter` within the tests specifically for the `ErrorHandler`. 

I hope this makes sense. Any thoughts?

akramtexas: - Yep, I agree, @AnthonyMBonafide, that this is definitely a good catch by @michaelestrin in his observation (_"Should errorconcept.ErrorHandler be an actual instance or a mock?"_). 
- It led me, during implementing the approach of mocking `errorconcept.ErrorHandler`, to investigate the root cause of the following scenario wherein I would get test failures due to mismatches on the expected HTTP response code, for example:
```
--- FAIL: TestGetAllAddressables/Error_Limit_Exceeded (0.00s)
    rest_addressable_test.go:133: status code mismatch -- expected 413 got 200
--- FAIL: TestGetAllAddressables/Error_Unknown (0.00s)
    rest_addressable_test.go:133: status code mismatch -- expected 500 got 200
```
- And which is when I had reached out to bounce ideas off you, get some ideas. Thanks again for making the time.
- Exactly, tracing back to the fact that the response code is not explicitly written by `http.ResponseWriter` in the following three relevant functions for ErrorHandler:
```
- Handle(w http.ResponseWriter, err error, ec ErrorConceptType)
- HandleManyVariants(w http.ResponseWriter, err error, allowableErrors []ErrorConceptType, defaultError ErrorConceptType)
- HandleOneVariant(w http.ResponseWriter, err error, allowableError ErrorConceptType, defaultError ErrorConceptType)
```
- Basically, the HTTP status code is swallowed when the following is called: `rr := httptest.NewRecorder()`
- To add to your point, @AnthonyMBonafide, that had led to us creating an anonymous function to be executed when the mock was invoked that would pass the response to the `http.ResponseWriter`, an approach that may not be desirable, as you note, since we are writing the "hard-coded" status code value to the `http.ResponseWriter`. Basically, we achieve that by passing an anonymous function during set up of mocking as a side-effect, something like the following (partially-complete snippet), in outline:
```
func createMockWithVariations(
variations []mockVariations,
w http.ResponseWriter) errorconcept.ErrorHandler {
  ehMock := errorHandlerMock.ErrorHandler{}

  for _, o := range variations {
    ehMock.On(o.methodName, o.args...).Run(func(args mock.Arguments){
      writer := args[0].(http.ResponseWriter)
      writer.WriteHeader()})
  }

  return &ehMock
}
```

michaelestrin: I agree with @AnthonyMBonafide's suggested approach.

akramtexas:
Recheck

akramtexas:
Recheck

akramtexas:
Recheck

internal/core/metadata/rest_addressable_test.go
```diff
@@ -46,15 +46,18 @@ var TestTopic = "TestTopic"
...
 var ErrorPathParam = "%zz"
 
-// ErrorPortPathParam path parameter used to trigger an error in the `restGetAddressableByPort` function where the port variable is expected to be a number.
+// ErrorPortPathParam path parameter used to trigger an error in the `restGetAddressableByPort` function where the
+// port variable is expected to be a number.
 var ErrorPortPathParam = "abc"
 
-func TestGetAllAddressables(t *testing.T) {
+// Instantiate one logger mock that is used throughout
+var loggerMock = logger.NewMockClient()
```
michaelestrin: I'm sorry I didn't communicate my ask clearly.

I didn't intend for you to create a global variable.  I intended for you to create an instance of the logger.NewMockClient() once per test invocation and use that instance wherever a logger is required within that test.

For example, the code that I commented on in rest_addressable_test looked like this before:

``` go
	for _, tt := range tests {

			rr := httptest.NewRecorder()
			restGetAllAddressables(
				rr,
				logger.NewMockClient(),
				tt.dbMock,
				errorconcept.NewErrorHandler(logger.NewMockClient()))
			response := rr.Result()
			if response.StatusCode != tt.expectedStatus {
				t.Errorf("status code mismatch -- expected %v got %v", tt.expectedStatus, response.StatusCode)
				return
			}
		})
	}
```

I was looking for an implementation that looked like this:

``` go
	for _, tt := range tests {

			rr := httptest.NewRecorder()
			loggerMock := logger.NewMockClient()
			restGetAllAddressables(
				rr,
				loggerMock,
				tt.dbMock,
				errorconcept.NewErrorHandler(loggerMock))
			response := rr.Result()
			if response.StatusCode != tt.expectedStatus {
				t.Errorf("status code mismatch -- expected %v got %v", tt.expectedStatus, response.StatusCode)
				return
			}
		})
	}
```

Ask is to remove the global and refactor your tests to instantiate a single logger mock instance per test.

akramtexas: Gotcha, I like your approach better, and will make it so. All good, thanks 👍

akramtexas:
Rebased.

akramtexas:
Rebased.

akramtexas:
Rebased.

akramtexas:
Recheck

PR #2141, opened 2019-11-15
fix issue-2089 for master : Running security-proxy-setup multiple times now fails.

tingyuz:
Fix issue-2089 that has been merged into Fuji. 

Based on the previous discussion  this PR is slightly different from the one merged into Fuji. Instead of returning two variables (error and bool) , postCert() returns one custom error and Init() checks the error type for further action.

Signed-off-by: Tingyu Zeng <Tingyu.Zeng@rsa.com>


tingyuz:
> Don't see anything wrong with the code but the commit message is terrible.

updated the description section.

bnevis-i:
Specific suggestion is

```sh
git reset HEAD~1  # Undo the 2nd commit
git add (files)
git commit --amend -s -S  # Fix the commit message
git push --force   # Force push
```

PR #2145, opened 2019-11-18
Add metrics endpoint explanation in sys-mgmt service API raml

OdysLam:
Added some background information on the metrics returned from the sys-mgmt endpoint. 

t is based on the [official docs](https://golang.org/pkg/runtime/#MemStats) for the runtime pkg. It will help developers make sense of the API and integrate it into other monitoring solutions.

Signed-off-by: OdysLam <odyslam@gmail.com>


tsconn23:
Closing due to inactivity

PR #2149, opened 2019-11-19
Remove dbClient global variable from core-data

### LINKED ISSUE #2025: Data: refactor to remove dbClient global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2024

AnthonyMBonafide:
Fix #2025

Remove dbClient global variable from the core-data service/package(s)
and update the routers to get the dbClient from th DI container and pass
it to the necessary REST handlers. Also, update any functions/methods
which were referring to the dbClient global variable to rather accept an
instance of the DBClient interface.

Update tests to conform to the changes to the production code.

### DIFF
--- a/internal/core/data/event.go
+++ b/internal/core/data/event.go
@@ -25,6 +25,7 @@ import (
 	msgTypes "github.com/edgexfoundry/go-mod-messaging/pkg/types"
 
 	"github.com/edgexfoundry/edgex-go/internal/core/data/errors"
+	"github.com/edgexfoundry/edgex-go/internal/core/data/interfaces"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/correlation"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/correlation/models"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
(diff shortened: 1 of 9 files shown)


AnthonyMBonafide:
I tested these changes with the blackbox tests to ensure all functionality still works. All the `core-data` tests are passing.

internal/core/data/event.go
```diff
@@ -88,7 +89,7 @@ func getEvents(limit int) ([]contract.Event, error) {
 	return events, err
 }
 
-func addNewEvent(e models.Event, ctx context.Context, loggingClient logger.LoggingClient) (string, error) {
+func addNewEvent(e models.Event, ctx context.Context, loggingClient logger.LoggingClient, dbClient interfaces.DBClient) (string, error) {
```
michaelestrin: Nitpick - lines > ~120 characters should be folded; i.e. 

``` go
func addNewEvent(
	e models.Event, 
	ctx context.Context,
	dbClient interfaces.DBClient) (string, error) {

	err := checkDevice(e.Device, ctx)
	if err != nil {
		return "", err
	}
        ...
```

Comment once; occurs elsewhere.

AnthonyMBonafide: I am going to create an issue to add this to an `editorconfig` so that this is consistent rather than doing this by hand.

PR #2152, opened 2019-11-19
Scheduler: Remove LoggingClient global variable

### LINKED ISSUE #2035: Scheduler: refactor to remove LoggingClient global variable

brandonforster:
Fixes #2035

### DIFF
--- a/internal/support/scheduler/rest_interval_test.go
+++ b/internal/support/scheduler/rest_interval_test.go
@@ -324,8 +324,9 @@ func createMockIntervalLoaderAllErr() interfaces.DBClient {
 func createMockIntervalLoaderAddSuccess() interfaces.DBClient {
 	myMock := mocks.DBClient{}
 	interval := createIntervals(1)[0]
-	b, _ := json.Marshal(intervalForAdd)
-	_ = intervalForAdd.UnmarshalJSON(b)
+
+	validateInterval(&intervalForAdd)
+	validateInterval(&interval)
 
 	myMock.On("IntervalByName", intervalForAdd.Name).Return(interval, nil)
 	myMock.On("AddInterval", intervalForAdd).Return(intervalForAdd.ID, nil)
--- a/internal/support/scheduler/rest_interval_test.go
+++ b/internal/support/scheduler/rest_interval_test.go
@@ -17,20 +17,21 @@ package scheduler
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
+	goErrors "errors"
 	"net/http"
 	"net/http/httptest"
 	"os"
 	"testing"
 
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
+	errors "github.com/edgexfoundry/go-mod-core-contracts/clients/types"
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+	"github.com/gorilla/mux"
+
 	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
 	"github.com/edgexfoundry/edgex-go/internal/support/scheduler/interfaces"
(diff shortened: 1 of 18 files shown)


internal/support/scheduler/utils.go
```diff
@@ -104,11 +106,12 @@ func parseInt64(value string) int64 {
...
 }
 
 // NewClient
-func NewSchedulerQueueClient() *QueueClient {
-	queueClient := &QueueClient{}
+func NewSchedulerQueueClient(loggingClient logger.LoggingClient) *QueueClient {
+	queueClient := &QueueClient{loggingClient}
 	currentQueueClient = queueClient // Set the singleton
 	return queueClient
 }
```
michaelestrin: Acknowledge you didn't write this code.

It's a mess.  Can you improve it?  Create a separate GitHub issue.  See individual comments.

internal/support/scheduler/utils.go
```diff
@@ -104,11 +106,12 @@ func parseInt64(value string) int64 {
 // Scheduler Queue Client
 var currentQueueClient *QueueClient // Singleton used so that queueClient can use it to de-reference readings
```
michaelestrin: (for new GitHub issue) My IDE indicates this global isn't referenced.  Should probably be removed.

internal/support/scheduler/utils.go
```diff
@@ -104,11 +106,12 @@ func parseInt64(value string) int64 {
...
 type QueueClient struct {
+	loggingClient logger.LoggingClient
 }
```
michaelestrin: (for new GitHub issue) QueueClient receiver should be moved to schedule.go where it's methods reside.

internal/support/scheduler/utils.go
michaelestrin: (for new GitHub issue) QueueClient factory method should be moved to schedule.go where it's implementation lives.

internal/support/scheduler/rest_interval_test.go
```diff
@@ -403,24 +423,24 @@ func createMockIntervalLoaderAddNameInUse() interfaces.DBClient {
...
 	myMock.On("IntervalActionsByIntervalName", intervalForAdd.Name).Return([]contract.IntervalAction{}, nil)
 	myMock.On("UpdateInterval", intervalForAdd).Return(nil)
 	return &myMock
 }
 
 func validateInterval(interval *contract.Interval) {
 	b, _ := json.Marshal(interval)
-	interval.UnmarshalJSON(b)
+	_ = interval.UnmarshalJSON(b)
 }
```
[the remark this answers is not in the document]
brandonforster: The purpose of this is to set the unexported field `isValidated` which can only be done by marshalling and unmarshalling.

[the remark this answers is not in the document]
michaelestrin: Makes sense.  Wish it wasn't so.  Appreciate the explanation -- definitely isn't self-evident.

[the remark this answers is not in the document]
brandonforster: I went ahead and added a comment to note this for the next time it comes up 👍

internal/support/scheduler/schedule.go
```diff
@@ -90,33 +93,33 @@ func (qc *QueueClient) Connect() (string, error) {
...
 	mutex.Lock()
 	defer mutex.Unlock()
 
 	intervalContext, exists := intervalIdToContextMap[intervalId]
 	if !exists {
-		logMsg := fmt.Sprintf("scheduler could not find a interval context with interval id : %s", intervalId)
-		LoggingClient.Info(logMsg)
-		return contract.Interval{}, errors.New(logMsg)
+		return contract.Interval{},
+			errors.New(fmt.Sprintf("scheduler could not find a interval context with interval id : %s", intervalId))
```
michaelestrin: Nitpick: `errors.New(fmt.Sprintf(...))` == `fmt.Errorf`

Commented once; occurs elsewhere.

brandonforster:
recheck

PR #2146, opened 2019-11-18
Split customized vault into two docker images- part 1

jim-wang-intel:
This PR attempts to address issues #1716.  There are a couple steps to accomplish that:

1. Split the docker image file from inhering the Vault
2. Update the newly generated TLS paths (Related PRs: https://github.com/edgexfoundry/docker-edgex-mongo/pull/71, )
3. Update the consul checking scripts (Related PR: https://github.com/edgexfoundry/docker-edgex-consul/pull/10)

The focus of this PR is on the **step 1**.

  The original customized edgex-vault docker image is broken into two separate images:
    the official Vault docker image from HashiCorp and the `security-secrets-setup` docker image

 - add customized docker entrypoint script for starting vault
 - use sentinel files for readiness check of TLS cert of vault because vault image does not have curl command available
 - prepare cross-mount volume /vault/init for customized startup script
 - remove creation of /vault/config/assets folder from security-secrets-setup and
   create that folder from entrypoint of security-secretstore-setup (aka vault-worker)

## Testing steps
1. Build the new docker images for both `security-secrets-setup` and `security-secretstore-setup`:
```sh
$ make docker_security_secrets_setup
$ make docker_security_secretstore_setup
```
2. Run the secrets-setup docker container:
```sh
$ docker run --tmpfs /run  -v /run/edgex/secrets:/run/edgex/secrets -v vault-init:/vault/init edgexfoundry/docker-edgex-secrets-setup-go:1.1.1-dev
```
If running ok, one should expect to see console outputs like this:
```sh
XDG_RUNTIME_DIR /run/user/0
BASE_DIR /edgex/security-secrets-setup
Executing security-secrets-setup generate
level=ERROR ts=2019-11-18T23:40:17.197784157Z app=edgex-security-secrets-setup source=logger.go:73 msg="logTarget cannot be blank, using stdout only"
Creating directory: /edgex/security-secrets-setup/logs
{
    "CACertFile": "/run/user/0/edgex/security-secrets-setup/scratch/config/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem",
    "CACountry": "US",
    "CAKeyFile": "/run/user/0/edgex/security-secrets-setup/scratch/config/pki/EdgeXFoundryCA/EdgeXFoundryCA.priv.key",
    "CALocality": "San Francisco",
    "CAName": "EdgeXFoundryCA",
    "CAOrg": "EdgeXFoundry",
    "CAState": "CA",
    "DumpKeys": false,
[...]

Issue #1716, opened 2019-08-28
Split edgex-vault Docker container into edgex-vault and edgex-pki-init


bnevis-i:
Please add to https://github.com/edgexfoundry/edgex-go/projects/13

EdgeX is currently in the position that we are distributing a custom version of Vault.
This means that we are inheriting security vulnerabilities in Vault's container, and we have to do extra releases whenever the upstream Vault is released. Now that the PKI-init code is merged into security-secrets-setup we should go back to using a stack version of Vault, and move security-secrets-setup into its own container.

 * Modify docker-compose to modify edgex-vault microservice, deloyed using sha256 hash tag from dockerhub.
 * New vault to bind mount /run/edgex/secrets/vault volume
 * Remove /vault/config mount from the vault service (security-secrets-worker still needs to to create legacy tokens)
 * New vault uses same backend data store as old vault (see existing config file in notes)
 * Set VAULT_LOCAL_CONFIG environment variable on container to bind TLS assets and consul backend. Set the following keys
 * tls_cert_file = /run/edgex/secrets/edgex-vault/server.crt
 * tls_key_file = /run/edgex/secrets/edgex-vault/server.key
 * tls_perfer_server_cipher_suites = true
 * Block block container startup on availability of /run/edgex/secrets/vault/.pki-init.complete
 * Init-block example run command override (for vault)
 * /usr/bin/dumb-init -- /bin/sh -c "while test \! -f /run/edgex/secrets/edgex-vault/.pki-init.complete; do sleep 1; done; exec  /usr/local/bin/docker-entrypoint.sh"

##### https://github.com/edgexfoundry/developer-scripts/issues/150
Issue 150 places additional constraints on how this key material can be shared across containers in a MacOS environment.  Specifically:
1. MacOS doesn't have a /run
2. On MacOS, bind mounts are white-listed and /run is not in the white-list by default.

Proposed solution must accommodate differences between Linux and MacOS deployments.  One  possible solution is to share via /tmp by default, and recommend that Linux users use /run.

jim-wang-intel:
Will re-visit this after Fuji

PR #2164, opened 2019-11-21
Split customized vault into two docker images- part 2

jim-wang-intel:
This PR addresses the step 2 of PR # https://github.com/edgexfoundry/edgex-go/pull/2146 and is related to issue #1716.

Step 2 is **Update the docker configuration for TLS paths** 

Steps: 
1. Split the docker image file from inhering the Vault (#2146 merged already)
2. Update the newly generated TLS paths (Related PRs: https://github.com/edgexfoundry/docker-edgex-mongo/pull/71, and this one)
3. Update the consul checking scripts (Related PR: https://github.com/edgexfoundry/docker-edgex-consul/pull/12)

  - the TLS paths of ca, edgex-vault, and edgex-kong in docker containers are changed due to the changes from splitting edgex-vault images
- the old TLS path `/vault/config/pki` is removed and the new one is `/tmp/edgex/secrets/<service_name>`, where the `service_name` can be ca if it is for root CA TLS,  edgex-vault if it is for Vault's TLS, and edgex-kong if it is for Kong's TLS
- the default path of that is also changed from /vault/config/pki to **/tmp/edgex/secrets** for better compatibility (see related issue https://github.com/edgexfoundry/developer-scripts/issues/150)


jim-wang-intel:
> 
> 
> Should the reference to `./run` [here](https://github.com/edgexfoundry/edgex-go/blob/7b27dc6800fcd038bf6097b97a63cab550f6d76d/internal/security/secrets/helper/helper_test.go#L69) be updated to `/tmp` for consistency?
> 
> Otherwise LGTM. Not explicitly approving to allow security-related reviewers to review/comment/approve.

Good point, changed.

Issue #2163, opened 2019-11-21
Failing blackbox tests - core-data when using Redis


AnthonyMBonafide:
# 🐞 Bug Report

### Affected Services
core-data, specifically events

### Is this a regression?
Not sure when the problem was introduced.

### Description and Minimal Reproduction

1. Configure core-data to use Redis for persistence, either by TOML or Consul
2. Start core-data
3. Run core-data blackbox tests against the running instance of core-data using Redis

## 🔥 Exception or Error

Newman failures:

<pre><code>
 ↳ 30 http://localhost:48080/api/v1/event/removeold/age/:age
 DELETE http://localhost:48080/api/v1/event/removeold/age/0 [500 Internal Server Error, 198B, 18ms]
 2. Status code is 200
 ✓  Response time is less than 800ms

 ↳ 32 http://localhost:48080/api/v1/event/scrub
   DELETE http://localhost:48080/api/v1/event/scrub [500 Internal Server Error, 198B, 12ms]
   1. Status code is 200
   ✓  Response time is less than 800ms

</code></pre>

## 🌍  Your Environment
**Deployment Environment:**
OS: Linux Fedora 31
Persistence: Redis running locally and natively
Service: core-data with an updated TOML file to use the locally running instance of Redis 
Blackbox tests: Updated the blackbox tests to use the local postman environments rather than the docker versions.

**EdgeX Version:**
Geneva(tip of master branch)

**Anything else relevant?**
I have not determined the root cause of the error, so that will need to be determined. If I figure out a root cause I will provide a comment to this issue.

michaelestrin:
Blackbox tests currently execute using Mongo.  In the Geneva cycle, I anticipate they'll be updated to execute using Redis.  However, the project currently supports both databases and I do not expect this to change going forward.  

Because we're not running blackbox tests against all of the supported database variations, we're missing issues that would otherwise be caught.  [This particular GitHub issue](https://github.com/edgexfoundry/edgex-go/issues/2163) is a representative example.

I recognize the project intends to change the testing framework and test implementation away from the existing Newman/JSON blackbox tests to the TAF framework.

The ask here is to update the build environments to incorporate the two currently support databases as variations when we run the tests (whatever framework we're currently using and those that we intend to use in the future).  Expect this will result in a discussion topic in either the DevOps WG, the Q/A WG, or both.

@frq92290 @cloudxxx8 @jamesrgregg @lranjbar @tsconn23

[...]

PR #2165, opened 2019-11-21
Fix orphaned event checksum data for Redis

### LINKED ISSUE #2143: Delete event func for Redis misses the deletion of the checksum collection for binary data
# 🐞 Bug Report

### Affected Services

### Is this a regression?
No, it also happens in Edinburgh release

When creating the event, there is an additional collections added to reference to the event id for binary data
https://github.com/edgexfoundry/edgex-go/blob/d7fd3d05ee7535716585419652c1d825b0693da7/internal/pkg/db/redis/data.go#L1060-L1062

However, the deleteEvent func doesn't delete it
https://github.com/edgexfoundry/edgex-go/blob/d7fd3d05ee7535716585419652c1d825b0693da7/internal/pkg/db/redis/data.go#L1091-L1119

## 🔥 Exception or Error
The data will keep growing until out of memory

## 🌍  Your Environment
**Deployment Environment:**
Ubuntu

**EdgeX Version:**
Edinburgh and Fuji

**Anything else relevant?**
it is similar to https://github.com/edgexfoundry/edgex-go/issues/2076

AnthonyMBonafide:
Fix #2143

Update the deleteEvent function to perform an extra lookup for the
checksum value associated with the event to be deleted and remove the
lingering data. The additional lookup is necessary since there are three
different types of Events(correlation,contract,and redisEvent) being
used within the Redis implementation of the DBClient and the one being
used for common lookups(contract) does not have the checksum information
so it is lost during serialization. The approach is the most direct and
non-invasive solution. Other attempts to clean-up and reduce the
divergence of the types being used resulted in unrelated Blackbox test
failures.


AnthonyMBonafide:
This fix will also go into `fuji`, I will cherry-pick this commit once it is approved.

internal/pkg/db/redis/data.go
```diff
@@ -1097,13 +1097,25 @@ func deleteEvent(conn redis.Conn, id string) error {
 		return err
 	}
 
+	// Obtain the checksum associated with the event.
+	// This is necessary since the helper functions result in an Event type which does not contain the checksum
```
brandonforster: Would it be better to modify the helper functions to return a Redis model instead of a contract?

AnthonyMBonafide:
These changes were verified manually by creating, then deleting an event, and ensuring the data was no longer available in Redis. The blackbox tests were also used to confirm these changes. However, the tests are failing when run against `master` using Redis, so the tests are not completely reliable, but I verified these changes result in the same failures as witnessed when executing against master and all other tests pass.

An issue has been created to fix the failures for the blackbox tests: #2163

internal/pkg/db/redis/data.go
AnthonyMBonafide: That was my original intent, but that caused other things to break. Rather than digging for a root cause, I went with this direct approach to provide a fix for the bug rather than re-working the whole structure of the Redis DBClient. The problem is that the Redis DBClient is using 3 different representations of the Event type and its not clear which is being used  and we got into this odd situation. Another problem is that the Blackbox tests are not being run, and they seem to not be working so I am very hesitant to start restructuring everything. When I attempted to clean things up the blackbox tests showed even more failures than it is already reporting. I am going to create an issue to go back and re-work the Redis client.

AnthonyMBonafide: Created the issue to clean this up: #2166

michaelestrin:
> tests are failing when run against `master` using Redis, so the tests are not completely reliable

Are the tests "not completely reliable" or are the test failures valid?

PR #2167, opened 2019-11-21
Scheduler: Remove dbClient global variable

### LINKED ISSUE #2036: Scheduler: refactor to remove dbClient global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2035

brandonforster:
Fixes #2036 

- Removed several functions that were marked as orphaned by my IDE.

- Moved all interval action handlers from the router to `rest_intervalaction`

Requires #2152 to be merged first.


michaelestrin:
Please rebase/update so I can review.

brandonforster:
recheck

PR #2172, opened 2019-11-22
Implement cmd/security-file-token-provider scaffolding

bnevis-i:
This is the main program bootstrap skeleton for
cmd/security-file-token-provider as documented in
edgex-docs/security/token-file-provider.1.rst

### DIFF
--- a/cmd/security-file-token-provider/main.go
+++ b/cmd/security-file-token-provider/main.go
@@ -1,0 +1,71 @@
+/*******************************************************************************
+ * Copyright 2019 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ *******************************************************************************/
+
+package main
+
+import (
+	"flag"
+
+	"github.com/edgexfoundry/edgex-go/internal"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap"
+	bootstrapContainer "github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/container"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/interfaces"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/startup"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/usage"
+	"github.com/edgexfoundry/edgex-go/internal/security/fileprovider"
+	"github.com/edgexfoundry/edgex-go/internal/security/fileprovider/config"
+	"github.com/edgexfoundry/edgex-go/internal/security/fileprovider/container"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+)
+
+func main() {
+	startupTimer := startup.NewStartUpTimer(internal.BootRetrySecondsDefault, internal.BootTimeoutSecondsDefault)
+
+	var configDir, profileDir string
+	var useRegistry bool
+
+	flag.BoolVar(&useRegistry, "registry", false, "Indicates the service should use registry service.")
+	flag.BoolVar(&useRegistry, "r", false, "Indicates the service should use registry service.")
+	flag.StringVar(&profileDir, "profile", "", "Specify a profile other than default.")
+	flag.StringVar(&profileDir, "p", "", "Specify a profile other than default.")
+	flag.StringVar(&configDir, "confdir", "", "Specify local configuration directory")
+
+	flag.Usage = usage.HelpCallbackSecurityFileTokenProvider
+	flag.Parse()
+
+	configuration := &config.ConfigurationStruct{}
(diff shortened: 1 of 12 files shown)


bnevis-i:
@michaelestrin  This is my first time using the new DI-capable bootstrap.  Would you be able to take a close look at this and tell me how to integrate the additional dependencies in init.go into the new structure?

cmd/security-file-token-provider/main.go
```diff
@@ -0,0 +1,70 @@
...
+		bootstrapContainer.ConfigurationInterfaceName: func(get di.Get) interface{} {
+			return get(container.ConfigurationName)
+		},
+	})
+	bootstrap.Run(
+		configDir,
+		profileDir,
+		internal.ConfigFileName,
+		useRegistry,
+		internal.SecuritySecretStoreSetupServiceKey,
```
michaelestrin: I think you'll want to replace `internal.SecuritySecretStoreSetupServiceKey` with this service's unique key (which may need to be created).

bnevis-i: Good catch

internal/security/fileprovider/config/config.go
```diff
@@ -0,0 +1,100 @@
...
+	LogLevel string
+	Title    string
+}
+
+type TokenFileProviderInfo struct {
+	// Path to Vault authorization token to be used by the service (default: /run/edgex/secrets/token-file-provider/secrets-token.json)
+	PrivilegedTokenPath string
+	// Configuration file used to control token creation (default: res/token-config.json)
+	ConfigFile string
+	// Base directory for token file output (default: /run/edgex/secrets)
```
jim-wang-intel: Should the default be  `/tmp/edgex/secrets` instead of `/run/edgex/secrets`?

internal/security/fileprovider/provider.go
```diff
@@ -0,0 +1,60 @@
...
+// Set configuration
+func (p *fileTokenProvider) SetConfiguration(secretConfig secretstoreclient.SecretServiceInfo, tokenConfig config.TokenFileProviderInfo) {
+	p.secretConfig = secretConfig
+	p.tokenConfig = tokenConfig
+}
+
+// Do whatever is needed
+func (p *fileTokenProvider) Run() error {
+	p.logger.Info("Implementation to be provided later")
+	return nil
```
jim-wang-intel: Should this do fail-fast first to throw an error when it is not implemented yet?

internal/security/fileprovider/config/config.go
```diff
@@ -0,0 +1,100 @@
...
+	TokenFileProvider TokenFileProviderInfo
+}
+
+type WritableInfo struct {
+	LogLevel string
+	Title    string
+}
+
+type TokenFileProviderInfo struct {
+	// Path to Vault authorization token to be used by the service (default: /run/edgex/secrets/token-file-provider/secrets-token.json)
```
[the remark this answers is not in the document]
bnevis-i: Actually, I removed the defaults setting code entirely,  but I need to go back end update the comments.

internal/security/fileprovider/provider.go
bnevis-i: Sure.

internal/security/fileprovider/config/config.go
bnevis-i: Removed comment; code no longer has a default.

internal/security/fileprovider/provider.go
bnevis-i: Code now throws an error.

bnevis-i:
recheck

cmd/security-file-token-provider/res/docker/configuration.toml
```diff
@@ -0,0 +1,19 @@
+[SecretService]
+Scheme = "https"
+Server = "edgex-vault"
+ServerName = "edgex-vault"
```
tsconn23: @bnevis-i I notice `ServerName` is missing from the non-docker `configuration.toml`. Is that an oversight?

cmd/security-file-token-provider/res/docker/configuration.toml
```diff
@@ -0,0 +1,19 @@
...
+Port = 8200
+# CaFilePath = "EdgeXFoundryCA.pem"
```
tsconn23: Since this is commented out in both versions of `configuration.toml` would recommend to just delete it.

cmd/security-file-token-provider/res/docker/configuration.toml
bnevis-i: In fact, I forgot to git add that change.  Fixed.

cmd/security-file-token-provider/res/docker/configuration.toml
bnevis-i: Changed both to `CaFilePath = '/tmp/edgex/secrets/ca/ca.pem'`

bnevis-i: This is to align with #1716

bnevis-i:
Re-pushed due to signature change on bootstrap handler to put Context argument first.

Code review needed, @tinyuz.  See @michaelestrin's on correct implementation of bootstrap handler.

internal/security/fileprovider/init.go
```diff
@@ -0,0 +1,77 @@
...
+
+type Bootstrap struct {
+}
+
+func NewBootstrapHandler() *Bootstrap {
+	return &Bootstrap{}
+}
+
+// BootstrapHandler fulfills the BootstrapHandler contract and performs initialization needed by the data service.
+func (b *Bootstrap) Handler(ctx context.Context, wg *sync.WaitGroup, startupTimer startup.Timer, dic *di.Container) bool {
```
michaelestrin: No reason for `Bootstrap` receiver to exist.

bnevis-i: Good point.  Removed.

PR #2178, opened 2019-11-22
Refactor metadata to remove the nc (i.e. NotificationsClient) global variable.

### LINKED ISSUE #2021: Metadata: refactor to remove nc global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2020

akramtexas:
Fix #2021

### DIFF
--- a/internal/core/metadata/container/notifications.go
+++ b/internal/core/metadata/container/notifications.go
@@ -1,0 +1,15 @@
+package container
+
+import (
+	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/notifications"
+)
+
+// NotificationsClientName contains the name of the NotificationsClient's implementation in the DIC.
+var NotificationsClientName = di.TypeInstanceToName((*notifications.NotificationsClient)(nil))
+
+// NotificationsClientFrom helper function queries the DIC and returns the NotificationsClient's implementation.
+func NotificationsClientFrom(get di.Get) notifications.NotificationsClient {
+	return get(NotificationsClientName).(notifications.NotificationsClient)
+}
--- a/internal/core/metadata/init.go
+++ b/internal/core/metadata/init.go
@@ -18,12 +18,14 @@ import (
 	"context"
 	"sync"
 
+	metadataContainer "github.com/edgexfoundry/edgex-go/internal/core/metadata/container"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/container"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/startup"
 	errorContainer "github.com/edgexfoundry/edgex-go/internal/pkg/container"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/endpoint"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/errorconcept"
+
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/coredata"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/notifications"
@@ -32,7 +34,6 @@ import (
 
 // Global variables
 var Configuration = &ConfigurationStruct{}
-var nc notifications.NotificationsClient
 var vdc coredata.ValueDescriptorClient
 
 // BootstrapHandler fulfills the BootstrapHandler contract and performs initialization needed by the metadata service.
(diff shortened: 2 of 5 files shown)


internal/core/metadata/init.go
```diff
@@ -26,29 +26,19 @@ import (
...
 	registryClient := container.RegistryFrom(dic.Get)
-	nc = notifications.NewNotificationsClient(
-		types.EndpointParams{
-			ServiceKey:  clients.SupportNotificationsServiceKey,
-			Path:        clients.ApiNotificationRoute,
-			UseRegistry: registryClient != nil,
-			Url:         Configuration.Clients["Notifications"].Url() + clients.ApiNotificationRoute,
-			Interval:    Configuration.Service.ClientMonitor,
-		},
-		endpoint.Endpoint{RegistryClient: &registryClient})
```
michaelestrin: You've removed the instantiation of the notifications client.  You've created the helpers (in containers/notifications.go) to provide a unique name and getter for the notifications client in the dependency injection container.  You haven't actually added the service to the dependency injection container.  This code will compile and run but when you invoke an endpoint that requires the notification client, the service will panic.

This mirrors what you needed to do (and ultimately did) in a prior PR for the errorHandler.

akramtexas: @michaelestrin Good catch, thanks. I need to run integration tests more often, too.

internal/core/metadata/init.go
```diff
@@ -26,29 +27,21 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/pkg/errorconcept"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/coredata"
-	"github.com/edgexfoundry/go-mod-core-contracts/clients/notifications"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/notifications"
```
brandonforster: Why was this import separated?

akramtexas: - I have removed the blank line, and separated the imports for consistency. The convention followed (for ordering the imports) is as follows:
`language libraries`
`blank line`
`edgex-go libraries`
`blank line`
`go-mod-core-contracts libraries`
`blank line`
`third-party libraries`
`blank line`

akramtexas:
Rebased.

internal/core/metadata/container/notifications.go
```diff
@@ -0,0 +1,14 @@
+package container
+
+import (
+	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/notifications"
+)
```
AnthonyMBonafide: Can these imports be sorted? I think this would match the rest of the code-base:
```go
import (
	"github.com/edgexfoundry/go-mod-core-contracts/clients/notifications"

	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
)

```

michaelestrin: FWIW, I sort edge-go before go-mod-core-contracts and put empty lines between each of the edgexfoundry modules' imports.

AnthonyMBonafide: Lets go with what @michaelestrin suggested. Ignore my suggestion.

akramtexas:
> This looks and functions well. However, I added one very minor comment regarding import organization.
> 
> Also, the Blackbox tests have some failures. These failures also are present in master, and after looking at them it looks like they are same and are not affected by these changes. I am curious if you are witnessing these issues as well, and if so I will look a little more into this and possibly create an issue to resolve this. Here are the failing requests within the Blackbox tests for reference:
> 
> ```
> 334 http://localhost:48081/api/v1/addressable
>   PUT http://localhost:48081/api/v1/addressable [405 Method Not Allowed, 91B, 4ms]
>   1. Status code is 404
>   ✓  Response time is less than 800ms
> 
> ↳ 134 http://localhost:48081/api/v1/deviceprofile/name/{profileName}
>   DELETE http://localhost:48081/api/v1/deviceprofile/name/DeviceProfileDeleteByName [405 Method Not Allowed, 91B, 3ms]
>   1. Status code is 200
> 
> ↳ 140 http://localhost:48081/api/v1/deviceprofile/uploadfile
>   ┌
>   │ 'Form param `file`, file load error: "sample_profile.y
>   │ aml", no such file'
>   └
>   POST http://localhost:48081/api/v1/deviceprofile/uploadfile [500 Internal Server Error, 194B, 6ms]
>   2. Status code is 200
>   ✓  Response time is less than 800ms
>   ✓  Has Content-Type
>   ✓  Content-Type is text/plain
> 
> ❏ deviceprofile_error_4xx
> ↳ 136 http://localhost:48081/api/v1/deviceprofile/name/{profileName}
>   DELETE http://localhost:48081/api/v1/deviceprofile/name/DeviceProfileDeleteByNameNp [405 Method Not Allowed, 91B, 49ms]
>   1. Status code is 404
>   ✓  Response time is less than 800ms
>   2. Has Content-Type
> ```

- Yes, same observation on my end: These failures also are present in master when I ran them, but they are not affected by the changes in this PR. Please feel free to look into possibly creating an issue to resolve the BB issues 👍

PR #2179, opened 2019-11-22
Scheduler: Remove Configuration global variable

### LINKED ISSUE #2039: Scheduler: refactor to remove Configuration global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2038

brandonforster:
Fixes #2039.

### DIFF
--- a/internal/support/scheduler/rest_intervalaction.go
+++ b/internal/support/scheduler/rest_intervalaction.go
@@ -27,6 +27,7 @@ import (
 	"github.com/gorilla/mux"
 
 	"github.com/edgexfoundry/edgex-go/internal/pkg"
+	"github.com/edgexfoundry/edgex-go/internal/support/scheduler/config"
 	"github.com/edgexfoundry/edgex-go/internal/support/scheduler/errors"
 	"github.com/edgexfoundry/edgex-go/internal/support/scheduler/interfaces"
 	"github.com/edgexfoundry/edgex-go/internal/support/scheduler/operators/intervalaction"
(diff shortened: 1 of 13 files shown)


internal/support/scheduler/rest_intervalaction.go
```diff
@@ -36,12 +37,13 @@ func restGetIntervalAction(
 	w http.ResponseWriter,
 	r *http.Request,
 	loggingClient logger.LoggingClient,
-	dbClient interfaces.DBClient) {
+	dbClient interfaces.DBClient,
+	configuration *config.ConfigurationStruct) {
```
AnthonyMBonafide: I am just noticing that we are sometimes using pointers, but not always, when we are getting instances(objects) from the dependency injection container. I think what you are doing here is correct by passing a pointer(reference) down the call stack, so that  the changes are propagated to all consumers. I am wondering if our previous work here and in other services are passing references and values at the correct times. For example, the `dbClient` and `loggingClient`in this function is a passing by value. Any thoughts?

brandonforster: [Michael called out configuration specifically as an object that should be passed by reference and not passed by value](https://github.com/edgexfoundry/edgex-go/pull/2118#discussion_r345405845). I truly do not know what his reasoning for that distinction is, but I know this is the pattern he wants established.

michaelestrin: Intent was/is to pass these global variables around as references.  `dbClient` and `loggingClient` are both interfaces.  For some reason, I thought interfaces were passed around as references by default.  After a quick look at the literature, I'm not clear whether or not this is actually the case.  

I'd suggest we move forward with the patterns established to date.  If our implementation is actually passing around an interface's concrete implementation by value, we can/should come along afterwards and clean that up.

AnthonyMBonafide: I'll look deeper into this and create issues if needed.

PR #2181, opened 2019-11-25
Fix orphaned event checksum data for Redis

### LINKED ISSUE #2143: Delete event func for Redis misses the deletion of the checksum collection for binary data

AnthonyMBonafide:
Fix #2143

Update the deleteEvent function to perform an extra lookup for the
checksum value associated with the event to be deleted and remove the
lingering data. The additional lookup is necessary since there are three
different types of Events(correlation,contract,and redisEvent) being
used within the Redis implementation of the DBClient and the one being
used for common lookups(contract) does not have the checksum information
so it is lost during serialization. The approach is the most direct and
non-invasive solution. Other attempts to clean-up and reduce the
divergence of the types being used resulted in unrelated Blackbox test
failures.

### DIFF
--- a/internal/pkg/db/redis/data.go
+++ b/internal/pkg/db/redis/data.go
@@ -1097,6 +1097,14 @@ func deleteEvent(conn redis.Conn, id string) error {
 		return err
 	}
 
+	// Obtain the checksum associated with the event.
+	// This is necessary since the helper functions result in an Event type which does not contain the checksum
+	// information, so we need to do a separate lookup to get the information.
+	checksum, err := checksumByEventID(conn, o.ID)
+	if err != nil {
+		return err
+	}
+
 	_ = conn.Send("MULTI")
 	_ = conn.Send("UNLINK", id)
 	_ = conn.Send("UNLINK", db.EventsCollection+":readings:"+id)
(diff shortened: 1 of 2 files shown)


AnthonyMBonafide:
This commit was acherry-picked off of master. This fix has been applied to master(Geneva)

PR #2185, opened 2019-11-26
Remove chEvent global variable from core-data

### LINKED ISSUE #2026: Data: refactor to remove chEvents global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2025

AnthonyMBonafide:
Fix #2026

Remove the global variable chEvent and instead use the DI container to
pass the reference to the channel to different parts of the system.

Update function arguments to accept either publishing or consuming Go
channels so that the intent of the function is clear and also to avoid
errors at compile time.

Update tests to work with the new function signatures.

### DIFF
--- a/internal/core/data/init.go
+++ b/internal/core/data/init.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"sync"
 
+	dataContainer "github.com/edgexfoundry/edgex-go/internal/core/data/container"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/container"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/startup"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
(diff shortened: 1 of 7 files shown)


AnthonyMBonafide:
These changes were verified using the Blackbox tests(all core-data tests pass), and manually verifying that when an Event is created via the REST API the data was sent and received over the channel.

internal/core/data/init.go
```diff
@@ -48,9 +48,16 @@ var httpErrorHandler errorconcept.ErrorHandler
 func BootstrapHandler(wg *sync.WaitGroup, ctx context.Context, startupTimer startup.Timer, dic *di.Container) bool {
 	// update global variables.
 	loggingClient := container.LoggingClientFrom(dic.Get)
-
 	httpErrorHandler = errorconcept.NewErrorHandler(loggingClient)
 
+	chEvents := make(chan interface{}, 100)
+	dic.Update(di.ServiceConstructorMap{dataContainer.EventsChannelName: func(get di.Get) interface{} {
```
michaelestrin: Formatting.  Move service name and closure value definition to its own line.

AnthonyMBonafide:
Rebased

PR #2186, opened 2019-11-26
Logging: Remove dbClient global variable

### LINKED ISSUE #2070: support-logging: refactor to remove dbClient global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2069

brandonforster:
Fixes #2070 

- Refactored support-logging by creating two new interfaces: `Criteria` and `Persistence`. These interfaces are necessary because the container needs an external interface to reference (`Persistence`) and `Persistence` depends on a parameter called a `MatchCriteria` which then itself needs an interface to not have a dependency on the main `logging` package.

- Moved all non-superflous controller logic out of `router.go`. Since this is a huge change and the primary driver of the source lines of code change, I have made one commit where I move code without changing anything, and the second commit with functional changes. This should allow the changes to be parsed.

- Removed `dbClient` `Persistence` global.


internal/support/logging/interfaces/persistence.go
```diff
@@ -0,0 +1,35 @@
...
+
+type Persistence interface {
+	Add(logEntry models.LogEntry) error
+	CloseSession()
+	Remove(criteria Criteria) (int, error)
+	Find(criteria Criteria) ([]models.LogEntry, error)
+
+	// Needed for the tests. Reset the instance (closing files, sessions...)
+	// and clear the logs.
+	Reset()
```
AnthonyMBonafide: Do we need to have this method in the interface, and/or is there a better way to achieve this? My concerns here are:  

1. We are now requiring implementations to provide functionality that is only useful during testing.
2. The perceived intention of this function is to purge all data to allow for a clean testing slate. An accidental call in production is more likely and unrecoverable, that is if the implementation is indeed purging data. For example, the Mongo implementation will permanently delete everything.

internal/support/logging/mongo.go
```diff
@@ -124,29 +129,32 @@ func (ml *mongoLog) remove(criteria matchCriteria) (int, error) {
...
+	le := []models.LogEntry{}
+	if err := q.Limit(matchCriteria.Limit).All(&le); err != nil {
 		return le, err
 	}
 
 	return le, nil
 }
 
-func (ml *mongoLog) reset() {
+func (ml *mongoLog) Reset() {
```
AnthonyMBonafide: See my last comment regarding this now being exported and more accessible.

internal/support/logging/interfaces/persistence.go
```diff
@@ -0,0 +1,35 @@
...
+const (
+	PersistenceDB   = "database"
+	PersistenceFile = "file"
+)
+
+type Persistence interface {
+	Add(logEntry models.LogEntry) error
+	CloseSession()
+	Remove(criteria Criteria) (int, error)
+	Find(criteria Criteria) ([]models.LogEntry, error)
```
AnthonyMBonafide: Can we add GoDocs to these exported entities?

michaelestrin:
> These interfaces are necessary because the container needs an external interface

FWIW the dependency injection container does not require interfaces; it does support concrete types.

brandonforster:
> > These interfaces are necessary because the container needs an external interface
> 
> FWIW the dependency injection container does not require interfaces; it does support concrete types.

Right, but before this PR, there is an unexported interface and two implementations both in the `logging` package. There is no way to reference that interface in the container without a cyclical dependency unless the interface is moved to its own package. That's what I was trying to say; sorry I was unclear.

michaelestrin:
There's some questions around whether or not we should be logging to a database; it's useful for development, less so for production.  Thinking here is ELK stack (or similar) would substitute for support-logging.  Ask is to hold off merging this work until I return, review, and resolve the database logging question.

brandonforster:
@AnthonyMBonafide @michaelestrin Rebased and feedback addressed.

michaelestrin:
Approved.  Setting aside until https://github.com/edgexfoundry/edgex-go/pull/2196 is merged.  Expect merge conficts that will require additional work.  Closing PR to ensure it is not accidentally merged.  Will use this work for follow-on PR that is rebased over merged #2196 and that addresses the anticipated merge conflicts.

PR #2192, opened 2019-11-28
Fix event deleting func with redis DB.

### LINKED ISSUE #2187: [Redis]Delete event failed when the event contains 2 or above readings
# 🐞 Bug Report

### Affected Services

### Is this a regression?

1. Deploy EdgeX with **Redis** database.
2. Create an event that contains 2 readings

> API: {{baseUrl}}/api/v1/event
> Method: POST
> Body:
> {  
>    "origin":1471806386919,
>    "device":"cartachometer",
>    "readings":[  
>       {  
>         "origin":1471806386919,
> 		"device":"cartachometer",
>         "name":"luminousIntensity",
>         "value":"40"
>       },
>       {  
>         "origin":1471806386919,
> 		"device":"cartachometer",
>         "name":"luminousIntensity",
>         "value":"99"
>       }
>    ]
> }

3. Delete the above event.

> API: {{baseUrl}}/api/v1/event/id/${EventId}
> Method: DELETE

## 🔥 Exception or Error
<pre><code>
Got response code 500 with an error message "unexpected end of JSON input"
The following APIs got the same error.
{{baseUrl}}/api/v1/event/scrub
{{baseUrl}}/api/v1/event/removeold/age/0

</code></pre>

## 🌍  Your Environment
**Deployment Environment:** Ubuntu 18.04

**EdgeX Version:** Fuji

**Anything else relevant?**

weichou1229:
The issue where Redis events were not being deleted correctly. 
The nil object will cause the reading unmarshal error "unexpected end of JSON input", so check every object and then return the result.

Fix #2187

### DIFF
--- a/internal/pkg/db/redis/queries.go
+++ b/internal/pkg/db/redis/queries.go
@@ -132,15 +132,18 @@ func getObjectsBySomeRange(conn redis.Conn, command string, key string, start in
 		return nil, err
 	}
 
+	var result [][]byte
 	if len(ids) > 0 {
-		objects, err = redis.ByteSlices(conn.Do("MGET", ids...))
+		result, err = redis.ByteSlices(conn.Do("MGET", ids...))
 		if err != nil {
 			return nil, err
 		}
 	}
 
-	if len(objects) == 1 && objects[0] == nil {
-		objects = [][]byte{}
+	for _, obj := range result {
+		if obj != nil {
+			objects = append(objects, obj)
+		}
 	}
 
 	return objects, nil


weichou1229:
@brandonforster Could you please help review this PR?

brandonforster:
I'm working on running this code, but my immediate concern is that this is a regression by not including the logic I introduced here:

https://github.com/edgexfoundry/edgex-go/pull/2094/files#r343803666

I'm going to run those verification steps and make sure this does not reintroduce that problem while fixing this one.

Issue #2072, opened 2019-11-05
Update loggingClient parameter name to lc


michaelestrin:
Update name from `loggingClient` to `lc` where global variable has been replaced with an injected logger.LoggingClient instance.  @jdharms made this suggestion after several of the services had already been refactored to remove their LoggingClient global.  Instead of making this change in-flight, decided to postpone it until the LoggingClient global had been removed from all services and do the renaming in one follow on PR.

Depends on completion of remaining open LoggingClient global variable removal issues:

- https://github.com/edgexfoundry/edgex-go/issues/2018
- https://github.com/edgexfoundry/edgex-go/issues/2035
- https://github.com/edgexfoundry/edgex-go/issues/2071

PR #2196, opened 2019-12-03
support-logging: remove LoggingClient global

### LINKED ISSUE #2069: support-logging: refactor to remove LoggingClient global variable

judwhite:
Fixes #2069

### DIFF
--- a/internal/support/logging/file.go
+++ b/internal/support/logging/file.go
@@ -93,16 +93,21 @@ func (fl *fileLog) remove(criteria matchCriteria) (int, error) {
 		}
 	}
 
+	// Close old file to open the new one when writing next log
 	tmpFile.Close()
+	f.Close()
+
+	if fl.out != nil {
+		fl.out.Close()
+		fl.out = nil
+	}
+
 	err = os.Rename(tmpFilename, fl.filename)
 	if err != nil {
 		//fmt.Printf("Error renaming %s to %s: %v", tmpFilename, fl.filename, err)
 		return 0, err
 	}
 
-	// Close old file to open the new one when writing next log
-	fl.out.Close()
-	fl.out = nil
 	return count, nil
 }
 
@@ -113,6 +118,7 @@ func (fl *fileLog) find(criteria matchCriteria) ([]models.LogEntry, error) {
 		//fmt.Println("Error opening log file: ", fl.filename, err)
 		return nil, err
 	}
+	defer f.Close()
 	scanner := bufio.NewScanner(f)
 	for scanner.Scan() {
 		var le models.LogEntry
--- a/internal/support/logging/file.go
+++ b/internal/support/logging/file.go
@@ -93,21 +93,16 @@ func (fl *fileLog) remove(criteria matchCriteria) (int, error) {
 		}
 	}
 
-	// Close old file to open the new one when writing next log
 	tmpFile.Close()
-	f.Close()
-
-	if fl.out != nil {
-		fl.out.Close()
-		fl.out = nil
-	}
-
 	err = os.Rename(tmpFilename, fl.filename)
 	if err != nil {
 		//fmt.Printf("Error renaming %s to %s: %v", tmpFilename, fl.filename, err)
 		return 0, err
 	}
 
+	// Close old file to open the new one when writing next log
+	fl.out.Close()
+	fl.out = nil
 	return count, nil
 }
 
@@ -118,7 +113,6 @@ func (fl *fileLog) find(criteria matchCriteria) ([]models.LogEntry, error) {
 		//fmt.Println("Error opening log file: ", fl.filename, err)
 		return nil, err
 	}
-	defer f.Close()
 	scanner := bufio.NewScanner(f)
 	for scanner.Scan() {
 		var le models.LogEntry
--- a/internal/support/logging/init.go
+++ b/internal/support/logging/init.go
@@ -64,7 +64,7 @@ func (s ServiceInit) BootstrapHandler(
 		},
 	})
 
-	lc := container.LoggingClientFrom(dic.Get)
+	loggingClient := container.LoggingClientFrom(dic.Get)
 
 	// get database credentials.
 	var credentials config.Credentials
@@ -74,7 +74,7 @@ func (s ServiceInit) BootstrapHandler(
 		if err == nil {
 			break
 		}
-		lc.Warn(fmt.Sprintf("couldn't retrieve database credentials: %v", err.Error()))
+		loggingClient.Warn(fmt.Sprintf("couldn't retrieve database credentials: %v", err.Error()))
 		startupTimer.SleepForInterval()
 	}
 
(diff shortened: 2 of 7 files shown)


judwhite:
@michaelestrin RFR

michaelestrin:
@brandonforster @AnthonyMBonafide - please review and provide comments/feedback but do not approve and do not merge.

internal/support/logging/router_test.go
```diff
@@ -62,9 +66,17 @@ func (dp dummyPersist) reset() {
...
+		bootstrapContainer.LoggingClientInterfaceName: func(get di.Get) interface{} {
+			return logger.NewMockClient()
+		},
+	})
+}
+
 func TestPing(t *testing.T) {
 	// create test server with handler
-	ts := httptest.NewServer(LoadRestRoutes())
+	ts := httptest.NewServer(LoadRestRoutes(mockDIC()))
```
AnthonyMBonafide: Can we make these tests more effective and to-the-point by simply targeting the REST handler. for example:

https://github.com/edgexfoundry/edgex-go/blob/8d215e236cccce0aabd5398d23074e7883ede7ba/internal/core/data/router_test.go#L185

This also means we can remove the function `func mockDIC()` and simply just use a `LoggingClient` mock to call the REST handler.

internal/support/logging/router_test.go
```diff
@@ -92,7 +104,7 @@ func TestAddLog(t *testing.T) {
 			http.StatusBadRequest},
 	}
 	// create test server with handler
-	ts := httptest.NewServer(LoadRestRoutes())
+	ts := httptest.NewServer(LoadRestRoutes(mockDIC()))
```
AnthonyMBonafide: See my other comment regarding simplifying this test to target the REST handler rather than loading all the routes and making a mock HTTP request.

internal/support/logging/router_test.go
```diff
@@ -204,7 +216,7 @@ func TestGetLogs(t *testing.T) {
 			3},
 	}
 	// create test server with handler
-	ts := httptest.NewServer(LoadRestRoutes())
+	ts := httptest.NewServer(LoadRestRoutes(mockDIC()))
```
AnthonyMBonafide: See my other comment regarding simplifying this test to target the REST handler rather than loading all the routes and making a mock HTTP request.

internal/support/logging/router_test.go
```diff
@@ -295,7 +307,7 @@ func TestRemoveLogs(t *testing.T) {
 			matchCriteria{LogLevels: logLevels, OriginServices: services, Start: 1, End: 2}},
 	}
 	// create test server with handler
-	ts := httptest.NewServer(LoadRestRoutes())
+	ts := httptest.NewServer(LoadRestRoutes(mockDIC()))
```
AnthonyMBonafide: See my other comment regarding simplifying this test to target the REST handler rather than loading all the routes and making a mock HTTP request.

internal/support/logging/file.go
```diff
@@ -93,16 +93,21 @@ func (fl *fileLog) remove(criteria matchCriteria) (int, error) {
...
 
+	// Close old file to open the new one when writing next log
 	tmpFile.Close()
+	f.Close()
+
+	if fl.out != nil {
+		fl.out.Close()
+		fl.out = nil
+	}
+
```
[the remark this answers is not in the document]
judwhite: @AnthonyMBonafide sure, np

internal/support/logging/init.go
```diff
@@ -67,8 +64,7 @@ func (s ServiceInit) BootstrapHandler(
 		},
 	})
 
-	// update global variables.
-	LoggingClient = container.LoggingClientFrom(dic.Get)
+	lc := container.LoggingClientFrom(dic.Get)
```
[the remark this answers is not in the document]
judwhite: @AnthonyMBonafide Yep. Note also #2072 which will come after all the LoggingClient globals have been removed.

internal/support/logging/router_test.go
judwhite: @AnthonyMBonafide It gets a little challenging for the handlers which use `mux.Vars(r)`. To set the vars on the `*http.Request` directly relies on knowledge of Gorilla's internals, and knowing that the router is `gorilla/mux`.

I looked at the discussion in #2042 before doing this change since I had some questions about how to approach this. I couldn't find a test for a handler which used `mux.Vars(r)` that was refactored to being invoked directly. I spoke with @tsconn23 about this Monday and we agreed on this approach for the scope of this PR, specifically that a mock DI Container would be needed and `httptest.NewServer` would stay for now (@tsconn23, please correct me if I'm wrong).

If there's a handler which uses `mux.Vars(r)` that's had its tests refactored please let me know, I'll use that pattern.

internal/support/logging/init.go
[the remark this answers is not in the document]
michaelestrin: We're looking for consistency across refactoring (i.e. loggingClient) with the intent to make the change across all services (i.e. lc) in #2072.   The change to use `lc` was made mid-refactor and we opted for consistency short-term until the deferred change could be made.

internal/support/logging/router_test.go
michaelestrin: I absolutely don't want the DIC (real, mock, or otherwise) added to any of the tests.  I looked briefly at the code and see `mux.Vars()` is called in `getCriteria()`.  I'd recommend bubbling it up into the affected controller functions and passing the result of `mux.Vars()` into those affected functions.  This will allow you to pass in (and vary) the `map[string]string` value returned by `mux.Vars()` in the tests (and not have the tests take a dependency on the DIC).

internal/support/logging/router_test.go
michaelestrin: See above.

internal/support/logging/router_test.go
michaelestrin: See above.

internal/support/logging/router_test.go
michaelestrin: See above.

internal/support/logging/file.go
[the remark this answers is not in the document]
michaelestrin: Please cross-reference the issue you create to this one.

internal/support/logging/router.go
```diff
@@ -32,8 +34,8 @@ func pingHandler(w http.ResponseWriter, _ *http.Request) {
 	w.Write([]byte("pong"))
 }
 
-func configHandler(w http.ResponseWriter, _ *http.Request) {
-	pkg.Encode(Configuration, w, LoggingClient)
+func configHandler(w http.ResponseWriter, _ *http.Request, loggingClient logger.LoggingClient) {
```
michaelestrin: Called out once; occurs elsewhere.

Removed unused parameters from the handler signature.  With the creation of a closure in the route's definition, the unused parameters are no longer needed to fulfill HandleFunc's handler func(ResponseWriter, *Request) signature.

judwhite:
@michaelestrin RFR, will rebase and squash if ready

Issue #2199, opened 2019-12-05
Refactor support-logging to smaller packages


michaelestrin:
Break monolithic implementation package into smaller packages.

Simplify code.

Address race conditions in file logging implementation.

Clean up resulting from #2196.

Prerequisite for https://github.com/edgexfoundry/edgex-go/issues/2070.

PR #2201, opened 2019-12-05
Refactor support-logging to smaller packages

michaelestrin:
Break monolithic implementation package into smaller packages.

Simplify code.

Address race conditions in file logging implementation.

Clean up resulting from #2196.

Prerequisite for #2070.

Fixes: https://github.com/edgexfoundry/edgex-go/issues/2199


internal/support/logging/logger/file/file.go
```diff
@@ -13,69 +13,92 @@ import (
...
+func NewLogger(filename string) *Logger {
+	return &Logger{
+		filename: filename,
 	}
 }
 
-func (fl *fileLog) add(le models.LogEntry) error {
-	if fl.out == nil {
-		var err error
+func (l *Logger) openSession() (err error) {
```
tingyuz: This function never returns error. Is it on purpose?

michaelestrin: It actually does.  It uses a named return parameter.  Whatever the value of that parameter is when  return is called is what is actually returned to the caller.  Functions/methods with named parameters can -- but don't have to -- have explicit values specified as part of the `return` statement.

tingyuz: Thanks, I missed that language feature.

PR #2202, opened 2019-12-06
Business logic for security-file-token-provider

### LINKED ISSUE #1677: Create Vault access tokens for different roles of EdgeX in Security-secretstore-setup
Currently all the micro services in EdgeX is utilizing master token of Vault to access credentials in Vault. It is needed to create different tokens based on different roles in Security-secretstore-setup.

bnevis-i:
This is the main business logic for
security-file-token-provider as documented in
edgex-docs/security/token-file-provider.1.rst

This component is responsible for looping through
the configuration file and creating Vault policies
and security tokens to enable retrieval of
service-specific secrets.

Fixes #1677 

# Testing instructions

## Test

Copy the included scripts into the `testdata` directory 
and run the `integrationtest.sh` script.
(Create `testdata` if it does not exist.)
(Note that this script requires the `jq` utility to be installed.)

```sh
cd testdata
./integrationtest.sh
```

Verify the resulting output for correctness.
The script will automatically fail if a subcommand
returns a nonzero exit status.

Items to look for in the output:

- The AFTER token list contains one more token than the BEFORE token list.
- There is a new policy created: edgex-service-service-name
- The policy dump has the nonstandard path in the KV store with special permisisons, as was configured in the token-config.json
- The dump of the new token has `meta map[edgex-service-name:service-name]` and `policies [default edgex-service-service-name]`

Afterwards, shut down the docker composition

```sh
docker-compose down
```

## Details

The integration test script first starts Vault and Consul via a docker-compose.yml script.
The script copies out the CA certificate and root token JSON file to the current directory.
It then overwrites `res/configuration.toml` with some values to allow the integration test to run.
The script then dumps a list of all policies, and a list of current Vault tokens.

Next, the script runs `security-file-token-provider`

Afterwards, the script then dumps the policies, Vault tokens,
the policy it just created, and calls the token lookup command
on the newly-created Vault token.

## Testing with TLS

In order to test with TLS, it is necessary to edit `/etc/hosts` and add
an alias for edgex-vault:

```
127.0.0.1 edgex-vault
```

Then edit `integrationtest.sh` and uncomment `CaFilePath`.

Re-run `integrationtest.sh` and verify that it makes a TLS connection to Vault.

_BE SURE TO REMOVE THE EDITS FROM `/etc/hosts` WHEN YOU ARE DONE!_

### res/token-config.json
```
{
    "service-name": {
      "edgex_use_defaults": true,
      "custom_policy": {
        "path": {
          "secret/non/standard/location/*": {
            "capabilities": [ "list", "read" ]
          }
        }
      },
      "custom_token_parameters": { }
    }
  }
```

### integrationtest.sh
Run from the testdata directory
```
#!/bin/sh -e

make -C ../../.. cmd/security-file-token-provider/security-file-token-provider

docker-compose -f docker-compose.yml up -d

echo 'Waiting for stack to come up...'
sleep 1

vaultid=`docker ps | grep 'vault:1.' | cut -d' ' -f1`
echo "Vault container id $vaultid"

while ! docker exec $vaultid cat /vault/config/assets/resp-init.json > resp-init.json; do
  echo "Waiting for vault to unlock"
  sleep 1
done

vault_token=`jq -r .root_token ./resp-init.json`
echo "Vault root token is $vault_token"

cat <<EOH > res/configuration.toml
[SecretService]
Scheme = "https"
Server = "localhost"
ServerName = "edgex-vault"
Port = 8200
CaFilePath = "/tmp/edgex/secrets/ca/ca.pem"

[TokenFileProvider]
PrivilegedTokenPath = "resp-init.json"
ConfigFile = "res/token-config.json"
OutputDir = "/tmp"
OutputFilename = "secrets-token.json"

[Writable]
LogLevel = 'DEBUG'

[Logging]
EnableRemote = false
File = './logs/security-file-token-provider.log'
EOH

echo "---==[ BEFORE policies ]==---"
docker exec -e "VAULT_ADDR=https://edgex-vault:8200" -e "VAULT_TOKEN=$vault_token" $vaultid vault policy list -tls-skip-verify

echo "---==[ BEFORE tokens ]==---"
docker exec -e "VAULT_ADDR=https://edgex-vault:8200" -e "VAULT_TOKEN=$vault_token" $vaultid vault list -tls-skip-verify auth/token/accessors

../security-file-token-provider

echo "---==[ AFTER policies ]==---"
docker exec -e "VAULT_ADDR=https://edgex-vault:8200" -e "VAULT_TOKEN=$vault_token" $vaultid vault policy list -tls-skip-verify

echo "---==[ AFTER policy (dump edgex-service-service-name) ]==---"
docker exec -e "VAULT_ADDR=https://edgex-vault:8200" -e "VAULT_TOKEN=$vault_token" $vaultid vault policy read -tls-skip-verify edgex-service-service-name

echo "---==[ AFTER tokens ]==---"
docker exec -e "VAULT_ADDR=https://edgex-vault:8200" -e "VAULT_TOKEN=$vault_token" $vaultid vault list -tls-skip-verify auth/token/accessors

new_token=`jq -r .auth.client_token /tmp/service-name/secrets-token.json`

echo "---==[ INFO on new token ]==---"
docker exec -e "VAULT_ADDR=https://edgex-vault:8200" -e "VAULT_TOKEN=$new_token" $vaultid vault token lookup -tls-skip-verify

```

### docker-compose.yml
```

version: '3.4'

volumes:
  db-data:
  log-data:
  consul-config:
  consul-data:
  consul-scripts:
  portainer_data:
  vault-init:
  vault-config:
  vault-file:
  vault-logs:
  secrets-setup-cache:

services:
  volume:
    image: nexus3.edgexfoundry.org:10004/docker-edgex-volume:master
    container_name: edgex-files
    networks:
      - edgex-network
    volumes:
      - db-data:/data/db
      - log-data:/edgex/logs
      - consul-config:/consul/config
      - consul-data:/consul/data

  consul:
    image: nexus3.edgexfoundry.org:10004/docker-edgex-consul:master
    ports:
      - "8400:8400"
      - "8500:8500"
    container_name: edgex-core-consul
    hostname: edgex-core-consul
    networks:
      edgex-network:
        aliases:
            - edgex-core-consul
    volumes:
      - db-data:/data/db
      - log-data:/edgex/logs
      - consul-config:/consul/config
      - consul-data:/consul/data
      - consul-scripts:/consul/scripts
      - vault-config:/vault/config
      - /tmp/edgex/secrets/ca:/tmp/edgex/secrets/ca:ro
      - /tmp/edgex/secrets/edgex-vault:/tmp/edgex/secrets/edgex-vault:ro
      - /tmp/edgex/secrets/edgex-kong:/tmp/edgex/secrets/edgex-kong:ro
    depends_on:
      - volume

  vault:
    image: vault:1.0.3
    container_name: edgex-vault
    hostname: edgex-vault
    networks:
      edgex-network:
        aliases:
            - edgex-vault
    ports:
      - "8200:8200"
    cap_add:
      - "IPC_LOCK"
    entrypoint: ["/vault/init/start_vault.sh"]
    environment:
      - VAULT_ADDR=https://edgex-vault:8200
      - VAULT_CONFIG_DIR=/vault/config
      - VAULT_UI=true
    volumes:
      - vault-config:/vault/config
      - vault-file:/vault/file
      - vault-logs:/vault/logs
      - vault-init:/vault/init:ro
      - /tmp/edgex/secrets/edgex-vault:/tmp/edgex/secrets/edgex-vault:ro
    depends_on:
      - security-secrets-setup
      - consul

  security-secrets-setup:
    image: nexus3.edgexfoundry.org:10004/docker-edgex-secrets-setup-go:master
    container_name: edgex-secrets-setup
    hostname: edgex-secrets-setup
    tmpfs:
      - /tmp
    command: "generate"
    volumes:
      - secrets-setup-cache:/etc/edgex/pki
      - vault-init:/vault/init
      - /tmp/edgex/secrets:/tmp/edgex/secrets

  vault-worker:
    image: nexus3.edgexfoundry.org:10004/docker-edgex-security-secretstore-setup-go:master
    container_name: edgex-vault-worker
    hostname: edgex-vault-worker
    networks:
      edgex-network:
        aliases:
            - edgex-vault-worker
    volumes:
      - vault-config:/vault/config
      - consul-scripts:/consul/scripts
      - /tmp/edgex/secrets/ca/ca.pem:/tmp/edgex/secrets/ca/ca.pem:ro
      - /tmp/edgex/secrets/edgex-kong:/tmp/edgex/secrets/edgex-kong:ro
    depends_on:
      - consul
      - vault

networks:
  edgex-network:
    driver: "bridge"

```


internal/security/fileprovider/defaults_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDefaultTokenPolicy(t *testing.T) {
+	// Arrange
```
jim-wang-intel: Is this comment still valid? Remove it if nothing to be done.

internal/security/fileprovider/defaults_test.go
```diff
@@ -0,0 +1,53 @@
...
+	bytes, err := json.Marshal(policies)
+	assert.NoError(t, err)
+
+	expected := `{"path":{"secret/edgex/service-name/*":{"capabilities":["create","update","delete","list","read"]}}}`
+	actual := string(bytes)
+	assert.Equal(t, expected, actual)
+}
+
+func TestDefaultTokenParameters(t *testing.T) {
+	// Arrange
```
jim-wang-intel: Is this comment still valid? Remove it if nothing to be done.

internal/security/fileprovider/provider.go
```diff
@@ -57,6 +61,124 @@ func (p *fileTokenProvider) SetConfiguration(secretConfig secretstoreclient.Secr
...
+			if serviceConfig.FilePermissions != nil &&
+				(serviceConfig.FilePermissions).Uid != nil &&
+				(serviceConfig.FilePermissions).Gid != nil {
+				err := permissionable.Chown(*(serviceConfig.FilePermissions).Uid, *(serviceConfig.FilePermissions).Gid)
+				if err != nil {
+					p.logger.Error(fmt.Sprintf("failed to set file user/group on %s: %s", outputTokenFilename, err.Error()))
+					return err
+				}
+			}
+		}
```
jim-wang-intel: What happens if it is not ok?  i.e. if it is not `permissible` type compatible? Should we skip the rest of execution?

internal/security/fileprovider/types.go
```diff
@@ -0,0 +1,28 @@
...
+// or implied. See the License for the specific language governing permissions and limitations under
+// the License.
+//
+// SPDX-License-Identifier: Apache-2.0'
+//
+
+package fileprovider
+
+type TokenFileProviderInfo struct {
+	// Path to Vault authorization token to be used by the service (default: /run/edgex/secrets/token-file-provider/secrets-token.json)
```
jim-wang-intel: now default should be `/tmp/edgex/secrets`

internal/security/fileprovider/types.go
```diff
@@ -0,0 +1,28 @@
...
+	PrivilegedTokenPath string
+	// Configuration file used to control token creation (default: res/token-config.json)
+	ConfigFile string
+	// Base directory for token file output (default: /run/edgex/secrets)
```
jim-wang-intel: change `default: /run/edgex/secrets` to `default: /tmp/edgex/secrets`

cmd/security-file-token-provider/testdata/docker-compose.yml
```diff
@@ -0,0 +1,119 @@
...
+      - 'VAULT_ADDR=https://edgex-vault:8200'
+      - 'VAULT_CONFIG_DIR=/vault/config'
+      - 'VAULT_UI=true'
+    volumes:
+      - vault-config:/vault/config
+      - vault-file:/vault/file
+      - vault-logs:/vault/logs
+    depends_on:
+      - volume
+      - consul
```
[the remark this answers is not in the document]
bnevis-i: Updated to :master version of nightly containers versus edinburgh version of containers.

cmd/security-file-token-provider/testdata/docker-compose.yml
```diff
@@ -0,0 +1,119 @@
...
+      - vault-config:/vault/config
+    depends_on:
+      - volume
+      - consul
+      - vault  
+
+networks:
+  edgex-network:
+    driver: "bridge"
+...
```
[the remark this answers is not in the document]
bnevis-i: resolved.

internal/security/fileprovider/provider.go
bnevis-i: If it is not permissible, then it isn't backed by a file, and setting file permissions is not applicable.

internal/security/fileprovider/types.go
bnevis-i: Removed incorrect documentation.

internal/security/fileprovider/types.go
bnevis-i: removed incorrect documentation

cmd/security-file-token-provider/README.md
```diff
@@ -12,16 +12,25 @@ It should create an executable named `security-file-token-provider` in the curre
 
 ## Run
 
-Run the `integrationtest.sh` script.
+Change to the testdata directory and run the `integrationtest.sh` script.
 
 ```sh
+cd testdata
 ./integrationtest.sh
```
jim-wang-intel: You probably want to mention the prerequisite of `jq` in the documentation. I.e. jq has to be installed first before run the `./integrationtest.sh` as I was having issue on run this locally without `jq`.

jim-wang-intel:
```sh
Creating edgex-secrets-setup ... done
Creating edgex-files         ... done
Creating edgex-core-consul   ... done
Creating edgex-vault         ... done
Creating edgex-vault-worker  ... done
Waiting for stack to come up...
Vault container id b6576557cf3e
./integrationtest.sh: 1: ./integrationtest.sh: jq: not found
```

bnevis-i: Resolved.

bnevis-i:
Applied 'hold' label - working with Jim to get the integration test to run properly.

cmd/security-file-token-provider/testdata/docker-compose.yml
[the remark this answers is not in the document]
bnevis-i: networks is still there in the nightly docker-compose.  Are you sure it is obsolete?

internal/security/fileprovider/defaults.go
```diff
@@ -0,0 +1,44 @@
...
+// or implied. See the License for the specific language governing permissions and limitations under
+// the License.
+//
+// SPDX-License-Identifier: Apache-2.0'
+//
+
+package fileprovider
+
+func makeDefaultTokenPolicy(serviceName string) map[string]interface{} {
+	protectedPath := "secret/edgex/" + serviceName + "/*"
```
jim-wang-intel: Should we validate the input `serviceName` is something non-empty?

internal/security/fileprovider/defaults.go
```diff
@@ -0,0 +1,44 @@
...
+			}
+		  }
+	*/
+}
+
+func makeDefaultTokenParameters(serviceName string) map[string]interface{} {
+	return map[string]interface{}{
+		"display_name": serviceName,
+		"no_parent":    true,
+		"policies":     []string{"edgex-service-" + serviceName},
```
jim-wang-intel: Should we validate the input `serviceName` is something non-empty?

internal/security/fileprovider/provider_test.go
```diff
@@ -0,0 +1,420 @@
...
+	// - All other expectations met
+	assert.NoError(t, err)
+	mockFileIoPerformer.AssertExpectations(t)
+	mockAuthTokenLoader.AssertExpectations(t)
+	mockSecretStoreClient.AssertExpectations(t)
+	assert.Equal(t, expectedTokenFile("service1"), service1Buffer.Bytes())
+	assert.Equal(t, expectedTokenFile("service2"), service2Buffer.Bytes())
+}
+
+func setCreateTokenResponse(serviceName string, retval *interface{}) {
```
jim-wang-intel: Should `serviceName` be used? If not, remove `serviceName`.

internal/security/fileprovider/provider_test.go
```diff
@@ -0,0 +1,420 @@
...
+	mockAuthTokenLoader.AssertExpectations(t)
+	mockSecretStoreClient.AssertExpectations(t)
+	assert.Equal(t, expectedTokenFile("myservice"), service1Buffer.Bytes())
+}
+
+// TestTokenUsingDefaults
+func TestTokenUsingDefaults(t *testing.T) {
+	// Arrange
+	privilegedTokenPath := "/dummy/privileged/token.json"
+	configFile := "token-config.json"
```
jim-wang-intel: I saw this repeats in several places in the test code, should consider to make it as a constant and use it to reduce the maintenance if wanna change in future.

internal/security/fileprovider/provider_test.go
```diff
@@ -0,0 +1,420 @@
...
+	outputDir := "/outputdir"
+	outputFilename := "secrets-token.json"
```
jim-wang-intel: I saw this repeats in several places in the test code, should consider to make it as a constant and use it to reduce the maintenance if wanna change in future.

internal/security/fileprovider/config.go
```diff
@@ -0,0 +1,93 @@
...
+// UpdateFromRaw converts configuration received from the registry to a service-specific configuration struct which is
+// then used to overwrite the service's existing configuration struct.
+func (c *ConfigurationStruct) UpdateFromRaw(rawConfig interface{}) bool {
+	configuration, ok := rawConfig.(*ConfigurationStruct)
+	if ok {
+		// Check that information was successfully read from Registry
+		if configuration.SecretService.Server == "" {
+			return false
+		}
+		c = configuration
```
jim-wang-intel: Not sure this is going to have a deep copy of `configuration` into `c`. it might be ineffective and lose the assigned values after this call. I think you should do 
```go
*c = *configuration
```

internal/security/fileprovider/defaults.go
bnevis-i: The ultimate input is coming from token-config.json where the service name is a key in a key-value map.  It is invalid JSON for this to be empty.  Making the assumption that this is a non-malicous path.

internal/security/fileprovider/config.go
bnevis-i: Bug is legit.  This was part of the boilerplate and was moved to config/config.go. This file needs to be deleted.

internal/security/fileprovider/defaults.go
bnevis-i: As above argument, servicename is a json key and cannot be empty for the json to be valid.

cmd/security-file-token-provider/testdata/docker-compose.yml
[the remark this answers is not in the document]
jim-wang-intel: no, i meant the `...` part :-)

cmd/security-file-token-provider/README.md
jim-wang-intel: Ran and validated that it is working as it intends to do

internal/security/fileprovider/defaults.go
jim-wang-intel: ok, sounds good to me then

internal/security/fileprovider/provider_test.go
bnevis-i: Removed.  Root cause is I though that I wanted to set the metadata property on the response.  In reality, I check that I set the metadata property on the call to the secretstore client and don't care about the actual response.

internal/security/fileprovider/provider_test.go
bnevis-i: Moved these to a const block at the top.

internal/security/fileprovider/provider_test.go
bnevis-i: Moved these to a const block at the top.

internal/security/fileprovider/config.go
bnevis-i: File removed.

internal/security/fileprovider/defaults_test.go
bnevis-i: Removed

internal/security/fileprovider/defaults_test.go
bnevis-i: Removed

cmd/security-file-token-provider/testdata/docker-compose.yml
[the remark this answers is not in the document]
bnevis-i: The ... is no longer in the current version of the file.

cmd/security-file-token-provider/testdata/docker-compose.yml
```diff
@@ -0,0 +1,129 @@
+# /*******************************************************************************
```
michaelestrin: Why are we embedding docker-compose files inside edgex-go for testing?  Is there an alternative that doesn't involve taking a dependency on Docker?

bnevis-i: Most of these security features require the cooperation of multiple different components in order to test that the feature actually does what it says.  In a lot of cases, some nonstandard configuration is needed.  In this particular case, we need to be able to inspect the state of the vault (which is normally uninspectable for security reasons) in order to tell that the code did something that we want.

bnevis-i: I am open to alternatives, but the two supported runtime environments are supposed to be snaps and docker.

bnevis-i:
Rebased and resolved merge conflicts created with go-mod-bootstrap introduction. Requesting re-review.

cmd/security-file-token-provider/testdata/docker-compose.yml
bnevis-i: The other alternative is to not check these files in and include them verbatim in the PR testing instructions.

bnevis-i: @michaelestrin Your call as to whether I pull the scripts or leave them in.  Tingyu approved in https://github.com/edgexfoundry/edgex-go/pull/2202#pullrequestreview-332629740 but I need an approving review to merge since I had to resolve a merge conflict.

michaelestrin: My concern stems from the community direction to coalesce on a single referential set of docker-compose files in the https://github.com/edgexfoundry/developer-scripts/ repository.  

We've experienced difficulty in keeping the docker-compose file in https://github.com/edgexfoundry/blackbox-testing/ in sync; so much so that we've created several issues (https://github.com/edgexfoundry/developer-scripts/issues/154, https://github.com/edgexfoundry/blackbox-testing/issues/295) and are working towards eliminating it altogether.

If I understand correctly, these docker-compose files exist to facilitate integration testing of this PR (and are ultimately useful for ongoing integration testing).  My preference is to remove them from your PR (i.e. not include them in the edgex-go repository).  I would strongly encourage you to leverage them in an integration test suite -- perhaps a conversation with @frq92290 and the Q/A working group is needed to understand where and how to implement such a thing.

@bnevis-i

bnevis-i: @michaelestrin   Once there are users of the new functionality there would be no longer a need for the scripts.  However, this will not occur for several more weeks.   I will remove the scripts.

bnevis-i:
Moved test scripts into PR description.  Rebased and force pushed.  Requesting re-review.

internal/security/fileprovider/provider.go
```diff
@@ -59,6 +69,124 @@ func (p *fileTokenProvider) SetConfiguration(secretConfig secretstoreclient.Secr
...
+
+		defer writeCloser.Close()
+
+		json.NewEncoder(writeCloser).Encode(createTokenResponse) // Write resulting token
+
+		if err := writeCloser.Close(); err != nil {
```
jim-wang-intel: this is a duplicate execution of line 181 
```go
defer writeCloser.Close()
```
should consider combine that line with this one.
Suggested change:

1. Remove line 181
2. make it as defer func:
```suggestion
		defer func() {
		    if err := writeCloser.Close(); err != nil {
```

bnevis-i: @jim-wang-intel 

This article specifically recommends the pattern that I am using:

https://www.joeshaw.org/dont-defer-close-on-writable-files/

This file contains the vault authentication token.  I want to be absolutely sure that it is written without issues.  Moreover, I cannot `return err` inside of the deferred function to indicate to the caller than an error occurred if I defer the explicit Close().

Other suggestions?

internal/security/fileprovider/provider.go
```diff
@@ -59,6 +69,124 @@ func (p *fileTokenProvider) SetConfiguration(secretConfig secretstoreclient.Secr
...
+			return err
+		}
+		permissionable, ok := writeCloser.(permissionable)
+		if ok {
+			if serviceConfig.FilePermissions != nil &&
+				(serviceConfig.FilePermissions).ModeOctal != nil {
+				mode, err := strconv.ParseInt(*(serviceConfig.FilePermissions).ModeOctal, 8, 32)
+				if err != nil {
+					p.logger.Error(fmt.Sprintf("invalid file mode %s: %s", *(serviceConfig.FilePermissions).ModeOctal, err.Error()))
+					return err
```
michaelestrin: Should writeCloser be explicitly closed before this `return`?

internal/security/fileprovider/provider.go
```diff
@@ -59,6 +69,124 @@ func (p *fileTokenProvider) SetConfiguration(secretConfig secretstoreclient.Secr
...
+				}
+				if err := permissionable.Chmod(os.FileMode(mode)); err != nil {
+					p.logger.Error(fmt.Sprintf("failed to set file mode on %s: %s", outputTokenFilename, err.Error()))
+					return err
```
michaelestrin: Should writeCloser be explicitly closed before this `return`?

internal/security/fileprovider/provider.go
```diff
@@ -59,6 +69,124 @@ func (p *fileTokenProvider) SetConfiguration(secretConfig secretstoreclient.Secr
...
+				}
+			}
+			if serviceConfig.FilePermissions != nil &&
+				(serviceConfig.FilePermissions).Uid != nil &&
+				(serviceConfig.FilePermissions).Gid != nil {
+				err := permissionable.Chown(*(serviceConfig.FilePermissions).Uid, *(serviceConfig.FilePermissions).Gid)
+				if err != nil {
+					p.logger.Error(fmt.Sprintf("failed to set file user/group on %s: %s", outputTokenFilename, err.Error()))
+					return err
```
michaelestrin: Should writeCloser be explicitly closed before this `return`?

internal/security/fileprovider/provider.go
michaelestrin: Could you do a defer that only closes the writeCloser if it hasn't been closed already (e.g. isn't nil).  This would catch the various error exits that might occur before you get to your explicit close (which could immediately/subsequently set writeCloser to nil to ensure the deferred close isn't executed..?

internal/security/fileprovider/provider.go
bnevis-i: File contents are not written until line 183.  defer Close() should be moved to line 156.

internal/security/fileprovider/provider.go
bnevis-i: File contents are not written until line 183.  defer Close() should be moved to line 156.

internal/security/fileprovider/provider.go
bnevis-i: This is possible and has been done.

internal/security/fileprovider/provider.go
bnevis-i: defered Close() is moved to before this line.

internal/security/fileprovider/provider.go
bnevis-i: defered Close() is moved to before this line.

cmd/security-file-token-provider/res/token-config.json
```diff
@@ -0,0 +1,64 @@
...
+  },
+  "edgex-core-metadata": {
+    "edgex_use_defaults": true
+  },
+  "edgex-export-client": {
+    "edgex_use_defaults": true
+  },
+  "edgex-export-distro": {
+    "edgex_use_defaults": true
+  },
```
michaelestrin: Omit these.  I expect to merge https://github.com/edgexfoundry/edgex-go/pull/2219 later today; it will remove all traces of the export services from edgex-go.

internal/security/fileprovider/provider.go
```diff
@@ -59,6 +69,130 @@ func (p *fileTokenProvider) SetConfiguration(secretConfig secretstoreclient.Secr
...
+				}
+			}
+		}
+
+		json.NewEncoder(writeCloser).Encode(createTokenResponse) // Write resulting token
+
+		if err := writeCloser.Close(); err != nil {
+			p.logger.Error(fmt.Sprintf("failed to close %s: %s", outputTokenFilename, err.Error()))
+			return err
+		} else {
```
michaelestrin: FWIW, the `else` isn't necessary.  You could omit it and simply have `writeCloser=nil` as the next statement.

cmd/security-file-token-provider/res/token-config.json
bnevis-i: Removed.

internal/security/fileprovider/provider.go
bnevis-i: Simplified.

internal/security/fileprovider/provider.go
jim-wang-intel: can you use err scope closure to get the error returned from inside the defer func()?  the article also mentions that as well.  like cerr := f.close() and assign the closure err = cerr? so err changes its state right before its return

michaelestrin: I'd strive for readability and code that's easy to understand.  My interpretation of the above is anything but.

bnevis-i: It is possible with a named result parameter so instead of 

```
		defer func() {
			if writeCloser != nil {
				writeCloser.Close()
			}
		}()
		// code
		if err := writeCloser.Close(); err != nil {
			p.logger.Error(fmt.Sprintf("failed to close %s: %s", outputTokenFilename, err.Error()))
			return err
		}
		writeCloser = nil
```

```
	defer func() {
		if newerr: = writeCloser.Close() {
			p.logger.Error(fmt.Sprintf("failed to close %s: %s", outputTokenFilename, err.Error()))
			err= newerr
			return
		}
	}()
	// code
```

bnevis-i: I would prefer to go with @michaelestrin 's suggestion as I've been burned too many times by named result arguments.

jim-wang-intel: I do not agree that the first one is more readable. It looks strange to me that if you already got  the close error ,why close again the second time no matter what?

bnevis-i: This PR is 2 weeks old.  I don't care which way, as long as the code gets in.  Can @jim-wang-intel and @michaelestrin  come to agreement on which way you want it done and I will do it that way.

michaelestrin: Your frustration is palatable and I totally understand and empathize with you.

To quote the article:

> Still, I find this pattern to be a little too magical. I dislike using named return values, and modifying the return value after the core function finishes is not intuitively clear even to experienced Go programmers.

A return value that's set in a closure isn't exactly clear or "readable."  "Too magical."  But I don't particularly care to argue the point with @jim-wang-intel and reserve my right to my opinion.

As far as I'm concerned, the current implementation looks good to me and I'll approve it.   Whether you change it or not is up to you.  :)

bnevis-i:
recheck

bnevis-i:
recheck

PR #2204, opened 2019-12-06
support-logging: refactor to remove dbClient global variable

michaelestrin:
Dependency on #2199 (PR #2201)

Fixes: #2070

### DIFF
--- a/internal/support/logging/init.go
+++ b/internal/support/logging/init.go
@@ -10,18 +10,19 @@ package logging
 import (
 	"context"
 	"fmt"
-	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/logging"
-	"github.com/edgexfoundry/edgex-go/internal/support/logging/logger/mongo"
 	"sync"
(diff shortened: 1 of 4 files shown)


michaelestrin:
Rebased.

michaelestrin:
Supercedes https://github.com/edgexfoundry/edgex-go/pull/2186

michaelestrin:
Rebased.

internal/support/logging/init.go
```diff
@@ -86,6 +80,7 @@ func (s ServiceInit) BootstrapHandler(
 	}
 
 	// initialize database.
+	var dbClient interfaces.Logger
```
AnthonyMBonafide: I wonder if should name this something different, something like `persistenceClient`, since this can be something other than a database. Just a thought.

michaelestrin: Created https://github.com/edgexfoundry/edgex-go/issues/2211

Issue #2206, opened 2019-12-06
Remove Unnecessary DIC definition in service's main.go


michaelestrin:
Remove redundant definition of `ConfigurationInterfaceName` in dependency injection container instantiation in each core service's main.go.

PR #2207, opened 2019-12-06
Remove Unnecessary DIC definition in service's main.go

michaelestrin:
Removed redundant definition of ConfigurationInterfaceName in
dependency injection container instantiation in each core
service's main.go.

Fixes: https://github.com/edgexfoundry/edgex-go/issues/2206


michaelestrin:
recheck

PR #2209, opened 2019-12-06
Remove msgClient global variable for core-data

### LINKED ISSUE #2027: Data: refactor to remove msgClient global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2026

AnthonyMBonafide:
Fix #2027

Remove the msgClient global variable in favor of using the DI container
for passing around an instance during bootstrap. Update functions which
were referring to the global variable to instead accept an instance as
function arguments.

Update tests to work with the changes to the production code.

### DIFF
--- a/internal/core/data/container/messaging.go
+++ b/internal/core/data/container/messaging.go
@@ -1,0 +1,29 @@
+/********************************************************************************
+ *  Copyright 2019 Dell Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package container
+
+import (
+	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
+
+	"github.com/edgexfoundry/go-mod-messaging/messaging"
+)
+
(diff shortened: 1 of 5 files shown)


AnthonyMBonafide:
These changes where verified with passing blackbox tests.

internal/core/data/init.go
```diff
@@ -96,5 +89,13 @@ func BootstrapHandler(ctx context.Context, wg *sync.WaitGroup, startupTimer star
 		loggingClient.Error(fmt.Sprintf("failed to create messaging client: %s", err.Error()))
 	}
 
+	dic.Update(di.ServiceConstructorMap{
+		dataContainer.MessagingClientName: func(get di.Get) interface{} {
+			return msgClient
+		},
```
michaelestrin: You're putting a potentially invalid or nil reference for msgClient into the DIC.  Would it be better to fail-fast and stop the service here?

internal/core/data/container/messaging.go
```diff
@@ -0,0 +1,29 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package container
+
+import (
+	"github.com/edgexfoundry/go-mod-messaging/messaging"
+
+	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
```
michaelestrin: I've tried to be consistent in putting edgex-go imports before the go-mod-* imports.

internal/core/data/init.go
AnthonyMBonafide: Yes, I think failing fast is the best option here.

michaelestrin: I don't know if it makes sense contextually (or the way the underlying code that returns msgClient is written) -- you might consider implementing multiple retries before failing.  See [here](https://github.com/edgexfoundry/edgex-go/blob/333829e89581c14c68f5ecbcfda36ed5096dfa50/internal/pkg/bootstrap/handlers/database/database.go#L118) for an example that uses the startup timer to determine when to give up.

Issue #2215, opened 2019-12-09
Refactor to integrate go-mod-bootstrap


michaelestrin:
When go-mod-bootstrap lands (and has a semver-based build pipeline), refactor existing core and system management service implementations to reference go-mod-bootstrap (instead of the existing embedded bootstrap/dependency injection container implementation).

michaelestrin:
Related to https://github.com/edgexfoundry/edgex-go/issues/2169

michaelestrin:
Dependent on https://github.com/edgexfoundry/ci-management/issues/523

PR #2223, opened 2019-12-11
Refactor to integrate go-mod-bootstrap

michaelestrin:
Fixes: https://github.com/edgexfoundry/edgex-go/issues/2215


michaelestrin:
As of this writing, https://github.com/edgexfoundry/go-mod-bootstrap does not have a build pipeline or semver versioning in place.  See https://github.com/edgexfoundry/ci-management/issues/523

michaelestrin:
Rebased.

internal/core/command/router.go
```diff
@@ -17,13 +17,16 @@ package command
 import (
 	"net/http"
 
-	"github.com/edgexfoundry/edgex-go/internal/core/command/container"
+	commandContainer "github.com/edgexfoundry/edgex-go/internal/core/command/container"
 	"github.com/edgexfoundry/edgex-go/internal/pkg"
-	bootstrapContainer "github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/container"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/container"
```
lenny-intel: Why is this still in edgex-go?

internal/pkg/bootstrap/container/database.go
```diff
@@ -16,7 +16,8 @@ package container
 
 import (
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db/interfaces"
-	"github.com/edgexfoundry/edgex-go/internal/pkg/di"
+
+	"github.com/edgexfoundry/go-mod-bootstrap/di"
```
lenny-intel: So DB boot strapping isn't moved to go-mod-bootstrap since it is service specific, correct?

internal/pkg/config/types.go
```diff
@@ -33,65 +35,6 @@ func ListDefaultServices() map[string]string {
...
-	File         string
-}
-
-// StartupInfo provides the startup timer values which are applied to the StartupTimer created at boot.
-type StartupInfo struct {
-	Duration int
-	Interval int
-}
-
 // MessageQueueInfo provides parameters related to connecting to a message queue
```
lenny-intel: Why isn't MessageQueueInfo also moved?

lenny-intel: Why isn't MessageQueueInfo also moved?

internal/support/notifications/router.go
```diff
@@ -19,11 +19,13 @@ import (
 	"net/http"
 
 	"github.com/edgexfoundry/edgex-go/internal/pkg"
-	bootstrapContainer "github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/container"
+	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/container"
```
lenny-intel: This is for DB correct? Should it be renamed to something like /pkg/database/container?
and aliased as dbContainer for consistency...

internal/pkg/bootstrap/container/database.go
michaelestrin: That's correct.

internal/pkg/config/types.go
michaelestrin: It isn't referenced by the common bootstrap implementation or its common handlers.

internal/pkg/bootstrap/container/database.go
lenny-intel: What about changing the package name  to something like /pkg/database/container?

michaelestrin: Maybe but I see that as out of scope for this PR.

lenny-intel: Why? Isn't this about refactoring out the bootstrap package? Seems strange to leave something in it that really isn't bootstrap....

michaelestrin: It is bootstrap-related.  Just happens to be specific to edgex-go.  This file contains the helper functions related to the edgex-go-specific database handler (which is also under internal/pkg/bootstrap).  It makes sense to me to keep the same pattern in edgex-go for the service-specific bootstrap handlers that we're using in go-mod-bootstrap for the globally common bootstrap handlers.

lenny-intel: Ok, good with how it is now.

tsconn23:
FYI -- On MacOS I get the following error running unit tests. I assume this is OS specific since Michael and Lenny are on Windows and Anthony and Jenkins are on Linux.
```
--- 
FAIL: TestTrue (0.00s)
    tokenprovider_test.go:64: 
        	Error Trace:	tokenprovider_test.go:64
        	Error:      	Expected nil, but got: &exec.Error{Name:"/bin/true", Err:(*os.PathError)(0xc00011a8d0)}
        	Test:       	TestTrue
FAIL
coverage: 22.7% of statements
FAIL	github.com/edgexfoundry/edgex-go/internal/security/secretstore	0.056s
```
Tried more than once. Tagging @bnevis-i as the author of the test.

michaelestrin:
> On MacOS I get the following error running unit tests. I assume this is OS specific since Michael and Lenny are on Windows, Anthony and Jenkins are on Linux.

The test you reference assumes the operating system has a `true` command/executable.  There is another in the same file/suite -- TestEcho -- that assumes the operating system has an `echo` command/executable.  These are Linux-centric tests and assume a Linux environment.  I've run across other tests that have similar assumptions about the operating system.

As I'm currently configured, both of the aforementioned tests fail on my personal Windows machine.

I think we should avoid writing OS-specific tests (or tests that make assumptions about what is available within the environment a process is running).  IMO, these two tests should be rewritten; though I also believe doing so is outside the scope of this PR.

@tsconn23 @bnevis-i

PR #2225, opened 2019-12-12
Add guard to security-secrets-setup against replacing active PKI

### LINKED ISSUE #2224: security-secrets-setup/docker infrastructure does not guard against replacing an active PKI
# 🐞 Bug Report

### Affected Services
Most directly affects blackbox testing; can come up in other scenarios as well.

### Is this a regression?
Introduced in Geneva when secrets-setup moved to its own container.

This error in introduced when containers are started one-by-one, such as in blackbox-testing.  Security-secrets-setup is a one-shot container that generates a PKI and terminates.  Suppose someone starts vault-worker and its dependencies (one of which is secrets-setup to generate the PKI).  Then supposed edgex-proxy is started (again, secrets-setup is a dependency).  On the second launch, the container is run a second time and the PKI regenerated because the container is not running.  This will replace the current PKI with a new CA and leaf certificates.  As a result, edgex-proxy will verify edgex-vault against a different PKI than what was used when edgex-vault started. This will cause TLS certificate verification failures.

## 🔥 Exception or Error
<pre><code>
Get https://edgex-vault:8200/v1/sys/health: x509: certificate signed by unknown authority (possibly because of \"x509: ECDSA verification failure\" while trying to verify candidate authority certificate \"EdgeXFoundryCA\")"
</code></pre>

## 🌍  Your Environment
**Deployment Environment:**
Ubuntu 18.10

**EdgeX Version:**
master

**Anything else relevant?**

bnevis-i:
Fixes #2224 

Prevent security-secrets-setup script from terminating
by tailing /dev/null at the end of execution.

This mitigation is needed because blackbox tests
start services one-by-one.  Because dependencies that
are not running are restarted, this causes re-execution
of security-secrets setup, which replaces the active PKI,
which results in fatal TLS verification failures.
Such activity could also be triggered via SMA as well.

In addition, installation of the Vault entrypoint script
is moved up in the secution flow for further robsustness.


cmd/security-secrets-setup/Dockerfile
```diff
@@ -86,4 +86,6 @@ RUN chmod 755 security-secrets-setup
 
 ENTRYPOINT ["entrypoint.sh"]
 
+STOPSIGNAL SIGTERM
```
jim-wang-intel: FYI: This gives exit code `143` if docker-compose down.  Not sure if any blackbox testing is checking the exit code of services or not.

bnevis-i: Not sure.  This was blocking blackbox testing and I haven't gotten all the way through getting blackbox testing to see if this is of consequence.

michaelestrin: I don't know if this is desirable.  `STOPSIGNAL SIGTERM` sets the signal used to terminate the container.  On receipt, this would collapse the container.  I suspect it would be better to allow the signal to pass through to the application.  If the application is using the common bootstrap, the signal should result in a graceful shutdown.

bnevis-i: Not quite right.  STOPSIGNAL is the signal that is set to a docker container when `docker stop` (or `docker-compose down`) is issued.  Actually the default STOPSIGNAL is already SIGTERM so I am going to remove it.

bnevis-i:
rebased and code review changes applied.

cmd/security-secrets-setup/Dockerfile
bnevis-i: (steady state of the container is that it is hung on /bin/tail after setting up PKI.)

PR #2240, opened 2019-12-16
Remove platform dependencies from TokenProvider unit tests.

### LINKED ISSUE #2233: Update security/secretstore tests to be OS-agnostic
> On MacOS I get the following error running unit tests. I assume this is OS specific since Michael and Lenny are on Windows, Anthony and Jenkins are on Linux.
>```
>--- 
>FAIL: TestTrue (0.00s)
>    tokenprovider_test.go:64: 
>        	Error Trace:	tokenprovider_test.go:64
>        	Error:      	Expected nil, but got: &exec.Error{Name:"/bin/true", Err:(*os.PathError)(0xc00011a8d0)}
>        	Test:       	TestTrue
>FAIL
>coverage: 22.7% of statements
>FAIL	github.com/edgexfoundry/edgex-go/internal/security/secretstore	0.056s
>```

The test you reference assumes the operating system has a `true` command/executable.  There is another in the same file/suite -- TestEcho -- that assumes the operating system has an `echo` command/executable.  These are Linux-centric tests and assume a Linux environment.  I've run across other tests that have similar assumptions about the operating system.

As I'm currently configured, both of the aforementioned tests fail on my personal Windows machine.

I think we should avoid writing OS-specific tests (or tests that make assumptions about what is available within the environment a process is running).  IMO, these two tests should be rewritten; though I also believe doing so is outside the scope of this PR.

@tsconn23 @bnevis-i

_Originally posted by @michaelestrin in https://github.com/edgexfoundry/edgex-go/pull/2223#issuecomment-565724270_

bnevis-i:
Remove platform dependencies from TokenProvider unit tests.

Fixes #2233.

Remove platform dependencies (assumption of /bin/false, /bin/true,
and /bin/echo) from the unit tests by injecting a dependency
that abstracts the concrete implementation of os.exec and os.Cmd.

### DIFF
--- a/internal/security/secretstore/tokenprovider_test.go
+++ b/internal/security/secretstore/tokenprovider_test.go
@@ -18,12 +18,16 @@ package secretstore
 
 import (
 	"context"
+	"errors"
+	"os"
+	"os/exec"
 	"testing"
 
 	"github.com/edgexfoundry/edgex-go/internal/security/secretstoreclient"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
 )
 
 func TestInvalidProvider(t *testing.T) {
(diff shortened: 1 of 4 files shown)


michaelestrin:
Removing myself as reviewer.  Leaving for @brandonforster to review.

bnevis-i:
Request re-review.

* Testing Instructions
Go to `internal/security/secretstore` and run `go test` and verify unit tests pass.
Also run the same unit tests on Windows and make sure that the Linux specific test does not run.

brandonforster:
Verified that all 9 compatible tests run and pass on macOS. Verified that `make test` passes.

internal/security/secretstore/tokenprovider_test.go
```diff
@@ -18,78 +18,123 @@ package secretstore
...
-		TokenProvider:     "/bin/true",
+		TokenProvider:     "success-executable",
 		TokenProviderType: OneShotProvider,
+		TokenProviderArgs: []string{"arg1", "arg2"},
 	}
-	cancel, err := testCommon(config)
+	mockExecRunner := mockExecRunner{}
+	mockCmd := mockCmd{}
+	mockExecRunner.On("LookPath", "success-executable").Return("success-executable", nil)
+	mockExecRunner.On("CommandContext", mock.Anything, "success-executable", []string{"arg1", "arg2"}).Return(&mockCmd)
```
brandonforster: Can you fold this line out so it isn't so long?

internal/security/secretstore/tokenprovider_test.go
```diff
@@ -18,78 +18,123 @@ package secretstore
...
-		TokenProvider:     "/bin/false",
+		TokenProvider:     "failure-executable",
 		TokenProviderType: OneShotProvider,
+		TokenProviderArgs: []string{"arg1", "arg2"},
 	}
-	cancel, err := testCommon(config)
+	mockExecRunner := mockExecRunner{}
+	mockCmd := mockCmd{}
+	mockExecRunner.On("LookPath", "failure-executable").Return("failure-executable", nil)
+	mockExecRunner.On("CommandContext", mock.Anything, "failure-executable", []string{"arg1", "arg2"}).Return(&mockCmd)
```
brandonforster: Can you fold this line out so it isn't so long?

internal/security/secretstore/tokenprovider_test.go
```diff
@@ -18,78 +18,123 @@ package secretstore
...
+	mockCmd.On("Start").Return(nil)
+	mockCmd.On("Wait").Return(&exec.ExitError{ProcessState: &os.ProcessState{}})
+	cancel, err := testCommon(config, &mockExecRunner)
 	defer cancel()
 	assert.Error(t, err)
+	mockExecRunner.AssertExpectations(t)
+	mockCmd.AssertExpectations(t)
 }
 
-func TestEcho(t *testing.T) {
```
brandonforster: Would this be valuable in the new Linux file?

internal/security/secretstore/tokenprovider_test.go
```diff
@@ -18,78 +18,123 @@ package secretstore
...
 	config := secretstoreclient.SecretServiceInfo{
-		TokenProvider:     "/bin/false",
+		TokenProvider:     "failure-executable",
 		TokenProviderType: OneShotProvider,
+		TokenProviderArgs: []string{"arg1", "arg2"},
 	}
-	cancel, err := testCommon(config)
+	mockExecRunner := mockExecRunner{}
+	mockCmd := mockCmd{}
+	mockExecRunner.On("LookPath", "failure-executable").Return("failure-executable", nil)
```
brandonforster: Please use `config.TokenProvider` instead of retyping the constant.

internal/security/secretstore/tokenprovider_test.go
```diff
@@ -18,78 +18,123 @@ package secretstore
...
 	config := secretstoreclient.SecretServiceInfo{
-		TokenProvider:     "/bin/true",
+		TokenProvider:     "success-executable",
 		TokenProviderType: OneShotProvider,
+		TokenProviderArgs: []string{"arg1", "arg2"},
 	}
-	cancel, err := testCommon(config)
+	mockExecRunner := mockExecRunner{}
+	mockCmd := mockCmd{}
+	mockExecRunner.On("LookPath", "success-executable").Return("success-executable", nil)
```
brandonforster: Please use `config.TokenProvider` instead of retyping the constant.

internal/security/secretstore/tokenprovider_test.go
bnevis-i: The original intent of TestEcho was to check command line argument processing.  `TestCreatesFile()` does the equivalent  testing.

internal/security/secretstore/tokenprovider_test.go
bnevis-i: folded.

internal/security/secretstore/tokenprovider_test.go
bnevis-i: reused (cc: many many places)

internal/security/secretstore/tokenprovider_test.go
bnevis-i: foled

internal/security/secretstore/tokenprovider_test.go
bnevis-i: reused

PR #2241, opened 2019-12-16
Remove global variable mdc from core data

### LINKED ISSUE #2028: Data: refactor to remove mdc global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2027

AnthonyMBonafide:
Fix #2028

Remove the global variable mdc(metadata client), and instead use the DI
container to create and pass an instance to functions/methods which need
it.

Update function/method signatures to accept the metadata client as a
parameter.

Update tests to work with changes to the production code.

### DIFF
--- a/internal/core/data/container/metadata.go
+++ b/internal/core/data/container/metadata.go
@@ -1,0 +1,29 @@
+/********************************************************************************
+ *  Copyright 2019 Dell Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package container
+
+import (
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata"
+
+	"github.com/edgexfoundry/go-mod-bootstrap/di"
+)
+
+// MetadataDeviceClientName contains the name of the Metadata device client instance in the DIC.
+var MetadataDeviceClientName = "MetadataDeviceClient"
+
+// MetadataDeviceClientFrom helper function queries the DIC and returns the Metadata device client instance.
+func MetadataDeviceClientFrom(get di.Get) metadata.DeviceClient {
+	return get(MetadataDeviceClientName).(metadata.DeviceClient)
+}
--- a/internal/core/data/event_test.go
+++ b/internal/core/data/event_test.go
@@ -22,17 +22,16 @@ import (
 	"testing"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
-	"github.com/edgexfoundry/go-mod-core-contracts/models"
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 	"github.com/edgexfoundry/go-mod-messaging/messaging"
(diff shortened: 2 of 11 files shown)


internal/core/data/container/metadata.go
```diff
@@ -0,0 +1,29 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package container
+
+import (
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata"
+
+	"github.com/edgexfoundry/go-mod-bootstrap/di"
```
michaelestrin: Nitpick -- these should be reversed based on alphabetical ordering.

internal/core/data/event_test.go
```diff
@@ -22,17 +22,16 @@ import (
...
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 	"github.com/edgexfoundry/go-mod-messaging/messaging"
+	msgTypes "github.com/edgexfoundry/go-mod-messaging/pkg/types"
+	"github.com/globalsign/mgo/bson"
+	"github.com/stretchr/testify/mock"
 
 	"github.com/edgexfoundry/edgex-go/internal/core/data/errors"
 	dbMock "github.com/edgexfoundry/edgex-go/internal/core/data/interfaces/mocks"
 	correlation "github.com/edgexfoundry/edgex-go/internal/pkg/correlation/models"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
```
michaelestrin: Nitpick.  I'd restructure this to:

edgex-go references.

go-mod-core-contracts references.

go-mod-messaging references.

public github.com references.

internal/core/data/event_test.go
```diff
@@ -56,7 +55,7 @@ func TestCountByDevice(t *testing.T) {
 	dbClientMock := &dbMock.DBClient{}
 	dbClientMock.On("EventCountByDeviceId", mock.Anything).Return(2, nil)
 
-	count, err := countEventsByDevice(testEvent.Device, context.Background(), dbClientMock)
+	count, err := countEventsByDevice(testEvent.Device, context.Background(), dbClientMock, newMockDeviceClient())
```
michaelestrin: You're referencing a mock from a different test file albeit from the same package.  This is problematic and couples event_test.go to device_test.go making future refactoring of these tests (and their underlying implementation) to separate packages more difficult.  A more flexible implementation would extract the mock factory function into its own package.

I see you didn't write newMockDeviceClient(); you're simply using it.  

Commented here; applies elsewhere.

internal/core/data/event_test.go
```diff
@@ -66,12 +65,12 @@ func TestCountByDevice(t *testing.T) {
...
+func buildEvents() []contract.Event {
+	events := []contract.Event{}
+	events = append(events, contract.Event{
 		ID:     "1",
 		Device: testDeviceName,
-		Readings: []models.Reading{
+		Readings: []contract.Reading{
 			{Id: "1"},
 			{Id: "2"},
 		},
```
michaelestrin: Nitpick.  This could be reduced to:

```
func buildEvents() []contract.Event {
	return []contract.Event{
		{
			ID:     "1",
			Device: testDeviceName,
			Readings: []contract.Reading{
				{Id: "1"},
				{Id: "2"},
			},
		},
	}
}
```

internal/core/data/router.go
```diff
@@ -86,8 +88,21 @@ func LoadRestRoutes(dic *di.Container) *mux.Router {
 				bootstrapContainer.LoggingClientFrom(dic.Get),
 				container.DBClientFrom(dic.Get),
 				dataContainer.PublisherEventsChannelFrom(dic.Get),
-				dataContainer.MessagingClientFrom(dic.Get))
+				dataContainer.MessagingClientFrom(dic.Get),
+				dataContainer.MetadataDeviceClientFrom(dic.Get))
+
```
michaelestrin: Nitpick.  Empty line should be removed.

Commented here; applies elsewhere.

PR #2242, opened 2019-12-16
Feature/issue 2029 coredata remove msc global

### LINKED ISSUE #2029: Data: refactor to remove msc global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2028

AnthonyMBonafide:
Fix #2029
    
    Remove mdc(Metadata DeviceService client) global and instead pass an
    instance to the functions/methods which need it.
    
    Update functions/methods referring to the mdc global variable to accept
    the client as arguments.
    
### DIFF
--- a/internal/core/data/device.go
+++ b/internal/core/data/device.go
@@ -61,7 +61,8 @@ func updateDeviceLastReportedConnected(device string, loggingClient logger.Loggi
 func updateDeviceServiceLastReportedConnected(
 	device string,
 	loggingClient logger.LoggingClient,
-	mdc metadata.DeviceClient) {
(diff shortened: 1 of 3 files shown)


michaelestrin:
Why was this issue closed @AnthonyMBonafide?

michaelestrin:
> Fix #2029

Include full GitHub issue URL in commit messages -- it creates links within the GitHub PR/issue (which the non-URL version does not).

AnthonyMBonafide:
@michaelestrin, I closed this issue due to the diff containing changes from this PR and the dependant issue. We talked about this being OK and I will be doing this for the following dependant PRs.

PR #2244, opened 2019-12-17
Feature/issue 2030 coredata remove http error handler global

### LINKED ISSUE #2030: Data: refactor to remove httpErrorHandler global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2029

AnthonyMBonafide:
Fix #2030 

Remove the `httpErrorHandler` global variable in favor of using the DI container to pass an instance to the necessary functions.

Update the loadRoutes function to pull the `httpErrorHandler` out of the DI container and pass it to the REST handlers which require it.


internal/core/data/container/http.go
```diff
@@ -0,0 +1,29 @@
+/********************************************************************************
```
michaelestrin: Use the existing implementation in /internal/pkg/container/errorHandler.go

internal/core/data/init.go
```diff
@@ -95,7 +94,11 @@ func BootstrapHandler(ctx context.Context, wg *sync.WaitGroup, startupTimer star
 		},
 		dataContainer.EventsChannelName: func(get di.Get) interface{} {
 			return chEvents
-		}})
+		},
+		dataContainer.HttpErrorHandlerName: func(get di.Get) interface{} {
+			return httpErrorHandler
```
michaelestrin: Why not eliminate the httpErrorHandler variable?

`return errorconcept.NewErrorHandler(loggingClient)`

internal/core/data/router.go
```diff
@@ -109,19 +111,35 @@ func LoadRestRoutes(dic *di.Container) *mux.Router {
 	e.HandleFunc(
 		"/"+SCRUB,
 		func(w http.ResponseWriter, r *http.Request) {
-			scrubHandler(w, r, bootstrapContainer.LoggingClientFrom(dic.Get), container.DBClientFrom(dic.Get))
+			scrubHandler(
+				w,
+				r,
+				bootstrapContainer.LoggingClientFrom(dic.Get), container.DBClientFrom(dic.Get),
```
michaelestrin: One parameter per line.

Comment here; occurs elsewhere.

internal/core/data/router.go
```diff
@@ -109,19 +111,35 @@ func LoadRestRoutes(dic *di.Container) *mux.Router {
...
 	e.HandleFunc(
 		"/"+SCRUBALL,
 		func(w http.ResponseWriter, r *http.Request) {
-			scrubAllHandler(w, r, bootstrapContainer.LoggingClientFrom(dic.Get), container.DBClientFrom(dic.Get))
+			scrubAllHandler(
+				w,
+				r,
+				bootstrapContainer.LoggingClientFrom(dic.Get), container.DBClientFrom(dic.Get),
+				dataContainer.HttpErrorHandlerFrom(dic.Get))
+
```
michaelestrin: Empty line should be removed (https://github.com/edgexfoundry/edgex-go/pull/2241)

Commented here; occurs elsewhere.

internal/core/data/router_test.go
```diff
@@ -179,10 +179,9 @@ func TestRestValueDescriptorsUsageHandler(t *testing.T) {
 			Configuration = &ConfigurationStruct{Service: bootstrapConfig.ServiceInfo{MaxResultCount: 10}}
 			defer func() { Configuration = &ConfigurationStruct{} }()
 
-			httpErrorHandler = errorconcept.NewErrorHandler(logger.NewMockClient())
 			Configuration.Service = tt.config
 			rr := httptest.NewRecorder()
-			restValueDescriptorsUsageHandler(rr, tt.request, logger.NewMockClient(), tt.dbMock)
+			restValueDescriptorsUsageHandler(rr, tt.request, logger.NewMockClient(), tt.dbMock, errorconcept.NewErrorHandler(logger.NewMockClient()))
```
michaelestrin: Use a single logger mock.

AnthonyMBonafide:
Sorry, forgot to address the PR comments from other PR.

AnthonyMBonafide:
Addressed the code review comments for this PR and for #2241

AnthonyMBonafide:
Recheck

michaelestrin:
recheck

michaelestrin:
Build appears stuck.  @lranjbar

jamesrgregg:
recheck

MightyNerdEric:
recheck

PR #2246, opened 2019-12-17
Feature/issue 2031 coredata remove configuration global

### LINKED ISSUE #2031: Data: refactor to remove Configuration global variable

Depends on https://github.com/edgexfoundry/edgex-go/issues/2030

AnthonyMBonafide:
Fix #2031 

Remove the 'Configuration' global variable from coredata and instead use
the DI container to pass a reference to which functions/methods require.
   
Update REST handlers and loadRoutes function to refer to the
configuration provided in the DI container
    
Update tests to work with the changes to production code.
    
Fix test which was incorrectly performing a DeepEqual rather than a type
verification check to ensure the factory was returning the correct type
 of instance.
    
### DIFF
--- a/cmd/core-data/toml_test.go
+++ b/cmd/core-data/toml_test.go
@@ -10,12 +10,12 @@ package main
 import (
 	"testing"
 
-	"github.com/edgexfoundry/edgex-go/internal/core/data"
+	dataConfig "github.com/edgexfoundry/edgex-go/internal/core/data/config"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/config"
 )
 
 func TestToml(t *testing.T) {
-	configuration := &data.ConfigurationStruct{}
+	configuration := &dataConfig.ConfigurationStruct{}
 	if err := config.VerifyTomlFiles(configuration); err != nil {
 		t.Fatalf("%v", err)
 	}
--- a/internal/core/data/event_test.go
+++ b/internal/core/data/event_test.go
@@ -21,6 +21,7 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/edgexfoundry/edgex-go/internal/core/data/config"
 	"github.com/edgexfoundry/edgex-go/internal/core/data/errors"
 	dbMock "github.com/edgexfoundry/edgex-go/internal/core/data/interfaces/mocks"
 	dataMocks "github.com/edgexfoundry/edgex-go/internal/core/data/mocks"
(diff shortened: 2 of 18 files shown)


internal/core/data/event_test.go
```diff
@@ -55,7 +56,7 @@ func TestCountByDevice(t *testing.T) {
 	dbClientMock := &dbMock.DBClient{}
 	dbClientMock.On("EventCountByDeviceId", mock.Anything).Return(2, nil)
 
-	count, err := countEventsByDevice(testEvent.Device, context.Background(), dbClientMock, newMockDeviceClient())
+	count, err := countEventsByDevice(testEvent.Device, context.Background(), dbClientMock, newMockDeviceClient(), &config.ConfigurationStruct{})
```
michaelestrin: Extends beyond 120 characters;  wrap (one parameter per line).

Commented here; occurs elsewhere.

internal/core/data/io.go
```diff
@@ -54,10 +55,12 @@ func (jsonReader) Read(reader io.Reader, ctx *context.Context) (models.Event, er
 }
 
 // cborReader handles unmarshaling of a CBOR request body payload
-type cborReader struct{}
+type cborReader struct {
+	configuration *config.ConfigurationStruct
+}
```
michaelestrin: I struggled with whether or not to provide a factory function to explicitly populate and return a cborReader instance.  On one hand, "consistency."  On the other, this struct is package private.  Then again, the package currently contains pretty much the entirety of core-data's implementation.  

I'd err on the side of consistency and recommend you create a factory function that takes the explicit dependencies, populates a cborReader instance with them, and returns the cborReader instance.

internal/core/data/reading_test.go
```diff
@@ -28,17 +30,22 @@ func newReadingsMockDB() interfaces.DBClient {
 
 func TestGetAllReadings(t *testing.T) {
 	reset()
-	Configuration.Service.MaxResultCount = 5
-	_, err := getAllReadings(logger.NewMockClient(), newReadingsMockDB())
+	_, err := getAllReadings(logger.NewMockClient(), newReadingsMockDB(), &config2.ConfigurationStruct{
+		Service: config.ServiceInfo{
+			MaxResultCount: 5,
+		},
+	})
```
michaelestrin: For consistency's sake, break an entire folded line; i.e.

```
	_, err := getAllReadings(
		logger.NewMockClient(),
		newReadingsMockDB(),
		&config2.ConfigurationStruct{
			Service: config.ServiceInfo{
				MaxResultCount: 5,
			},
		})
```

internal/core/data/io.go
AnthonyMBonafide: There is a factory function for creating Reader instances:

https://github.com/edgexfoundry/edgex-go/blob/41e418c1fbeacadce475f5c89f03f5ec7ee74cf7/internal/core/data/io.go#L91-L102

Are you suggesting making another factory method?

michaelestrin: Reader is the contract, I think.  cborReader is an implementation of that contract.  What I'm after is avoiding creating a cborReader instance directly as a struct.  If I recall, we're doing that somewhere in the tests.

cmd/core-data/toml_test.go
```diff
@@ -10,12 +10,12 @@ package main
 import (
 	"testing"
 
-	"github.com/edgexfoundry/edgex-go/internal/core/data"
+	config2 "github.com/edgexfoundry/edgex-go/internal/core/data/config"
```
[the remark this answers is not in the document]
michaelestrin: You still have this -- see data/reading_test.go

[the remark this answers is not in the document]
AnthonyMBonafide: Sorry, I missed that.

PR #2254, opened 2019-12-29
fix:not delete schedule completely

worldmaomao:
Fix the issue that : not delete the keys of scheduler action completely in redis. 
Avoid the memory leak

### DIFF
--- a/internal/pkg/db/redis/models/interval_action.go
+++ b/internal/pkg/db/redis/models/interval_action.go
@@ -38,7 +38,6 @@ func NewIntervalAction(from contract.IntervalAction) (ia IntervalAction) {
 }
 
 func (ia IntervalAction) Add() (cmds []DbCommand) {
-	cmds = make([]DbCommand, len(intervalActionKeys))
 	for _, key := range intervalActionKeys {
 		switch key {
 		case IntervalActionKey:
@@ -51,12 +50,11 @@ func (ia IntervalAction) Add() (cmds []DbCommand) {
 			cmds = append(cmds, DbCommand{Command: "SADD", Hash: key + ":" + ia.Target, Key: ia.ID})
 		}
 	}
-	return cmds
+	return
 }
 
 func (ia IntervalAction) Remove() (cmds []DbCommand) {
-	cmds = make([]DbCommand, len(intervalActionKeys))
-	for _, key := range intervalKeys {
+	for _, key := range intervalActionKeys {
 		switch key {
 		case IntervalActionKey:
 			cmds = append(cmds, DbCommand{Command: "ZREM", Hash: key, Key: ia.ID})
--- a/internal/pkg/db/redis/queries.go
+++ b/internal/pkg/db/redis/queries.go
@@ -242,7 +242,9 @@ func deleteObject(remover models.Remover, id string, conn redis.Conn) {
 	for _, cmd := range remover.Remove() {
 		switch cmd.Command {
 		case "ZREM":
+			_ = conn.Send(cmd.Command, cmd.Hash, cmd.Key)
 		case "SREM":
+			_ = conn.Send(cmd.Command, cmd.Hash, cmd.Key)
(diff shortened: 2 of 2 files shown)


internal/pkg/db/redis/queries.go
```diff
@@ -242,7 +242,9 @@ func deleteObject(remover models.Remover, id string, conn redis.Conn) {
 	for _, cmd := range remover.Remove() {
 		switch cmd.Command {
 		case "ZREM":
+			_ = conn.Send(cmd.Command, cmd.Hash, cmd.Key)
 		case "SREM":
+			_ = conn.Send(cmd.Command, cmd.Hash, cmd.Key)
 		case "HDEL":
 			_ = conn.Send(cmd.Command, cmd.Hash, cmd.Key)
```
michaelestrin: Why duplicate these lines?  Why not `fallthrough`?

internal/pkg/db/redis/models/interval_action.go
```diff
@@ -38,7 +38,6 @@ func NewIntervalAction(from contract.IntervalAction) (ia IntervalAction) {
 }
 
 func (ia IntervalAction) Add() (cmds []DbCommand) {
-	cmds = make([]DbCommand, len(intervalActionKeys))
```
michaelestrin: Why did you remove the slice initialization (it's explicitly initializing the slice for the proper size)?

internal/pkg/db/redis/models/interval_action.go
```diff
@@ -51,12 +50,11 @@ func (ia IntervalAction) Add() (cmds []DbCommand) {
 			cmds = append(cmds, DbCommand{Command: "SADD", Hash: key + ":" + ia.Target, Key: ia.ID})
 		}
 	}
-	return cmds
+	return
 }
 
 func (ia IntervalAction) Remove() (cmds []DbCommand) {
-	cmds = make([]DbCommand, len(intervalActionKeys))
```
michaelestrin: Why did you remove the slice initialization (it's explicitly initializing the slice for the proper size)?

internal/pkg/db/redis/queries.go
worldmaomao: The original code is  only available for  redis command "HDEL".  
Command "SREM" and "SREM" is not available and will not send to redis.

internal/pkg/db/redis/models/interval_action.go
worldmaomao: If length of intervalActionKeys is 3, after slice initialization  there are 3 empty DbCommand structs. According to the below code, it will append more DbCommand struct  which will not replace the empty struct in the slice.
111WX20191231-134922@2x

internal/pkg/db/redis/models/interval_action.go
worldmaomao: If length of intervalActionKeys is 3, after slice initialization  there are 3 empty DbCommand structs. According to the below code, it will append more DbCommand struct  which will not replace the empty struct in the slice.
111WX20191231-134922@2x

internal/pkg/db/redis/queries.go
worldmaomao: That's a good idea.   
I will fix it.

internal/pkg/db/redis/models/interval_action.go
michaelestrin: You are correct.  However, I wonder if a more efficient solution would be to initialize cmds using make once (since we know its required size in advance) and subsequently assign its individual elements directly (as opposed to using append).

internal/pkg/db/redis/models/interval_action.go
michaelestrin: You are correct.  However, I wonder if a more efficient solution would be to initialize cmds using make once (since we know its required size in advance) and subsequently assign its individual elements directly (as opposed to using append).

PR #2257, opened 2020-01-02
Explicit Close() when security-file-token provider writes tokens

### LINKED ISSUE #2248: Refactor security-file-token-provider's provider.go to perform explicit closes
Pertains to this comment
https://github.com/edgexfoundry/edgex-go/pull/2202#pullrequestreview-333460337

The ideal logic we want here is
```
writeCloser, err := p.fileOpener.OpenFileWriter(outputTokenFilename, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, os.FileMode(0600))
// err check here
// chmod() here
if err
  close()
  return err
// chown here
if err
  close()
  return err
// create encoder/writer
if encoder nil
  close()
  return err
// encode
if err
  close
  return err
close()
```

This is to adopt best-known-practice to avoid deferred closes when writing files.

bnevis-i:
This is a followup to a previous code review to clean up
the logic to explicitly Close() a file being written that
contains the authentication token to the secret store.
Deferred Close()s are removed and replaced with
explicit closes before leaving the function and an error-
reporting Close() at the end of each loop iteration.

Fixes #2248

### DIFF
--- a/internal/security/fileprovider/provider.go
+++ b/internal/security/fileprovider/provider.go
@@ -153,11 +153,7 @@ func (p *fileTokenProvider) Run() error {
 			p.logger.Error(fmt.Sprintf("failed open token file for writing %s: %s", outputTokenFilename, err.Error()))
 			return err
 		}
-		defer func() {
-			if writeCloser != nil {
-				writeCloser.Close()
-			}
-		}()
+		// writeCloser is writable file -- explicitly close() to ensure we catch errors writing to it
 
 		permissionable, ok := writeCloser.(permissionable)
 		if ok {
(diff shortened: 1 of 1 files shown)


bnevis-i:
This PR addresses the issues raised in https://github.com/edgexfoundry/edgex-go/pull/2202

internal/security/fileprovider/provider.go
```diff
@@ -153,22 +153,20 @@ func (p *fileTokenProvider) Run() error {
...
-		}()
+		// writeCloser is writable file -- explicitly close() to ensure we catch errors writing to it
 
 		permissionable, ok := writeCloser.(permissionable)
 		if ok {
 			if serviceConfig.FilePermissions != nil &&
 				(serviceConfig.FilePermissions).ModeOctal != nil {
 				mode, err := strconv.ParseInt(*(serviceConfig.FilePermissions).ModeOctal, 8, 32)
 				if err != nil {
+					writeCloser.Close()
```
michaelestrin: Recommend explicitly swallowing the error return value; i.e. `_ = writeCloser.Close()`

Mentioned once, occurs elsewhere.

bnevis-i: Implemented as suggested 5X.

bnevis-i:
Rebased with fixes.

Issue #2277, opened 2020-01-13
Add Skeletal Support for API V2 Implementation


michaelestrin:
Dependent on https://github.com/edgexfoundry/edgex-go/issues/2273

michaelestrin:
Dependent on https://github.com/edgexfoundry/go-mod-bootstrap/issues/29

PR #2307, opened 2020-01-17
Add In-Memory DB Implementation for Addressable

### LINKED ISSUE #2286: Implement db/memory/addressables.go behavior
Write an in-memory implementation of the `DBClient` interface for the file mentioned in this issue's title.  Include unit tests that cover the implementation.  

This in-memory persistence implementation will be used for Golang-based acceptance tests.

Related to https://github.com/edgexfoundry/edgex-go/issues/2277 and https://github.com/edgexfoundry/edgex-go/issues/2273.

AnthonyMBonafide:
Fix #2286

Implement previously noop functionality for the memory.Client's methods
for dealing with persisting contract.Addressable types, and
memory.Client constructor.

Update memory.Client methods with noop implementation to panic rather
than return empty data. The reason is that the tests are using the newly
implemented constructor to create instances of the client for easier
maintainability. Previously, we were panicking at the constructor to
protect users from creating an instance of the memory.Client. Now we are
panicking at when an unimplemented method is invoked.

Add tests covering newly added implementation.

### DIFF
--- a/internal/pkg/db/memory/addressables_test.go
+++ b/internal/pkg/db/memory/addressables_test.go
@@ -1,0 +1,737 @@
+/********************************************************************************
+ *  Copyright 2020 Dell Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package memory
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
+
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+
+	"github.com/google/uuid"
+)
+
+// allTestAddressables a slice containing all the pre-constructed addressables for testing.
+var allTestAddressables = []contract.Addressable{addressable1, addressable2, addressable3}
+
+// addressable1 is a pre-constructed addressable with unique properties which does not match any other addressable when
+//  invoking functions which can return multiple addressables such as GetAddressablesByPort
+var addressable1 = contract.Addressable{
+	Timestamps: contract.Timestamps{
+		Created:  123,
+		Modified: 456,
+		Origin:   789,
+	},
+	// This is a valid UUID which allows us to ensure that the ID used is not replaced.
+	Id:         "6d40520f-035b-4be9-bacc-152d4f4f37a9",
+	Name:       "FirstAddressable",
+	Protocol:   "tcp",
+	HTTPMethod: "POST",
+	Address:    "0.0.0.0",
+	Port:       8000,
+	Path:       "/",
+	Publisher:  "Mushroom Kingdom",
+	User:       "Yoshi",
+	Password:   "MarioIsGettingHeavy",
+	Topic:      "dinosaur-things",
+}
+
+// addressable2 is a pre-constructed addressable with non-unique properties. The properties which can lead to multiple
+// results when invoking the client, such as GetAddressablesByPort, have matching properties with addressable3. This
+// helps with testing situations where multiple addressables are returned
+var addressable2 = contract.Addressable{
+	Timestamps: contract.Timestamps{
+		Created:  789,
+		Modified: 456,
+		Origin:   123,
+	},
+	// This is a valid UUID which allows us to ensure that the ID used is not replaced.
+	Id:         "1d874125-df04-4718-bd81-b550019b74aa",
+	Name:       "SecondAddressable",
+	Protocol:   "http",
+	HTTPMethod: "GET",
+	Address:    "127.0.0.1",
+	Port:       8080,
+	Path:       "/home",
+	Publisher:  "Nintendo",
+	User:       "Link the hero time",
+	Password:   "RescueZelda",
+	Topic:      "hyrule-events",
+}
+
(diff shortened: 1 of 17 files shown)


internal/pkg/db/memory/factory.go
```diff
@@ -14,10 +14,49 @@
...
+	"sync"
 
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+)
+
+// UnimplementedMethodPanicMessage is the common panic message used for unimplemented methods of memory.Client.
+const UnimplementedMethodPanicMessage = "This method has not been implemented."
+
+// addressableMemoryStore encapsulates the data needed to store addressables.
+type addressableMemoryStore struct {
```
michaelestrin: Eliminate.

Add global mutex to guard access across entire db/memory implementation.  We shouldn't need a reentrant implementation.

Add addressableMap to Client struct.

Work out how to reuse NewClient in CloseSession; you'll need to capture your mutex before calling NewClient and ensure you unlock the older mutex instance.

internal/pkg/db/memory/addressables_test.go
```diff
@@ -0,0 +1,737 @@
...
+
+	"github.com/google/uuid"
+)
+
+// allTestAddressables a slice containing all the pre-constructed addressables for testing.
+var allTestAddressables = []contract.Addressable{addressable1, addressable2, addressable3}
+
+// addressable1 is a pre-constructed addressable with unique properties which does not match any other addressable when
+//  invoking functions which can return multiple addressables such as GetAddressablesByPort
+var addressable1 = contract.Addressable{
```
michaelestrin: Commented once; occurs elsewhere.  

Don't use global variables for your tests.  Think about creating a single factory function that returns a structure with random data to ensure variance across calls.

internal/pkg/db/memory/addressables_test.go
```diff
@@ -0,0 +1,737 @@
...
+	Port:       8080,
+	Path:       "/castle",
+	Publisher:  "Nintendo",
+	User:       "Gannon",
+	Password:   "GetTriforce",
+	Topic:      "hyrule-events",
+}
+
+func TestClient_AddAddressableUUID(t *testing.T) {
+	client := NewClient()
```
michaelestrin: Create a private `newSUT` method that returns an instance of the system under test -- in this case, the result of `NewClient`.  Call `newSUT` from each of your tests.

internal/pkg/db/memory/addressables_test.go
```diff
@@ -0,0 +1,737 @@
...
+			"No Addressables",
+			[]contract.Addressable{},
+
+			[]contract.Addressable{},
+			false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c, err := createAndPopulateClient(tt.prePopulatedAddressables)
```
michaelestrin: Commented once; occurs elsewhere.

Separate `createAndPopulateClient` into `newSUT` and `populateClient`.

internal/pkg/db/memory/addressables_test.go
```diff
@@ -0,0 +1,737 @@
...
+			if err != nil {
+				t.Errorf("Unexpected error during test setup: %v", err)
+				return
+			}
+
+			got, err := c.GetAddressables()
+			if (err != nil) != tt.wantErr {
```
michaelestrin: Commented once; occurs elsewhere.

Spatially separate test setup, test run, and follow-on assertions into three separate sections; empty lines in-between.

internal/pkg/db/memory/addressables_test.go
```diff
@@ -0,0 +1,737 @@
...
+			if !tt.wantErr && !verifySliceEqualIgnoreOrder(got, tt.want) {
+				t.Errorf("GetAddressables() got = %v, want %v", got, tt.want)
+
+			}
+		})
+	}
+}
+
+func TestClient_UpdateAddressable(t *testing.T) {
+
```
michaelestrin: Commented once; occurs elsewhere.

Remove blank lines between function/method definition and first line of code (unless function/method definition spans multiple lines).

internal/pkg/db/memory/addressables_test.go
```diff
@@ -0,0 +1,737 @@
...
+func TestClient_GetAddressables(t *testing.T) {
+
+	tests := []struct {
+		name                     string
+		prePopulatedAddressables []contract.Addressable
+		want                     []contract.Addressable
+		wantErr                  bool
+	}{
+		{
+			"One Addressable",
```
michaelestrin: Commented once, occurs elsewhere.

For readability and consistency, add field names to each statically defined instance.

internal/pkg/db/memory/addressables_test.go
```diff
@@ -0,0 +1,737 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return client, nil
+}
+
+// verifySliceEqualIgnoreOrder verfies the contents of two slices of Addressables are equal ignoring order.
+func verifySliceEqualIgnoreOrder(a, b []contract.Addressable) bool {
```
michaelestrin: I just wrote something roughly equivalent (but generic) for #2277.  Think about how you can write this generically and do so (so it can be reused by the other memory/db types' tests.  Generic version belongs in its own file.

internal/pkg/db/memory/factory.go
AnthonyMBonafide: I originally avoided that approach because I didn't like the idea of the whole client being blocked and possibly preventing unrelated operations from processing. I am fine if we go down that path, I just wanted to bring up the performance impact. Another way would be to create another layer which abstracts locking and ensures access to data which is interconnected to be guarded. That would be more difficult, but much faster.

michaelestrin: I'm aware of the performance trade-off.  This implementation is specific to acceptance tests only -- for that I'll trade performance for simplicity.

AnthonyMBonafide: Ahhh! I thought this would possibly be used in production, under some odd circumstances. I misunderstood. Thank you for the clarification.

AnthonyMBonafide:
Closing this issue since we have decided to do away with the in memory implementation. See #2340 for more details

PR #2311, opened 2020-01-19
#2310 Mongo core reading's created is not work very well (master)

### LINKED ISSUE #2310: Mongo core reading's created is not work very well (fuji, master)
# 🐞 Bug Report

### Affected Services

 The issue is located in: Data service

### Is this a regression?

 NO!!

redis version's `Created` time is created from `e.Created`, so all Readings have the same Created Time.
But, the mongo version's Created time is created every Reading, so some time(millisecond) the created time is not the same.
I use created time to aggregate readings, it work not ok.
image

mongo
https://github.com/edgexfoundry/edgex-go/blob/d939fa2a280896df5d7ca92c81329f4401ab2857/internal/pkg/db/mongo/data.go#L56-L78

redis
https://github.com/edgexfoundry/edgex-go/blob/d939fa2a280896df5d7ca92c81329f4401ab2857/internal/pkg/db/redis/data.go#L1066-L1082

## 🔥 Exception or Error
No Error Here or Exception

## 🌍  Your Environment
**Deployment Environment:**

**EdgeX Version:**
fuji

**Anything else relevant?**
redis is work ok.

kaisawind:
Fixes #2310 

create timestamp out of for `e.Readings`


tsconn23:
@kaisawind All PRs require an issue to be associated with them. If you believe there's a bug here, please create an issue using the "Bug Report" template and then create a new PR referencing that issue in your commit message.

tsconn23:
Re-opening after commit message above was edited to include issue number

michaelestrin:
@kaisawind - issue specifies this as a problem with Fuji.  Merge request targets master.  We'll need a separate PR to target Fuji.

@AnthonyMBonafide - please don't merge master without having a separate Fuji PR to review/merge.

internal/pkg/db/mongo/data.go
```diff
@@ -55,14 +55,16 @@ func (mc MongoClient) AddEvent(e correlation.Event) (string, error) {
...
+		timestamp := db.MakeTimestamp()
 		for i, reading := range e.Readings {
 			var r models.Reading
 			id, err := r.FromContract(reading)
 			if err != nil {
 				return "", err
 			}
 
-			r.TimestampForAdd()
+			r.Created = timestamp
```
AnthonyMBonafide: The Redis implementation checks to see if the value is provided before generating a timestamp and setting it. Should we be doing that here for consistency? See the code block here:

https://github.com/edgexfoundry/edgex-go/blob/f1941eeb15de8730bc21d10e7aa8954d0a8f4e76/internal/pkg/db/redis/data.go#L1035-L1037

kaisawind: The mongo implementation generate an event's timestamp after readings is set.So no need to check it.
https://github.com/edgexfoundry/edgex-go/blob/7ca3df59c01c7f56d3c83a840c5995cdb7718c20/internal/pkg/db/mongo/data.go#L80-L87

AnthonyMBonafide: I apologize, I referred to different entities in my last comment and even confused myself when I went back to look at it. Allow me to start over and provide concrete examples for clarification.

When an event is created via the HTTP API, the call can optionally contain a `created` field. If it is not provided in the HTTP body or is `0` the `json.Marshal` function will create an `event` with the `created` field of `0`. When the Mongo or Redis gets passed an unmarshalled `event` the `AddEvent` logic for each implementation is handling the `created` field differently, resulting in an inconsistency.

The Mongo implementation is calling this function right before persisting to the database:
https://github.com/edgexfoundry/edgex-go/blob/eec8e1f3ce6d9a9c8c1e1098fbe9df49c355ee97/internal/pkg/db/mongo/models/event.go#L97-L100

Which is always setting the `created` value to the current time in milliseconds regardless if the original content has already been provided with a `created` value.

On the other hand, the Redis implementation is performing this check and only updating the `created` field if the original value is `0`, which means the caller either did not provide it or it was explicitly set to `0`. This logic is calling the same method to update the created time, but it is doing it only if the value was not set:
https://github.com/edgexfoundry/edgex-go/blob/f1941eeb15de8730bc21d10e7aa8954d0a8f4e76/internal/pkg/db/redis/data.go#L1035-L1037

I was able to verify this by starting `core-data` using a Mongo database and provided this request to create an event, notice that the `created` field is provided:

```json
{  
   "origin":1471806386919,
   "device":"cartachometer",
   "created": 62388,
   "readings":[  
      {  
        "origin":1471806386919,
		"device":"cartachometer",
        "name":"luminousIntensity",
        "value":"40"
      }
   ]
}
```
And then looked up the event with the ID that was returned from the creation API and got, notice the original `created` value was ignored and over written:
```json
{
    "id": "1d60e893-3021-4948-8e86-502c8638a022",
    "device": "cartachometer",
    "created": 1579877020172,
    "modified": 1579877020172,
    "origin": 1471806386919,
    "readings": [
        {
            "id": "ef33dacc-bd76-44a6-ad49-078598b19f2d",
            "created": 1579877020169,
            "origin": 1471806386919,
            "modified": 1579877020169,
            "device": "cartachometer",
            "name": "luminousIntensity",
            "value": "40"
        }
    ]
}
```

I then did the same with `core-data` using Redis. My creation request was:
```json
{  
   "origin":1471806386919,
   "device":"cartachometer",
   "created": 62388,
   "readings":[  
      {  
        "origin":1471806386919,
		"device":"cartachometer",
        "name":"luminousIntensity",
        "value":"40"
      }
   ]
}
```

And my lookup for the newly created event response was, notice that the `created` value provided was persisted:
```json
{
    "id": "d20eea1d-efcf-491f-98b9-15f8409ece51",
    "device": "cartachometer",
    "created": 62388,
    "origin": 1471806386919,
    "readings": [
        {
            "id": "37feee2c-fb45-4990-bde8-ecf76533db79",
            "created": 62388,
            "origin": 1471806386919,
            "device": "cartachometer",
            "name": "luminousIntensity",
            "value": "40"
        }
    ]
}
```

You can see that the data for the `created` field that was persisted in the database differs depending on which database implementation you are using.

So my suggestion is to add logic to the Mongo implementation so that it behaves like the Redis implementation with regards to adding an event. I am assuming that the Redis implementation is correct since it makes sense to let the user provide a `created` value if they are storing data that has been created previously. For example, if the system is down and data gets backed up, we should honor the original created times. That is just my thought.

internal/pkg/db/mongo/data.go
```diff
@@ -52,6 +52,15 @@ func (mc MongoClient) AddEvent(e correlation.Event) (string, error) {
...
 	defer s.Close()
 
+	if e.Created == 0 {
+		e.Created = db.MakeTimestamp()
+		e.Modified = e.Created
+	}
+
+	if e.Modified == 0 {
+		e.Modified = db.MakeTimestamp()
+	}
```
AnthonyMBonafide: Should we be setting the `modified` value or leaving it to whatever the caller has specified? I only ask because the Redis implementation does not seem to be doing this, and it would be nice if this is consistent. If we do want to keep this logic, then I think we may need to update this. With this implementation we will overwrite the `modified` value if the `created` value is not set, regardless if the `modified` value has been provided. I think line `57` could be removed. and line `61` could simply be:
```go
e.modified = e.created
```

that way we guarantee that the `created` and `modified` times are the same in the case where neither of the values are provided and are generated by the system. I think that would make more sense to a user who is looking at the data rather than possibly having two different times.

internal/pkg/db/mongo/data.go
```diff
@@ -62,7 +71,8 @@ func (mc MongoClient) AddEvent(e correlation.Event) (string, error) {
 				return "", err
 			}
 
-			r.TimestampForAdd()
```
AnthonyMBonafide: Can we revert this change? This function handles the setting the times and we can leverage that in a central location and not duplicate the logic.

internal/pkg/db/mongo/data.go
kaisawind: Yes.it's great.

internal/pkg/db/mongo/data.go
kaisawind: No. This PR is to fix it. Read more #2310.
The creation of time is in a for loop, which causes the creation time to be different in milliseconds.

AnthonyMBonafide: You are correct. Ignore my previous comment.

Issue #2321, opened 2020-01-24
Update OpenAPI Docs for V2 API -- Ping, Metrics, Version, Configuration


michaelestrin:
Support for `/ping`, `/metrics`, `/version`, and `/config` endpoints is currently specified for the HTTP `GET` method.  That method is cachable.  These are not resource retrievals that should be cached; they are commands.  Update to specify support via the HTTP `POST` method.

Add a corresponding request DTO for each of these endpoints.

PR #2322, opened 2020-01-24
Update OpenAPI Docs for V2 API -- Ping, Metrics, Version, Configuration

michaelestrin:
Support for /ping, /metrics, /version, and /config endpoints is currently
specified for the HTTP GET method. That method is cachable. These are not
resource retrievals that should be cached; they are commands. Update to
specify support via the HTTP POST method.

Add a corresponding request DTO for each of these endpoints.

Update to support single and array of use-case requests and responses 
for these endpoints.

Fixes: https://github.com/edgexfoundry/edgex-go/issues/2321

### DIFF
--- a/api/openapi/v2/core-command.yaml
+++ b/api/openapi/v2/core-command.yaml
@@ -95,7 +95,12 @@ components:
         - $ref: '#/components/schemas/BaseResponse'
       description: "A response type for returning a generic error to the caller."
       type: object
-    GetConfigurationResponse:
+    ConfigRequest:
+      description: "A request associated with the /config endpoint."
+      allOf:
+        - $ref: '#/components/schemas/BaseRequest'
+      type: object
+    ConfigResponse:
       allOf:
         - $ref: '#/components/schemas/BaseResponse'
       description: "Provides a response containing the configuration for the targeted service."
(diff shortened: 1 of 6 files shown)


api/openapi/v2/core-command.yaml
```diff
@@ -347,8 +367,16 @@ paths:
...
   /config:
-    get:
+    post:
       summary: "Returns the current configuration of the service."
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: array
```
tsconn23: I don't think we need an array of requests here. Probably will only service one request at a time and no array of responses is currently returned.

Comment applies to /config, /ping, /version and /metrics endpoints on all services.

michaelestrin:
Updated to be dependent on #2323.

api/openapi/v2/core-command.yaml
michaelestrin: Any use-case endpoint takes either a single use-case request or an array of use-case requests (all of the same type).  I believe both the request and response sections of the specification need to be modified to reflect this (`anyOf` per conversation [here](https://github.com/edgexfoundry/edgex-go/pull/2324#discussion_r371374159)).

tsconn23: We discussed offline rationale for changing endpoints from GET to POST (these endpoints do not uniquely identify a specific resource and we don't want their output to be cacheable). In changing the endpoints to POST, the above was done for consistency's sake.

Issue #2323, opened 2020-01-24
Update OpenAPI Docs for V2 API -- Generic Errors


michaelestrin:
Remove ErrorEnvelope (batch endpoint DTO -- wraps use-case DTO).

Add ErrorResponse (use-case DTO).

Add ErrorResponse to all relevant use-case-request paths.

Related conversation:
![image](https://user-images.githubusercontent.com/464327/73100384-0f00cc00-3eb3-11ea-9f4b-0cb1238875a0.png)

PR #2324, opened 2020-01-24
Update OpenAPI Docs for V2 API -- Generic Errors

michaelestrin:
Remove ErrorEnvelope (batch endpoint DTO -- wraps use-case DTO).

Add ErrorResponse (use-case DTO).

Add ErrorResponse to all relevant use-case-request paths.

Fixes: https://github.com/edgexfoundry/edgex-go/issues/2323

### DIFF
--- a/api/openapi/v2/core-command.yaml
+++ b/api/openapi/v2/core-command.yaml
@@ -90,21 +90,11 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/Command'
-    ErrorEnvelope:
+    ErrorResponse:
       allOf:
-        - $ref: '#/components/schemas/ResponseEnvelope'
-      description: "A type utilized by the /batch endpoint for returning an error applicable to an individual request."
+        - $ref: '#/components/schemas/BaseResponse'
+      description: "A response type for returning a generic error to the caller."
       type: object
-      properties:
-        message:
-          description: "A field that can contain a free-form message, such as an error message."
-          type: string
-        statusCode:
-          description: "A numeric code signifying the operational status of the response."
-          type: integer
-      required:
-        - message
-        - statusCode
     GetConfigurationResponse:
       allOf:
         - $ref: '#/components/schemas/BaseResponse'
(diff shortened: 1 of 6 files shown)


api/openapi/v2/core-command.yaml
```diff
@@ -345,7 +337,9 @@ paths:
               schema:
                 type: array
                 items:
-                  $ref: '#/components/schemas/IssueCommandResponse'
+                  oneOf:
```
tsconn23: I believe this should be `allOf`. Referring to docs [here](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/)

```
The following JSON object is valid against both schemas, so the response body is incorrect – it should be valid against only one of the schemas, since we are using the oneOf keyword.
```
Both response schemas for this endpoint are valid if the `IssueCommandResponse` is returned
```
[
  {
    "requestId": "e6e8a2f4-eb14-4649-9e2b-175247911369",
    "statusCode": 0,
    "message": "string"
  },
  {
    "requestId": "e6e8a2f4-eb14-4649-9e2b-175247911369",
    "statusCode": 0,
    "message": "string",
    "device": "string",
    "command": "string"
  }
]
```
They both share the requestId, statusCode and message properties. If `oneOf` is used, returning `IssueCommandResponse` will be invalid since it fulfills the ErrorResponse contract as well.

The comment probably applies for other endpoints returning multiple responses (207)

michaelestrin: Acknowledging our (@tsconn23) out-of-band conversation on this topic yesterday.

I believe `oneOf` is appropriate.

`allOf` validates the value against _all_ of the subschemas.  It takes an array of object definitions that are used for independent validation but together compose a single object.  

Using the example from yesterday's conversation, where either `IssueCommandResponse` or `ErrorResponse` are valid response schemas.  While `ErrorResponse` is a subset of `IssueCommandResponse`, this is coincidental only.  Further, an `IssueCommandResponse` would not logically validate against `ErrorResponse` -- it has required fields that `ErrorResponse` does not include.  Likewise, `ErrorResponse` would not validate against `IssueCommandResponse` as it would not include the latter's required fields.  As I understand it, `allOf` requires the response to satisfy _all_ of the defined schemas.

The difference between `oneOf` and `anyOf` is `oneOf` matches exactly one subschema, and `anyOf` can match one or more subschemas.  The latter is more permissive (i.e. "looser") than the former.  However, we're only returning one of the subschemas in the example scenario and they will -- by definition -- be distinctly different.  That is, we're only returning one.

Do you concur that `oneOf` is the appropriate qualifier?

tsconn23: Yep, sounds good.

PR #2332, opened 2020-02-03
Create a unit test for Update Device Service LastReported by ID.

### LINKED ISSUE #1737: Unit Test -- Core-Metadata: Create a unit test for Update Device Service LastReported by ID.
Create a unit test at the REST layer level (add to `internal/core/metadata/rest_deviceservice_test.go`).

akramtexas:
Fix #1737

### DIFF
--- a/internal/core/metadata/rest_deviceservice_test.go
+++ b/internal/core/metadata/rest_deviceservice_test.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"net/http"
 	"net/http/httptest"
+	"strconv"
 	"testing"
 
 	metadataConfig "github.com/edgexfoundry/edgex-go/internal/core/metadata/config"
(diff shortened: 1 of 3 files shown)


internal/core/metadata/rest_deviceservice_test.go
```diff
@@ -594,3 +691,20 @@ func createDeviceServiceRequestWithBody(
...
+	return dbMock
+}
+
+func newMockDBClientSecond() interfaces.DBClient {
+	dbMock := &mocks.DBClient{}
+	dbMock.On("GetDeviceServiceById", testDeviceServiceLastUpdatedId).Return(testDeviceServiceLastUpdated, nil)
+	dbMock.On("UpdateDeviceService", mock.Anything).Return(testError)
+
+	return dbMock
+}
```
AnthonyMBonafide: Can this be removed in favor of refactoring `createMockWithOutlines` to support mocking methods with varying number or return entities? It seems like we have outgrown the original function and it needs to be updated to support this. If we do want to keep these functions I think we should rename them better so it is clear what the difference is between them.

If we do refactor the `createMockWithOutlines` I think we can than 
 remove `TestUpdateServiceLastReportedByIdServiceUnavailable` in favor of a test table.

internal/core/metadata/rest_deviceservice_test.go
```diff
@@ -560,6 +576,87 @@ func TestUpdateAdminStateByName(t *testing.T) {
...
+			createMockWithOutlines([]mockOutline{
+				{"GetDeviceServiceById", testDeviceServiceLastUpdatedNotExistId, contract.DeviceService{}, db.ErrNotFound},
+			}),
+			http.StatusInternalServerError,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			rr := httptest.NewRecorder()
+			var loggerMock = logger.NewMockClient()
```
AnthonyMBonafide: For consistency, can we use the `:=` operator here. Our code base tends to favor `:=` over `var`. This also applies to other uses of `var` in these changes.

internal/core/metadata/rest_deviceservice_test.go
akramtexas: I like the suggestion and will pursue the implementation 👍

akramtexas:
@AnthonyMBonafide 
- Following up on your suggestion, I have refactored the func `createMockWithOutlines()` to support mocking methods with varying number of arguments and/or varying number of return entities. This resulted in updating about 50+ locations wherefrom `createMockWithOutlines()` is called. 
- Removed TestUpdateServiceLastReportedByIdServiceUnavailable() in favor of a table-driven test.

PR #2333, opened 2020-02-03
Add OpenAPI spec for System Agent API v2

### LINKED ISSUE #2315: Create OpenAPI Doc for V2 API Specification (System Management API)
Follow pattern and conventions established by https://github.com/edgexfoundry/edgex-go/issues/2308 to create an OpenAPI-based specification for system management's V2 API.

Add me as sole reviewer to your PR.

AnthonyMBonafide:
Fix #2315


AnthonyMBonafide:
I took some creative liberties here and  diverged from some of the things that were in the other API specs for V2. The biggest ones are response types which encapsulate all the information from the response. This allows for a cleaner API and more usable when dealing with responses programmatically. For example, having requests which returned an array of multiple types would require the caller to perform some logic when deserializing the response. The other major change was getting rid of the `207` response codes from the non-batch endpoints. To me this makes more sense as the `207` response code is meant for WebDav and I think it makes more sense to respond to the caller with a `200` indicating we were able to process the whole request and obtain all the information successfully. Then it is up to the user to inspect the details of the request.

I was also toying with the idea of having `metrics`, `config`, and `health` by default return information from all services and the user can optionally provide query parameters to filter further. I am not sure where the best place to capture that idea, either here or in a new issue.

AnthonyMBonafide:
After a discussion with @michaelestrin , I realized I was headed in the wrong direction. I will re-work this to better align with the vision of the other existing OpenAPI documents. Please ignore my previous comment.

AnthonyMBonafide:
I updated the spec to match the other service's APIs for the V2.

api/openapi/v2/system-agent.yaml
```diff
@@ -0,0 +1,625 @@
...
+        '400':
+          description: "Bad request"
+          headers:
+            X-Correlation-ID:
+              $ref: '#/components/headers/correlatedResponseHeader'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+    put:
```
michaelestrin: This should be POST.  This isn't an API over a normal resource (e.g. create, read, update, delete).  Rather, when you invoke this endpoint you're commanding the service to undertake an action.

api/openapi/v2/system-agent.yaml
```diff
@@ -0,0 +1,625 @@
...
+      description: "Response containing the health status for the targeted service(s)."
+      additionalProperties:
+        type: string
+    MetricsRequest:
+      allOf:
+        - $ref: '#/components/schemas/BaseRequest'
+      properties:
+        services:
+          description: "The services from which to obtain metrics."
+          type: array
```
michaelestrin: Since it'd be possible in v2 to send multiple use-case requests in a single transport request, I wonder if we still need to support requesting for more than one service in a use-case request.

api/openapi/v2/system-agent.yaml
```diff
@@ -0,0 +1,625 @@
...
+          type: string
+      required:
+        - config
+    HealthRequest:
+      allOf:
+        - $ref: '#/components/schemas/BaseRequest'
+      properties:
+        services:
+          description: "The services from which to obtain health information."
+          type: array
```
michaelestrin: Since it'd be possible in v2 to send multiple use-case requests in a single transport request, I wonder if we still need to support requesting for more than one service in a use-case request.

api/openapi/v2/system-agent.yaml
```diff
@@ -0,0 +1,625 @@
...
+        - $ref: '#/components/schemas/BaseRequest'
+      description: "An Operation for SMA processing."
+      properties:
+        action:
+          description: "The operation to execute on the service."
+          type: string
+          enum: [start, stop, restart]
+        services:
+          description: "The services which to apply the action."
+          type: array
```
michaelestrin: Since it'd be possible in v2 to send multiple use-case requests in a single transport request, I wonder if we still need to support requesting for more than one service in a use-case request.

api/openapi/v2/system-agent.yaml
```diff
@@ -0,0 +1,625 @@
...
+        - $ref: '#/components/schemas/BaseRequest'
+      type: object
+      properties:
+        service:
+          description: "The service for which to obtain metrics"
+          type: string
+          example: "edgex-core-data"
+        configuration:
+          description: "Service's configuration"
+          type: object
```
michaelestrin: Since it'd be possible in v2 to send multiple use-case requests in a single transport request, I wonder if we still need to support requesting for more than one service in a use-case request.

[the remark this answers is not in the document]
AnthonyMBonafide: I am a little confused. The API today only allows the caller to specify one key/value pair, but allows for multiple services to be provided in the URI. This OpenAPI representation aligns with the API as it is today and only allows for one key/value pair. As I am seeing it, the thing that is missing is that the endpoint does not allow the user to specify multiple services in a single use-case request, but that seems desirable since we would be allowing that capability through the 
 multiple use-case requests functionality to this endpoint since the request structure contains the targeted service and its desired configuration. Are my thoughts correct? I am pretty sure I am missing something here.

api/openapi/v2/system-agent.yaml
```diff
@@ -0,0 +1,625 @@
...
+            description: "The name of the service from which to obtain health information."
+            type: string
+            example: "edgex-core-data"
+    HealthResponse:
+      allOf:
+        - $ref: '#/components/schemas/BaseRequest'
+      type: object
+      description: "Response containing the health status for the targeted service(s)."
+      additionalProperties:
+        type: string
```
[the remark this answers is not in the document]
AnthonyMBonafide: The service currently returns a series of key/value pairs where the key is the service's name and the value is the service's health. Isn't that the delineation, or does that need to be contained in with a sub-structure?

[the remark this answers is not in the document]
michaelestrin: I'm looking for more structure than this:

image

Where it's not clear what exactly the `<*> string` being returned contains.

api/openapi/v2/system-agent.yaml
AnthonyMBonafide: If I make this a `POST` then I will have to change the endpoint for obtaining configurations, since that is currently a `POST` under the same URI. Should I make the other endpoint a `GET`? That makes the most sense, but does not align with the other endpoints in this spec which would also make sense as a `GET` like obtaining health and metrics.

api/openapi/v2/system-agent.yaml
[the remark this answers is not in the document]
AnthonyMBonafide: I do not see a way to achieve more structure with the OpenAPI spec with how the service is currently responding. The service is returning a dynamic amount of service names to health information(key/value) attributes. The only way I see how to represent this in the OpenAPI format is by using the Hashmap/Dictionary data type which results in what we already have. I could be missing something. 

Also, I am updating this response to only return one service and its health information, as per your other comment, which might simplify this a little more.

api/openapi/v2/system-agent.yaml
AnthonyMBonafide: No need to answer this question. I think it worked itself out.

AnthonyMBonafide:
Addressed code review comments, except for one unresolved comment which I was not clear what the suggested solution is.

api/openapi/v2/system-agent.yaml
```diff
@@ -0,0 +1,634 @@
...
+          example: "metrics"
+        service:
+          description: "The service assocaited with the metrics"
+          type: string
+          example: "edgex-core-data"
+        result:
+          description: "The metric information retireved from the targeted service"
+          type: object
+          additionalProperties:
+            type: object
```
michaelestrin: There is a standard response with a non-standard component.  Include the standard response; you can "object" the non-standard object.

image

(`raw` object is non-standard)

api/openapi/v2/system-agent.yaml
```diff
@@ -0,0 +1,634 @@
...
+      allOf:
+        - $ref: '#/components/schemas/BaseResponse'
+      description: "Provides a response containing the configuration for the targeted service."
+      type: object
+      properties:
+        config:
+          description: "A string-ified representation of the service's configuration. For purposes of this specification, a string has been used since configuration structure differs from service to service."
+          type: string
+      required:
+        - config
```
michaelestrin: image

Note structure, not string.

api/openapi/v2/system-agent.yaml
```diff
@@ -0,0 +1,634 @@
...
+          description: "Bad request"
+          headers:
+            X-Correlation-ID:
+              $ref: '#/components/headers/correlatedResponseHeader'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+  /config/{services}:
+    get:
```
michaelestrin: POST.  It's a command, not a resource retrieval.  You'll need to differentiate URL.

api/openapi/v2/system-agent.yaml
```diff
@@ -0,0 +1,634 @@
...
+      summary: "Retrieves the current configuration of the targeted service(s)."
+      parameters:
+        - $ref: '#/components/parameters/correlatedRequestHeader'
+        - name: services
+          in: path
+          required: true
+          schema:
+            type: string
+          description: "A comma separated list of the services for which to obtain configuration settings."
```
michaelestrin: Simplify to support retrieval of single service's configuration.  Can use multiple use-case requests to retrieve configuration across services within a single transport request.

api/openapi/v2/system-agent.yaml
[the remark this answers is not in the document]
michaelestrin: Endpoint should target single service. Caller can update multiple services via multiple use-case requests in a single request.

I'd tend towards specifying the capability to update multiple settings for a single service within a single use-case request.  Not the way it's implemented today.

api/openapi/v2/system-agent.yaml
```diff
@@ -0,0 +1,634 @@
...
+    limitParam:
+      in: query
+      name: limit
+      required: false
+      schema:
+        type: integer
+        minimum: 1
+        maximum: 50
+        default: 20
+      description: "The numbers of items to return."
```
michaelestrin: These should be removed; they're not referenced anywhere.

api/openapi/v2/system-agent.yaml
```diff
@@ -0,0 +1,637 @@
...
+        service:
+          description: "The service associated with the metrics"
+          type: string
+          example: "edgex-core-data"
+        result:
+          description: "The metric information retrieved from the targeted service"
+          type: object
+          properties:
+            cpuBusyAvg:
+              type: integer
```
michaelestrin: This is part of the raw response.

api/openapi/v2/system-agent.yaml
```diff
@@ -0,0 +1,637 @@
...
+              $ref: '#/components/headers/correlatedResponseHeader'
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  anyOf:
+                    - $ref: '#/components/schemas/ErrorResponse'
+                    - $ref: '#/components/schemas/ResponseEnvelope'
+  /config/settings:
```
michaelestrin: Prefer `/config/set`

api/openapi/v2/system-agent.yaml
```diff
@@ -0,0 +1,637 @@
...
+        '400':
+          description: "Bad request"
+          headers:
+            X-Correlation-ID:
+              $ref: '#/components/headers/correlatedResponseHeader'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+  /config:
```
michaelestrin: Prefer `/config/get`

PR #2337, opened 2020-02-05
Fix the issue of all lower-case spellings of properties in configuration.toml

### LINKED ISSUE #1762: Reorganize the properties in configuration file of secretstore to be consistent with EdgeX-Go
Currently the properties in the configuration file ( under /res folder) have lower case, and the structure is not consistent with others in the EdgeX-Go. It is suggested to make them consistent.

tingyuz:
change the names of properties in configuration.toml of security-secretstore-setup project from all lower cases to upper case of first letter so that they are consistent with others in edgex-go project

fix #1762

### DIFF
--- a/cmd/security-secretstore-setup/res/configuration.toml
+++ b/cmd/security-secretstore-setup/res/configuration.toml
@@ -23,20 +23,22 @@ EnableRemote = false
 File = './logs/security-secretstore-setup.log'
 
 [SecretService]
-scheme = "https"
-server = "edgex-vault"
-port = 8200
-certpath = "v1/secret/edgex/edgex-security-proxy-setup/kong-tls"
-cafilepath = "/vault/config/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem"
-certfilepath = "/vault/config/pki/EdgeXFoundryCA/edgex-kong.pem"
-keyfilepath = "/vault/config/pki/EdgeXFoundryCA/edgex-kong.priv.key"
-tokenfolderpath = "/vault/config/assets"
-tokenfile = "resp-init.json"
-vaultsecretshares = 5
-vaultsecretthreshold = 3
-tokenprovider = ""
-tokenproviderargs = [ "-confdir", "res/token-file-provider" ]
-tokenprovidertype = "oneshot"
+Scheme = "https"
+Server = "edgex-vault"
+Port = 8200
+CertPath = "v1/secret/edgex/edgex-security-proxy-setup/kong-tls"
+CaFilePath = "/tmp/edgex/secrets/ca/ca.pem"
+CertFilePath = "/tmp/edgex/secrets/edgex-kong/server.crt"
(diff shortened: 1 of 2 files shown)


cmd/security-secretstore-setup/res/configuration.toml
```diff
@@ -23,20 +23,22 @@ EnableRemote = false
...
+KeyFilePath = "/tmp/edgex/secrets/edgex-kong/server.key"
+TokenFolderPath = "/vault/config/assets"
+TokenFile = "resp-init.json"
+VaultSecretShares = 5
+VaultSecretThreshold = 3
+TokenProvider = "/security-file-token-provider"
+TokenProviderArgs = [ "-confdir", "res-file-token-provider" ]
+TokenProviderType = "oneshot"
+TokenProviderAdminTokenPath = "/run/edgex/secrets/tokenprovider/secrets-token.json"
+RevokeRootTokens = true
```
bnevis-i: Do not change values of TokenProvider , TokenProviderArgs, TokenProviderType, TokenProviderAdminTokenPath, and RevokeRootTokens as per-service vault tokens and token revocation has not been enabled for the non-Docker configuration.

cmd/security-secretstore-setup/res/configuration.toml
```diff
@@ -23,20 +23,22 @@ EnableRemote = false
...
-keyfilepath = "/vault/config/pki/EdgeXFoundryCA/edgex-kong.priv.key"
-tokenfolderpath = "/vault/config/assets"
-tokenfile = "resp-init.json"
-vaultsecretshares = 5
-vaultsecretthreshold = 3
-tokenprovider = ""
-tokenproviderargs = [ "-confdir", "res/token-file-provider" ]
-tokenprovidertype = "oneshot"
+Scheme = "https"
+Server = "edgex-vault"
```
bnevis-i: I'm wondering if this even worked before this change in the non-docker configuration...  Usually this is "localhost".  Suggest change to "localhost" if this is indeed broken.

cmd/security-secretstore-setup/res/configuration.toml
bnevis-i: Note: Docker `configuration.toml` changes appear correct.

tingyuz: In the past we made these two files identical and Ian stepped in later to adjust the values that worked on snap so wondering if we should follow such routine?

cmd/security-secretstore-setup/res/configuration.toml
tingyuz: The reason we have configuration.toml file outside the res folder is to make snap project easier to covert to whatever formats/values that work in the snap environment, and we just keep them identical first, then another PR will be followed to address snap issues - we don't have such environment to test/verify.

tingyuz:
Per request #2338 is created to track the efforts for snap environment.

Issue #2380, opened 2020-02-19
All microservices should authenticate themselves when using mongo database


difince:
# 🐞 Bug Report
All microservices should authenticate themselves when using mongo database no matter if they are started up via docker or via snap. Once issue  https://github.com/edgexfoundry/docker-edgex-mongo/issues/77 is fixed (mongo is started with enabled access control), services started up using configuration.toml files that are directly under **/res** directory will got authentication exception when manipulating the database. 
### Affected Services
All microservices that  use mongo database and are run usually locally or via snap
- core-command
- core-data
- core-metadata
- support-logging
- support-notifications
- support-sheduler

 The issue is located in these files: 
- [core-command](https://github.com/edgexfoundry/edgex-go/blob/e5ebbe2778b17324c446436c8bb6947fa319c3bf/cmd/core-command/res/configuration.toml#L35-L43)
- [core-data](https://github.com/edgexfoundry/edgex-go/blob/e5ebbe2778b17324c446436c8bb6947fa319c3bf/cmd/core-data/res/configuration.toml#L41-L49)
- [ core-metadata](https://github.com/edgexfoundry/edgex-go/blob/e5ebbe2778b17324c446436c8bb6947fa319c3bf/cmd/core-metadata/res/configuration.toml#L40-L48)
- [support-logging](https://github.com/edgexfoundry/edgex-go/blob/e5ebbe2778b17324c446436c8bb6947fa319c3bf/cmd/support-logging/res/configuration.toml#L24-L32)
- [support-notifications](https://github.com/edgexfoundry/edgex-go/blob/e5ebbe2778b17324c446436c8bb6947fa319c3bf/cmd/support-notifications/res/configuration.toml#L31-L39)
- [support-scheduler](https://github.com/edgexfoundry/edgex-go/blob/e5ebbe2778b17324c446436c8bb6947fa319c3bf/cmd/support-scheduler/res/configuration.toml#L31-L39)

### Is this a regression?

 -

### Description and Minimal Reproduction
All /edgex-go/cmd/*/res/configuration.toml files have empty `Password` and `Username` filed under [**Databases**]
Example with edgex-go/cmd/metadata/configuration.tolm file 

> [Databases]
  [Databases.Primary]
  Host = 'localhost'
  Name = 'metadata'
  Password = ''
  Port = 27017
  Username = ''
  Timeout = 5000
  Type = 'mongodb'

Username and Password files need to be updated accordingly.  

## 🔥 Exception or Error
[...]

PR #2381, opened 2020-02-19
Provide mongo credentials in all configuration.toml files

difince:
Mongo is going to be started with enabled access control.(see [here](https://github.com/edgexfoundry/docker-edgex-mongo/issues/77))
This means that each service that connects to mongo will need to authenticate itself even when
EDGEX_SECURITY_SECRET_STORE = false, or the service is run by snap.

## PR Checklist

- [ ] The commit message follows our guidelines: https://wiki.edgexfoundry.org/display/FA/Contributor%27s+Guide

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/edgexfoundry/edgex-go/issues/2380

- [ ] Yes
- [X] No

## Other information


michaelestrin:
Added security-related reviewers.  Larger question -- should we be propagating "standard" persistence credentials into individual service configuration files?

jim-wang-intel:
> 
> 
> Added security-related reviewers. Larger question -- should we be propagating "standard" persistence credentials into individual service configuration files?

This is a good question to discuss:  I thought the current Mongo's credentials for services are generated on the fly and uploaded to Vault via Vault worker.  Is this credentials from toml file mongo's ROOT username and password? If so, it looks insecure via this approach...

difince:
@jim-wang-intel No, these are not  mongo's ROOT username and password.
You are correct, Mongo credentials are generated on the fly and uploaded into Vault via vault-worker but only when the security is enabled (i.e when EDGEX_SECURITY_SECRET_STORE=True )
If the security is disabled, microservices should nevertheless authenticate itself in front of mongo. This authentication will become mandatory when mongo access control is enabled

difince:
> should we be propagating "standard" persistence credentials into individual service configuration files

Is you question equal to: 

- "Is the configuration file the right place for storing these "standard" credentials ? "

accepting that "Standard Credential" = The credentials used when security is disabled

PR #2387, opened 2020-02-19
Deleted spurious file.

### LINKED ISSUE #2379: Spurious git diff caused by unit testing (security services)
Forward from bforster over Slack:

Is this something that belongs in the repo or is it an artifact that shouldn’t have been committed? Every time I do a make test make deletes it and then git marks it as a deleted file
https://github.com/edgexfoundry/edgex-go/blob/master/internal/security/secrets/helper/validDirectory/test.file

brandonforster:
Fixes #2379

Signed-off-by: Brandon Forster <me@brandonforster.com>


bnevis-i:
It appears that this file was mistakenly introduced in https://github.com/edgexfoundry/edgex-go/commit/0a580629264ee501ee3be2916c9cbbe81faf93f2.

However, the root cause is the following code in helper_test.go:
```
+       f, err := os.Create(filepath.Join(certConfigDir, testFileName))
+       if err != nil {
                assert.Fail(t, "unable to create file")
        }
-
+       defer f.Close()
```

Because the file is created and not closed with os.Close(), the deferred os.RemoveAll() is not able to cleanup the temp directory.

brandonforster:
@bnevis-i I'll take a look at the root cause. Closing for now, will reopen with more.

PR #2389, opened 2020-02-20
Fix provision watcher ID not added to callback object bug

### LINKED ISSUE #2358: Assign id back to provisionwatcher after adding to database
# 🐞 Bug Report

### Affected Services

 The issue is located in: Metadata

### Is this a regression?

 No

After we post provisionwatcher onto metadata, it will then use callback path to update back to device service. 
https://github.com/edgexfoundry/edgex-go/blob/0c2784e56b6ef2dced463a7975227a56906f5104/internal/core/metadata/rest_provisionwatcher.go#L404
but the `Id` field in argument `pw` is not updated, will eventually affect the callback handler in device service:

https://github.com/edgexfoundry/device-sdk-go/blob/83e9a03e857076317f35131868562e7812717f60/internal/handler/callback/general.go#L19

## 🔥 Exception or Error
<pre><code>

level=ERROR ts=2020-02-12T09:36:03.703436Z app=device-simple source=general.go:21 msg="Missing callback parameters"
</code></pre>

## 🌍  Your Environment
**Deployment Environment:**
macOS Catalina

**EdgeX Version:**
fuji release

**Anything else relevant?**

### LINKED ISSUE #2388: Update provision watcher cannot correctly identify duplicate IDs
# 🐞 Bug Report

### Affected Services
the issue is located in: core-metadata

### Is this a regression?
no

[The provision watcher update code will never check for duplicate IDs. The code that checks for duplicate IDs is inside the error guard, which in most implementations of the interface would be expected to not return an ID.](https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/metadata/rest_provisionwatcher.go#L480-L488)

brandonforster:
Fixes #2358

This PR ensures that after a provision watcher is updated, the ID returned by the database is propagated back into the provision watcher object before its `notify` method is called.

Note: A previous version of this PR included work for #2388. A new PR will be cut for that.


brandonforster:
@hahattan please verify this fixes your issue.

hahattan:
Yes it fixed the issue .

tingyuz:
This PR does fix the bug mentioned in the original issue ticket. However when considering a bigger picture of device provisioning some form of security mechanism is needed. The details are captured in a newly create issue here: https://github.com/edgexfoundry/edgex-go/issues/2415

PR #2392, opened 2020-02-20
Fixed provision watcher duplicate update bug

### LINKED ISSUE #2388: Update provision watcher cannot correctly identify duplicate IDs

brandonforster:
Fixes #2388. Additionally, performed some linter warning cleanup.

This PR is in a draft state because while I believe this changeset addresses a defect in the code and ensures there is no orphan code, I am not sure this is actually the approach we want.

[With this changeset (implemented with the perspective of "what is the code trying to do") an update cannot truly be an update, because any provision watcher update cannot have the same name as an extant one.](https://github.com/edgexfoundry/edgex-go/pull/2392/files#diff-f3ed5f54b83a24a229f301e696cae8bdL479-L489)

Seeking perspective from @michaelestrin, @akramtexas, and @jbonafide623, the most recent authors of these files to describe the code's intent.

Signed-off-by: Brandon Forster <me@brandonforster.com>

### DIFF
--- a/internal/core/metadata/rest_provisionwatcher.go
+++ b/internal/core/metadata/rest_provisionwatcher.go
@@ -50,7 +50,7 @@ func restGetProvisionWatchers(
 	}
 
 	w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
-	json.NewEncoder(w).Encode(&res)
+	_ = json.NewEncoder(w).Encode(&res)
 }
 
 func restDeleteProvisionWatcherById(
@@ -61,7 +61,7 @@ func restDeleteProvisionWatcherById(
 	errorHandler errorconcept.ErrorHandler) {
 
 	vars := mux.Vars(r)
-	var id string = vars[ID]
+	var id = vars[ID]
 
 	// Check if the provision watcher exists
 	pw, err := dbClient.GetProvisionWatcherById(id)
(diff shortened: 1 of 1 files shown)


brandonforster:
Spoke with @akramtexas. He doesn't recall any specific authorial intent to reject duplicate IDs and concurred that there may be serious flaws in current implementation. He concurred that it is possible that the current implementation only updates provision watchers because none of the current implementation is doing what it "intends" to do, and this PR may not be the correct approach.

akramtexas:
> Spoke with @akramtexas. He doesn't recall any specific authorial intent to reject duplicate IDs and concurred that there may be serious flaws in current implementation. He concurred that it is possible that the current implementation only updates provision watchers because none of the current implementation is doing what it "intends" to do, and this PR may not be the correct approach.

- @brandonforster That's correct: As and when I had touched this part of the code, as I mentioned in our chat a few minutes ago, these likely were part of a large refactoring that was undertaken at that time.

brandonforster:
Superseded by #2411.

Issue #2272, opened 2020-01-10
Response of issueDeviceCommand is not providing CBOR Content-Type header


FelixTing:
# 🐞 Bug Report

### Affected Services

 The issue is located in: command

### Is this a regression?

 No

### Description and Minimal Reproduction
Function [issueDeviceCommand](https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/command/restDevice.go#L54) and [issueDeviceCommandByNames](https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/command/restDevice.go#L126) is still applying the application/json Content-Type header when the response body is a CBOR event.

akramtexas:
@FelixTing 
- As part of recreating (more on that in a minute) the bug that you've reported, I get to the point of issuing the following REST call, using my device name of "countcamera1":
`http://localhost:48082/api/v1/device/name/countcamera1`
- And here is the response I get: 
`http://:0/api/v1/device/name/%7B%7BgetDeviceByName%7D%7D: dial tcp :0: connect: can't assign requested address`
- Can you please confirm that you're seeing the same response as well?
- Can you please some detail into your setup where the response body, as you have noted, is a CBOR event? And how do you arrive at the conclusion that the functions `issueDeviceCommand` and `issueDeviceCommandByNames` are still applying the application/json Content-Type header in such a scenario (i.e. when the response body is a CBOR event)?

akramtexas:
@FelixTing As I referred to in the previous comment, the steps I have following in my attempt to recreate the reported issue are these:
- I start with creating reference information in EdgeX, making two calls to Core Metadata, to create the addressables for the Device Service.
- I then define Value Descriptors by calls to Core Data.
- Create a Device Profile via Core Metadata.
- Use Core Metadata to create the Device Service.
- Use Core Metadata to provision the device ("countcamera1").
- Use Core Command to request a list of the commands supported by the device ("countcamera1") via the following API call:
`http://localhost:48082/api/v1/device/name/countcamera1`
- And this is the point at which I get (as mentioned in previous comment) the following response:
`http://:0/api/v1/device/name/countcamera1: dial tcp :0: connect: can't assign requested address`

michaelestrin:
> * And this is the point at which I get (as mentioned in previous comment) the following response:
>   `http://:0/api/v1/device/name/countcamera1: dial tcp :0: connect: can't assign requested address`

[...]

PR #2446, opened 2020-03-20
Fix core-command response when handling a CBOR result.

akramtexas:
- This is the second (thought first in sequence of being reviewed) of two `PR`s that will address https://github.com/edgexfoundry/edgex-go/issues/2272 ("Response of issueDeviceCommand is not providing CBOR Content-Type header"). 
- The second `PR` (in `device-sdk-go`) was previously submitted, with [details on sequencing here](https://github.com/edgexfoundry/device-sdk-go/pull/464#issuecomment-598815862).
- `Blackbox Integration tests` were run, to validate the updates per this `PR` (Results are provided in the comment below).
- The set of steps used for testing are provided in the comment below.
- Immediately after submitting this `PR`, on realizing that we would be better served by marking this as a `Draft PR`, I sought to reclassify this as a `Draft PR`, but don't see an option to do so - The thinking here is that as we work through the best approach to resolving a challenge that was encountered (and related details into that insight below) during the work done per this `PR`.
- The insight referred to above is thanks entirely to @tsconn23: The handling of `PUT` and `GET` is done significantly down a deeply nested call to functions (as follows) and we may be better served by making them (i.e. `PUT` and `GET`) first class citizens by creating a type abstraction for them, which can be tapped into at a much higher level in the call chain, enabling the request and client to be assembled higher up the chain:
```
github.com/edgexfoundry/edgex-go/internal/core/command.serviceCommand.Execute at types.go:47
github.com/edgexfoundry/edgex-go/internal/core/command.executeCommandByDevice at device.go:128
github.com/edgexfoundry/edgex-go/internal/core/command.executeCommandByDeviceID at device.go:72
github.com/edgexfoundry/edgex-go/internal/core/command.issueDeviceCommand at restDevice.go:79
github.com/edgexfoundry/edgex-go/internal/core/command.restGetDeviceCommandByCommandID at restDevice.go:42
github.com/edgexfoundry/edgex-go/internal/core/command.TestRestGetDeviceCommandByCommandID.func1 at restDevice_test.go:272
```

### DIFF
--- a/internal/core/command/device.go
+++ b/internal/core/command/device.go
@@ -15,43 +15,54 @@
 package command
 
 import (
+	"bytes"
 	"context"
+	"fmt"
 	"net/http"
 
-	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
-	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata"
-	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
-	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
-
+	"github.com/edgexfoundry/edgex-go/internal"
 	"github.com/edgexfoundry/edgex-go/internal/core/command/config"
 	"github.com/edgexfoundry/edgex-go/internal/core/command/errors"
 	"github.com/edgexfoundry/edgex-go/internal/core/command/interfaces"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata"
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+
+	"github.com/gorilla/mux"
 )
 
 func executeCommandByDeviceID(
-	ctx context.Context,
-	deviceID string,
-	commandID string,
+	originalRequest *http.Request,
 	body string,
-	queryParams string,
-	isPutCommand bool,
 	lc logger.LoggingClient,
 	dbClient interfaces.DBClient,
-	deviceClient metadata.DeviceClient) (string, error) {
+	deviceClient metadata.DeviceClient,
+	httpCaller internal.HttpCaller) (deviceServiceResponse *http.Response, theResponseBody string, failure error) {
+
+	if originalRequest == nil {
+		return nil, "", errors.NewErrExtractingInfoFromRequest()
+	}
+
+	ctx := originalRequest.Context()
+	deviceID, commandID, err := extractDeviceIdAndCommandIdFromRequest(originalRequest)
+	if err != nil {
+		return nil, "", err
+	}
 
 	d, err := deviceClient.Device(ctx, deviceID)
 	if err != nil {
-		return "", err
+		return nil, "", err
 	}
 
 	if d.AdminState == contract.Locked {
-		return "", errors.NewErrDeviceLocked(d.Name)
+		return nil, "", errors.NewErrDeviceLocked(d.Name)
 	}
 
 	// once command service have its own persistence layer this call will be changed.
 	commands, err := dbClient.GetCommandsByDeviceId(d.Id)
 	if err != nil {
-		return "", err
+		return nil, "", err
 	}
 
 	var c contract.Command
@@ -63,38 +74,53 @@ func executeCommandByDeviceID(
 	}
 
 	if c.String() == (contract.Command{}).String() {
-		return "", errors.NewErrCommandNotAssociatedWithDevice(commandID, deviceID)
+		return nil, "", errors.NewErrCommandNotAssociatedWithDevice(commandID, deviceID)
 	}
 
-	return executeCommandByDevice(ctx, d, c, body, queryParams, isPutCommand, lc)
+	return executeCommandByDevice(ctx, d, c, body, lc, originalRequest, httpCaller)
+}
+
+// extractDeviceIdAndCommandIdFromRequest extracts deviceID and commandID from r, which
+// is the HTTP request parameter, and returns the deviceID, commandID to caller, or, if not
+// successfully extracted, the associated error is returned.
+func extractDeviceIdAndCommandIdFromRequest(r *http.Request) (string, string, error) {
+	vars := mux.Vars(r)
+	deviceID := vars[ID]
+	commandID := vars[COMMANDID]
+
+	if deviceID == "" || commandID == "" {
+		return deviceID, commandID, errors.NewErrExtractingInfoFromRequest()
+	}
+
+	return deviceID, commandID, nil
 }
 
 func executeCommandByName(
+	originalRequest *http.Request,
 	ctx context.Context,
 	dn string,
 	cn string,
 	body string,
-	queryParams string,
-	isPutCommand bool,
 	lc logger.LoggingClient,
 	dbClient interfaces.DBClient,
-	deviceClient metadata.DeviceClient) (string, error) {
+	deviceClient metadata.DeviceClient,
+	httpCaller internal.HttpCaller) (deviceServiceResponse *http.Response, theResponseBody string, failure error) {
 
 	d, err := deviceClient.DeviceForName(ctx, dn)
 	if err != nil {
-		return "", err
+		return nil, "", err
 	}
 
 	if d.AdminState == contract.Locked {
-		return "", errors.NewErrDeviceLocked(d.Name)
+		return nil, "", errors.NewErrDeviceLocked(d.Name)
 	}
 
 	command, err := dbClient.GetCommandByNameAndDeviceId(cn, d.Id)
 	if err != nil {
-		return "", err
+		return nil, "", err
 	}
 
-	return executeCommandByDevice(ctx, d, command, body, queryParams, isPutCommand, lc)
+	return executeCommandByDevice(ctx, d, command, body, lc, originalRequest, httpCaller)
 }
 
 func executeCommandByDevice(
@@ -102,31 +128,43 @@ func executeCommandByDevice(
 	device contract.Device,
 	command contract.Command,
 	body string,
-	queryParams string,
-	isPutCommand bool,
-	lc logger.LoggingClient) (string, error) {
+	lc logger.LoggingClient,
+	originalRequest *http.Request,
+	httpCaller internal.HttpCaller) (deviceServiceResponse *http.Response, theResponseBody string, failure error) {
 
+	var method string
 	var ex Executor
 	var err error
-	if isPutCommand {
-		ex, err = NewPutCommand(device, command, body, ctx, &http.Client{}, lc)
-	} else {
-		ex, err = NewGetCommand(device, command, queryParams, ctx, &http.Client{}, lc)
+
+	if originalRequest == nil {
+		return nil, "", errors.NewErrParsingOriginalRequest("method")
+	}
+
+	switch originalRequest.Method {
+	case http.MethodPut:
+		ex, err = NewPutCommand(device, command, body, ctx, httpCaller, lc, originalRequest)
+	case http.MethodGet:
+		ex, err = NewGetCommand(device, command, ctx, httpCaller, lc, originalRequest)
+	default:
+		lc.Error(fmt.Sprintf("unknown method: %s", method))
 	}
 
 	if err != nil {
-		return "", err
+		return nil, "", err
 	}
 
-	responseBody, responseCode, err := ex.Execute()
+	deviceServiceResponse, err = ex.Execute()
 	if err != nil {
-		return "", err
-	}
-	if responseCode != http.StatusOK {
-		return "", types.NewErrServiceClient(responseCode, []byte(responseBody))
+		return nil, "", err
 	}
 
-	return responseBody, nil
+	responseBody := new(bytes.Buffer)
+	_, readErr := responseBody.ReadFrom(deviceServiceResponse.Body)
+	if readErr != nil {
+		return nil, "", readErr
+	}
+
+	return deviceServiceResponse, responseBody.String(), nil
 }
 
 func getAllCommands(
--- a/internal/core/command/device_test.go
+++ b/internal/core/command/device_test.go
@@ -15,138 +15,172 @@
 package command
 
 import (
-	"context"
 	goErrors "errors"
 	"net/http"
 	"net/url"
 	"testing"
-
-	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
-	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
-	"github.com/edgexfoundry/go-mod-core-contracts/models"
-	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 
 	"github.com/edgexfoundry/edgex-go/internal/core/command/errors"
 	"github.com/edgexfoundry/edgex-go/internal/core/command/interfaces"
 	"github.com/edgexfoundry/edgex-go/internal/core/command/interfaces/mocks"
 	mdMocks "github.com/edgexfoundry/edgex-go/internal/mocks"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/db"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
+	"github.com/edgexfoundry/go-mod-core-contracts/models"
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+
+	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 )
 
 var (
-	status400 = "400"
-	status404 = "404"
-	status500 = "500"
-	status423 = "423"
-
-	mismatch = "d200-c200-mismatch"
-	d200c404 = "d200-c404"
-	d200c500 = "d200-c500"
-
-	TestCommandId = "TestCommandID"
+	ExistingDeviceID                       = "existing device id"
+	DeviceIDWithAssociatedInvalidObjectID  = "device id with associated invalid object id"
+	NonExistentDeviceID                    = "non existent device id"
+	DeviceIDResultingInInternalServerError = "device id resulting in internal server error"
+	DeviceIDForLockedResource              = "device id for locked resource"
+	MismatchedDeviceID                     = "mismatched device id"
+	DeviceIDd200c404                       = "device id d200-c404"
+	DeviceIDd200c500                       = "device id d200-c500"
+	DeviceIDd200c200                       = "device id d200-c200"
+	TestCommandID                          = "test command id"
+	TestDeviceID                           = "test device id"
 )
 
+func createTestDeviceWithPathUrl(commandID string, deviceID string) models.Device {
+
+	return models.Device{
+		AdminState: models.Unlocked,
+		Service: models.DeviceService{
+			Addressable: models.Addressable{
+				Path: "/api/v1/device/" + deviceID + "/command/" + commandID}}}
+}
+
+func createDeviceRequestWithPathParameters(
+	sampleDevice models.Device,
+	params map[string]string) *http.Request {
+
+	req, _ := http.NewRequest(http.MethodGet, cmdURI, nil)
+	req.URL.Path = sampleDevice.Service.Addressable.Path
+
+	return mux.SetURLVars(req, params)
+}
+
 // commandByDeviceID
-func TestExecuteGETCommandByDeviceIdAndCommandId(t *testing.T) {
+func TestExecuteGETCommandByDeviceIDAndCommandID(t *testing.T) {
 	tests := []struct {
 		name        string
-		deviceId    string
-		commandId   string
+		status      string
+		request     *http.Request
 		expectedErr error
 	}{
 		{
-			"Device-InvalidObjectId",
-			status400,
-			TestCommandId,
+			"device with invalid object id",
+			DeviceIDWithAssociatedInvalidObjectID,
+			createDeviceRequestWithPathParameters(
+				createTestDeviceWithPathUrl(TestCommandID, DeviceIDWithAssociatedInvalidObjectID),
+				map[string]string{ID: DeviceIDWithAssociatedInvalidObjectID, COMMANDID: TestCommandID}),
 			types.NewErrServiceClient(400, []byte("Invalid object ID")),
 		},
 		{
-			"Device-NotFound",
-			status404,
-			TestCommandId,
+			"device not found",
+			NonExistentDeviceID,
+			createDeviceRequestWithPathParameters(
+				createTestDeviceWithPathUrl(TestCommandID, NonExistentDeviceID),
+				map[string]string{ID: NonExistentDeviceID, COMMANDID: TestCommandID}),
 			types.NewErrServiceClient(http.StatusNotFound, []byte{}),
 		},
 		{
-			"Device-InternalServerErr",
-			status500,
-			TestCommandId,
+			"device resulting in internal server error",
+			DeviceIDResultingInInternalServerError,
+			createDeviceRequestWithPathParameters(
+				createTestDeviceWithPathUrl(TestCommandID, DeviceIDResultingInInternalServerError),
+				map[string]string{ID: DeviceIDResultingInInternalServerError, COMMANDID: TestCommandID}),
 			goErrors.New("unexpected error"),
 		},
 		{
-			"Device-Locked",
-			status423,
-			TestCommandId,
+			"device was locked",
+			DeviceIDForLockedResource,
+			createDeviceRequestWithPathParameters(
+				createTestDeviceWithPathUrl(TestCommandID, DeviceIDForLockedResource),
+				map[string]string{ID: DeviceIDForLockedResource, COMMANDID: TestCommandID}),
 			errors.NewErrDeviceLocked(""),
 		},
 		{
-			"Command-NotFound",
-			d200c404,
-			status400,
+			"command was not found",
+			DeviceIDd200c404,
+			createDeviceRequestWithPathParameters(
+				createTestDeviceWithPathUrl(TestCommandID, DeviceIDd200c404),
+				map[string]string{ID: DeviceIDd200c404, COMMANDID: DeviceIDWithAssociatedInvalidObjectID}),
 			db.ErrNotFound,
 		},
 		{
-			"Command-InternalServerErr",
-			d200c500,
-			status500,
+			"command could not be handled",
+			DeviceIDd200c500,
+			createDeviceRequestWithPathParameters(
+				createTestDeviceWithPathUrl(TestCommandID, DeviceIDd200c500),
+				map[string]string{ID: DeviceIDd200c500, COMMANDID: DeviceIDResultingInInternalServerError}),
 			goErrors.New("unexpected error"),
 		},
 		{
-			"Command-NotBelongToDevice",
-			mismatch,
-			mismatch,
-			errors.NewErrCommandNotAssociatedWithDevice(TestCommandId, mismatch),
-		},
-		{
-			"Command-500URLCannotBeParsed",
-			"d200c200",
-			"200",
+			"command did not belong to device",
+			MismatchedDeviceID,
+			createDeviceRequestWithPathParameters(
+				createTestDeviceWithPathUrl(TestCommandID, MismatchedDeviceID),
+				map[string]string{ID: MismatchedDeviceID, COMMANDID: TestCommandID}),
+			errors.NewErrCommandNotAssociatedWithDevice(TestCommandID, MismatchedDeviceID),
+		}, {
+			"command could not be parsed",
+			DeviceIDd200c200,
+			createDeviceRequestWithPathParameters(
+				createTestDeviceWithPathUrl(ExistingDeviceID, DeviceIDd200c200),
+				map[string]string{ID: DeviceIDd200c200, COMMANDID: ExistingDeviceID}),
 			&url.Error{Op: "parse", URL: "://:0?", Err: goErrors.New("missing protocol scheme")},
 		},
 	}
+	httpCaller := createMockHttpCaller()
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			_, actualErr := executeCommandByDeviceID(
-				context.Background(),
-				tt.deviceId,
-				TestCommandId,
+			_, _, actualErr := executeCommandByDeviceID(
+				tt.request,
 				"",
-				"",
-				false,
 				logger.NewMockClient(),
 				newMockDBClient(),
-				newMockDeviceClient())
+				newMockDeviceClient(),
+				httpCaller)
 			if actualErr == nil {
 				t.Fatal("expected error")
 			}
-			if tt.expectedErr.Error() != actualErr.Error() {
-				t.Fatalf("error value mismatch -- expected %v got %v", tt.expectedErr, actualErr)
-			}
+			require.Equal(t, tt.expectedErr.Error(), actualErr.Error())
 		})
 	}
 }
 
 func newMockDeviceClient() *mdMocks.DeviceClient {
 	client := mdMocks.DeviceClient{}
-	client.On("Device", context.Background(), status404).Return(contract.Device{}, types.NewErrServiceClient(http.StatusNotFound, []byte{}))
-	client.On("Device", context.Background(), status400).Return(contract.Device{}, types.NewErrServiceClient(400, []byte("Invalid object ID")))
-	client.On("Device", context.Background(), status500).Return(contract.Device{}, goErrors.New("unexpected error"))
-	client.On("Device", context.Background(), status423).Return(contract.Device{Id: status423, AdminState: "LOCKED"}, nil)
-	client.On("Device", context.Background(), d200c404).Return(contract.Device{Id: status400}, nil)
-	client.On("Device", context.Background(), d200c500).Return(contract.Device{Id: status500}, nil)
-	client.On("Device", context.Background(), mismatch).Return(contract.Device{Id: mismatch}, nil)
-
-	client.On("Device", context.Background(), "d200c200").Return(contract.Device{Id: TestDeviceId}, nil)
+	client.On("Device", mock.Anything, DeviceIDWithAssociatedInvalidObjectID).Return(contract.Device{}, types.NewErrServiceClient(400, []byte("Invalid object ID")))
+	client.On("Device", mock.Anything, NonExistentDeviceID).Return(contract.Device{}, types.NewErrServiceClient(http.StatusNotFound, []byte{}))
+	client.On("Device", mock.Anything, DeviceIDResultingInInternalServerError).Return(contract.Device{}, goErrors.New("unexpected error"))
+	client.On("Device", mock.Anything, DeviceIDForLockedResource).Return(contract.Device{Id: DeviceIDForLockedResource, AdminState: "LOCKED"}, nil)
+	client.On("Device", mock.Anything, DeviceIDd200c404).Return(contract.Device{Id: DeviceIDWithAssociatedInvalidObjectID}, nil)
+	client.On("Device", mock.Anything, DeviceIDd200c500).Return(contract.Device{Id: DeviceIDResultingInInternalServerError}, nil)
+	client.On("Device", mock.Anything, MismatchedDeviceID).Return(contract.Device{Id: MismatchedDeviceID}, nil)
+	client.On("Device", mock.Anything, TestDeviceID).Return(contract.Device{Id: TestDeviceID}, nil)
+	client.On("Device", mock.Anything, DeviceIDd200c200).Return(contract.Device{Id: DeviceIDd200c200}, nil)
 	return &client
 }
 
 func newMockDBClient() interfaces.DBClient {
 	dbMock := &mocks.DBClient{}
-	dbMock.On("GetCommandsByDeviceId", status400).Return(nil, db.ErrNotFound)
-	dbMock.On("GetCommandsByDeviceId", status500).Return(nil, goErrors.New("unexpected error"))
-	dbMock.On("GetCommandsByDeviceId", mismatch).Return([]models.Command{{Id: "dummy"}}, nil)
-
-	dbMock.On("GetCommandsByDeviceId", TestDeviceId).Return([]models.Command{{Id: TestCommandId}}, nil)
-
+	dbMock.On("GetCommandsByDeviceId", DeviceIDWithAssociatedInvalidObjectID).Return(nil, db.ErrNotFound)
+	dbMock.On("GetCommandsByDeviceId", DeviceIDResultingInInternalServerError).Return(nil, goErrors.New("unexpected error"))
+	dbMock.On("GetCommandsByDeviceId", MismatchedDeviceID).Return([]models.Command{{Id: "dummy"}}, nil)
+	dbMock.On("GetCommandsByDeviceId", TestDeviceID).Return([]models.Command{{Id: TestCommandID}}, nil)
+	dbMock.On("GetCommandsByDeviceId", ExistingDeviceID).Return([]models.Command{{Id: ExistingDeviceID}}, nil)
+	dbMock.On("GetCommandsByDeviceId", DeviceIDd200c200).Return([]models.Command{{Id: ExistingDeviceID}}, nil)
 	return dbMock
 }
--- a/internal/core/command/errors/types.go
+++ b/internal/core/command/errors/types.go
@@ -26,3 +26,36 @@ func (e ErrCommandNotAssociatedWithDevice) Error() string {
 func NewErrCommandNotAssociatedWithDevice(commandID string, deviceID string) error {
 	return ErrCommandNotAssociatedWithDevice{commandID, deviceID}
 }
+
+// ErrExtractingInfoFromRequest is a struct that serves as the value
+// receiver for Error as defined for NewErrExtractingInfoFromRequest
+type ErrExtractingInfoFromRequest struct {
+}
+
+// Error returns a meaningful string message describing error details.
+func (e ErrExtractingInfoFromRequest) Error() string {
+	return fmt.Sprintf("error extracting command id and device id.")
+}
+
+// NewErrExtractingInfoFromRequest returns the relevant, properly-
+// constructed error type.
+func NewErrExtractingInfoFromRequest() error {
+	return ErrExtractingInfoFromRequest{}
+}
+
+// ErrBadRequest is a struct that serves as the value receiver
+// for Error as defined for NewErrParsingOriginalRequest
+type ErrBadRequest struct {
+	value string
+}
+
+// Error returns a meaningful string message describing error details.
+func (e ErrBadRequest) Error() string {
+	return fmt.Sprintf("error in parsing related to '%s'", e.value)
+}
+
+// NewErrParsingOriginalRequest returns the relevant, properly-
+// constructed error type.
+func NewErrParsingOriginalRequest(invalid string) error {
+	return ErrBadRequest{value: invalid}
+}
--- a/internal/core/command/get.go
+++ b/internal/core/command/get.go
@@ -21,7 +21,8 @@ import (
 	"strings"
 
 	"github.com/edgexfoundry/edgex-go/internal"
-	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+	"github.com/edgexfoundry/edgex-go/internal/core/command/errors"
+
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 )
@@ -30,24 +31,30 @@ import (
 func NewGetCommand(
 	device contract.Device,
 	command contract.Command,
-	queryParams string,
 	context context.Context,
 	httpCaller internal.HttpCaller,
-	lc logger.LoggingClient) (Executor, error) {
-	urlResult := device.Service.Addressable.GetBaseURL() + strings.Replace(command.Get.Action.Path, DEVICEIDURLPARAM, device.Id, -1) + "?" + queryParams
+	lc logger.LoggingClient,
+	originalRequest *http.Request) (Executor, error) {
+
+	queryParams := originalRequest.URL.RawQuery
+	urlResult := device.Service.Addressable.GetBaseURL() + strings.Replace(
+		command.Get.Action.Path,
+		DEVICEIDURLPARAM,
+		device.Id,
+		-1) + "?" + queryParams
 	validURL, err := url.ParseRequestURI(urlResult)
 	if err != nil {
 		return serviceCommand{}, err
 	}
-	request, err := http.NewRequest(http.MethodGet, validURL.String(), nil)
+	deviceServiceProxiedRequest, err := http.NewRequest(http.MethodGet, validURL.String(), nil)
 	if err != nil {
 		return serviceCommand{}, err
 	}
 
-	correlationID := context.Value(clients.CorrelationHeader)
-	if correlationID != nil {
-		request.Header.Set(clients.CorrelationHeader, correlationID.(string))
+	err = addHeadersToRequest(originalRequest, deviceServiceProxiedRequest, context)
+	if err != nil {
+		return serviceCommand{}, errors.NewErrParsingOriginalRequest("header")
 	}
 
-	return newServiceCommand(device, httpCaller, request, lc), nil
+	return newServiceCommand(device, httpCaller, deviceServiceProxiedRequest, lc), nil
 }
--- a/internal/core/command/get_test.go
+++ b/internal/core/command/get_test.go
@@ -16,6 +16,8 @@ package command
 
 import (
 	"context"
+	"net/http"
+	"reflect"
 	"strconv"
 	"testing"
 
@@ -25,10 +27,11 @@ import (
 )
 
 const (
-	TestProtocol = "http"
-	TestDeviceId = "TestDeviceID"
-	TestAddress  = "example.com"
-	TestPort     = 8080
+	TestProtocol        = "http"
+	TestDeviceId        = "TestDeviceID"
+	TestAddress         = "example.com"
+	TestPort            = 8080
+	NonPropagatedHeader = "NonPropagatedHeader"
 )
 
 // Device which can be used as a basis for test setup. By default this is constructed for happy path testing.
@@ -45,7 +48,6 @@ var testDevice = contract.Device{
 }
 
 // Command which can be used as a basis for test setup. By default this is constructed for happy path testing.
-
 var testCommand = contract.Command{
 	Get: contract.Get{
 		Action: contract.Action{
@@ -61,9 +63,11 @@ var testCommand = contract.Command{
 
 func TestNewGetCommandWithCorrelationId(t *testing.T) {
 	expectedCorrelationIDHeaderValue := "Testing"
+	req := newRequestWithHeaders(map[string]string{clients.ContentType: clients.ContentTypeCBOR}, http.MethodGet)
 	testContext := context.WithValue(context.Background(), clients.CorrelationHeader, expectedCorrelationIDHeaderValue)
-	getCommand, _ := NewGetCommand(testDevice, testCommand, "", testContext, nil, logger.NewMockClient())
+	getCommand, _ := NewGetCommand(testDevice, testCommand, testContext, nil, logger.NewMockClient(), req)
 	actualCorrelationIDHeaderValue := getCommand.(serviceCommand).Request.Header.Get(clients.CorrelationHeader)
+
 	if actualCorrelationIDHeaderValue == "" {
 		t.Errorf("The populated GetCommand's request should contain a correlation ID header value")
 	}
@@ -74,32 +78,37 @@ func TestNewGetCommandWithCorrelationId(t *testing.T) {
 }
 
 func TestNewGetCommandWithQueryParams(t *testing.T) {
+	req := newRequestWithHeaders(map[string]string{clients.ContentType: clients.ContentTypeCBOR}, http.MethodGet)
 	queryParams := "test=value1&test2=value2"
-	getCommand, _ := NewGetCommand(testDevice, testCommand, queryParams, context.Background(), nil, logger.NewMockClient())
-	req := getCommand.(serviceCommand).Request.URL
-	if req.Scheme != TestProtocol {
+	req.URL.RawQuery = queryParams
+	getCommand, _ := NewGetCommand(testDevice, testCommand, context.Background(), nil, logger.NewMockClient(), req)
+	r := getCommand.(serviceCommand).Request.URL
+	if r.Scheme != TestProtocol {
 		t.Errorf("Unexpected protocol")
 	}
 	expectedHost := TestAddress + ":" + strconv.Itoa(TestPort)
-	if req.Host != expectedHost {
+	if r.Host != expectedHost {
 		t.Errorf("Unexpected host address and port")
 	}
-	if req.Path != testCommand.Get.Action.Path {
+	if r.Path != testCommand.Get.Action.Path {
 		t.Errorf("Unexpected path")
 	}
-	if req.RawQuery != queryParams {
+	if r.RawQuery != queryParams {
 		t.Errorf("Unexpected Raw Query Value")
 	}
 }
 func TestNewGetCommandWithMalformedQueryParams(t *testing.T) {
+	req := newRequestWithHeaders(map[string]string{clients.ContentType: clients.ContentTypeJSON}, http.MethodGet)
 	queryParams := "!@#$%"
-	_, err := NewGetCommand(testDevice, testCommand, queryParams, context.Background(), nil, logger.NewMockClient())
+	req.URL.RawQuery = queryParams
+	_, err := NewGetCommand(testDevice, testCommand, context.Background(), nil, logger.NewMockClient(), req)
 	if err == nil {
 		t.Errorf("Expected error for malformed query parameters")
 	}
 }
 func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
-	getCommand, _ := NewGetCommand(testDevice, testCommand, "", context.Background(), nil, logger.NewMockClient())
+	req := newRequestWithHeaders(map[string]string{clients.ContentType: clients.ContentTypeJSON}, http.MethodGet)
+	getCommand, _ := NewGetCommand(testDevice, testCommand, context.Background(), nil, logger.NewMockClient(), req)
 	actualCorrelationIDHeaderValue := getCommand.(serviceCommand).Request.Header.Get(clients.CorrelationHeader)
 	if actualCorrelationIDHeaderValue != "" {
 		t.Errorf("No correlation ID should be specified")
@@ -108,9 +117,67 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
 
 func TestNewGetCommandInvalidBaseUrl(t *testing.T) {
 	device := testDevice
+	req := newRequestWithHeaders(map[string]string{clients.ContentType: clients.ContentTypeCBOR}, http.MethodGet)
 	device.Service.Addressable.Address = "!@#$"
-	_, err := NewGetCommand(device, testCommand, "", context.Background(), nil, logger.NewMockClient())
+	_, err := NewGetCommand(device, testCommand, context.Background(), nil, logger.NewMockClient(), req)
 	if err == nil {
 		t.Errorf("The invalid URL error was not properly propagated to the caller")
+	}
+}
+
+func TestNewGetCommandContentType(t *testing.T) {
+	tests := []struct {
+		name            string
+		originalHeaders map[string]string
+		expectedHeaders map[string]string
+	}{
+		{
+			name:            "cbor content type header propagated",
+			originalHeaders: map[string]string{clients.ContentType: clients.ContentTypeCBOR},
+			expectedHeaders: map[string]string{clients.ContentType: clients.ContentTypeCBOR},
+		},
+		{
+			name:            "json content type header propagated",
+			originalHeaders: map[string]string{clients.ContentType: clients.ContentTypeJSON},
+			expectedHeaders: map[string]string{clients.ContentType: clients.ContentTypeJSON},
+		},
+		{
+			name:            "no content type header provided",
+			originalHeaders: map[string]string{clients.ContentType: ""},
+			expectedHeaders: map[string]string{},
+		},
+		{
+			name:            "cbor content type propagated, random header not propagated",
+			originalHeaders: map[string]string{clients.ContentType: clients.ContentTypeCBOR, NonPropagatedHeader: "NonPropagatedHeader"},
+			expectedHeaders: map[string]string{clients.ContentType: clients.ContentTypeCBOR},
+		},
+		{
+			name:            "json content type propagated, random header not propagated",
+			originalHeaders: map[string]string{clients.ContentType: clients.ContentTypeJSON, NonPropagatedHeader: "NonPropagatedHeader"},
+			expectedHeaders: map[string]string{clients.ContentType: clients.ContentTypeJSON},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var loggerMock = logger.NewMockClient()
+			ctx := context.Background()
+			proxiedRequest := newRequestWithHeaders(tt.originalHeaders, http.MethodGet)
+			getCommand, _ := NewGetCommand(
+				testDevice,
+				testCommand,
+				ctx,
+				nil,
+				loggerMock,
+				proxiedRequest)
+			actualHeaders := map[string]string{}
+			for headerName, headerValues := range getCommand.(serviceCommand).Request.Header {
+				// Extract the first element only from slice.
+				actualHeaders[headerName] = headerValues[0]
+			}
+			if !reflect.DeepEqual(actualHeaders, tt.expectedHeaders) {
+				t.Errorf("expected %s does not match the observed %s", tt.expectedHeaders, actualHeaders)
+			}
+		})
 	}
 }
--- a/internal/core/command/put.go
+++ b/internal/core/command/put.go
@@ -20,7 +20,8 @@ import (
 	"strings"
 
 	"github.com/edgexfoundry/edgex-go/internal"
-	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+	"github.com/edgexfoundry/edgex-go/internal/core/command/errors"
+
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 )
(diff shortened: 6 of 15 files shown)


akramtexas:
- Sample results for `Blackbox Integration tests` are as follows:
- Issued the following command for the registered device:
```
curl --location --request GET 'http://localhost:48082/api/v1/device/9f872d68-2281-4af4-959d-29e4d51c2192'
```
- And got this response as expected:
```
{
    "id": "9f872d68-2281-4af4-959d-29e4d51c2192",
    "name": "Random-Binary-Device",
    "adminState": "UNLOCKED",
    "operatingState": "ENABLED",
    "labels": [
        "device-virtual-example"
    ],
    "commands": [
        {
            "created": 1583181972178,
            "modified": 1583181972178,
            "id": "b349df4a-6c3d-4218-b8bc-7ff8d51ea50d",
            "name": "Binary",
            "get": {
                "path": "/api/v1/device/{deviceId}/Binary",
                "responses": [
                    {
                        "code": "200",
                        "expectedValues": [
                            "Binary"
                        ]
                    },
                    {
                        "code": "503",
                        "description": "service unavailable"
                    }
                ],
                "url": "http://localhost:48082/api/v1/device/9f872d68-2281-4af4-959d-29e4d51c2192/command/b349df4a-6c3d-4218-b8bc-7ff8d51ea50d"
            },
            "put": {
                "path": "/api/v1/device/{deviceId}/Binary",
                "responses": [
                    {
                        "code": "200"
                    },
                    {
                        "code": "503",
                        "description": "service unavailable"
                    }
                ],
                "url": "http://localhost:48082/api/v1/device/9f872d68-2281-4af4-959d-29e4d51c2192/command/b349df4a-6c3d-4218-b8bc-7ff8d51ea50d",
                "parameterNames": [
                    "EnableRandomization_Binary"
                ]
            }
        }
    ]
}
```

akramtexas:
Here are the set of steps used for DEV testing:

- Launch all `edgex-go` core and support services locally, running natively, without containerization (This will be needed for the next step).
- Locally build this `device-sdk-go` branch.
- Point `device-virtual-go` to use this `device-sdk-go` branch (i.e. the local build) and launch device-virtual-go locally.
- At this point, the virtual device named "Random-Binary-Device" will get registered with EdegeX.
- Issue the following command:
```
curl --location --request GET 'http://localhost:48082/api/v1/device/name/Random-Binary-Device' \
--header 'Content-Type: application/cbor' \
--data-raw ''
```
- The preceding command should respond with two URLs, one each to use for PUT and GET, for example as follows:
```
{
   "id":"9f872d68-2281-4af4-959d-29e4d51c2192",
   "name":"Random-Binary-Device",
   "adminState":"UNLOCKED",
   "operatingState":"ENABLED",
   "labels":[
      "device-virtual-example"
   ],
   "commands":[
      {
         "created":1583181972178,
         "modified":1583181972178,
         "id":"b349df4a-6c3d-4218-b8bc-7ff8d51ea50d",
         "name":"Binary",
         "get":{
            "path":"/api/v1/device/{deviceId}/Binary",
            "responses":[
               {
                  "code":"200",
                  "expectedValues":[
                     "Binary"
                  ]
               },
               {
                  "code":"503",
                  "description":"service unavailable"
               }
            ],
            "url":"http://localhost:48082/api/v1/device/9f872d68-2281-4af4-959d-29e4d51c2192/command/b349df4a-6c3d-4218-b8bc-7ff8d51ea50d"
         },
         "put":{
            "path":"/api/v1/device/{deviceId}/Binary",
            "responses":[
               {
                  "code":"200"
               },
               {
                  "code":"503",
                  "description":"service unavailable"
               }
            ],
            "url":"http://localhost:48082/api/v1/device/9f872d68-2281-4af4-959d-29e4d51c2192/command/b349df4a-6c3d-4218-b8bc-7ff8d51ea50d",
            "parameterNames":[
               "EnableRandomization_Binary"
            ]
         }
      }
   ]
}
```
- Use the URL for **PUT** from above response as follows, where the `--data-binary` is supplied via a local file, whose contents are `°xEnableRandomization_Binarydtrue`, so you'll want to replace the path `'@/Users/username/Decoded_Binary'` by a suitably-located local file on your filesystem:
```
curl --location --request PUT 'http://localhost:48082/api/v1/device/9f872d68-2281-4af4-959d-29e4d51c2192/command/b349df4a-6c3d-4218-b8bc-7ff8d51ea50d' \
--header 'Content-Type: application/cbor' \
--data-binary '@/Users/username/Decoded_Binary'


```
- And use the URL for **GET** from the same response  above as follows:
```
curl --location --request GET 'http://localhost:48082/api/v1/device/9f872d68-2281-4af4-959d-29e4d51c2192/command/b349df4a-6c3d-4218-b8bc-7ff8d51ea50d' \
--header 'Content-Type: application/cbor'
```

akramtexas:
Rebased

akramtexas:
Rebased

internal/core/command/put.go
```diff
@@ -44,5 +50,7 @@ func NewPutCommand(
 		request.Header.Set(clients.CorrelationHeader, correlationID.(string))
 	}
 
+	request.Header.Set(clients.ContentType, r.Header.Get(clients.ContentType))
```
AnthonyMBonafide: Should we be passing more headers than just `Content-type` and `Correlation`? I am thinking we pass on all the headers from the original request since we are not sure which ones might be useful. For example, `Accept`, `Content-MD5`, `Date`, `Host`, etc. We discussed this previously, did you do any digging and find if there are any headers you thought were important to propagate? I think we maybe able to introduce the [Forwarded](https://tools.ietf.org/html/rfc7239#section-4) header in our request to properly inform the requester as of the origin of the request, just a thought.

internal/core/command/device.go
```diff
@@ -104,14 +109,16 @@ func executeCommandByDevice(
 	body string,
 	queryParams string,
 	isPutCommand bool,
-	lc logger.LoggingClient) (string, error) {
+	lc logger.LoggingClient,
+	r *http.Request,
```
AnthonyMBonafide: We talked previously about redundant fields and reducing the arguments to the functions which now accept `r *http.Request`. For example, you can derive the following from the `http.Request`:

1. `isPutCommand`
1. `queryParams`
1. `body`

internal/core/command/device.go
```diff
@@ -104,14 +109,16 @@ func executeCommandByDevice(
...
+	httpCaller internal.HttpCaller) (string, error) {
 
 	var ex Executor
 	var err error
 	if isPutCommand {
-		ex, err = NewPutCommand(device, command, body, ctx, &http.Client{}, lc)
+		ex, err = NewPutCommand(device, command, body, ctx, httpCaller, lc, r)
```
AnthonyMBonafide: See my previous comment regarding reducing the number of redundant arguments.

internal/core/command/device.go
```diff
@@ -37,7 +38,9 @@ func executeCommandByDeviceID(
 	isPutCommand bool,
 	lc logger.LoggingClient,
 	dbClient interfaces.DBClient,
-	deviceClient metadata.DeviceClient) (string, error) {
+	deviceClient metadata.DeviceClient,
+	r *http.Request,
+	httpCaller internal.HttpCaller) (string, error) {
```
AnthonyMBonafide: These arguments can be reduced since `http.Request` is now being passed in and some of the other arguments can be obtained from the request. Please apply this comment to the other functions down the call stack.

internal/core/command/device.go
```diff
@@ -78,7 +81,9 @@ func executeCommandByName(
 	isPutCommand bool,
 	lc logger.LoggingClient,
 	dbClient interfaces.DBClient,
-	deviceClient metadata.DeviceClient) (string, error) {
+	deviceClient metadata.DeviceClient,
```
AnthonyMBonafide: See my other comment regarding reducing the number of function arguments.

internal/core/command/restDevice.go
```diff
@@ -35,9 +37,10 @@ func restGetDeviceCommandByCommandID(
 	lc logger.LoggingClient,
 	dbClient interfaces.DBClient,
 	deviceClient metadata.DeviceClient,
-	httpErrorHandler errorconcept.ErrorHandler) {
+	httpErrorHandler errorconcept.ErrorHandler,
+	httpCaller internal.HttpCaller) {
 
-	issueDeviceCommand(w, r, false, lc, dbClient, deviceClient, httpErrorHandler)
+	issueDeviceCommand(w, r, false, lc, dbClient, deviceClient, httpErrorHandler, httpCaller)
```
AnthonyMBonafide: See my other comment regarding reducing the number of arguments.

internal/core/command/restDevice.go
```diff
@@ -96,7 +104,18 @@ func issueDeviceCommand(
...
+	switch r.Header.Get(clients.ContentType) {
+	case clients.ContentTypeCBOR:
+		// Content-type CBOR will be sent to the caller.
+		w.Header().Set(clients.ContentType, clients.ContentTypeCBOR)
+	case clients.ContentTypeJSON:
+		// Content-type JSON will be sent to the caller.
+		w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
+	default:
+		lc.Error(fmt.Sprintf("header value neither JSON nor CBOR, but: %s", r.Header.Get(clients.ContentType)))
+	}
```
AnthonyMBonafide: Can we just set the response header `Content-type` to whatever the request header `Content-type` is without having to inspect it? This way we do not have to update this logic to reflect supported content types downstream. Also we can rely on the lower level functionality to fail if the content type is not supported.

internal/core/command/restDevice.go
```diff
@@ -228,7 +252,15 @@ func restGetCommandsByDeviceName(
...
 
-	w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
+	// Check the header value on the response
+	if r.Header.Get(clients.ContentType) == clients.ContentTypeCBOR {
+		// Content-type CBOR will be sent to the caller.
+		w.Header().Set(clients.ContentType, clients.ContentTypeCBOR)
+	} else if r.Header.Get(clients.ContentType) == clients.ContentTypeJSON {
+		// Content-type JSON will be sent to the caller.
+		w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
+	}
```
AnthonyMBonafide: See my other comment regarding adding `Content-type` response header from the request header.

internal/core/command/device.go
```diff
@@ -104,14 +109,16 @@ func executeCommandByDevice(
...
+	httpCaller internal.HttpCaller) (string, error) {
 
 	var ex Executor
 	var err error
 	if isPutCommand {
-		ex, err = NewPutCommand(device, command, body, ctx, &http.Client{}, lc)
+		ex, err = NewPutCommand(device, command, body, ctx, httpCaller, lc, r)
 	} else {
-		ex, err = NewGetCommand(device, command, queryParams, ctx, &http.Client{}, lc)
+		ex, err = NewGetCommand(device, command, queryParams, ctx, httpCaller, lc)
```
AnthonyMBonafide: I think we should add the same functionality for propagating headers for GET commands as well. This will make the service consistent. This will require adding tests for this as well.

internal/core/command/device_test.go
```diff
@@ -53,56 +55,65 @@ func TestExecuteGETCommandByDeviceIdAndCommandId(t *testing.T) {
 		deviceId    string
 		commandId   string
 		expectedErr error
+		request     *http.Request
```
AnthonyMBonafide: Depending on if we add the header functionality for GET commands, we should either remove this test argument since it is always nil, or pass in an actual request and assert that the `HttpCaller` was invoked with a request with the correct headers.

internal/core/command/put_test.go
```diff
@@ -73,3 +80,9 @@ func TestNewPutCommandBody(t *testing.T) {
 		t.Error("Failed to verify the request body contents")
 	}
 }
+
+func newRequestWithContentType(contentType string) *http.Request {
```
AnthonyMBonafide: Even though we are now including a `Content-type` in the requests, we are never verifying the resulting request has the correct headers. Can we add these tests to ensure the newly added functionality?

internal/core/command/device.go
akramtexas: Yes, looks like I had implemented this (reduction of arguments to the said functions) in the corresponding `device-sdk-go` `PR` _only_, and will apply them here as well 👍

AnthonyMBonafide:
@akramtexas, I forgot to mention one caveat regarding the reduction of arguments, with the `body` argument I think you can only read the contents of the request body one time. If that is the case then reading the request body and passing it as arguments alongside the `*http.Request` might be desirable. Please verify that is the case before reducing the arguments.

akramtexas:
> @akramtexas, I forgot to mention one caveat regarding the reduction of arguments, with the `body` argument I think you can only read the contents of the request body one time. If that is the case then reading the request body and passing it as arguments alongside the `*http.Request` might be desirable. Please verify that is the case before reducing the arguments.

@AnthonyMBonafide Thanks, appreciated. Will do 👍

akramtexas:
> > @akramtexas, I forgot to mention one caveat regarding the reduction of arguments, with the `body` argument I think you can only read the contents of the request body one time. If that is the case then reading the request body and passing it as arguments alongside the `*http.Request` might be desirable. Please verify that is the case before reducing the arguments.
> 
> @AnthonyMBonafide Thanks, appreciated. Will do 👍

- @AnthonyMBonafide What I found was that indeed, one can only read the contents of the request body one time (So reading the request body and passing it as an argument alongside the *http.Request is the thing to do). Plus, attempting to to read the request body earlier in the deeply nested call chain, in `func NewPutCommand(...) {...}` introduces its own problems, such as introducing error-handling at a much-lower-level than desirable, for example:
```
	if r != nil {
		body, err := ioutil.ReadAll(r.Body)
		if body == nil && err != nil {
			httpErrorHandler.Handle(w, err, errorconcept.Common.InvalidRequest_StatusBadRequest)
			return "", err
		}
	}
```
- In summary, reading the request body and passing it as an argument alongside the *http.Request is the thing to do 👍

akramtexas:
Rebased, plus addressed nearly all comments. Will request review _after_ remaining comments have also been addressed.

internal/core/command/put.go
akramtexas: - Yes, the point regarding whether we should be passing more headers than just `Content-type` and `Correlation` is very pertinent, especially given the open-ended nature of which headers may be useful to end users. I did do some digging,  most recently in connection with the possibility of introducing the [Forwarded header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) in our request.
- So far, it seems like the golang community is grappling with [an issue in this area](https://github.com/golang/go/issues/20526), _without_ a clear outcome or direction.

akramtexas:
Rebased.

akramtexas:
Rebased

internal/core/command/put.go
tsconn23: Per Core WG 26-Mar-2020 we will not add these additional headers for pass-through. A separate task would be to audit passing of Content-Length on all requests. This may be handled already by the Go http implementation.

akramtexas: - To add a data point regarding the mention above of the `Content-Length` part of the header (i.e. `This may be handled already by the Go http implementation.`), on looking up the `golang` reference online for [Package http](https://golang.org/pkg/net/http/):
```
    // For client requests, certain headers such as Content-Length
    // and Connection are automatically written when needed and
    // values in Header may be ignored. See the documentation
    // for the Request.Write method.
```

akramtexas: - As @AnthonyMBonafide pointed out during the aforementioned `Core WG` 26-Mar-2020 call, we're specifically talking here in the context of the `core-command` talking to a device service. Noting, too, that device service won't do anything with those headers; they are not "materialized".

AnthonyMBonafide: This is duplicated in `get.go`. Can we eliminate code duplication here by introducing a function that adds the necessary headers to the request? Furthermore, I think we can do this in a way that allows us to add more request headers easily in the future. For example, one way would be to have a constant slice of request headers we want to propagate, and the function responsible for adding the headers can iterate over the slice adding any header values that are present in that slice. That way the code will be clean and very easy to add functionality to. I hope that makes sense.

AnthonyMBonafide: Were you able to verify if the `Content-Length` header was being added to the request by the underlying HTTP client?

internal/core/command/get_test.go
```diff
@@ -73,33 +76,9 @@ func TestNewGetCommandWithCorrelationId(t *testing.T) {
...
-		t.Errorf("Unexpected Raw Query Value")
-	}
-}
-func TestNewGetCommandWithMalformedQueryParams(t *testing.T) {
-	queryParams := "!@#$%"
-	_, err := NewGetCommand(testDevice, testCommand, queryParams, context.Background(), nil, logger.NewMockClient())
-	if err == nil {
-		t.Errorf("Expected error for malformed query parameters")
-	}
-}
```
AnthonyMBonafide: Should we be removing these tests? They seem valid.

internal/core/command/get_test.go
```diff
@@ -108,9 +87,38 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+	req := newRequestForGetWithContentType(clients.ContentTypeCBOR)
 	device.Service.Addressable.Address = "!@#$"
-	_, err := NewGetCommand(device, testCommand, "", context.Background(), nil, logger.NewMockClient())
+	_, err := NewGetCommand(device, testCommand, context.Background(), nil, logger.NewMockClient(), req)
 	if err == nil {
 		t.Errorf("The invalid URL error was not properly propagated to the caller")
 	}
 }
+
+func TestNewGetCommandCborContentType(t *testing.T) {
```
AnthonyMBonafide: Can this test and `TestNewGetCommandJsonContentType` be condensed into a single test table since there is some code duplication? Also, can we create negative tests to ensure we are not passing  unexpected headers? 

This also applies to the tests in `put_test.go`

internal/core/command/put_test.go
```diff
@@ -26,18 +28,21 @@ import (
 func TestNewPutCommandWithCorrelationId(t *testing.T) {
 	expectedCorrelationIDHeaderValue := "Testing"
 	testContext := context.WithValue(context.Background(), clients.CorrelationHeader, expectedCorrelationIDHeaderValue)
-	putCommand, _ := NewPutCommand(testDevice, testCommand, "Test body", testContext, nil, logger.NewMockClient())
+	req := newRequestForPutWithContentType(clients.ContentTypeCBOR)
+	putCommand, _ := NewPutCommand(testDevice, testCommand, "Test body", testContext, nil, logger.NewMockClient(), req)
 	actualCorrelationIDHeaderValue := putCommand.(serviceCommand).Request.Header.Get(clients.CorrelationHeader)
 	if actualCorrelationIDHeaderValue == "" {
-		t.Errorf("The populated PutCommand's request should contain a correlation ID header value")
```
AnthonyMBonafide: Can we revert this change since this is less informative then it was previously? This goes for the other changes to error messages that change `request` to `req`.

internal/core/command/put_test.go
```diff
@@ -48,28 +53,58 @@ func TestNewPutCommandInvalidBaseUrl(t *testing.T) {
...
+	}
+}
+
+func TestNewPutCommandCborContentType(t *testing.T) {
+	req := newRequestForPutWithContentType(clients.ContentTypeCBOR)
+	putCommand, _ := NewPutCommand(testDevice, testCommand, "Test Body", context.Background(), nil, logger.NewMockClient(), req)
+	actualContentType := putCommand.(serviceCommand).Request.Header.Get(clients.ContentType)
+
+	expectedContentType := clients.ContentTypeCBOR
+	if actualContentType != expectedContentType {
```
AnthonyMBonafide: Can we leverage the `require` package and use `require.Equal`. The test failure message is nicer and the code is simpler. This also applied to other test functions we are changing in this PR to keep the function consistent. I would say just keep these changes to the functions we are changing and not to change this everywhere as that would be a big effort. I would like to start leveraging the `require` and `assert` packages to have better consistency within the project, and better failure messages.

internal/core/command/restDevice_test.go
```diff
@@ -0,0 +1,395 @@
...
+		})
+	}
+}
+
+func TestRestPutDeviceCommandByCommandID(t *testing.T) {
+	tests := []struct {
+		name           string
+		dcMock         *mocks.DeviceClient
+		dbMock         interfaces.DBClient
+		request        *http.Request
```
AnthonyMBonafide: This test argument is always `nil`. Can we actually leverage this and pass a request with different headers and assert the request sent to the device service has the correct headers? As I see it we are not really testing the newly added functionality and not verifying that the `Content-type` header is present. I think we need those tests as well as negative tests which ensure we are not passing additional unwanted headers.

internal/core/command/restDevice_test.go
```diff
@@ -0,0 +1,395 @@
...
+		})
+	}
+}
+
+func TestRestGetDeviceCommandByCommandID(t *testing.T) {
+	tests := []struct {
+		name           string
+		dcMock         *mocks.DeviceClient
+		dbMock         interfaces.DBClient
+		request        *http.Request
```
AnthonyMBonafide: See my previous comment regarding the `http.Request` always being `nil`.

internal/core/command/put.go
akramtexas: ✅ Yes, I verified that the Content-Length part of the header was being added to the request by the underlying HTTP client. Here, for handy reference, is the `golang` reference online for [Package http](https://golang.org/pkg/net/http/):
```
    // For client requests, certain headers such as Content-Length
    // and Connection are automatically written when needed and
    // values in Header may be ignored. See the documentation
    // for the Request.Write method.
```

internal/core/command/device.go
```diff
@@ -16,28 +16,31 @@ package command
...
-	"github.com/edgexfoundry/edgex-go/internal/core/command/config"
-	"github.com/edgexfoundry/edgex-go/internal/core/command/errors"
-	"github.com/edgexfoundry/edgex-go/internal/core/command/interfaces"
 )
 
 func executeCommandByDeviceID(
+	r *http.Request,
 	ctx context.Context,
 	deviceID string,
 	commandID string,
```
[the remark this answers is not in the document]
akramtexas: Done ✅ Further refactorings were done to accomplish this. These resulted in overhauling several unit tests.

internal/core/command/device.go
```diff
@@ -16,28 +16,45 @@ package command
...
+	var deviceID string
+	var commandID string
+	var ctx context.Context
+
+	if r != nil {
+		ctx = r.Context()
+		deviceID, commandID = extractDeviceIdAndCommandIdFromRequest(r)
+	} else {
+		return "", errors.NewErrExtractingInfoFromRequest(commandID, deviceID)
+	}
```
AnthonyMBonafide: Can we invert this conditional logic? This would make this simpler to read, and eliminate the need for side effects. This stood out more so since there seems to me a not so obvious bug with the `commandID` and `deviceID` within the `else` portion of the conditional. Those variables do not seem to be set and will always be empty strings.

By inverting the logic we would essentially be eliminating the `else` portion. We can check if the request is `nil` and if it is we can return the error. This will allow us to remove the declarations (`var`) and instead use `:=` which eliminates side effects and by proxy bugs.

internal/core/command/device.go
```diff
@@ -66,19 +83,35 @@ func executeCommandByDeviceID(
...
+		if s[deviceIdPositionInUrlPath] != "" {
+			deviceID = s[deviceIdPositionInUrlPath]
+		}
+		if s[commandIdPositionInUrlPath] != "" {
+			commandID = s[commandIdPositionInUrlPath]
+		}
+	}
+
+	return deviceID, commandID
 }
```
AnthonyMBonafide: Is there an easier way to do this? I think I saw something like this in one of your previous iterations in this PR:

```go
vars := mux.Vars(r)
did := vars[DEVICEID]
cid := vars[COMMANDID]
```

Also, would it be wise to add and `error` to the returned parameters since this function is now accessible to all other REST handlers in the function and not all requests will have that information in the request?  Can we also add a Godoc comment at the beginning of the function?

internal/core/command/device.go
```diff
@@ -94,24 +127,34 @@ func executeCommandByName(
...
-		ex, err = NewGetCommand(device, command, queryParams, ctx, &http.Client{}, lc)
+		method = r.Method
+		switch method {
+		case http.MethodPut:
+			ex, err = NewPutCommand(device, command, body, ctx, httpCaller, lc, r)
+		case http.MethodGet:
+			ex, err = NewGetCommand(device, command, ctx, httpCaller, lc, r)
+		default:
+			lc.Error(fmt.Sprintf("unknown method: %s", method))
+		}
```
AnthonyMBonafide: See my other comment regarding inverting the logic here for simplicity, and elimination of side effects. Also, there seems to be a bug here with the `error` in the `if` portion. The `error` uses `r.Method` directly after ensuring that it is blank, which means it is always empty.

internal/core/command/device_test.go
```diff
@@ -42,101 +46,132 @@ var (
...
 		expectedErr error
 	}{
 		{
 			"Device-InvalidObjectId",
 			status400,
-			TestCommandId,
+			createDeviceRequestWithPathParameters(
+				createTestDeviceWithPathUrl(TestCommandId, status400),
+				map[string]string{ID: status400, COMMANDID: TestCommandId}),
```
AnthonyMBonafide: All of these tests specify `ID` as `statusxxx` to test invalid data. Can we use a better named variable or string that states that it is invalid for clarity? 

This applies to the other test arguments for this test and the associated mocks.

internal/core/command/errors/types.go
```diff
@@ -26,3 +26,28 @@ func (e ErrCommandNotAssociatedWithDevice) Error() string {
 func NewErrCommandNotAssociatedWithDevice(commandID string, deviceID string) error {
 	return ErrCommandNotAssociatedWithDevice{commandID, deviceID}
 }
+
+type ErrExtractingInfoFromRequest struct {
+	commandID string
+	deviceID  string
```
AnthonyMBonafide: Are these useful? If you cannot extract the information from the request then I do not see how we can provide this information. This goes with my existing comment regarding this `error` always being constructed with empty strings

internal/core/command/errors/types.go
```diff
@@ -26,3 +26,28 @@ func (e ErrCommandNotAssociatedWithDevice) Error() string {
...
+	return fmt.Sprintf("error extracting command with id '%v' for device with id '%v'.", e.commandID, e.deviceID)
+}
+
+func NewErrExtractingInfoFromRequest(commandID string, deviceID string) error {
+	return ErrExtractingInfoFromRequest{commandID, deviceID}
+}
+
+type RequestParameterSupplied struct {
+	methodType string
+}
```
AnthonyMBonafide: Is this error struct needed? Can we simply use: https://github.com/edgexfoundry/edgex-go/blob/42aa357818f7845ac1db35d24861146831d3085c/internal/core/metadata/errors/types.go#L86-L88

internal/core/command/get.go
```diff
@@ -30,11 +32,23 @@ import (
...
-	urlResult := device.Service.Addressable.GetBaseURL() + strings.Replace(command.Get.Action.Path, DEVICEIDURLPARAM, device.Id, -1) + "?" + queryParams
+	lc logger.LoggingClient,
+	r *http.Request) (Executor, error) {
+
+	var queryParams string
+	if r != nil {
+		queryParams = r.URL.RawQuery
+	} else {
+		return serviceCommand{}, errors.New("error in request parameters that was passed in")
+	}
```
AnthonyMBonafide: See my previous comment regarding inverting the conditional logic for simplicity, and elimination of side effects.

internal/core/command/get.go
```diff
@@ -30,11 +32,23 @@ import (
...
-	lc logger.LoggingClient) (Executor, error) {
-	urlResult := device.Service.Addressable.GetBaseURL() + strings.Replace(command.Get.Action.Path, DEVICEIDURLPARAM, device.Id, -1) + "?" + queryParams
+	lc logger.LoggingClient,
+	r *http.Request) (Executor, error) {
+
+	var queryParams string
+	if r != nil {
+		queryParams = r.URL.RawQuery
+	} else {
+		return serviceCommand{}, errors.New("error in request parameters that was passed in")
```
AnthonyMBonafide: Can we use an error struct rather than a "random" error? Can we simply use: https://github.com/edgexfoundry/edgex-go/blob/42aa357818f7845ac1db35d24861146831d3085c/internal/core/metadata/errors/types.go#L86-L88

internal/core/command/get.go
```diff
@@ -49,5 +63,11 @@ func NewGetCommand(
 		request.Header.Set(clients.CorrelationHeader, correlationID.(string))
 	}
 
+	addHeadersToRequest(request, r.Header.Get(clients.ContentType))
+
 	return newServiceCommand(device, httpCaller, request, lc), nil
 }
+
+func addHeadersToRequest(r *http.Request, value string) {
```
AnthonyMBonafide: The idea for my previous comment in favor of this helper function was so that this function would encapsulate all the business logic within a single place which handles all the headers that need to be added to the out going request. Could we make this function add `Content-type` and `Correlation`? If so, like my previous comment about this helper function, I think we should do this in a smart way which allows us to extend functionality within requiring modification. The reason I say this is because we very well may want to add or remove headers from the "proxied" request in the future. Therefore, I would like to do this a little smarter and avoid hard-coding which headers are being propagated. See my previous comment from a previous iteration for more details and a starting idea as to how to accomplish this.

internal/core/command/put.go
```diff
@@ -44,5 +50,7 @@ func NewPutCommand(
 		request.Header.Set(clients.CorrelationHeader, correlationID.(string))
 	}
 
+	addHeadersToRequest(request, r.Header.Get(clients.ContentType))
```
AnthonyMBonafide: See my previous comment regarding this function.

internal/core/command/get.go
AnthonyMBonafide: Could be move this function to a more common place? This is being used by `put.go` as well and may better serve in a common location.

akramtexas:
Rebased

internal/core/command/device.go
```diff
@@ -16,28 +16,41 @@ package command
...
+	deviceClient metadata.DeviceClient,
+	httpCaller internal.HttpCaller) (string, error) {
+
+	var ctx context.Context
+
+	if r == nil {
+		return "", errors.NewErrExtractingInfoFromRequest()
+	}
+
+	ctx = r.Context()
```
AnthonyMBonafide, line 49: Can we use `:=` here and eliminate the declaration(`var`) above to simplify the code and make it consistent with the rest of the code base?

internal/core/command/rest_device.go
```diff
@@ -96,7 +87,9 @@ func issueDeviceCommand(
 		return
 	}
 
-	w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
+	// Set the response header Content-type (No need to inspect it).
+	w.Header().Set(clients.ContentType, r.Header.Get(clients.ContentType))
```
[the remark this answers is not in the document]
akramtexas: @AnthonyMBonafide I will follow up on creating a new issue to handle different response types for core-command and the device service, basically like we are doing here for the request 👍

internal/core/command/device_test.go
```diff
@@ -15,138 +15,173 @@
...
+	client.On("Device", mock.Anything, d200c200).Return(contract.Device{Id: d200c200}, nil)
 	return &client
 }
 
 func newMockDBClient() interfaces.DBClient {
 	dbMock := &mocks.DBClient{}
-	dbMock.On("GetCommandsByDeviceId", status400).Return(nil, db.ErrNotFound)
-	dbMock.On("GetCommandsByDeviceId", status500).Return(nil, goErrors.New("unexpected error"))
+	dbMock.On("GetCommandsByDeviceId", invalidStatus400).Return(nil, db.ErrNotFound)
+	dbMock.On("GetCommandsByDeviceId", invalidStatus500).Return(nil, goErrors.New("unexpected error"))
```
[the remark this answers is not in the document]
akramtexas: Renamed variables to make them intention-revealing 👍

akramtexas:
Recheck

akramtexas:
- Noticed an error, unrelated to this `PR`, in the `Jenkins` build:
```
[2020-04-02T22:50:09.777Z] go: finding gopkg.in/yaml.v2 v2.2.8
[2020-04-02T22:50:09.777Z] build github.com/edgexfoundry/edgex-go/cmd/config-seed: cannot load golang_org/x/crypto/chacha20poly1305: malformed module path "golang_org/x/crypto/chacha20poly1305": missing dot in first path element
[2020-04-02T22:50:09.777Z] Makefile:74: recipe for target 'test' failed
[2020-04-02T22:50:09.777Z] make: *** [test] Error 1
script returned exit code 2
```

internal/core/command/device_test.go
```diff
@@ -15,138 +15,173 @@
...
 		expectedErr error
 	}{
 		{
 			"Device-InvalidObjectId",
-			status400,
-			TestCommandId,
+			badRequestStatus400,
+			createDeviceRequestWithPathParameters(
+				createTestDeviceWithPathUrl(TestCommandId, badRequestStatus400),
+				map[string]string{ID: badRequestStatus400, COMMANDID: TestCommandId}),
```
AnthonyMBonafide, line 86: I missed this in my last comment, but can we change this `ID` variable to be something more clear. See my previous round of comments for more information.

internal/core/command/device_test.go
```diff
@@ -15,138 +15,173 @@
...
 			types.NewErrServiceClient(400, []byte("Invalid object ID")),
 		},
 		{
 			"Device-NotFound",
-			status404,
-			TestCommandId,
+			notFoundStatus404,
+			createDeviceRequestWithPathParameters(
+				createTestDeviceWithPathUrl(TestCommandId, notFoundStatus404),
+				map[string]string{ID: notFoundStatus404, COMMANDID: TestCommandId}),
```
AnthonyMBonafide, line 94: See my other comment regarding changing the `ID` variable name.

internal/core/command/device_test.go
```diff
@@ -15,138 +15,173 @@
...
 			types.NewErrServiceClient(http.StatusNotFound, []byte{}),
 		},
 		{
 			"Device-InternalServerErr",
-			status500,
-			TestCommandId,
+			internalServerErrorStatus500,
+			createDeviceRequestWithPathParameters(
+				createTestDeviceWithPathUrl(TestCommandId, internalServerErrorStatus500),
+				map[string]string{ID: internalServerErrorStatus500, COMMANDID: TestCommandId}),
```
AnthonyMBonafide, line 102: See my other comment regarding changing the `ID` variable name.

internal/core/command/device.go
```diff
@@ -94,24 +119,33 @@ func executeCommandByName(
...
+	var method string
 	var ex Executor
 	var err error
-	if isPutCommand {
-		ex, err = NewPutCommand(device, command, body, ctx, &http.Client{}, lc)
-	} else {
-		ex, err = NewGetCommand(device, command, queryParams, ctx, &http.Client{}, lc)
+
+	if r == nil || r.Method == "" {
+		return "", errors.NewErrRequestParameterSupplied("")
```
AnthonyMBonafide, line 139: Is this empty string intentional? If it is not used should we remove it?

internal/core/command/utils.go
```diff
@@ -0,0 +1,28 @@
...
+	"net/http"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+)
+
+// addHeadersToRequest populates deviceServiceRequest with matching headers from
+// originalRequest to retain information that can be used by the device service.
+func addHeadersToRequest(originalRequest *http.Request, deviceServiceRequest *http.Request, context context.Context) {
+
+	var propagatedHeader = []string{clients.ContentType}
```
AnthonyMBonafide, line 14: Can we move this outside of this function so it acts more like a constant? Then we can comment what it is used for and future developers can simply just add or remove from this slice to get the desired affect.

internal/core/command/device.go
```diff
@@ -66,19 +77,33 @@ func executeCommandByDeviceID(
 		return "", errors.NewErrCommandNotAssociatedWithDevice(commandID, deviceID)
 	}
 
-	return executeCommandByDevice(ctx, d, c, body, queryParams, isPutCommand, lc)
+	return executeCommandByDevice(ctx, d, c, body, lc, r, httpCaller)
+}
+
+// Extract Device ID and Command ID from the HTTP request, returning an
+// error if either (Device ID or Command ID) is not successfully extracted.
```
AnthonyMBonafide, line 84: Since we have other changes that need to happen before we merge, can we update this to be a proper GoDoc comment?

internal/core/command/device_test.go
akramtexas: No worries, great feedback. I have refactored the names of this as well as the other `ID` variables for conciseness and clarity. Since refactoring variable names is necessarily subjective, should this update not align with what you have in mind, I'll be glad to refactor these variable names again so as to make the unit test as readable and as easy as possible for other users to follow along 👍

internal/core/command/get_test.go
```diff
@@ -61,9 +64,11 @@ var testCommand = contract.Command{
...
 	expectedCorrelationIDHeaderValue := "Testing"
+	req := newRequestForGetWithContentType(clients.ContentTypeCBOR)
 	testContext := context.WithValue(context.Background(), clients.CorrelationHeader, expectedCorrelationIDHeaderValue)
-	getCommand, _ := NewGetCommand(testDevice, testCommand, "", testContext, nil, logger.NewMockClient())
+	getCommand, _ := NewGetCommand(testDevice, testCommand, testContext, nil, logger.NewMockClient(), req)
 	actualCorrelationIDHeaderValue := getCommand.(serviceCommand).Request.Header.Get(clients.CorrelationHeader)
+
 	if actualCorrelationIDHeaderValue == "" {
 		t.Errorf("The populated GetCommand's request should contain a correlation ID header value")
 	}
```
AnthonyMBonafide, line 74: This looks like it is duplicated in the test table for `TestNewGetCommandContentType`. If so can we remove this test?

internal/core/command/get_test.go
```diff
@@ -108,9 +118,51 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+	req := newRequestForGetWithContentType(clients.ContentTypeCBOR)
 	device.Service.Addressable.Address = "!@#$"
-	_, err := NewGetCommand(device, testCommand, "", context.Background(), nil, logger.NewMockClient())
+	_, err := NewGetCommand(device, testCommand, context.Background(), nil, logger.NewMockClient(), req)
 	if err == nil {
 		t.Errorf("The invalid URL error was not properly propagated to the caller")
 	}
 }
+
+func TestNewGetCommandContentType(t *testing.T) {
```
AnthonyMBonafide, line 129: Can we add negative test cases so that we ensure headers that we do not expect to be propagated are not? Also, to my other comment regarding the behavior of the function which copies headers to the proxied request, can we add a test which does not have a `Content-type` header and ensure that it is not present in the outgoing request?

internal/core/command/put_test.go
```diff
@@ -17,16 +17,21 @@ package command
...
+	"strings"
 	"testing"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestNewPutCommandWithCorrelationId(t *testing.T) {
```
AnthonyMBonafide, line 30: This looks like it is duplicated in the test `TestNewPutCommandContentType`. If so can we remove this?

internal/core/command/put_test.go
```diff
@@ -73,3 +82,45 @@ func TestNewPutCommandBody(t *testing.T) {
 		t.Error("Failed to verify the request body contents")
 	}
 }
+
+func TestNewPutCommandContentType(t *testing.T) {
```
AnthonyMBonafide, line 86: Please refer to my other comment regarding adding test coverage for different negative test cases.

internal/core/command/router.go
```diff
@@ -103,7 +103,8 @@ func loadDeviceRoutes(b *mux.Router, dic *di.Container) {
 				bootstrapContainer.LoggingClientFrom(dic.Get),
 				container.DBClientFrom(dic.Get),
 				commandContainer.MetadataDeviceClientFrom(dic.Get),
-				errorContainer.ErrorHandlerFrom(dic.Get))
+				errorContainer.ErrorHandlerFrom(dic.Get),
+				&http.Client{})
```
AnthonyMBonafide, line 107: Would it be better to put this into the dependency injection container(dic) and retrieve the value like we are doing for the other components? However, I am wondering if this would create a race condition since we can only have one instance of `http.Client` in the `dic` and each time this REST handler is invoked it would be on a different go-routine and each of them would be using the same instance of `http.Client` and some of the methods look like there is some state being accessed for each call. Any thoughts?

internal/core/command/get_test.go
```diff
@@ -108,9 +118,78 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+	req, _ := http.NewRequest(http.MethodGet, "/", strings.NewReader("Test body"))
+	req.Header.Set("User-Agent", headerType)
+	return req
+}
+
+func newRequestForGetWithUnexpectedContentMD5Header(headerType string) *http.Request {
+	req, _ := http.NewRequest(http.MethodGet, "/", strings.NewReader("Test body"))
+	req.Header.Set("Content-MD5", headerType)
+	return req
+}
```
AnthonyMBonafide, line 195: Can we consolidate these functions to one general function which accepts the header name and value? Furthermore, I think you can make the tests cover more functional cases by having the general function accept multiple headers that way we can test cases where there are multiple headers. For example, requests with headers we want to propagate, and ones we do not want. The tests as they are only test that cases where only one header present, we have not really verified that the functionality properly handles multiple headers.

internal/core/command/put_test.go
```diff
@@ -73,3 +82,72 @@ func TestNewPutCommandBody(t *testing.T) {
...
+	req, _ := http.NewRequest(http.MethodGet, "/", strings.NewReader("Test body"))
+	req.Header.Set("User-Agent", headerType)
+	return req
+}
+
+func newRequestForPutWithUnexpectedContentMD5Header(headerType string) *http.Request {
+	req, _ := http.NewRequest(http.MethodGet, "/", strings.NewReader("Test body"))
+	req.Header.Set("Content-MD5", headerType)
+	return req
+}
```
AnthonyMBonafide, line 153: Please see my previous comment regarding handling all these cases in one more general function.

Also, there is code duplication between this file and `get_test.go` can we move this logic to a common place and leverage that to eliminate code duplication?

internal/core/command/router.go
AnthonyMBonafide: Maybe we can add a comment in the code stating that so our thoughts are documented.

internal/core/command/get_test.go
```diff
@@ -108,9 +116,99 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+				testDevice,
+				testCommand,
+				ctx,
+				nil,
+				loggerMock,
+				tt.req)
+			actualHeaders := map[string]string{}
+			for k := range tt.expectedHeaders {
+				actualHeaders[k] = getCommand.(serviceCommand).Request.Header.Get(k)
+			}
```
AnthonyMBonafide, line 208: This looks like we are not capturing all the `actualHeaders`. We are only getting the header values specified in the `expectedHeaders` Should we get all the headers from `getCommand` and inspect that the expected values are all present and not additional ones are added?

internal/core/command/get_test.go
```diff
@@ -108,9 +116,99 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+			"no content type header provided",
+			map[string]string{
+				clients.ContentType: ""},
+			newRequestWithHeaders(map[string]string{
+				clients.ContentType: ""}, http.MethodGet),
+		},
+		{
+			"user-agent header not propagated",
+			map[string]string{
+				"User-Agent": ""},
```
AnthonyMBonafide, line 157: For these negative cases could we use some uncommon header name, say `NonPropegatedHeaderName`? This way if we add `User-Agent`, `Content-MD5` in the future we will not have a series of tests which break.

internal/core/command/get_test.go
```diff
@@ -108,9 +116,99 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+			newRequestWithHeaders(map[string]string{
+				"User-Agent": "User-Agent",
```
AnthonyMBonafide, line 159: See my other comment regarding uncommon header names.

internal/core/command/get_test.go
```diff
@@ -108,9 +116,99 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+			}, http.MethodGet),
+		},
+		{
+			"content-md5 header not propagated",
+			map[string]string{
+				"Content-MD5": ""},
```
AnthonyMBonafide, line 165: See my other comment regarding uncommon header names.

internal/core/command/get_test.go
```diff
@@ -108,9 +116,99 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+			newRequestWithHeaders(map[string]string{
+				"Content-MD5": "Content-MD5",
```
AnthonyMBonafide, line 167: See my other comment regarding uncommon header names.

internal/core/command/get_test.go
```diff
@@ -108,9 +116,99 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+			}, ""),
+		},
+		{
+			"json content type propagated, content-md5 header not propagated",
+			map[string]string{
+				clients.ContentType: clients.ContentTypeJSON,
+				"Content-MD5":       ""},
```
AnthonyMBonafide, line 174: See my other comment regarding uncommon header names.

internal/core/command/get_test.go
```diff
@@ -108,9 +116,99 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+			newRequestWithHeaders(map[string]string{
+				clients.ContentType: clients.ContentTypeJSON,
+				"Content-MD5":       "Content-MD5"}, http.MethodGet),
```
AnthonyMBonafide, line 177: See my other comment regarding uncommon header names.

internal/core/command/get_test.go
```diff
@@ -108,9 +116,99 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+		},
+		{
+			"cbor content type propagated, user-agent header not propagated, content-md5 header not propagated",
+			map[string]string{
+				clients.ContentType: clients.ContentTypeCBOR,
+				"User-Agent":        "",
+				"Content-MD5":       "",
```
AnthonyMBonafide, line 184: See my other comment regarding uncommon header names.

internal/core/command/get_test.go
```diff
@@ -108,9 +116,99 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+			},
+			newRequestWithHeaders(map[string]string{
+				clients.ContentType: clients.ContentTypeCBOR,
+				"User-Agent":        "User-Agent",
+				"Content-MD5":       "Content-MD5",
```
AnthonyMBonafide, line 189: See my other comment regarding uncommon header names.

internal/core/command/put_test.go
```diff
@@ -73,3 +80,92 @@ func TestNewPutCommandBody(t *testing.T) {
...
+			"no content type header provided",
+			map[string]string{
+				clients.ContentType: ""},
+			newRequestWithHeaders(map[string]string{
+				clients.ContentType: ""}, http.MethodPut),
+		},
+		{
+			"user-agent header not propagated",
+			map[string]string{
+				"User-Agent": ""},
```
AnthonyMBonafide, line 114: See my other comment regarding uncommon header names.

internal/core/command/put_test.go
```diff
@@ -73,3 +80,92 @@ func TestNewPutCommandBody(t *testing.T) {
...
+			newRequestWithHeaders(map[string]string{
+				"User-Agent": "User-Agent",
```
AnthonyMBonafide, line 116: See my other comment regarding uncommon header names.

internal/core/command/put_test.go
```diff
@@ -73,3 +80,92 @@ func TestNewPutCommandBody(t *testing.T) {
...
+			}, http.MethodPut),
+		},
+		{
+			"content-md5 header not propagated",
+			map[string]string{
+				"Content-MD5": ""},
```
AnthonyMBonafide, line 122: See my other comment regarding uncommon header names.

internal/core/command/put_test.go
```diff
@@ -73,3 +80,92 @@ func TestNewPutCommandBody(t *testing.T) {
...
+			newRequestWithHeaders(map[string]string{
+				"Content-MD5": "Content-MD5",
```
AnthonyMBonafide, line 124: See my other comment regarding uncommon header names.

internal/core/command/put_test.go
```diff
@@ -73,3 +80,92 @@ func TestNewPutCommandBody(t *testing.T) {
...
+			}, ""),
+		},
+		{
+			"json content type propagated, content-md5 header not propagated",
+			map[string]string{
+				clients.ContentType: clients.ContentTypeJSON,
+				"Content-MD5":       ""},
```
AnthonyMBonafide, line 131: See my other comment regarding uncommon header names.

internal/core/command/put_test.go
```diff
@@ -73,3 +80,92 @@ func TestNewPutCommandBody(t *testing.T) {
...
+			newRequestWithHeaders(map[string]string{
+				clients.ContentType: clients.ContentTypeJSON,
+				"Content-MD5":       "Content-MD5"}, http.MethodPut),
```
AnthonyMBonafide, line 134: See my other comment regarding uncommon header names.

internal/core/command/put_test.go
```diff
@@ -73,3 +80,92 @@ func TestNewPutCommandBody(t *testing.T) {
...
+		},
+		{
+			"cbor content type propagated, user-agent header not propagated, content-md5 header not propagated",
+			map[string]string{
+				clients.ContentType: clients.ContentTypeCBOR,
+				"User-Agent":        "",
+				"Content-MD5":       "",
```
AnthonyMBonafide, line 141: See my other comment regarding uncommon header names.

internal/core/command/put_test.go
```diff
@@ -73,3 +80,92 @@ func TestNewPutCommandBody(t *testing.T) {
...
+			},
+			newRequestWithHeaders(map[string]string{
+				clients.ContentType: clients.ContentTypeCBOR,
+				"User-Agent":        "User-Agent",
+				"Content-MD5":       "Content-MD5",
```
AnthonyMBonafide, line 146: See my other comment regarding uncommon header names.

internal/core/command/put_test.go
```diff
@@ -73,3 +80,92 @@ func TestNewPutCommandBody(t *testing.T) {
...
+				testCommand,
+				"Test Body",
+				ctx,
+				nil,
+				loggerMock,
+				tt.req)
+			actualHeaders := map[string]string{}
+			for k := range tt.expectedHeaders {
+				actualHeaders[k] = putCommand.(serviceCommand).Request.Header.Get(k)
+			}
```
AnthonyMBonafide, line 165: See my other comment regarding testing all the values present in `putCommand` rather than just the header present from the list of `expectedHeaders`

internal/core/command/get_test.go
akramtexas: My apologies for the confusing way this was set up... I have fixed it.

internal/core/command/get_test.go
```diff
@@ -108,9 +116,145 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+			newRequestWithHeaders(map[string]string{
+				clients.ContentType: ""}, http.MethodGet),
+		},
+		{
+			"user-agent header not propagated, if-modified-since not propagated, max-forwards not propagated",
+			[]string{"User-Agent", "If-Modified-Since", "Max-Forwards"},
+			map[string]string{
+				"User-Agent":        "",
+				"If-Modified-Since": "",
+				"Max-Forwards":      "",
```
AnthonyMBonafide, line 164: I think my last comment discouraging the use commonly used headers was not communicated properly. Can we avoid using common HTTP headers and instead use some irrelevant string? The reason for this is that the functionality which propagates headers is done in a way which very easily allows to add or remove headers in the future and if we decide to add one that is explicitly not expected in the tests we will have a lot of test break, which means low maintainability. You can effectively achieve the same functional test coverage by using a "random" header name, like `NonPropegatedHeader`, so that way we will unlikely cause issues in the future and it is clear the intent of the test. Please apply this comment to the rest of the test which specify headers like this including the tests in `put_test.go`

internal/core/command/get_test.go
```diff
@@ -108,9 +116,145 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+				testDevice,
+				testCommand,
+				ctx,
+				nil,
+				loggerMock,
+				tt.req)
+			actualHeaders := map[string]string{}
+			for _, h := range tt.requestedHeaders {
+				actualHeaders[h] = getCommand.(serviceCommand).Request.Header.Get(h)
+			}
```
AnthonyMBonafide, line 254: I think I was not clear in my last comment regarding testing all the headers that are present in the `getCommand` request. With this logic you are only verifying the `requestHeaders` contents from the `getCommand` which are present in `tt.requestedHeaders`. To ensure that not additional headers are present you can get all of the headers from the `getCommand` request and compare it to the tt.expectedHeaders. This will allow you to also remove the headers which have empty strings from the test arguments, since the test will now fail properly.

Please apply this comment to `put_test.go` as well.

internal/core/command/get_test.go
```diff
@@ -108,9 +117,83 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
+		{
+			"json content type propagated, if-modified-since not propagated, content-encoding not propagated, cookie not propagated, max-forwards not propagated, random header not propagated",
+			map[string]string{clients.ContentType: clients.ContentTypeJSON, NonPropagatedHeader: "NonPropagatedHeader"},
+			map[string]string{clients.ContentType: clients.ContentTypeJSON},
+			newRequestWithHeaders(map[string]string{
+				clients.ContentType:   clients.ContentTypeJSON,
+				"If-Modified-Since":   "If-Modified-Since",
+				"Content-Encoding":    "Content-Encoding",
+				"Cookie":              "Cookie",
+				"Max-Forwards":        "Max-Forwards",
```
AnthonyMBonafide, line 171: As per our conversation can we remove these common HTTP headers from our tests?

internal/core/command/put_test.go
```diff
@@ -73,3 +80,77 @@ func TestNewPutCommandBody(t *testing.T) {
...
+		{
+			"json content type propagated, if-modified-since not propagated, content-encoding not propagated, cookie not propagated, max-forwards not propagated, random header not propagated",
+			map[string]string{clients.ContentType: clients.ContentTypeJSON, NonPropagatedHeader: "NonPropagatedHeader"},
+			map[string]string{clients.ContentType: clients.ContentTypeJSON},
+			newRequestWithHeaders(map[string]string{
+				clients.ContentType:   clients.ContentTypeJSON,
+				"If-Modified-Since":   "If-Modified-Since",
+				"Content-Encoding":    "Content-Encoding",
+				"Cookie":              "Cookie",
+				"Max-Forwards":        "Max-Forwards",
```
AnthonyMBonafide, line 127: As per our conversation can we remove these common HTTP headers from our tests?

internal/core/command/put_test.go
```diff
@@ -73,3 +80,77 @@ func TestNewPutCommandBody(t *testing.T) {
...
+		name            string
+		originalHeaders map[string]string
+		expectedHeaders map[string]string
+		req             *http.Request
+	}{
+		{
+			"cbor content type header propagated",
+			map[string]string{clients.ContentType: clients.ContentTypeCBOR},
+			map[string]string{clients.ContentType: clients.ContentTypeCBOR},
+			newRequestWithHeaders(map[string]string{clients.ContentType: clients.ContentTypeCBOR}, http.MethodPut),
```
AnthonyMBonafide, line 95: I forgot to mention this in the previous round of iteration, but since we are going to have another round of development can we add names to each of these parameters. For example:
```go
name: "cbor content type header propagated",
originalHeaders: map[string]string{clients.ContentType: clients.ContentTypeCBOR},
expectedHeaders: map[string]string{clients.ContentType: clients.ContentTypeCBOR},
req: newRequestWithHeaders(map[string]string{clients.ContentType: clients.ContentTypeCBOR}, http.MethodPut),
```

akramtexas:
Rebased

internal/core/command/get_test.go
```diff
@@ -108,9 +117,79 @@ func TestNewGetCommandNoCorrelationIDInContext(t *testing.T) {
...
 		t.Errorf("The invalid URL error was not properly propagated to the caller")
 	}
 }
+
+func TestNewGetCommandContentType(t *testing.T) {
+	tests := []struct {
+		name            string
+		originalHeaders map[string]string
+		expectedHeaders map[string]string
+		req             *http.Request
```
AnthonyMBonafide, line 133: Can we remove this field from the test struct, since it is not being used?

internal/core/command/put_test.go
```diff
@@ -73,3 +80,73 @@ func TestNewPutCommandBody(t *testing.T) {
 		t.Error("Failed to verify the request body contents")
 	}
 }
+
+func TestNewPutCommandContentType(t *testing.T) {
+	tests := []struct {
+		name            string
+		originalHeaders map[string]string
+		expectedHeaders map[string]string
+		req             *http.Request
```
AnthonyMBonafide, line 89: Can we remove this field from the test struct, since it is not being used?

internal/core/command/device.go
```diff
@@ -94,24 +120,33 @@ func executeCommandByName(
...
+	var method string
 	var ex Executor
 	var err error
-	if isPutCommand {
-		ex, err = NewPutCommand(device, command, body, ctx, &http.Client{}, lc)
-	} else {
-		ex, err = NewGetCommand(device, command, queryParams, ctx, &http.Client{}, lc)
+
+	if r == nil {
+		return "", errors.NewErrRequestParameterSupplied("")
```
AnthonyMBonafide, line 140: I brought this up in a previous iteration and I think it got lost in all the comments, but I think we should change this to fit the situation better. The `ErrRequestParamenterSupplied` is never constructed with any arguments and therefore the resulting error message is not helpful. Can we either update the error struct to have a message that does not refer to the parameters in question(since they are never supplied anyway) or pass in the the parameters you are expecting so the error message is useful. Furthermore, I do not think this error fits this situation anyway since you are looking for the HTTP method and not any parameters.

internal/core/command/get.go
```diff
@@ -30,11 +31,21 @@ import (
...
-	queryParams string,
 	context context.Context,
 	httpCaller internal.HttpCaller,
-	lc logger.LoggingClient) (Executor, error) {
-	urlResult := device.Service.Addressable.GetBaseURL() + strings.Replace(command.Get.Action.Path, DEVICEIDURLPARAM, device.Id, -1) + "?" + queryParams
+	lc logger.LoggingClient,
+	r *http.Request) (Executor, error) {
+
+	if r == nil {
+		return serviceCommand{}, errors.NewErrRequestParameterSupplied("")
```
AnthonyMBonafide, line 40: Adding to other comment about this error. I do not think this error fits this situation since it looks like query parameters are not mandatory anyway and if they are we should be able to specify which one(s) specifically are missing.

akramtexas:
@AnthonyMBonafide Based on the helpful discussion earlier today, it makes sense to look at the two flows you outlined as follows:
```
Postman ----> Request  ----> core-command ----> Request  ----> device-sdk-go
Postman <---- Response <---- core-command <---- Response <---- device-sdk-go
```
- In relooking the flows, I think we have all the information we need to set the returned header `Content-type` properly, based on the header `Content-type` that is received in the Device Service request. The content itself (CBOR or JSON) is (and was) being returned properly, and it was only the header Content-type that was hard-coded to JSON, and which is fixed in this push.
- I have not squashed the commits so that it remains easy to diff the changes going forward, at the same time giving the flexibility to revert a change.

internal/core/command/rest_device.go
```diff
@@ -96,7 +88,9 @@ func issueDeviceCommand(
 		return
 	}
 
-	w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
+	// Set the returned header Content-type based on header Content-type received in
+	// the Device Service request (No need to inspect it).
+	w.Header().Set(clients.ContentType, r.Header.Get(clients.ContentType))
```
AnthonyMBonafide, line 93: This functionality does not match the intent of the comment above. Here we are copying the `Content-type` header from the original request, not the device service's response. In our meeting we discussed this and why this is not the solution we are after. To achieve what we are after we need to get the headers from the response from the device service's response, which at this point, only exists in the underlying operator:
https://github.com/edgexfoundry/edgex-go/blob/93a900b48c21df2a0611ee8b9bdbe214b0c3fa76/internal/core/command/device.go#L121

It seems like it is becoming hard to track which request is which since we have 2 in the context of this operation. Can we properly name each request to make it more obvious what is going on? Using `r`, `req`, or `request` does not help the situation. I think naming the request received by core-command to something like `originalRequest` , and the request to device service's to something like `deviceServiceProxiedRequest` will make these types of inconsistencies more obvious.

akramtexas:
- Following a realignment in the approach (as related to the associated `PR` in `device-sdk-go`), the work underway in that `PR` (i.e. in `device-sdk-go`) will be completed first, after which this current `PR` will be brought to closure.
- Meanwhile, pushing to the relevant branch (in `edgex-go`) the work completed for addressing how we need to get the headers from the response from the response of the `device service`, which, at this point, only exists in the underlying operator (as in `responseBody, responseCode, err := ex.Execute()`).

brandonforster:
@akramtexas Sorry if I missed the messaging in standup, but is this ready for review or does it need work done in the device services repo first?

akramtexas:
> @akramtexas Sorry if I missed the messaging in standup, but is this ready for review or does it need work done in the device services repo first?

@brandonforster Yes, this is ready for review, thanks for checking 👍 This `PR` does _not_ depend on the work done in `device-sdk-go`, which will be handled separately, at a later time, per the following: https://github.com/edgexfoundry/device-sdk-go/issues/488

akramtexas:
Rebased

internal/core/command/device.go
```diff
@@ -63,70 +74,97 @@ func executeCommandByDeviceID(
...
-		return "", err
+		return nil, err
 	}
-	if responseCode != http.StatusOK {
-		return "", types.NewErrServiceClient(responseCode, []byte(responseBody))
+
+	buf := new(bytes.Buffer)
+	_, readErr := buf.ReadFrom(deviceServiceResponse.Body)
+	if readErr != nil {
+		return nil, readErr
```
AnthonyMBonafide, line 164: I think you have a bug here. You are reading the response and not returning it and causing the caller to read the body from the response later, which will fail because you can only read it one. This is backed with the fact that these changes do not work. Issuing a GET command to device virtual for binary data though core-command results in an empty response body with a 200 status code, with the proper header. This was not noticed at first due to the generic name of `buf`. Can we please name variables properly so we can spot bugs a little easier?

akramtexas: Absolutely, I'm on it 👍

internal/core/command/types_test.go
```diff
@@ -67,41 +68,50 @@ func TestExecute(t *testing.T) {
...
-		t.Errorf("The response body was not properly propegated to the caller")
+	// Extract the headers from the device service's response.
+	headers := map[string]string{"Content-Type": ""}
+	m := make(map[string][]string)
+	m = deviceServiceResponse.Header
+	for name, values := range m {
+		for _, value := range values {
+			headers[name] = value
+		}
 	}
```
AnthonyMBonafide, line 85: As a follow up to this issue and not to be handled now, I think we need to clean this up. Using names like `m` makes this hard to read and --as we seen in the iterations on this PR-- causes bugs that are only evident during manual testing.

akramtexas:
- Completed the functional testing (after addressing how we should not only read the device service response body the first time, but also propagate it at that time, because the response body can evidently be read only once, during the first time only).
- I rebased so as to pull in Daniel's `PR` (Swap ugorji/go with fxamacker/cbor library) at https://github.com/edgexfoundry/edgex-go/pull/2490 
- After rebasing, I do not see any issues during functional testing, though I do see an error getting logged (to console logging) for the following PUT command, which has a CBOR payload attached to it: http://localhost:48082/api/v1/device/9f872d68-2281-4af4-959d-29e4d51c2192/command/b349df4a-6c3d-4218-b8bc-7ff8d51ea50d

> level=ERROR ts=2020-04-16T20:34:01.895958Z app=device-virtual source=command.go:318 msg="Handler - execWriteCmd: Put parameters parsing failed: \ufffdx\u001aEnableRandomization_BinarydtruedPathx\u001e/api/v1/device/9f872d68/BinarycUrlxBhttp://localhost:48082/api/v1/device/9f872d68/command/7ff8d51ea50dbidx$9f872d68-2281-4af4-959d-29e4d51c2192dnametRandom-Binary-DevicejadminStatehUNLOCKEDnoperatingStategENABLED"

- This would point to the fact that, at some point in time, we should also swap ugorji/go with fxamacker/cbor library for `device-sdk-go`.

akramtexas:
- Some sample test results:

GET with CBOR response

GET with CBOR Content-Type

akramtexas:
- The work done in this `PR` stands _independently_ of the work that was done in the `device-sdk-go` repo, which was for related functionality (to handle both `JSON` request bodies as well as `CBOR` request bodies).
- That other work can be picked up by anyone interested in running with it. To that intent, I captured relevant details in https://github.com/edgexfoundry/device-sdk-go/issues/488

brandonforster:
How are you verifying this? I'm trying to verify using the command black box tests, and while it passes the error case tests, it's hanging on the positive tests.

Here's my steps, which are the same as the standard command black box test run:

1. `redis-cli FLUSHALL`
2. start core-data
3. start core-metadata
4. start core-command
5. start device-virtual
6. run core-command importer
7. run collection

Doing this with the error collection works, doing this with the positive collection hangs. Any guidance here on what I'm doing wrong? Are you seeing this locally?

akramtexas:
> How are you verifying this? I'm trying to verify using the command black box tests, and while it passes the error case tests, it's hanging on the positive tests.
> 
> Here's my steps, which are the same as the standard command black box test run:
> 
> 1. `redis-cli FLUSHALL`
> 2. start core-data
> 3. start core-metadata
> 4. start core-command
> 5. start device-virtual
> 6. run core-command importer
> 7. run collection
> 
> Doing this with the error collection works, doing this with the positive collection hangs. Any guidance here on what I'm doing wrong? Are you seeing this locally?

Used the following steps for verifying: https://github.com/edgexfoundry/edgex-go/pull/2446#issuecomment-601930622

akramtexas:
@brandonforster Do please ping me as and when you have questions regarding the manual validation I've been doing - Also, and in addition to the manual verification steps (for which I've provided a link my reply above), I did also run the `core-command` black box tests:
black-box-command-tests

akramtexas:
@brandonforster @AnthonyMBonafide - With approvals from both you gents, I'll proceed to merge this feature branch before the need arises to go through another `rebase` cycle 😅

PR #2450, opened 2020-03-24
Race condition when initializing external clients using bootstrap

### LINKED ISSUE #2421: Empty reply from server when getting Core-Command device
# 🐞 Bug Report

### Affected Services

 The issue is located in: core-command

### Is this a regression?

 - 

1. download the [nightly-build docker-compose file](https://github.com/edgexfoundry/developer-scripts/blob/master/releases/nightly-build/compose-files/docker-compose-nexus-redis-no-secty.yml) and use docker-compose up
`docker-compose -f docker-compose-nexus-redis-no-secty.yml up -d`
2. use curl to send request to http://localhost:48082/api/v1/device
`curl http://localhost:48082/api/v1/device`
3. return `curl: (52) Empty reply from server`

Sending the curl request again, it works properly.
## 🔥 Exception or Error

There is no error log in edgex-core-command container.

## 🌍  Your Environment
**Deployment Environment:**

**EdgeX Version:**
Master

**Anything else relevant?**
It could be reproduced by simply restarting command service and sending curl request again.

related: [edgexfoundry/edgex-go#2399](https://github.com/edgexfoundry/edgex-go/issues/2399)

brandonforster:
## PR Checklist

- [X] The commit message follows our guidelines: https://wiki.edgexfoundry.org/display/FA/Contributor%27s+Guide

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

A race occurs during bootstrap external client initialization between the bootstrap handler and the bootstrap registry component. If the external client is initialized JIT, the registry variable will no longer contain valid data. On subsequent calls, the registry variable is updated and execution proceeds as normal.

Issue Number: Fixes #2421 

## What is the new behavior?
The external client is initialized before the bootstrap update code and the race does not occur.

- [ ] Yes
- [X] No

To verify:

Repro the bug on master:
Run (in order) consul, redis, config-seed, core-data, core-metadata, core-command. Ensure the core services are configured for consul usage and not local configuration. Make the following curl command: `curl http://localhost:48082/api/v1/device` .

Checkout this PR code, follow the steps above, and ensure that device data is returned.

## Other information
[The system management agent contains code that is very similar to the problematic code identified in this PR in command and metadata. I do not have the familiarity with the SMA to say if it is going to be an issue there.](https://github.com/edgexfoundry/edgex-go/blob/master/internal/system/agent/init.go#L68-L107)

### DIFF
--- a/cmd/core-command/res/configuration.toml
+++ b/cmd/core-command/res/configuration.toml
@@ -10,7 +10,7 @@ Port = 48082
 Protocol = 'http'
 MaxResultCount = 50000
 StartupMsg = 'This is the Core Command Microservice'
-Timeout = 5000
+Timeout = 45000
 
 [Registry]
 Host = 'localhost'
--- a/internal/core/command/init.go
+++ b/internal/core/command/init.go
@@ -49,6 +49,13 @@ func NewBootstrap(router *mux.Router) *Bootstrap {
 func (b *Bootstrap) BootstrapHandler(ctx context.Context, wg *sync.WaitGroup, _ startup.Timer, dic *di.Container) bool {
 	loadRestRoutes(b.router, dic)
 
+	// TODO: there is an outstanding known issue (https://github.com/edgexfoundry/edgex-go/issues/2462)
+	// 		that could be seemingly be solved by moving from JIT initialization of these external clients to static
+	// 		init on startup, like registryClient and configuration are initialized.
+	// 		Doing so would cover over the symptoms of the bug, but the root problem of server processing taking longer
+	// 		than the configured client time out would still be present.
+	// 		Until that problem is addressed by larger architectural changes, if you are experiencing a bug similar to
+	//		https://github.com/edgexfoundry/edgex-go/issues/2421, the correct fix is to bump up the client timeout.
 	registryClient := bootstrapContainer.RegistryFrom(dic.Get)
 	configuration := container.ConfigurationFrom(dic.Get)
 	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
--- a/internal/core/command/rest_device.go
+++ b/internal/core/command/rest_device.go
@@ -201,7 +201,7 @@ func restGetCommandsByDeviceID(
 
 	w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
 	w.WriteHeader(http.StatusOK)
-	json.NewEncoder(w).Encode(&device)
+	_ = json.NewEncoder(w).Encode(&device)
 }
 
 func restGetCommandsByDeviceName(
@@ -230,7 +230,7 @@ func restGetCommandsByDeviceName(
 
 	w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
 	w.WriteHeader(http.StatusOK)
-	json.NewEncoder(w).Encode(&devices)
+	_ = json.NewEncoder(w).Encode(&devices)
 }
 
 func restGetAllCommands(
@@ -257,5 +257,5 @@ func restGetAllCommands(
 
 	w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
 	w.WriteHeader(http.StatusOK)
-	json.NewEncoder(w).Encode(devices)
+	_ = json.NewEncoder(w).Encode(devices)
 }
--- a/internal/core/metadata/init.go
+++ b/internal/core/metadata/init.go
@@ -48,6 +48,13 @@ func NewBootstrap(router *mux.Router) *Bootstrap {
 func (b *Bootstrap) BootstrapHandler(ctx context.Context, wg *sync.WaitGroup, _ startup.Timer, dic *di.Container) bool {
 	loadRestRoutes(b.router, dic)
 
+	// TODO: there is an outstanding known issue (https://github.com/edgexfoundry/edgex-go/issues/2462)
+	// 		that could be seemingly be solved by moving from JIT initialization of these external clients to static
+	// 		init on startup, like registryClient and configuration are initialized.
+	// 		Doing so would cover over the symptoms of the bug, but the root problem of server processing taking longer
+	// 		than the configured client time out would still be present.
+	// 		Until that problem is addressed by larger architectural changes, if you are experiencing a bug similar to
+	//		https://github.com/edgexfoundry/edgex-go/issues/2421, the correct fix is to bump up the client timeout.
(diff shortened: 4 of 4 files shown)


brandonforster:
@michaelestrin We absolutely have, and it's a fair point, but then should we change core-data (which is set up like this already) to do JIT *right now*? I believe we should harmonize all the services to do JIT, but I think committing code that we know will fail black-box tests is irresponsible. To that end, I think the best course of action here is to merge this PR, acknowledge there is a race (but one we will always win) and work on rebuilding the registry client to work the same way the URL client does to address the underlying race. I do not think that leaving black box tests broken for any period of time (like we did for the URL client) should be the course of action here. I think the master branch should always be passing blackbox tests, even if it requires a bandaid that we need to replace with something permanent.

michaelestrin:
You're pushing code around and don't understand (or can't adequately describe) the root cause.  If you don't understand the root cause, you can't be sure you'll always win the race.  

This appears to be a result of a shortcoming with recent registry client changes.  Maybe find the root cause and fix it?

I stand by my prior statement.

michaelestrin:
FWIW, I'm not attacking you @brandonforster nor am I advocating for irresponsibility and broken black box tests.  I'm not saying you caused the problem or are somehow responsible for the current state of the code.  I'm simply saying the underlying root cause should be identified and addressed to prevent future similar issues from cropping up.  There are places in the code -- specifically in the agent -- where JIT can't be avoided.

brandonforster:
@michaelestrin I know you're not making a personal attack on me and I know you're not attacking my credentials or anything like that. You were correct about the best way to attack this problem from a root cause perspective on the URL client, and the code is now much better for it. All those points you made for the URL client apply 1:1 here and I have no doubt that undertaking that work you are pushing for would have the same dramatic improvement.

However, that work takes time I am just very concerned that we have not been passing black box tests on master since the bug that started the URL client work. After we added the URL client, there was a bug in the URL client code, and now there is this bug in registry that looks like a root cause solution could potentially be just as time consuming. I am not advocating for not doing the root cause work, au contraire, I would love to spearhead it. I just want the system to pass tests and not have these weird hiccups, and I don't think a `//TODO redo this` in service of passing our API tests is a bad thing.

AnthonyMBonafide:
I briefly looked into this and I think we should focus our efforts on finding the root cause and fixing it. To address @brandonforster concern regarding the time it would take to implement a long term fix, how about we start looking into correcting the issue in this PR and add new commits which address the long term issue. That way if we get to a point and we want the quick fix we can simply just take the original commit(s), and if we can come up with a long term fix within a reasonable time frame we can squash all the commits. This would require some additional organization on your part as the author. If you decide to take that path, you can squash you 2 commits into one and update the commit message to describe that it is a quick fix. Then for the following commits you can add to that and we can work from there once we need to make a decision as what to do. I think we can get this done within a reasonable time frame. Race conditions and concurrency problems are horrible to find and fix, if we have a handle on it now it would be nice to remove one more complex issue from the system. I hope that all makes sense.

AnthonyMBonafide:
Also, if you need any assistance with tracking down the issue and/or coming up with a solution please feel free to reach out. I am confident we can get this worked out.

brandonforster:
@AnthonyMBonafide Sounds like a good approach. If the scope starts to look like the URL client refactor, we can reassess, but you're right to not worry about complexity we don't necessarily need yet.

brandonforster:
I spoke with @AnthonyMBonafide, and it seems like the lazy initialization is causing a huge spin up penalty. The registry client has to be initialized, the URL client has to be initialized, we have to wait for the first URL poll, and we have to fetch that URL update on a successful poll and report it back down, all while the request is still in flight.

I've found that the underlying problem here isn't with the registry client, the bootstrap, the URL client or any of our code, it's that all that code and work that has to be done takes too long to accomplish while the request is in flight, and the HTTPResponseWriter times out while it's waiting for it to finish. Observing the current function in the debugger, everything returns successfully and with valid device data, but the endpoint function falls over on the call to `json.NewEncoder(w).Encode(devices)` and returns the following error on my machine:

`write tcp 127.0.0.1:48082->127.0.0.1:52314: i/o timeout`

This causes the `empty response from server`, which goes away after the first call because all the clients have already been initialized.


I believe that the best course of action is to remove lazy initialization of all clients, and take the time hit when it's least noticeable, during startup. This way, when a REST call is actually made, the HTTP stack is going to be the limiting factor in timing and not any of our business code, as it is today.

@michaelestrin, I'd really like your thoughts here since I know we were both pushing for lazy initialization.

brandonforster:
Note that as part of the current changeset, I have renamed restDevice.go to rest_device.go to bring it in line with the other core service's naming of similar files.

michaelestrin:
> I believe that the best course of action is to remove lazy initialization of all clients, and take the time hit when it's least noticeable, during startup.

As I said before, this may work with core and support services but will not work with the agent.  It's also limiting in that it prevents dynamic service discovery which will be a future issue when high-availability support is implemented (which seems to be gathering some steam).

@akramtexas ran into a similar issue recently with a new blackbox test he wrote taking longer than the socket timeout.  This was ultimately solved by a default configuration change to lengthen the timeout.  You might look into that as the long-term solution.  Not sure that a 5 second timeout lengthened to 6..9 seconds would be a big deal.  

I also wonder if this is a universal problem or is being contributed to by the work being done by the use case you're invoking that invokes a client call.

Anyway, that's my 2 cents @brandonforster.

AnthonyMBonafide:
@michaelestrin, I think I may be missing some important piece here. With @brandonforster proposed solution, I am not seeing how this is will cause and issue for dynamic service discovery. As I see it he is suggesting that we still use the dependency injection container to pass around the pointer to the underlying client, which can still be updated in the background when a URL update happens within Consul. However, we would kickoff the lookup for the URL during startup of the application so we are not doing it at the last possible moment, which is responding to the first request which indirectly needs the client. To me that seems like something we would want to do anyway since getting the client(s) ready for use is something that should be taken care of before the service has started. I hope my understanding is correct.

As for system management agent, is there a reason we are not using the same initialization and update strategy as the other core-services?

brandonforster:
Updated with changes @michaelestrin and @AnthonyMBonafide discussed on a call.

brandonforster:
Sorry about all the pushes. I updated the timeout value to 45,000 ms, because I realized that the URLClient polling interval is currently defined as 15,000 ms, and we should allow for a few polling updates before timeout.

Issue #2266, opened 2020-01-07
Ability to easily add additional Vault tokens for app services


bnevis-i:
# 🚀 Feature Request

### Relevant Package
security-file-token-provider

### Description
Application services (and device services) are easily added to EdgeX by adding a docker container running the app or device service.  The default list of Vault tokens is generated from a static configuration file that is bundled inside of the security-secretstore-setup container.  This makes it very difficult to add additional services that may need access to the secret store.  A mechanism is needed to EASILY specify additional services for which tokens need to be created.

### Describe the solution you'd like
security-file-token-provider should take an environment variable, `ADD_SECRETSTORE_TOKENS` that takes a comma-separated list of additional service keys for which to create tokens.  The tokens will be created with the default token parameters and policies.

cc: @lenny-intel, @rsdmike, @jim-wang-intel, @eno-intel

PR #2459, opened 2020-03-25
Implementation for adding addtional Vault tokens from the environment variable

jim-wang-intel:
- security-file-token-provider now can take the environment variable,
     ADD_SECRETSTORE_TOKENS that takes a comma-separated list of additional service keys
     for which to create tokens.
 - The tokens create with the default token parameters and policies.
 - Unit tests refactored to reduce some duplicate codes

Fix: #2266

## PR Checklist

- [x] The commit message follows our guidelines: https://wiki.edgexfoundry.org/display/FA/Contributor%27s+Guide

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The current security-token-file-provider can only take the pre-configured services from the TOML configuration file and it is asked for adding additional service token on the fly from the environment variable.

Issue Number: #2266 

## What is the new behavior?
With this PR, the additional Vault tokens can be created via the environment variable `ADD_SECRETSTORE_TOKENS`.  This suits the need for application function related services, which is more dynamic configured.

- [ ] Yes
- [x] No

No

## Other information
To test this locally, do the following steps:
1. Clone this branch
2. Build the docker image of `security_secretstore_setup` via make from the base of this repo: 
```
$ make docker_security_secretstore_setup
```
It should produce the docker image: edgexfoundry/docker-edgex-security-secretstore-setup-go:master-dev
3. Modify the docker-compose-nexus-mongo.yml from the developer-script to use the built docker image from local and add the entry in the environment section of `vault-worker` like this example:
```yml
vault-worker:
    #image: nexus3.edgexfoundry.org:10004/docker-edgex-security-secretstore-setup-go:master
    image: edgexfoundry/docker-edgex-security-secretstore-setup-go:master-dev
    container_name: edgex-vault-worker
    hostname: edgex-vault-worker
    environment:
      - "SECRETSTORE_SETUP_DONE_FLAG=/tmp/edgex/secrets/edgex-consul/.secretstore-setup-done"
      - "ADD_SECRETSTORE_TOKENS=test-env-add-service"
    networks:
      edgex-network:
        aliases:
            - edgex-vault-worker


```
Here we are adding `test-env-add-service` as an additional or extra service for creating Vault token of it.
4. Run docker-compose up of `vault-worker`  eg.:
```sh
$ docker-compose -f docker-compose-nexus-mongo.yml up vault-worker
edgex-files is up-to-date
edgex-secrets-setup is up-to-date
edgex-core-consul is up-to-date
edgex-vault is up-to-date
Recreating edgex-vault-worker ... done
Attaching to edgex-vault-worker
edgex-vault-worker        | Clearing secretstore-setup completion flag
edgex-vault-worker        | creating /vault/config/assets
edgex-vault-worker        | + exec /consul/scripts/health security-secrets-setup
edgex-vault-worker        | starting vault-worker...
edgex-vault-worker        | passing
edgex-vault-worker        | Initializing secret store
edgex-vault-worker        | level=INFO ts=2020-03-25T23:23:09.456239697Z app=edgex-security-secretstore-setup source=file.go:60 msg="Loaded configuration from ./res/configuration.toml"
edgex-vault-worker        | Creating directory: /logs
edgex-vault-worker        | level=INFO ts=2020-03-25T23:23:09.456365624Z app=edgex-security-secretstore-setup source=init.go:67 msg="using certificate verification for secret store connection"
edgex-vault-worker        | level=INFO ts=2020-03-25T23:23:09.460556844Z app=edgex-security-secretstore-setup source=requestor.go:63 msg="successful loading the rootCA certificate."
edgex-vault-worker        | level=INFO ts=2020-03-25T23:23:09.50545955Z app=edgex-security-secretstore-setup source=common.go:97 msg="successfully made request to health check"
edgex-vault-worker        | level=INFO ts=2020-03-25T23:23:09.50704265Z app=edgex-security-secretstore-setup source=vault.go:64 msg="vault health check HTTP status: StatusCode: 200"
<.... Omit due to lengthy printout ...>
edgex-vault-worker        | Executing custom command: 
edgex-vault-worker        | Signaling secretstore-setup completion
edgex-vault-worker        | Waiting for termination signal
```
5. Inspect the `/tmp/edgex/secrets` directory, and it should create a new directory for that named service from the environment variable. ie. `test-env-add-service` like this:
```sh
$ sudo ls -al /tmp/edgex/secrets/test-env-add-service
total 12
drwx------  2 root root 4096 Mar 25 16:23 .
drwxr-xr-x 21 root root 4096 Mar 25 16:23 ..
-rw-------  1 root root  515 Mar 25 16:23 secrets-token.json
```
6. Compose down if needed: 
```
docker-compose -f docker-compose-nexus-mongo.yml down -v
```

### DIFF
--- a/internal/security/fileprovider/provider.go
+++ b/internal/security/fileprovider/provider.go
@@ -77,14 +77,25 @@ func (p *fileTokenProvider) Run() error {
 		return err
 	}
 
+	tokenConfEnv, err := GetTokenConfigFromEnv()
+	if err != nil {
+		p.logger.Error(fmt.Sprintf("failed to get token config from environment variable %s with error: %s", addSecretstoreTokensEnvKey, err.Error()))
+		return err
+	}
+
 	var tokenConf TokenConfFile
 	if err := LoadTokenConfig(p.fileOpener, p.tokenConfig.ConfigFile, &tokenConf); err != nil {
 		p.logger.Error(fmt.Sprintf("failed to read token configuration file %s: %s", p.tokenConfig.ConfigFile, err.Error()))
 		return err
 	}
 
-	for serviceName, serviceConfig := range tokenConf {
+	// merge the additional token configuration list from environment variable
+	// note that the configuration file takes precedence, as the tokenConf will override
+	// the tokenConfEnv with same duplicate keys
+	// The tokenConfEnv only uses default settings.
+	tokenConf = tokenConfEnv.mergeWith(tokenConf)
 
+	for serviceName, serviceConfig := range tokenConf {
 		p.logger.Info(fmt.Sprintf("generating policy/token defaults for service %s", serviceName))
 
 		servicePolicy := make(map[string]interface{})
--- a/internal/security/fileprovider/provider_test.go
+++ b/internal/security/fileprovider/provider_test.go
@@ -36,6 +36,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 )
 
 /*
@@ -142,7 +143,7 @@ func expectedTokenFile(serviceName string) []byte {
 	var tokenResponse interface{}
 	setCreateTokenResponse(&tokenResponse)
 	b := new(bytes.Buffer)
-	json.NewEncoder(b).Encode(tokenResponse)
+	_ = json.NewEncoder(b).Encode(tokenResponse)
 	// Debugging note: take care to not write out the buffer or it will disturb the read pointer
 	return b.Bytes()
 }
@@ -247,51 +248,36 @@ func TestNoDefaultsCustomTokenParameters(t *testing.T) {
 
 // TestTokenUsingDefaults
 func TestTokenUsingDefaults(t *testing.T) {
-	// Arrange
-	mockLogger := logger.MockLogger{}
+	// Good cases:
+	// case to mimic normal service name from config file
+	err := runTokensWithDefault("myservice", "", t)
+	require.NoError(t, err)
+	// case to mimic normal service name with one additional service from env list
+	err = runTokensWithDefault("myservice2", "additional-service1", t)
+	require.NoError(t, err)
+	// case to mimic normal service name with two additional services from env list
+	err = runTokensWithDefault("myservice3", "additional-service1,new-servcie-2", t)
+	require.NoError(t, err)
+	// case to mimic normal service name with two additional services and empty service name from env list
+	err = runTokensWithDefault("myservice", "additional-service1,,new-servcie-2", t)
+	require.NoError(t, err)
+	// case to mimic normal service name with one additional service and leading + trailing commas from env list
+	err = runTokensWithDefault("myservice", ",addtionalservice1,", t)
+	require.NoError(t, err)
+	// case to mimic normal service name with some additional services with special charater names from env list
+	err = runTokensWithDefault("myservice", "test.service,test~name", t)
+	require.NoError(t, err)
 
-	mockFileIoPerformer := &MockFileIoPerformer{}
-	expectedService1Dir := filepath.Join(outputDir, "myservice")
-	expectedService1File := filepath.Join(expectedService1Dir, outputFilename)
-	service1Buffer := new(bytes.Buffer)
-	mockFileIoPerformer.On("MkdirAll", expectedService1Dir, os.FileMode(0700)).Return(nil)
-	mockFileIoPerformer.On("OpenFileReader", configFile, os.O_RDONLY, os.FileMode(0400)).Return(strings.NewReader(`{"myservice":{"edgex_use_defaults":true}}`), nil)
-	mockFileIoPerformer.On("OpenFileWriter", expectedService1File, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, os.FileMode(0600)).Return(&writeCloserBuffer{service1Buffer}, nil)
-
-	mockAuthTokenLoader := &MockAuthTokenLoader{}
-	mockAuthTokenLoader.On("Load", privilegedTokenPath).Return("fake-priv-token", nil)
-
-	expectedService1Policy := `{"path":{"secret/edgex/myservice/*":{"capabilities":["create","update","delete","list","read"]}}}`
-	expectedService1Parameters := makeDefaultTokenParameters("myservice")
-	expectedService1Parameters["meta"] = makeMetaServiceName("myservice")["meta"]
-	mockSecretStoreClient := &MockSecretStoreClient{}
-	mockSecretStoreClient.On("InstallPolicy", "fake-priv-token", "edgex-service-myservice", expectedService1Policy).Return(http.StatusNoContent, nil)
-	mockSecretStoreClient.On("CreateToken", "fake-priv-token", expectedService1Parameters, mock.Anything).
-		Run(func(args mock.Arguments) {
-			setCreateTokenResponse(args.Get(2).(*interface{}))
-		}).
-		Return(http.StatusOK, nil)
-
-	p := NewTokenProvider(mockLogger, mockFileIoPerformer, mockAuthTokenLoader, mockSecretStoreClient)
-	p.SetConfiguration(secretstoreclient.SecretServiceInfo{}, config.TokenFileProviderInfo{
-		PrivilegedTokenPath: privilegedTokenPath,
-		ConfigFile:          configFile,
-		OutputDir:           outputDir,
-		OutputFilename:      outputFilename,
-	})
-
-	// Act
-	err := p.Run()
-
-	// Assert
-	// - {OutputDir}/myservice/{OutputFilename} w/proper contents
-	// - Correct token parameters for myservice
-	// - All other expectations met
-	assert.NoError(t, err)
-	mockFileIoPerformer.AssertExpectations(t)
-	mockAuthTokenLoader.AssertExpectations(t)
-	mockSecretStoreClient.AssertExpectations(t)
-	assert.Equal(t, expectedTokenFile("myservice"), service1Buffer.Bytes())
+	// Negative cases:
+	// case to mimic normal service name with an invalid service name from env list
+	err = runTokensWithDefault("myservice", "/service1,,\\new-servcie-2", t)
+	require.Error(t, err, "expect error due to invalid servcie name from the list in env")
+	// case to mimic normal service name with an invalid service name from env list
+	err = runTokensWithDefault("myservice", "../service1", t)
+	require.Error(t, err, "expect error due to invalid servcie name from the list in env")
+	// case to mimic normal service name with one additional service and URL unsafe characters from env list
+	err = runTokensWithDefault("myservice", "core:!@%#$&service*()+func[x]", t)
+	require.Error(t, err, "expect error due to invalid servcie name from the list in env")
 }
 
 // TestTokenFilePermissions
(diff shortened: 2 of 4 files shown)


jim-wang-intel:
cc: @eno-intel

internal/security/fileprovider/tokenconfig_test.go
```diff
@@ -78,3 +78,138 @@ func TestLoadTokenConfigError2(t *testing.T) {
...
+
+	merged = map1.mergeWith(map3)
+
+	assert.Equal(t, 4, len(merged), "expect 4 entries in merged map")
+	expectedKeys = []string{"key1", "key2", "key3", "key7"}
+	for _, k := range expectedKeys {
+		assert.True(t, merged.keyExists(k))
+	}
+
+	merged = merged.mergeWith(map2)
```
lenny-intel: Is this the duplicates test? Make it its own test and initialize ServiceKey{} differently for each dup so you can validate you get the one you expect.

internal/security/fileprovider/tokenconfig_test.go
```diff
@@ -78,3 +78,138 @@ func TestLoadTokenConfigError2(t *testing.T) {
...
+			os.Setenv(addSecretstoreTokensEnvKey, currentTest.serviceNameList)
+			tokenFileMap, err := GetTokenConfigFromEnv()
+
+			if !test.expectError && err != nil {
+				t.Errorf("Unexpected error: %s", err.Error())
+			}
+
+			if test.expectError && err == nil {
+				t.Errorf("Expect error but found nil")
+			}
```
lenny-intel: refactor to:
```
if test.expectError {
   require.Error(t,err)
   return  // test is complete
}

require.NoError(t, err)
....
```

internal/security/fileprovider/tokenconfig_test.go
```diff
@@ -78,3 +78,138 @@ func TestLoadTokenConfigError2(t *testing.T) {
...
+				found := false
+				for _, expectedServiceName := range currentTest.expectedKeys {
+					if expectedServiceName == k {
+						found = true
+					}
+				}
+				if !found {
+					t.Errorf("unexpected service name: %s", k)
+				}
+			}
```
lenny-intel: refactor as:
``` 
for _, expectedServiceName := range currentTest.expectedKeys {
   _,found :=tokenFileMap[expectedServiceName]
   assert.Truef(,t,found, "expected service name: %s", expectedServiceName )
   
}
```

The length check above handle any unexpected entries.

internal/security/fileprovider/provider_test.go
```diff
@@ -247,51 +247,12 @@ func TestNoDefaultsCustomTokenParameters(t *testing.T) {
...
-	// - All other expectations met
-	assert.NoError(t, err)
-	mockFileIoPerformer.AssertExpectations(t)
-	mockAuthTokenLoader.AssertExpectations(t)
-	mockSecretStoreClient.AssertExpectations(t)
-	assert.Equal(t, expectedTokenFile("myservice"), service1Buffer.Bytes())
+	_ = runTokensWithDefault("myservice", "", t)
+	_ = runTokensWithDefault("myservice", "additional-service1", t)
+	_ = runTokensWithDefault("myservice", "additional-service1,new-servcie-2", t)
+	if err := runTokensWithDefault("myservice", "additional-service1,,new-servcie-2", t); err == nil {
```
lenny-intel: use require.Error instead of conditional.
```
err := runTokensWithDefault("myservice", "additional-service1,,new-servcie-2", t)
require.Error(t,err, "expected error due to empty service name from the list in env")
```

internal/security/fileprovider/provider_test.go
```diff
@@ -406,3 +367,113 @@ func (wcb *writeCloserBuffer) Chmod(mode os.FileMode) error {
...
+		return errFromEnv
+	}
+
+	// Assert
+	// - {OutputDir}/myservice/{OutputFilename} w/proper contents
+	// - Correct policy for serviceName
+	// - Correct token parameters for serviceName
+	// - All other expectations met
+
+	assert.NoError(t, err)
```
lenny-intel: Should this be a require? i.e. does it make sense to continue the test if an error occurred?

internal/security/fileprovider/provider.go
```diff
@@ -77,12 +77,21 @@ func (p *fileTokenProvider) Run() error {
...
+	}
+
 	var tokenConf TokenConfFile
 	if err := LoadTokenConfig(p.fileOpener, p.tokenConfig.ConfigFile, &tokenConf); err != nil {
 		p.logger.Error(fmt.Sprintf("failed to read token configuration file %s: %s", p.tokenConfig.ConfigFile, err.Error()))
 		return err
 	}
 
+	// merge the addtional token configuration list from environment variable
+	tokenConf = tokenConfEnv.mergeWith(tokenConf)
```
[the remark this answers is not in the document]
jim-wang-intel: ok

internal/security/fileprovider/tokenconfig.go
```diff
@@ -80,3 +88,77 @@ func LoadTokenConfig(fileOpener fileioperformer.FileIoPerformer, path string, to
...
+// to return a new TokenConfFile.
+// The merging is based on the key of TokenConfFile's map
+// if the key of another has already been existing on the current one,
+// then the former will replaces the latter.
+func (tf TokenConfFile) mergeWith(another TokenConfFile) TokenConfFile {
+	if len(another) == 0 {
+		// nothing to be merged, return itself
+		return tf
+	}
+
```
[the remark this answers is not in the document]
jim-wang-intel: could do that and maybe that way is more efficient; functionality wise the current implementation is already taken care by the following few lines of code when `tf` is empty, though.

internal/security/fileprovider/tokenconfig_test.go
```diff
@@ -78,3 +78,138 @@ func TestLoadTokenConfigError2(t *testing.T) {
...
+
+	assert.Equal(t, 5, len(merged), "expect 5 entries in merged map")
+
+	// should get all keys in the merged map
+	expectedKeys := []string{"key1", "key2", "key3", "key4", "key5"}
+	for _, k := range expectedKeys {
+		assert.True(t, merged.keyExists(k))
+	}
+
+	// empty key case
```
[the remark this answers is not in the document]
jim-wang-intel: ok

internal/security/fileprovider/tokenconfig_test.go
jim-wang-intel: yes, it is, i'll make it as own test

internal/security/fileprovider/provider_test.go
jim-wang-intel: ok

internal/security/fileprovider/tokenconfig.go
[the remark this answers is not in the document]
lenny-intel: It isn't about efficiency. The way you are using this func, `tf`=from env and `another`=from config. `another` will never be empty since loaded from config, but `tf` could be.  All I am suggesting it to add a check of `tf` for empty like your are doing for `another` to cover if either are empty.

internal/security/fileprovider/provider.go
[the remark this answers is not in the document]
bnevis-i: The configuration file should take precedence, as it is more expressive than the environment.  The environment variable can only use default settings.  The on-disk configuration file can add custom policies and custom settings to the token.  If the token env var overrides the local configuration, then custom configuration will be lost.  Alternatively, we need to outright fail if there is a duplicate.

internal/security/fileprovider/provider_test.go
bnevis-i: Add a test case for `,addtionalservice1,`  (leading and trailing commas)

internal/security/fileprovider/provider_test.go
bnevis-i: This comment misplaced?  But agree that TestTokenUsingDefaults() has more than one test encoded inside of it.  Each test should be its own test function.  I don't see a point in this function returning error...

internal/security/fileprovider/tokenconfig.go
```diff
@@ -80,3 +88,77 @@ func LoadTokenConfig(fileOpener fileioperformer.FileIoPerformer, path string, to
...
+	// the list of service names is comma-separated
+	tokenConfigFromEnv := make(TokenConfFile)
+	serviceNameList := strings.Split(addTokenList, ",")
+	// regex for valid service name as key
+	// the service name eventually becomes part of the URL to Vault's API call
+	// / is not ok and backslack \ is not ok comma , is not ok
+	// and the the length of name is upto 512 characters
+	serviceNameRegx := regexp.MustCompile(`^[^.*\/\\\,]*[^\W]*[\w.]{1,512}$`)
```
bnevis-i: I suspect this regexp is more complicated than it needs to be.  \W is defined as [0-9A-Za-z_]  (we should also allow hyphens).  Better to list what IS allowed versus what is NOT allowed.  Doing both is very confusing.

internal/security/fileprovider/tokenconfig.go
```diff
@@ -80,3 +88,77 @@ func LoadTokenConfig(fileOpener fileioperformer.FileIoPerformer, path string, to
...
+
+	for _, name := range serviceNameList {
+		serviceName := strings.TrimSpace(name)
+		if serviceName == "" {
+			return emptyTokenConfig, errors.New("cannot have empty service name as key from environment variable")
```
bnevis-i: Why error out instead of just skipping empty service names?

internal/security/fileprovider/tokenconfig.go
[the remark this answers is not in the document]
bnevis-i: This function runs once per boot.  Is it really necessary to optimize it?

internal/security/fileprovider/provider_test.go
bnevis-i: NM about "require" -- didn't know that existed.  In any case, personally I like to see what mocks weren't called when an error is returned; it is a great debugging aid.

internal/security/fileprovider/provider.go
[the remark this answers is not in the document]
jim-wang-intel: I will add the comments as indicated above and the current implementation of `mergeWith` is acting as the desire, ie. the configuration file takes the precedence as the `tokenConf`(which is from configuration file) overrides the duplicate from the env. one

internal/security/fileprovider/tokenconfig.go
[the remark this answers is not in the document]
jim-wang-intel: it is just a simple check to do so I think we should do it.

internal/security/fileprovider/provider_test.go
jim-wang-intel: ok

internal/security/fileprovider/tokenconfig.go
jim-wang-intel: This is least complicated regex I can think of. and hyphens are allowed according to this regex.  You can use this online tool to verify hyphens working and understand each regex's group is doing. https://regex101.com/   
can list what is allowed

internal/security/fileprovider/tokenconfig.go
jim-wang-intel: can do skipping. it is just the requirement issue.

internal/security/fileprovider/provider_test.go
jim-wang-intel: The reason I added the return error for this helper function is that I'd like to cover the expecting error case and also at the same time the error returning from `GetTokenConfigFromEnv` function call is covered to increase the testing coverage.

internal/security/fileprovider/tokenconfig.go
bnevis-i: What does `^[\w.-]{1,512}$` allow that you don't want?

internal/security/fileprovider/tokenconfig.go
bnevis-i: The reason I mention in in case someone puts something like this in the docker-compose in order to build a larger list out of sublists:

EXTRA_SERVICES: $VENDOR1LIST,$VENDOR2LIST,$VENDOR3LIST

internal/security/fileprovider/provider_test.go
bnevis-i: ok

internal/security/fileprovider/tokenconfig.go
jim-wang-intel: that one only allows `\w` anything starting with word character and upto length 512. it is not sufficient for cases with some other special characters or white spaces like:
```
case::should be ok
this is  a test cases
edgex:core:service
```

bnevis-i:
```
^[\w.-: ]{1,512}$
```

jim-wang-intel: false positive on this case: `/ddd` which we want to avoid

internal/security/fileprovider/tokenconfig.go
jim-wang-intel: make sense

internal/security/fileprovider/tokenconfig.go
tingyuz: I am wondering if we should the regex part into a single function so that it can be tested and legal/malformed string can be included for future reference?

internal/security/fileprovider/tokenconfig_test.go
bnevis-i: We should also document the assumption being made here and why we are doing it the way we are doing it.

internal/security/fileprovider/tokenconfig.go
bnevis-i: The regexp site has a bug.  Go to https://play.golang.org/ and run the following:

```
package main

import (
	"fmt"
	"regexp"
)

func main() {
	serviceNameRegx := regexp.MustCompile(`^[\w.-]{1,512}$`)
	ismatch := serviceNameRegx.MatchString("/ddd")
	if ismatch {
		fmt.Println("is a match")
	} else {
		fmt.Println("no match")
	}
}
```

Output: no match

jim-wang-intel: ok that is a good catch but it still fails this `core:service` if just using `^[\w.-]{1,512}$`.  So it is not that simple.... see this: https://play.golang.org/p/E_Q69IZJTxc

bnevis-i: ... please give me some time to investigate ...

jim-wang-intel: > I am wondering if we should extra the regex part into a single function so that it can be tested and legal/malformed strings can be included for future reference?

@tingyuz  it is kind of covered already in this test function: https://github.com/jim-wang-intel/edgex-go/blob/2266-addtion-vault-tokens/internal/security/fileprovider/provider_test.go#L270 
and we can add more negative or positive test cases to this test function as many as we want.

tingyuz: Sounds good - I didn't review the test file before putting the comments. It would be helpful to add more tests to show what a required format would be.

internal/security/fileprovider/tokenconfig.go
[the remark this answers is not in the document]
lenny-intel: It isn't about optimizing it... ;-)

internal/security/fileprovider/tokenconfig.go
jim-wang-intel: Added two more cases just now.

internal/security/fileprovider/tokenconfig_test.go
jim-wang-intel: added comments.

internal/security/fileprovider/tokenconfig_test.go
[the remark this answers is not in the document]
jim-wang-intel: done

internal/security/fileprovider/tokenconfig.go
bnevis-i: Move the hyphen to the end of the regexp

`^[\w.: -]{1,512}$`

What was happening is that `.-:` is being interpreted as a character range.  The hyphen needs to be last.

jim-wang-intel: I think the list of allowed symbols or charaters in the regex  becomes unmanageable according to the following rule of allowance:
```go
	// anything containing the followings are NOT ok:
	// slash (/) is not ok [since it is a directory delimiter]
	// backslash (\) is not ok [since it is a directory delimiter]
	// comma (,) is not ok [since it is a separator in the environment list itself]
	// any other characters than the above is ok
	// and the the length of name is upto 512 characters
```

so instead of saying which are allowed, we should do something not allowed and that is original idea I had in mind when doing the regex of that. Not sure how to express this more precisely.

bnevis-i: Then we agree to disagree.   From a security perspective blacklisting is deeply flawed.  My advice is not to do it.

 https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html

"It is a common mistake to use black list validation in order to try to detect possibly dangerous characters and patterns like the apostrophe ' character, the string 1=1, or the <script> tag, but this is a massively flawed approach as it is trivial for an attacker to bypass such filters."

jim-wang-intel: ok, so how about this list? 
```^[\w.: \!\@\#\^\&\?\-=\+;\|\'\"\<\>\{\}\[\]]{1,512}$```
basically it covers most of characters (white listed in that point of view) and note that the slash / backslash \\ and comma, are not in the list...

Comment (bnevis-i) [reply internal/security/fileprovider/tokenconfig.go]: https://tools.ietf.org/html/rfc3986#page-12

The following characters are reserved characters for URI's and should not be allowed

```
      gen-delims  = ":" / "/" / "?" / "#" / "[" / "]" / "@"

      sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
                  / "*" / "+" / "," / ";" / "="
```

Comment (jim-wang-intel) [reply internal/security/fileprovider/tokenconfig.go]: ok, sounds good; will construct according to this guideline.

**Comment by jim-wang-intel** [comment]
Ready for re-review and merging.

File: internal/security/fileprovider/tokenconfig.go [hunk 11]
```diff
@@ -80,3 +87,89 @@ func LoadTokenConfig(fileOpener fileioperformer.FileIoPerformer, path string, to
...
+	// the service name eventually becomes part of the URL to Vault's API call
+	// Based upon the RFC 3986: https://tools.ietf.org/html/rfc3986#page-12,
+	// the following characters are reserved characters for URI's and thus NOT allowed:
+	// gen-delims  = ":" / "/" / "?" / "#" / "[" / "]" / "@"
+	// sub-delims  = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="
+	// backslash (\) also is not allowed due to being as a delimiter for URI directory in Windows
+	// and the regular alphanumeric characters like A to Z, a to z, 0 to 9, underscore (_),
+	// -, ~, %, ^,  {, }, |, <, >, . are allowed.
+	// and the the length of the name is upto 512 characters
+	serviceNameRegx := regexp.MustCompile(`^[\w. \~\%\^\-\|\<\>\{\}]{1,512}$`)
```
bnevis-i: @jim-wang-intel  Please remove "%"  as well, because that is used to encode the reserved characters and could be used to bypass the above filter -- for example, `%24100` would encode `$100`, bypassing the above filter.

jim-wang-intel: just saw this- good catch and changed.

jim-wang-intel:
@lenny-intel can you re-review to see if changes are ok and please remove the change request status.

@tingyuz request your review and 2nd approval from Non-Intel like Bryon mentioned. 

Thanks!

internal/security/fileprovider/provider.go
```diff
@@ -77,14 +77,25 @@ func (p *fileTokenProvider) Run() error {
...
 		p.logger.Error(fmt.Sprintf("failed to read token configuration file %s: %s", p.tokenConfig.ConfigFile, err.Error()))
 		return err
 	}
 
-	for serviceName, serviceConfig := range tokenConf {
+	// merge the additional token configuration list from environment variable
+	// note that the configuration file takes precedence, as the tokenConf will override
+	// the tokenConfEnv with same duplicate keys
+	// The tokenConfEnv only uses default settings.
+	tokenConf = tokenConfEnv.mergeWith(tokenConf)
```
mkbhanda: @jim-wang-intel I thought the point of @lenny-intel's comment was to do
tokenConf = tokenConf.mergeWith(tokenConfEnv)
where the first act of business is to determine whether tokenConfEnv is empty and return promptly out of merge. This way you could even just return an None/nil object instead of instantiating an empty map.

lenny-intel: @mkbhanda , nope it was to and the comment so it was clear that the current call is doing as we expected.

jim-wang-intel: I added comments in the code above already- let me know if this is clear.

lenny-intel: @jim-wang-intel , looks good.

jim-wang-intel:
Rebase

Issue #2486, opened 2020-04-10
SMA continually generates logs errors if support-* services are not running


tonyespy:
# 🐞 Bug Report

### Affected Services
The issue is located in: sys-mgmt-agent

### Is this a regression?
Not sure...

### Description and Minimal Reproduction
**docker**: edit one the nightly-built docker-compose and comment out one of the support services (e.g. support-logging; which will actually be commented out in the release docker-compose* files for Geneva). Start up EdgeX with a docker-compose up command, and wait for everything to come up. The following log message will be generated every ~15s:

`edgex-sys-mgmt-agent | unable to get service endpoint for edgex-support-logging: no matching service endpoint found`

**snap**: install the snap from latest/edgex (on a system w/snap support):
 * sudo snap install edgexfoundry
 * snap services edgexfoundry | grep support
Notice that by default, all the support-* services are disabled.
 * sudo snap logs edgexfoundry -f

watch the following log messages get output every ~15s:
```
sys-mgmt-agent[28503]: unable to get service endpoint for edgex-support-notifications: no matching service endpoint found
sys-mgmt-agent[28503]: unable to get service endpoint for edgex-support-scheduler: no matching service endpoint found
sys-mgmt-agent[28503]: unable to get service endpoint for edgex-support-logging: no matching service endpoint found
```
## 🌍  Your Environment
**Deployment Environment:**
docker or snap

**EdgeX Version:**
EdgeX built from latest master branch (i.e. nightly build as of Apr 10 '20)

**Anything else relevant?**
Please see the related issue #1662.

First, I disagree with the initial statement in that issue that states all of the support-* services should be considered always available to any EdgeX instance. AFAIK, neither support-scheduler nor support-notifications are used by any of the internal EdgeX services by default, so it doesn't make sense to require that they're always running. If an EdgeX instance actually requires them, then they should be enabled, otherwise they're just burning compute cycles... 

Also I'm not sure why the registry isn't being used to determine what services are available? Are these checks from the SMA considered health checks? If so, aren't they duplicating the Consul's health checks?

tsconn23:
This isn't a bug. It's working as designed up to the current point. The message you're seeing is coming from the endpoint monitor which is bootstrapped at service start time. The monitor attempts to reconnect with the service it's been instructed to contact at a configurable interval (15 seconds by default). The monitor in the SMA is being instructed to include the support services by the following list of default services.

[...]

Issue #1943, opened 2019-10-16
Configure password protection for postgres DB with Kong


anonymouse64:
# 🚀 Feature Request

### Relevant Package

 This feature request is for... the security services

### Description
The default docker-compose includes PostgreSQL as the database backend for Kong, but it does not specify a password for it, meaning that the PostgreSQL database is essentially wide open to access. 

### Describe the solution you'd like
Create a kong user account with postgresql that is protected via a password, ideally store this password in Vault and have Kong retrieve the password via Vault with a specific token issued to it.

This password would be ideally generated by security-secrets-setup.

Also provide the password to PostgreSQL when it is started in an appropriate mechanism (which would need another vault token to access the password if that password is stored inside Vault)

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?

We could also just specify a password in the config file that is randomly generated upon startup and stored inside Kong's filesystem so that only the Kong container can access it.

anonymouse64:
This is not yet a problem for the snap, but will be shortly as we transition to using Postgres in the snap as per #1222

bnevis-i:
Additionally, PostgreSQL must also be provided the same password during initialization.  This is done via passing `POSTGRES_PASSWORD` environment variable to the entrypoint script, or `POSTGRES_PASSWORD_FILE` and having the password passed via a file system secret.

I would also like to point out that `POSTGRES_USER` creates a superuser account, which is not an issue when kong is the only database on the server, but would be an issue if one wanted multiple databases on the PostgreSQL instance.  There is no current machinery in the Docker Hub postgres image for creating databases with explicit database owners that are not superusers.

anonymouse64:
updated the issue description to address that

hutchic:
Do you envision this using vault dynamic secrets ( https://www.vaultproject.io/docs/secrets/databases/postgresql.html ).

Just idly curious have no strong opinions

anonymouse64:
That seems like a useful feature within Vault, but I don't know if this will use that or not. 
If it's not a lot of work perhaps folks will go that route, but it seems pretty straight forward to manually do this in the scripts that start things up instead (i.e. `security-secrets-setup` et al).

bnevis-i:
@hutchic It would be great if Kong could delegate to Vault to retrieve its password without the project having to write any glue logic.  This would eliminate the need to manage yet another secret besides the Vault access token.

hutchic:
[...]

PR #2534, opened 2020-04-29
Snap geneva updates

tonyespy:
## PR Checklist

- [x] The commit messages follows our guidelines: https://wiki.edgexfoundry.org/display/FA/Contributor%27s+Guide
- [x] Docs have been updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe:

This PR includes multiple changes to the snap packaging for Geneva including:

 * update Consul, postgreSQL, and Vault to the latest versions used for Geneva
 * enable secure postgreSQL (issue #1943)
 * enable secure Redis
 * disable sys-mgmt-agent by default (see issue #2486)
 * include new device-virtual binary device profile
 * add Kuiper support
 * update release tags for non-edgex-go EdgeX services (e.g. device-virtual, support-rulesengine, ...)
 * fix configure hook logic to allow the database provider to be switched between mongo and redis (needed due to the removal of core-config-seed)
 * update the snap README file

**Note** - testing Kuiper just requires using 'snap set' to activate kuiper, then using the included kuiper-cli based on the upstream EdgeX instructions here:

https://github.com/emqx/kuiper/blob/master/docs/en_US/edgex/edgex_rule_engine_tutorial.md#kuiper-edgex-integration

- [ ] Yes
- [x] No

Full testing of security-proxy has been deferred, and will be undertaken as part of the snap kong update PR (pending).

### DIFF
--- a/snap/README.md
+++ b/snap/README.md
@@ -141,6 +141,7 @@ Vault can be also be disabled, but doing so will also disable Kong, as it depend
 ```bash
 $ sudo snap set edgexfoundry security-secret-store=off
 ```
+**Note** - Kong is currently not supported in the snap when installed on an arm64-based device, so it will be disabled on install.
 
 ## Limitations
 
--- a/snap/README.md
+++ b/snap/README.md
@@ -3,14 +3,13 @@
 
 This folder contains snap packaging for the EdgeX Foundry reference implementation.
 
-The snap contains Consul, MongoDB, all of the EdgeX Go-based micro services from
-this repository, device-virtual, as well as Vault, Kong, Cassandra, and the 
-two go-based security micro services. The snap also contains a single OpenJDK 
-JRE used to run device-virtual.
+The snap contains Consul, MongoDB, Redis, and all of the EdgeX Go-based micro services from
+this repository, device-virtual, as well as Vault, Kong, PostgreSQL. The snap also contains a
+single OpenJDK JRE used to run the legacy support-rulesengine (deprecated).
 
-The project maintains a rolling release of the snap on the `edge` channel that is rebuilt and published at least once daily through the jenkins jobs setup for the EdgeX project. You can see the jobs run [here](https://jenkins.edgexfoundry.org/view/Snap/) specifically looking at the `edgex-go-snap-{branch}-stage-snap`.
+The project maintains a rolling release of the snap on the `edge` channel that is rebuilt and published at least once daily through the jenkins jobs setup for the EdgeX project.
 
-The snap currently supports running on both `amd64` and `arm64` platforms. Once the project no longer depends on MongoDB as it's primary database, the snap should start working on `armhf` platforms that support snaps as well.
+The snap currently supports running on both `amd64` and `arm64` platforms.
 
 ## Installation
 
@@ -35,25 +34,25 @@ The snap can be installed using `snap install`. To install the snap from the edg
 $ sudo snap install edgexfoundry --edge
 ```
 
-You can specify install specific releases using the `--channel` option. For example to install the Delhi release of the snap:
+You can specify install specific releases using the `--channel` option. For example to install the Fuji release of the snap:
 
 ```bash
-$ sudo snap install edgexfoundry --channel=delhi
+$ sudo snap install edgexfoundry --channel=fuji
+
 ```
 
 Lastly, on a system supporting it, the snap may be installed using GNOME (or Ubuntu) Software Center by searching for `edgexfoundry`.
 
 **Note** - the snap has only been tested on Ubuntu Desktop/Server versions 18.04 and 16.04, as well as Ubuntu Core versions 16 and 18.
 
-**WARNING** - don't install the EdgeX snap on a system which is already running mongoDB or Consul.
+**WARNING** - don't install the EdgeX snap on a system which is already running one of the included services (e.g. Consul, Redis, Vault, ...).
 
 ## Using the EdgeX snap
 
 Upon installation, the following EdgeX services are automatically and immediately started:
 
 * consul
-* mongod
-* mongo-worker
+* redis
 * core-data
 * core-command
 * core-metadata
@@ -61,10 +60,14 @@ Upon installation, the following EdgeX services are automatically and immediatel
 
 The following services are disabled by default:
 
-* support-notifications
-* support-logging
-* support-scheduler
+* app-service-configurable (required for Kuiper and support-rulesengine)
 * device-virtual
+* kuiper
+* support-logging
+* support-notifications
+* support-rulesengine (deprecated)
+* support-scheduler
+* sys-mgmt-agent
 
 Any disabled services can be enabled and started up using `snap set`:
 
@@ -82,7 +85,11 @@ All services which are installed on the system as systemd units, which if enable
 
 ### Configuring individual services
 
-All default configuration files are shipped with the snap inside `$SNAP/config`, however because `$SNAP` isn't writable, all of the config files are copied during snap installation (specifically during the install hook, see `snap/hooks/install` in this repository) to `$SNAP_DATA/config`. The configuration files in `$SNAP_DATA` may then be modified. You may wish to restart the snap's services to take configuration into account with:
+All default configuration files are shipped with the snap inside `$SNAP/config`, however because `$SNAP` isn't writable, all of the config files are copied during snap installation (specifically during the install hook, see `snap/hooks/install` in this repository) to `$SNAP_DATA/config`.
+
+Note - as the core-config-seed was removed as part of the Geneva release, services self-seed their configuration on startup. This means that if a service is
+started by default in the snap, the only way to change configuration is to use the Consul UI or [kv REST API](https://www.consul.io/api/kv.html). Services that
+aren't started by default (see above) *will* pickup any changes made to their config files when enabled.
 
 ```bash
 $ sudo snap restart edgexfoundry
@@ -114,20 +121,25 @@ $ journalctl -u snap.edgexfoundry.consul
 
 Currently, the security services are enabled by default. The security services consitute the following components:
 
- * Vault
- * Cassandra
  * Kong
- * vault-worker (from [security-secret-store](https://github.com/edgexfoundry/security-secret-store))
- * kong-worker (from [security-api-gateway](https://github.com/edgexfoundry/security-api-gateway/))
+ * PostgreSQL
+ * Vault
+ * security-secrets-setup
+ * security-secretstore-setup
+ * security-proxy-setup
 
-All services are currently bundled in the singular service, `security-services` (see issue [#485](https://github.com/edgexfoundry/edgex-go/issues/485) for more details on why). 
+Vault is used for secret management, and Kong is used as an HTTPS proxy for all the services.
 
-When security is enabled, Consul is secured using Vault for secret management, and Kong is used as an HTTPS proxy for all the services. The HTTPS keys for Kong and Vault are placed in `$SNAP_DATA/vault/pki`. Kong needs a database to manage itself, and can use either Postgres or Cassandra. Because Postgres cannot run inside of a snap due to issues running as root (currently all snap services must run as root, see [this post](https://forum.snapcraft.io/t/multiple-users-and-groups-in-snaps/1461) for details), we use Cassandra. 
-
-To turn off security, use `snap set`:
+Kong can be disabled by using the following command:
 
 ```bash
-$ sudo snap set edgexfoundry security-services=off
+$ sudo snap set edgexfoundry security-proxy=off
+```
+
+Vault can be also be disabled, but doing so will also disable Kong, as it depends on Vault. Thus the following command will disable both:
+
+```bash
+$ sudo snap set edgexfoundry security-secret-store=off
 ```
 
 ## Limitations
@@ -234,8 +246,6 @@ After building the snap from one of the above methods, you will have a binary sn
 ```bash
 $ sudo snap install --devmode edgexfoundry*.snap
 ```
-
-**Note** You can try installing a locally built snap with the `--dangerous` flag (instead of the `--devmode` flag), but there is a race condition with this method. Specifically Cassandra, MongoDB, and other services require accesses not provided by default to the snap, and these are provided by connecting the interfaces detailed below. The race condition occurs because if the services fail to start because the accesses were denied (because the interfaces weren't connected soon enough), the installation may be entirely aborted by snapd.  If you do install with `--dangerous`, it is recommended to perform the connections detailed below in the same shell command to minimize the time between the installation (and hence service startup) and granting of accesses from interface connection. Note this race condition doesn't happen when installing the snap from the store because the interface connection automatically happens before starting the services.
 
 In addition, if you are using snapcraft with multipass VM's, you can speedup development by not creating a *.snap file and instead running in "try" mode . This is done by running `snapcraft try` which results in a `prime` folder placed in the root project directory that can then be "installed" using `snap try`. For example:
 
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -30,6 +30,8 @@ ALL_SERVICES="$ALL_SERVICES device-virtual"
 ALL_SERVICES="$ALL_SERVICES security-secret-store"
 ALL_SERVICES="$ALL_SERVICES security-proxy"
 
+DB_SERVICES=(core-command core-data core-metadata support-notifications support-scheduler)
+
 # handle_svc will either turn a service off or on and set the associated
 # config item
 # first arg is the service, second is the state to put it in
@@ -50,6 +52,28 @@ handle_svc () {
         echo "invalid setting $2 for service $1"
         exit 1;;
     esac
+}
+
+restart_db_services () {
+    # since the configuration files have been modified, those
+    # changes need to be progpogated into relevant services, so
+    # restart the services with updated configuration if they were
+    # already running
+    for svc in ${DB_SERVICES[*]}; do
+        status=`snapctl services "$SNAP_NAME.$svc" | grep -o "^active"`
+
+        if [ "$status" = "active" ]; then
+            snapctl restart "$SNAP_NAME.$svc"
+        fi
+    done
+}
+
+update_consul_db_type () {
+    curl -s --request PUT --data "$2" \
+        http://localhost:8500/v1/kv/edgex/core/1.0/edgex-"$1"/Databases/Primary/Type
+
+    curl -s --request PUT --data "$3" \
+        http://localhost:8500/v1/kv/edgex/core/1.0/edgex-"$1"/Databases/Primary/Port
 }
 
 for key in $ALL_SERVICES; do
@@ -108,11 +132,13 @@ for key in $ALL_SERVICES; do
             handle_svc "security-secretstore-setup" "$status"
             ;;
         support-rulesengine)
-            # if we are turning rulesengine on, make sure 
+            # if we are turning rulesengine on, make sure
             # app-service-configurable is on too
             if [ "$status" = "on" ]; then
                 handle_svc "app-service-configurable" "on"
             fi
+            # handle the service too
+            handle_svc "$key" "$status"
             ;;
         *)
             # default case for all other services just enable/disable the service using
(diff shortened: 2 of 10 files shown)


snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -ex
+
+security-secretstore-setup -confdir "$SNAP_DATA/config/security-secretstore-setup/res" --vaultInterval=10
+
+if [ -n "$REDIS5_PASSWORD_PATHNAME" ]; then
```
andresrinivasan, line 5: When would this be not true?

tonyespy: Nice catch, since the env var is always set prior to this script being launched, the length of the string is always non-zero. Instead this should check if the file exists, and the security-secretstore-read command should also get moved into this if block. Here's the diff I'm testing now:
```

diff --git a/snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh b/snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh
index 234bd58..5cfc20a 100755
--- a/snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh
+++ b/snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh
@@ -2,7 +2,8 @@
 
 security-secretstore-setup -confdir "$SNAP_DATA/config/security-secretstore-setup/res" --vaultInterval=10
 
-if [ -n "$REDIS5_PASSWORD_PATHNAME" ]; then
+# if redis pw file doesn't exist, handle initialization & read it
+if [ ! -f "$REDIS5_PASSWORD_PATHNAME" ]; then
 
     CONFIG_FILE="$SNAP_DATA/config/security-secretstore-read/res/configuration.toml"
     # grep -q returns 0 if a match is found
@@ -18,6 +19,8 @@ if [ -n "$REDIS5_PASSWORD_PATHNAME" ]; then
 
         mv "$CONFIG_FILE.new" "$CONFIG_FILE"
     fi
+
+    # read the pw from secretstore & write to disk
+    security-secretstore-read -confdir "$SNAP_DATA/config/security-secretstore-read/res"
 fi
 
-security-secretstore-read -confdir "$SNAP_DATA/config/security-secretstore-read/res"

```

snap/local/runtime-helpers/bin/redis-launch.sh
```diff
@@ -0,0 +1,16 @@
...
+# following edgex-go bug is fixed:
+#
+# https://github.com/edgexfoundry/edgex-go/issues/2503
+#
+REDIS5_PASSWORD=`cat "$SNAP_DATA/secrets/edgex-redis/redis5-password"`
+
+exec "$SNAP/bin/redis-server" \
+     --requirepass "$REDIS5_PASSWORD" \
+     --dir "$SNAP_DATA/redis" \
+     --save 900 1 \
```
anonymouse64, line 14: what do the --save args do?

snap/snapcraft.yaml
```diff
@@ -155,13 +155,23 @@ apps:
     plugs:
       - network
       - network-bind
+  vault-cli:
+    adapter: none
+    command: bin/vault
+    environment:
+      VAULT_CONFIG: "$SNAP_DATA/config/security-secret-store/vault-config.hcl"
+      VAULT_ADDR: "https://localhost:8200"
+      VAULT_SKIP_VERIFY: "true"
```
anonymouse64, line 164: why is this arg needed? If we need to skip verification when talking to a local vaul in the snap, I'd much rather the user have to provide an "skip-verify" arg than us always hard-coding it here

snap/snapcraft.yaml
```diff
@@ -155,13 +155,23 @@ apps:
     plugs:
       - network
       - network-bind
+  vault-cli:
+    adapter: none
+    command: bin/vault
+    environment:
+      VAULT_CONFIG: "$SNAP_DATA/config/security-secret-store/vault-config.hcl"
+      VAULT_ADDR: "https://localhost:8200"
```
anonymouse64, line 163: if you wanted to be fancy, you could probably combine this with the yaml above with a yaml anchor. no preference from me, but if we need it for a third app I would prefer it.

snap/snapcraft.yaml
```diff
@@ -327,7 +339,8 @@ apps:
     plugs: [network, network-bind]
   app-service-configurable:
     adapter: full
-    command: bin/app-service-configurable -confdir $SNAP_DATA/config/app-service-configurable/res -profile rules-engine --registry
+    # -s option disables the core-data version check; not needed in the snap since versions are known
```
anonymouse64, line 342: :+1:

snap/snapcraft.yaml
```diff
@@ -966,22 +1013,16 @@ parts:
...
     source-depth: 1
     override-build: |
-      # When the snap is built by Jenkins w/in a docker instance, the existence of
-      # the top level go.mod file causes the consul build to fail when using go 1.13.
-      # This workaround can probably be removed if/when consul is updated to a more
-      # recent version.
-      if [ -f /build/go.mod ]; then
-        mv /build/go.mod /build/go.mod.bk
-      fi
+      patch -p1 < "$SNAPCRAFT_PROJECT_DIR/snap/local/patches/0001-Makefile-rm-golangci-lint-tool.patch"
```
anonymouse64, line 1022: could you add a comment about needing this patch?

snap/snapcraft.yaml
```diff
@@ -1062,3 +1108,30 @@ parts:
...
+
+      install -DT "$SNAPCRAFT_PART_SRC/LICENSE" \
+         "$SNAPCRAFT_PART_INSTALL/usr/share/doc/kuiper/LICENSE"
+    organize:
+      bin/cli: bin/kuiper-cli
+      bin/server: bin/kuiper-server
+    stage:
+      - -etc/mqtt_source.yaml
+    stage-packages:
+      - libzmq3-dev
```
anonymouse64, line 1137: this is already staged by the edgex-go part. snapcraft is smart enough to not care since the file contents will be the same AFAIK, but just noticing this

anonymouse64: also relatedly, why are we staging the dev package here, when we also stage `libzmq5` in app-service-config ? something smells wrong

snap/snapcraft.yaml
```diff
@@ -400,19 +413,7 @@ apps:
...
-      LC_ALL: C.UTF-8
-      LANG: C.UTF-8
-      PGHOST: $SNAP_COMMON/sockets
-    command-chain:
-      - bin/perl5lib-launch.sh
-      # createuser should be run as the snap_daemon user, which is the user
-      # who is initially created as a role with postgres
-      - bin/drop
-    plugs: [home, removable-media, network]
+  # TODO: remove as part of secure postgres PR!
```
anonymouse64, line 416: what should be removed as part of the secure postgres PR?

anonymouse64: (and is that this PR?)

snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh
```diff
@@ -0,0 +1,23 @@
...
+
+    CONFIG_FILE="$SNAP_DATA/config/security-secretstore-read/res/configuration.toml"
+    # grep -q returns 0 if a match is found
+    grep -q "/tmp/edgex" "$CONFIG_FILE"
+    RETVAL="$?"
+
+    if [ $RETVAL == 0 ]; then
```
anonymouse64, line 12: why isn't this whole thing just 

```suggestion
    if grep -q "/tmp/edgex" "$CONFIG_FILE"; then
```

snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh
anonymouse64: also I have the same question, but for your new patch, under what conditions would security-secretstore-setup-launch.sh get run when this file doesn't exist?  Do we allow users to configure to use redis w/o a password

snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh
```diff
@@ -0,0 +1,23 @@
...
+        sed -e "s@/tmp/edgex/secrets/ca/ca.pem@$SNAP_DATA/pki/EdgeXFoundryCA/EdgeXFoundryCA.pem@" \
+            -e "s@/tmp/edgex/secrets/edgex-redis/secrets-token.json@$SNAP_DATA/secrets/edgex-redis/secrets-token.json@" \
+            -e "s@/tmp/edgex/secrets/edgex-redis/redis5-password@$REDIS5_PASSWORD_PATHNAME@" \
+            -e "s@edgex-vault@localhost@" \
```
anonymouse64, line 16: why are we performing this sed expansion at runtime? I'd much rather do this at build-time or if really needed install-hook time, not at app service runtime

snap/hooks/install
```diff
@@ -10,14 +10,15 @@ SNAP_CURRENT=${SNAP/%$SNAP_REVISION/current}
 # into $SNAP_DATA/config
 # note that app-service-configurable is handled separately
 mkdir -p "$SNAP_DATA/config"
-for service in edgex-mongo security-file-token-provider security-proxy-setup security-secrets-setup security-secretstore-setup core-command core-data core-metadata support-logging support-notifications support-scheduler sys-mgmt-agent device-virtual; do
+for service in edgex-mongo security-file-token-provider security-proxy-setup security-secrets-setup security-secretstore-setup core-command core-data core-metadata support-logging support-notifications support-scheduler sys-mgmt-agent device-virtual security-secretstore-read; do
```
anonymouse64, line 13: is this a new service? I don't see the service defined in the snapcraft.yaml ?

snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh
tonyespy: At the moment, security-secretstore-setup-launch.sh is always run in order to launch security-secretstore-setup, which is a one-shot service. The file should not exist when this script is run.  We currently do not allow users to configure mongo or redis without password authentication.

snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh
tonyespy: Cause my shell fu isn't as strong as yours... thanks for the suggestion!

snap/hooks/install
tonyespy: Yes, this is a new service in edgex-go which reads the redis password from vault, and writes it to the redis password file. It's included in the edgex-go part's build-override logic (see the first for loop which iterates over the all the services to copy the binaries and configuration files).

snap/hooks/configure
```diff
@@ -126,75 +144,134 @@ done
...
-                    json2toml --preserve-key-order > "$configFile.tmp"
-                mv "$configFile.tmp" "$configFile"
+            for svc in ${DB_SERVICES[*]}; do
+                updated="false"
+
+                exists=`curl -s http://localhost:8500/v1/kv/edgex/core/1.0/edgex-$svc/Databases/Primary/Type?raw`
+                if [ "$exists" = "redisdb" ]; then
+
+                    curl -s --request PUT --data "mongodb" \
+                        http://localhost:8500/v1/kv/edgex/core/1.0/edgex-$svc/Databases/Primary/Type
```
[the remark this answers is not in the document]
tonyespy: Is this what you were thinking? I'm not sure I have the cycles to create a single function that handles all of the logic for each db type?

```
exists=`curl -s http://localhost:8500/v1/kv/edgex/core/1.0/edgex-$svc/Databases/Primary/Type?raw`
if [ "$exists" = "redisdb" ]; then
    update_consul_db_type "$svc" "mongodb" "27017"
    updated="true"
fi
```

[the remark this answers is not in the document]
anonymouse64: yes

snap/hooks/install
```diff
@@ -156,33 +162,48 @@ snapctl start "$SNAP_NAME.postgres"
...
-        echo "failed to create kong db in postgres after $iter_num iterations"
+        echo "failed to create kong user in postgres after $iter_num iterations"
         exit 1
     fi
 done
 
 # stop postgres again in case the security services should be turned off
 snapctl stop "$SNAP_NAME.postgres"
 
+# modify postgres authentication config to use 'md5' (password)
```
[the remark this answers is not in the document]
tonyespy: I'm loathe to change this, as it doesn't make sense to configure postgres to require password authentication before creating the user that owns the credentials...

snap/snapcraft.yaml
tonyespy: leftover comment, will remove when I push my update

snap/hooks/install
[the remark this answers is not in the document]
anonymouse64: okay, fair enough

snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh
tonyespy: This has been addressed in my latest update

tonyespy:
I just pushed an update that addresses most of the comments. The only thing I didn't update was the install script logic which enables postgresql password auth, as it doesn't make sense to enable password auth before the user (with password) is added.

I also added one more commit which enables configure hook support for the SMA, as it's now disabled by default.

tonyespy:
And I just snuck in one additional commit which updates mongo to the latest stable patch release (4.2.6).

ernestojeda:
recheck

snap/hooks/configure
```diff
@@ -52,6 +61,28 @@ handle_svc () {
...
+        status=`snapctl services "$SNAP_NAME.$svc" | grep -o "active"`
+
+        if [ "$status" = "active" ]; then
+            snapctl restart "$SNAP_NAME.$svc"
+        fi
+    done
+}
+
+update_consul_db_type () {
+    curl -s --request PUT --data "$2" \
```
lenny-intel, line 79: Why Consul and not env overrides? 

Remember that env overrides are record of truth now, not consul.

snap/snapcraft.yaml
```diff
@@ -426,6 +425,27 @@ apps:
...
+  kuiper:
+    adapter: full
+    command: bin/kuiper-server
+    daemon: simple
+    environment:
+      KuiperBaseKey: $SNAP_DATA/kuiper
+      # KUIPER_DEBUG: "true"
+      KUIPER_CONSOLE_LOG: "true"
+      KUIPER_REST_PORT: 48075
+      EDGEX_SERVER: edgex-app-service-configurable-rules
```
lenny-intel, line 437: This should be `localhost` not the docker host name.

snap/snapcraft.yaml
```diff
@@ -526,43 +547,72 @@ parts:
   # from edgex-go which at this point will fail
   # eventually we would like these changes to be upstreamed but until then
   # they live here. see also https://github.com/edgexfoundry/edgex-go/issues/1109
-  go:
+  go112:
```
lenny-intel, line 550: why still need go112?

snap/snapcraft.yaml
```diff
@@ -426,6 +425,27 @@ apps:
...
+    command: bin/kuiper-server
+    daemon: simple
+    environment:
+      KuiperBaseKey: $SNAP_DATA/kuiper
+      # KUIPER_DEBUG: "true"
+      KUIPER_CONSOLE_LOG: "true"
+      KUIPER_REST_PORT: 48075
+      EDGEX_SERVER: edgex-app-service-configurable-rules
+      EDGEX_SERVICE_SERVER: http://localhost:48080
+      EDGEX_TOPIC: events
```
lenny-intel, line 439: Also don't see where you set `PublishTopic` on app-service-rules to be `events`. By default it is blank for the old rules engine.

snap/README.md
```diff
@@ -115,19 +122,19 @@ $ journalctl -u snap.edgexfoundry.consul
...
  * Vault
- * Cassandra
+ * PostgreSQL
  * Kong
- * vault-worker (from [security-secret-store](https://github.com/edgexfoundry/security-secret-store))
- * kong-worker (from [security-api-gateway](https://github.com/edgexfoundry/security-api-gateway/))
+ * security-secrets-setup
+ * security-secretstore-setup
+ * security-token-file-provider
+ * When
```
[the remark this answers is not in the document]
tonyespy: Whoops, nice catch.

snap/README.md
```diff
@@ -115,19 +122,19 @@ $ journalctl -u snap.edgexfoundry.consul
...
 
-All services are currently bundled in the singular service, `security-services` (see issue [#485](https://github.com/edgexfoundry/edgex-go/issues/485) for more details on why). 
-
-When security is enabled, Consul is secured using Vault for secret management, and Kong is used as an HTTPS proxy for all the services. The HTTPS keys for Kong and Vault are placed in `$SNAP_DATA/vault/pki`. Kong needs a database to manage itself, and can use either Postgres or Cassandra. Because Postgres cannot run inside of a snap due to issues running as root (currently all snap services must run as root, see [this post](https://forum.snapcraft.io/t/multiple-users-and-groups-in-snaps/1461) for details), we use Cassandra. 
+security-proxy-setupsecurity is enabled, Consul is secured using Vault for secret management, and Kong is used as an HTTPS proxy for all the services.
```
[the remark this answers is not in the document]
tonyespy: Looks like this whole section got mangled during a rebase. Fixed, and will push shortly.

snap/hooks/configure
tonyespy: Well... back when core-config-seed still was around, this code modified the configuration.toml files in **$SNAP_DATA**, stopped the affected services, ran core-config-seed w/the overwrite flag, and then restarted the services. Doing this on a per-service basis seemed a bit too complex. Yes, it would be possible to do this via env overrides, but that would require an additional wrapper script (see bin/security-secret-store-env-var.sh which is used to handle vault being enabled/disabled). When Ian and I first discussed how to do this, he suggested just modifying the values directly in Consul, so I just went with that approach.

My gut tells me that modifying the dbtype (which needs to be done *before* anything's been written to the db) is going to be a fairly uncommon operation. The only reason I've kept it around is that we still haven't officially dropped mongo yet.

If you feel strongly about this, I'll go ahead and create the wrapper scripts.

snap/snapcraft.yaml
tonyespy: Because the snap builds consul and vault from source, and vault 1.3.1 can only be built with go 1.12.

snap/hooks/configure
lenny-intel: Do you have a mode where Consul isn't used? If not, then I am fine with this approach.

snap/snapcraft.yaml
lenny-intel: FYI, this is the override we have on app-service-rules for Kuiper.       
```
Binding_PublishTopic: events
```

snap/snapcraft.yaml
```diff
@@ -329,8 +341,9 @@ apps:
     adapter: full
     command: bin/app-service-configurable -confdir $SNAP_DATA/config/app-service-configurable/res -profile rules-engine --registry
     daemon: simple
+    environment:
+      Binding_PublishTopic: events
```
lenny-intel, line 345: FYI, I started to have concern that this would break the ability to switch to old rules engine that subscribes to blank topic, but I think the blank topic is like a wild card topic for ZMQ. Thoughts?

lenny-intel: I just confirmed this is correct:
```
A consequence of this prefix matching behaviour is that you can receive all published messages by subscribing with an empty topic string.
```

snap/hooks/configure
```diff
@@ -52,6 +61,28 @@ handle_svc () {
     esac
 }
 
+restart_db_services () {
+    # since the configuration files have been modified, those
+    # changes need to be progpogated into relevant services, so
+    # restart the services with updated configuration if they were
+    # already running
+    for svc in ${DB_SERVICES[*]}; do
```
anonymouse64, line 69: I have a slight preference for helper functions to take arguments rather than use global variables but this is fine if you would rather change this later

snap/hooks/configure
anonymouse64: Since we don't yet support running without consul, I'd say leave it like this. Setting env vars for other services is also awkward because you need to pass state to those other services from here via either snap config or a file or something so it's not quite as simple as just setting the env var.

snap/hooks/configure
```diff
@@ -126,75 +170,103 @@ done
...
-                    jq -r '.Databases.Primary.Type = "redisdb" | .Databases.Primary.Port = 6379' | \
-                    json2toml --preserve-key-order > "$configFile.tmp"
-                mv "$configFile.tmp" "$configFile"
-            done
+            for svc in ${DB_SERVICES[*]}; do
+                updated="false"
+
+                exists=`curl -s http://localhost:8500/v1/kv/edgex/core/1.0/edgex-$svc/Databases/Primary/Type?raw`
+                if [ "$exists" = "true" ]; then
+                    update_consul_db_type "$svc" "redisdb" "6379"
```
anonymouse64, line 199: Thanks for this

snap/hooks/install
anonymouse64: I guess I was more wondering why it's not a proper app/service in the yaml?

snap/hooks/install
```diff
@@ -156,33 +162,48 @@ snapctl start "$SNAP_NAME.postgres"
...
+drop sed -i -e "s@trust@md5@g" "$SNAP_DATA/postgresql/10/main/pg_hba.conf"
+
 # finally, disable and turn off non-default services
 # by default, we want the support-*, device-*, and mongo services
```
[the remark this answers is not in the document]
anonymouse64: Would still be nice to see the sma here

snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -ex
+
+security-secretstore-setup -confdir "$SNAP_DATA/config/security-secretstore-setup/res" --vaultInterval=10
+
+# if redis pw file doesn't exist, handle initialization & read it
+if [ ! -f "$REDIS5_PASSWORD_PATHNAME" ]; then
```
anonymouse64, line 6: I'm still not clear on when this script is called when there is not a password defined?

snap/snapcraft.yaml
anonymouse64: Ping on this

snap/snapcraft.yaml
```diff
@@ -676,18 +720,36 @@ parts:
...
           install -DT "./cmd/$service/$service" "$SNAPCRAFT_PART_INSTALL/bin/$service"
 
-          if [ "$service" != "security-file-token-provider" ]; then
+          if [ "$service" = "security-secretstore-read" ]; then
+              mkdir -p "$SNAPCRAFT_PART_INSTALL/config/$service/res/"
+
+              # since security-secretstore-read doesn't support config env overrides, edit the
+              # configuration file at build time
+              CONFIG_FILE="./cmd/$service/res/configuration.toml"
+              if grep -q "/tmp/edgex" "$CONFIG_FILE"; then
```
anonymouse64, line 742: Thanks for moving this here

snap/hooks/install
tonyespy: Because it's only ever called once from the wrapper script to read the Redis password from Vault, and write it to disk. I treated this the same as security-file-token-provider which is launched from security-secret-store-setup. Ideally, Redis should just grow support for reading secrets directly from Vault, but that wasn't scoped for this release.

snap/snapcraft.yaml
tonyespy: OK, thanks!

snap/hooks/configure
tonyespy: I just followed your lead (e.g. `ALL_SERVICES`) on this one. Let's leave this one alone.

anonymouse64: ah right I had missed that this variable was just globally assigned once, I thought it was mutated in the other functions, nvm me this is fine

snap/hooks/install
anonymouse64: ok, thanks for the explanation

snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh
tonyespy: This script gets called once, on first boot when the redis password file doesn't exist. This is a workaround to the fact that redis can't read the password directly from Vault.

snap/snapcraft.yaml
tonyespy: What was the ping for?

anonymouse64: you missed a comment from my previous review about why we are hard-coding `VAULT_SKIP_VERIFY: "true"`

snap/local/patches/0001-Makefile-rm-golangci-lint-tool.patch
```diff
@@ -0,0 +1,28 @@
+From 27c697fdc0a1e01b781921a0748277feba430e22 Mon Sep 17 00:00:00 2001
+From: Tony Espy <espy@canonical.com>
+Date: Fri, 24 Apr 2020 08:33:04 -0400
+Subject: [PATCH] Makefile: rm golangci-lint tool
+
+Remopve golangci-lint from EXTERNAL_TOOLS as
```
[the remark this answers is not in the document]
tonyespy: https://github.com/hashicorp/vault/issues/8927

tonyespy:
All comments should be addressed, and I fixed a few other issues that cropped up during testing.

snap/hooks/configure
```diff
@@ -126,75 +174,103 @@ done
...
-                configFile=$SNAP_DATA/config/$svc/res/configuration.toml
-                toml2json --preserve-key-order "$configFile" | \
-                    jq -r '.Databases.Primary.Type = "redisdb" | .Databases.Primary.Port = 6379' | \
-                    json2toml --preserve-key-order > "$configFile.tmp"
-                mv "$configFile.tmp" "$configFile"
-            done
+            for svc in ${DB_SERVICES[*]}; do
+                updated="false"
+
+                exists=`curl -s http://localhost:8500/v1/kv/edgex/core/1.0/edgex-$svc/Databases/Primary/Type?raw`
```
anonymouse64, line 201: this logic doesn't quite work right, when I run this curl command locally I see this as returning the value of the `Type` key, which never ends up being "true" as you check below:

```
$ curl -s http://localhost:8500/v1/kv/edgex/core/1.0/edgex-core-data/Databases/Primary/Type?raw; echo
mongodb
```

snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh
anonymouse64, line 6: Sorry to be a bother about this, but I'm still rather confused, since AFAICT, the thing that will generate the password file is `security-secretstore-setup` which we literally call right before this check in this script, so the only situation that this condition would be false is if security-secretstore-setup is configured to not run generate a redis password? Is that a supported configuration for the snap to configure security-secretstore-setup to _not_ use a password for redis? That also won't be the default configuration since it seems to generate this password on initial install, so not only would the user need to configure `security-secretstore-setup` to not generate a password, but they would also need to _delete_ the existing password too. Seems like a lot of work and unlikely to ever be done by a user.

All that is to say, I think we could probably get away with not having this check.

snap/hooks/install
```diff
@@ -198,14 +230,3 @@ for svc in support-notifications support-scheduler support-logging app-service-c
...
-# if we are on arm64, disable the security-proxy because kong isn't 
-# properly supported on arm64 due to incorrect memory pointers used by lua and
-# openresty
-# see https://github.com/edgexfoundry/blackbox-testing/issues/185 for more 
-# details
-if [ "$SNAP_ARCH" == "arm64" ]; then
-    snapctl set security-proxy=off
-    snapctl stop --disable "$SNAP_NAME.kong-daemon"
-    snapctl stop --disable "$SNAP_NAME.security-proxy-setup"
-fi
```
anonymouse64, line 211: hmm I just tried my kong snap on my rpi on arm64 and it didn't seem to work, so I think we need to leave this in for users to be able to use the snap on arm64 at all

snap/snapcraft.yaml
anonymouse64, line 164: just fyi I still see this here

snap/snapcraft.yaml
```diff
@@ -966,22 +1035,16 @@ parts:
...
     source-depth: 1
     override-build: |
-      # When the snap is built by Jenkins w/in a docker instance, the existence of
-      # the top level go.mod file causes the consul build to fail when using go 1.13.
-      # This workaround can probably be removed if/when consul is updated to a more
-      # recent version.
-      if [ -f /build/go.mod ]; then
-        mv /build/go.mod /build/go.mod.bk
-      fi
+      patch -p1 < "$SNAPCRAFT_PROJECT_DIR/snap/local/patches/0001-Makefile-rm-golangci-lint-tool.patch"
```
anonymouse64, line 1044: I would still like to see a comment about why we need this patch and a link to the upstream issue here

snap/local/runtime-helpers/bin/security-secretstore-setup-launch.sh
tonyespy: In a perfect world, security-secretstore-setup would redis would read it's password directly from vault, but this isn't supported. In a less perfect world, `security-secretstore-setup` would read the password from vault and write it out to disk, but this isn't supported either. In the docker deployment, `security-secretstore-read` is called from the [entrypoint.sh](https://github.com/edgexfoundry/edgex-go/blob/master/cmd/security-secretstore-setup/entrypoint.sh#L54), which is why this wrapper script exists in the first place. So, when this script is first called, the password file will **not** exist, hence the check...

tonyespy:
> Really really close, but...
> 
>     * mongodb -> redis logic seems to be broken
> 
>     * I don't think we're ready to re-enable arm64 yet
> 
>     * I would still like to see the check on the redis password eliminated since it doesn't seem to be needed as it will always exist
> 
>     * the vault skip_verify env var is still set
> 
>     * still no comment where we apply the patch to vault to build it without the CI failure

@anonymouse64 All of your comments have been addressed, and just re-pushed.

PR #2552, opened 2020-05-11
Fix bug for removal-of-scheduler-interval causing the scheduler to crash.

### LINKED ISSUE #2520: Removing a scheduler interval when a interval action still using it would crash the scheduler
# 🐞 Bug Report

### Affected Services
support-scheduler

If I remove an interval when an interval action still using it, but leave the action untouched, the scheduler won't start any more next time. I had to manually delete keys associated with the action in the redis database to make the scheduler work again.

Is there a mechanism preventing an interval being removed when it's still used by some actions? something like that you can't remove a profile when it's used by some devices.

## 🔥 Exception or Error
<pre><code>
level=INFO ts=2020-04-27T08:03:21.127215Z app=edgex-support-scheduler source=loader.go:144 name:=scrub-aged-events message="scheduler could not find a interval with interval name : midnight" msg="problem adding support-scheduler interval action"
level=ERROR ts=2020-04-27T08:03:21.127264Z app=edgex-support-scheduler source=init.go:63 msg="Failed to load schedules and events scheduler could not find a interval with interval name : midnight"
level=INFO ts=2020-04-27T08:03:21.12741Z app=edgex-support-scheduler source=database.go:159 msg="Database disconnected"
</code></pre>

## 🌍  Your Environment
**Deployment Environment:**
go 1.14
**EdgeX Version:**
master

**Anything else relevant?**

akramtexas:
Fix #2520 

Removing a scheduler interval when a interval action still using it would crash the scheduler

### DIFF
--- a/internal/support/scheduler/rest_interval.go
+++ b/internal/support/scheduler/rest_interval.go
@@ -178,8 +178,8 @@ func restDeleteIntervalByID(
 	w http.ResponseWriter,
 	r *http.Request,
 	lc logger.LoggingClient,
-	dbClient interfaces.DBClient,
-	scClient interfaces.SchedulerQueueClient) {
+	scClient interfaces.SchedulerQueueClient,
+	intervalDeleter interval.IntervalDeleter) {
 
 	if r.Body != nil {
 		defer r.Body.Close()
@@ -188,23 +188,18 @@ func restDeleteIntervalByID(
 	// URL parameters
 	vars := mux.Vars(r)
 	id, err := url.QueryUnescape(vars["id"])
+
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		lc.Error("Error un-escaping the value id: " + err.Error())
 		return
 	}
 
-	op := interval.NewDeleteByIDExecutor(dbClient, scClient, id)
+	op := interval.NewDeleteByIDExecutor(intervalDeleter, scClient, id)
 	err = op.Execute()
 
 	if err != nil {
-		lc.Error(err.Error())
-		switch err.(type) {
-		case errors.ErrIntervalNotFound:
-			http.Error(w, err.Error(), http.StatusNotFound)
-		default:
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-		}
+		handleDeleteIntervalRestErrors(err, w, lc)
 		return
 	}
 
@@ -212,6 +207,7 @@ func restDeleteIntervalByID(
 	w.WriteHeader(http.StatusOK)
 	w.Write([]byte("true"))
 }
+
 func restGetIntervalByName(
 	w http.ResponseWriter,
 	r *http.Request,
@@ -253,35 +249,29 @@ func restDeleteIntervalByName(
 	w http.ResponseWriter,
 	r *http.Request,
 	lc logger.LoggingClient,
-	dbClient interfaces.DBClient,
-	scClient interfaces.SchedulerQueueClient) {
+	scClient interfaces.SchedulerQueueClient,
+	intervalDeleter interval.IntervalDeleter) {
 
-	defer r.Body.Close()
+	if r.Body != nil {
+		defer r.Body.Close()
+	}
 
+	// URL parameters
 	vars := mux.Vars(r)
 	name, err := url.QueryUnescape(vars["name"])
 
-	// Issues un-escaping
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		lc.Error("Error un-escaping the value name: " + err.Error())
 		return
 	}
-	op := interval.NewDeleteByNameExecutor(dbClient, scClient, name)
-	err = op.Execute()
-	if err != nil {
 
-		switch err.(type) {
-		case errors.ErrIntervalNotFound:
-			http.Error(w, err.Error(), http.StatusNotFound)
-			return
-		case errors.ErrIntervalStillUsedByIntervalActions:
-			http.Error(w, err.Error(), http.StatusBadRequest)
-			return
-		default:
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-			return
-		}
+	op := interval.NewDeleteByNameExecutor(intervalDeleter, scClient, name)
+	err = op.Execute()
+
+	if err != nil {
+		handleDeleteIntervalRestErrors(err, w, lc)
+		return
 	}
 
 	w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
@@ -290,6 +280,28 @@ func restDeleteIntervalByName(
 }
 
 // ************************ UTILITY HANDLERS ************************************
+
+func handleDeleteIntervalRestErrors(err error, w http.ResponseWriter, lc logger.LoggingClient) {
+
+	lc.Debug(err.Error())
+	switch err.(type) {
+	case errors.ErrIntervalNotFound:
+		http.Error(w, err.Error(), http.StatusNotFound)
+		return
+	case errors.ErrDbNotFound:
+		http.Error(w, err.Error(), http.StatusNotFound)
+		return
+	case errors.ErrIntervalStillUsedByIntervalActions:
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	case errors.ErrIntervalNameInUse:
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	default:
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
 
 // Scrub all the Intervals and IntervalActions
 func restScrubAllIntervals(
--- a/internal/support/scheduler/router.go
+++ b/internal/support/scheduler/router.go
@@ -101,12 +101,11 @@ func loadRestRoutes(r *mux.Router, dic *di.Container) {
 	interval.HandleFunc(
 		"/{"+ID+"}",
 		func(w http.ResponseWriter, r *http.Request) {
-			restDeleteIntervalByID(
-				w,
+			restDeleteIntervalByID(w,
 				r,
 				bootstrapContainer.LoggingClientFrom(dic.Get),
-				container.DBClientFrom(dic.Get),
-				schedulerContainer.QueueFrom(dic.Get))
+				schedulerContainer.QueueFrom(dic.Get),
+				container.DBClientFrom(dic.Get))
 		}).Methods(http.MethodDelete)
 	interval.HandleFunc(
 		"/"+NAME+"/{"+NAME+"}",
@@ -124,8 +123,8 @@ func loadRestRoutes(r *mux.Router, dic *di.Container) {
 				w,
 				r,
 				bootstrapContainer.LoggingClientFrom(dic.Get),
-				container.DBClientFrom(dic.Get),
-				schedulerContainer.QueueFrom(dic.Get))
+				schedulerContainer.QueueFrom(dic.Get),
+				container.DBClientFrom(dic.Get))
 		}).Methods(http.MethodDelete)
 	// Scrub "Intervals and IntervalActions"
 	interval.HandleFunc(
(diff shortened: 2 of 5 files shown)


akramtexas:
delete-interval-fix


Steps for functional DEV testing:

- Purge the local instance of Redis:
127.0.0.1:6379> FLUSHALL
OK
- START EdgeX core services (including core-scheduler) locally.
- GET the ID of the midnight interval, for example the ID `7901fc46-ba80-4293-85dd-c81a9dd728d1`:
```
level=INFO ts=2020-05-11T21:16:06.844548Z app=edgex-support-scheduler source=loader.go:172 name=midnight id=id msg="added interval to the support-scheduler database"
level=INFO ts=2020-05-11T21:16:06.84472Z app=edgex-support-scheduler source=schedule.go:162 msg="added the interval with id : 7901fc46-ba80-4293-85dd-c81a9dd728d1 into the scheduler queue"
```
- DELETE the midnight interval by ID using core-scheduler's API:
curl --location --request DELETE 'http://localhost:48085/api/v1/interval/7901fc46-ba80-4293-85dd-c81a9dd728d1' \
--data-raw ''
- RESTART EdgeX, which starts up without issue.
- Thus, EdgeX now prevents the user from deleting the midnight interval not only by NAME doing (with an error of "interval still in use by intervalAction(s) name: midnight"), but also by ID, as noted in the screenshot above of REST client.

akramtexas:
Rebased.

internal/support/scheduler/rest_interval.go
```diff
@@ -254,23 +263,30 @@ func restDeleteIntervalByName(
...
 	}
-	op := interval.NewDeleteByNameExecutor(dbClient, scClient, name)
+
+	op := interval.NewDeleteByNameExecutor(intervalLoader, intervalActionLoader, dbClient, scClient, name)
 	err = op.Execute()
-	if err != nil {
 
+	if err != nil {
+		// TODO Remove following logging statement if warranted
+		lc.Error(err.Error())
```
AnthonyMBonafide, line 289: Can we either remove this logging statement or the `TODO` comment?

internal/support/scheduler/router.go
```diff
@@ -101,12 +101,7 @@ func loadRestRoutes(r *mux.Router, dic *di.Container) {
 	interval.HandleFunc(
 		"/{"+ID+"}",
 		func(w http.ResponseWriter, r *http.Request) {
-			restDeleteIntervalByID(
-				w,
-				r,
-				bootstrapContainer.LoggingClientFrom(dic.Get),
-				container.DBClientFrom(dic.Get),
-				schedulerContainer.QueueFrom(dic.Get))
+			restDeleteIntervalByID(w, r, bootstrapContainer.LoggingClientFrom(dic.Get), container.DBClientFrom(dic.Get), schedulerContainer.QueueFrom(dic.Get), container.DBClientFrom(dic.Get), container.DBClientFrom(dic.Get))
```
AnthonyMBonafide, line 104: Can we put these arguments on new lines like it was previously?

internal/support/scheduler/router.go
```diff
@@ -120,12 +115,7 @@ func loadRestRoutes(r *mux.Router, dic *di.Container) {
 	interval.HandleFunc(
 		"/"+NAME+"/{"+NAME+"}",
 		func(w http.ResponseWriter, r *http.Request) {
-			restDeleteIntervalByName(
-				w,
-				r,
-				bootstrapContainer.LoggingClientFrom(dic.Get),
-				container.DBClientFrom(dic.Get),
-				schedulerContainer.QueueFrom(dic.Get))
+			restDeleteIntervalByName(w, r, bootstrapContainer.LoggingClientFrom(dic.Get), container.DBClientFrom(dic.Get), schedulerContainer.QueueFrom(dic.Get), container.DBClientFrom(dic.Get), container.DBClientFrom(dic.Get))
```
AnthonyMBonafide, line 118: Can we put these arguments on new lines like it was previously?

akramtexas:
> I think there is another bug somewhere in these changes. I performed the following steps and get an unexpected result:
> 
> 1. Started `support-scheduler` with clean database
> 2. Performed a `GET /interval`  to get the the ID of the `midnight` interval
> 3. Performed a `GET /interval/{ID}` to ensure the `midnight` ID was correct
> 4. Executed a `DELETE interval/{ID}` for the midnight interval
> 5. Executed another `DELETE interval/{ID}` for the midnight interval
> 
> The first time I executed the delete I got the expected response that the `Interval` is in use. However, the second request resulted in a response of `scheduler could not find a interval context with interval id : c1b4a2b5-aa93-4674-8af4-8b881e6008be` which is the ID of the `midnight` `Interval` which I was attempting to delete. I also can execute the `GET /interval/{ID}` (step 3) and get back the expected `Interval`, so there seems to be some disconnect where the `Interval` still exists but the DELETE operation is not able to retrieve it for deletion. I think we need to look into what we are assuming is the `source of truth` for holding `Interval` and `IntervalActions`

- I am able to recreate your observation above (for deleting interval `by ID`, as well as for deleting interval `by Name`). 
- So this may not be a bug introduced by the updates in this `PR`. 
- But you bring up a good point: There seems to be a disconnect between what we are assuming is the `source of truth` for holding `Interval` and `IntervalActions`.

AnthonyMBonafide:
@akramtexas , You are correct. The bug I mentioned in my last comment is not a regression and is present in the code base separate from this issue.

As we discussed in a separate meeting, we are going to ignore that "bug comment" I made previously and only address the code cleanup comments. The "bug comment" will be handled in another issue #2554.

akramtexas:
- The work undertaken in this `PR` will be completed (remaining unit failures fixed), but _not_ for the `Geneva` release. As such, converted it to a `Draft PR` for now ✅

akramtexas:
> * The work undertaken in this `PR` will be completed (remaining unit failures fixed), but _not_ for the `Geneva` release. As such, converted it to a `Draft PR` for now ✅

- The remaining unit failures have now been fixed, and this `PR` is _ready for review_. 
- Referring to the note above, and with the work undertaken being completed (i.e. remaining unit failures fixed), this `PR` will not be merged into the `Geneva` release.
- As such, it currently remains a `Draft PR`.

internal/support/scheduler/rest_interval.go
```diff
@@ -254,23 +263,30 @@ func restDeleteIntervalByName(
 	r *http.Request,
 	lc logger.LoggingClient,
 	dbClient interfaces.DBClient,
-	scClient interfaces.SchedulerQueueClient) {
+	scClient interfaces.SchedulerQueueClient,
+	intervalActionLoader interval.IntervalActionLoader,
+	intervalLoader interval.IntervalLoader) {
```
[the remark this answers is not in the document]
AnthonyMBonafide: I think this comment still applies. Can we look into this? If this is valid I think we can also applies this to the operator structs which required both of these constructs separately.

internal/support/scheduler/router.go
AnthonyMBonafide: Can we look into applying this previous comment?

internal/support/scheduler/router.go
AnthonyMBonafide: Can we look into applying this previous comment?

internal/support/scheduler/rest_interval.go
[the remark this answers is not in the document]
akramtexas: I apologize for having missed replying (earlier on) that I had looked into this, and while it may well be possible to combine those two arguments, condensing them into one, by using the type `IntervalDeleter`, the ripple from this is significant. This applies to both (production) code as well as, and especially, and even more so, to the numerous unit tests which would require refactoring. In that light, I _highly_ recommend that we don't undertake such refactorings at this time.

akramtexas:
- Referring to the steps outlined at the top (to be carried out for functional DEV testing), those were of course carried out again.
- In addition to that, and referring to #2554, I tested that consecutive attempts at deleting a scheduler interval in use (both by `Id` and by `Name`) is handled properly in that such attempts are disallowed, with the response being, for example: `interval name: midnight in use` ✅

internal/support/scheduler/rest_interval.go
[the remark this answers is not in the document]
AnthonyMBonafide: OK, I will create a new issue calling out this tech debt.

akramtexas:
Rebased.

akramtexas:
> Overall looks good. I did not manually verify these changes as this is still a draft PR. I will do a final once over once we are ready to merge in hopes of not wasting time.

@lranjbar and @jpwhitemn - Referring to the following update (in `EdgeX Slack` channel), I'm reaching out to you for help in ensuring that we're good with the the `Geneva 1.2 tag` aspects (in this `PR`'s feature branch), since I had branched _before_ that tagging was done:

Jim White  2:47 PM
@here - EdgeX developers, just had a Slack exchange with Lisa and the master branch of repositories is open for Hanoi development - with exception of Modbus and MQTT device services which are still having some issues addressed.  For safety sake, please make sure the repository you do work in has a 1.2 tag for Geneva in place and if it does not, please let @Lisa Ranjbar (Intel) or @jpwhite know.  Otherwise, happy coding!

Referring to the note above by @AnthonyMBonafide, I have taken this `PR` out of `Draft` state, but will hold off from requesting Anthony from doing a final once over once we are ready to merge.

internal/support/scheduler/rest_interval.go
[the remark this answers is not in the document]
AnthonyMBonafide: After taking a second pass at this, I can see that the arguments have duplicate functionality. For example:
https://github.com/edgexfoundry/edgex-go/blob/530c35f1cd71203c25442055d8d029b8fc27e873/internal/support/scheduler/operators/interval/delete.go#L146-L148

Here we are passing `IntervalActionLoader`,`IntervalLoader`, and `IntervalDeleter` to the same function where `IntervalDeleter` encapsulates both `IntervalActionLoader` and`IntervalLoader`. Furthermore, in that example we are not even using `IntervalActionLoader` nor`IntervalLoader` but rather only `IntervalDeleter` . I think this is confusing and significant enough where I do not think we should introduce this into the codebase to be handled later.  If this was not as prominent and less confusing then I would agree to a follow up tech debt issue.

As for the amount of effort to refactor and trying to avoid it, I think that really does not apply here since this PR is the one introducing all the tech debt. In this PR we are fixing a bug, which means introducing some missed functionality, which requires changes from the REST handler(s) up the call stack to the operator(s). So really there is no real refactoring that needs to be done, we simply need to make the changes in this PR fit better with the intended solution rather than taking the easiest route to get the functionality in.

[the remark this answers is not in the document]
akramtexas: - The point above about the overlap between a number of related struct types (`IntervalActionLoader`, `IntervalLoader`, and `IntervalDeleter`, etc.) is well-taken 👍
- I agree and will address the removal of this overlap as referenced above.

akramtexas:
Rebased.

internal/support/scheduler/rest_interval.go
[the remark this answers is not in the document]
akramtexas: The ask above has been addressed and a review of this `PR` requested 👍

internal/support/scheduler/router.go
akramtexas: I missed this comment, and the analogous one below and will address the both this minute ✅

internal/support/scheduler/router.go
akramtexas: Done (Addressed both reformatting-related comments) ✅

akramtexas:
Rebased.

AnthonyMBonafide:
I am going to merge these changes so this branch does not become stale. @akramtexas, lets make sure we add the proper tags to this PR and the associated issue.

jpwhitemn:
All, this fix should go into a dot release of Geneva that also incorporates some rules engine changes and UI fixes.  Tagging @lranjbar @lenny-intel for awareness on this item to include in our dot release.

PR #2555, opened 2020-05-12
snap: final geneva release items

tonyespy:
## PR Checklist

- [x] The commit message follows our guidelines: https://wiki.edgexfoundry.org/display/FA/Contributor%27s+Guide

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: packaging change

 * updates app-service-config source-tag to v1.1.0
 * updates device-virtual source-tag to v1.2.0
 * replace app-service-config `--registry` option with `-cp http://localhost:8500 `per @lenny-intel's comment
 * fixed the legacy rules-engine which was failing on startup due to a JRE problem
 * update the snap **grade** to **stable** (from **devel**), which allows the snap to be published to stable channels in the Snap store

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1090,13 +1090,14 @@ parts:
   device-virtual-go:
     source: https://github.com/edgexfoundry/device-virtual-go.git
     source-depth: 1
-    # TODO: update to v1.2.0 rel tag (or latest dev tag)
-    source-tag: v1.2.0-dev.12
+    source-tag: v1.2.0
     plugin: make
     after: [go113]
     override-build: |
       export PATH="$SNAPCRAFT_STAGE/go1.13/bin:$GOPATH/bin:$PATH"
       cd $SNAPCRAFT_PART_SRC
+      # create VERSION file (supposed to be created by jenkins pipeline...)
+      echo "1.2.0" > ./VERSION
       make build
 
       install -DT "./cmd/device-virtual" "$SNAPCRAFT_PART_INSTALL/bin/device-virtual"
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -336,10 +336,14 @@ apps:
     plugs: [network, network-bind]
   app-service-configurable:
     adapter: full
-    command: bin/app-service-configurable -confdir $SNAP_DATA/config/app-service-configurable/res -profile rules-engine --registry
+    command: >-
+      bin/app-service-configurable -cp -r
+      -confdir $SNAP_DATA/config/app-service-configurable/res
+      -profile rules-engine
     daemon: simple
     environment:
       Binding_PublishTopic: events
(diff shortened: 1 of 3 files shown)


snap/snapcraft.yaml
```diff
@@ -340,6 +340,7 @@ apps:
     daemon: simple
```
lenny-intel, line 340: Need to add the -cp or --configProvider option to specify using the configuration provider.  --registry is now just service registration. with no value specified it defaults to consul.http://localhost:8500.

 I recommend adding just  `--configProvider`

snap/snapcraft.yaml
```diff
@@ -336,10 +335,15 @@ apps:
     plugs: [network, network-bind]
   app-service-configurable:
     adapter: full
-    command: bin/app-service-configurable -confdir $SNAP_DATA/config/app-service-configurable/res -profile rules-engine --registry
+    command: >-
+      bin/app-service-configurable
+      -confdir $SNAP_DATA/config/app-service-configurable/res
+      -profile rules-engine
+      -cp http://localhost:8500
```
lenny-intel, line 342: Still need `--registry`  for service registration/discovery

tonyespy:
**Note** - the device-service source-tag commit is still marked as "WIP", as I'm waiting for the final release tag to get created...

snap/snapcraft.yaml
lenny-intel, line 342: if you use this form the value needs to be `consul.http://localhost:8500`, which is the default if you just use `-cp` with no value. I recommend just `-cp` with no value. keep it simple... ;-)

tonyespy: OK, it seemed to work fine w/out the `consul.`, but I'll change per your suggestion.

Why is the `consul.` needed? Is that a valid URI?

tonyespy: Hmmm, I thought you said if I added -cp, I wouldn't need to specify --registry???

lenny-intel: that is the spec we agreed on so the provider type can be specified in the URL. BootStrap parses it out.

lenny-intel: No, we separated them config provider is separate from registry provider.

tonyespy: Right, but the [config seed removal spec](https://github.com/edgexfoundry/edgex-docs/blob/master/docs_src/design/adr/0005-Service-Self-Config.md) interchangeably uses URI and URL, and never actually defines the supported schemes. What I was reacting to was the dotted scheme (eg. consul.http), as wasn't sure a scheme could include a "." character. Turns out it can per [RFC 3986 section 3.1](https://tools.ietf.org/html/rfc3986#section-3.1), so looks like we're OK. That said, it would be nice to enumerate the supported schemes someplace.

Issue #1919, opened 2019-10-15
Add hooks for hardware-assisted protection of Vault Master Key to security-secretstore-setup


bnevis-i:
# 🚀 Feature Request

Provide hooks into security-secretstore-setup for hardware-assisted protection secret store master key.

### Prerequisites / Conditions
In-scope for Geneva is the design of the feature.  Assuming the design is accepted, addition of the required hooks to security-secretstore-setup is a stretch goal.  Hardware-based implementations using any of the below mechanisms are out-of-scope for Geneva.

### Relevant Package
Affects security-secretstore-setup.

### Description
The master key that protects the secret store is currently left unencrypted on a docker volume where it could be copied by an attacker and used to decrypt a copy of the encrypted secret store.  [CWE-313](https://cwe.mitre.org/data/definitions/313.html).

This key should be protected by an implementer-defined mechanism that can leverage security hardware such as a PKCS11 token, a TPM, or a TEE running on-device that can provide an device-unique secret.

### Describe the solution you'd like
- This is an opt-in feature (must be explicitly enabled)
- Provide a migration option that will encrypt the secret store master key after-the-fact
- Create a salt from a cryptographically secure random number source, and persist it to disk.
- Spawn an executable that provides input key material to an [RFC-5689](https://tools.ietf.org/html/rfc5869) SHA-256 HKDF.
- Use the HKDF with supplied input key material to derive encryption keys for the master key shards.
- Use random number source to generate per-encryption initialization vectors.
- Encrypt the key shards with AES-256-GCM prior to being saved in persistent storage.

The security of the mechanism relies on keeping the input key private and the secret having sufficient entropy that it cannot be easily guessed.  At implementer would also need to mitigate against unauthorized software execution to avoid interception of the secret.

### Limitations
- There are no facilities for automatic rotation of any key (input key, derived key, secret store master key, or secret store data encryption key)

### Describe alternatives you've considered
- Depend on security hardware directly (rejected: not cross-platform portable)
- Use of go plugins (rejected: EdgeX has not ratified a plugin architecture)
- A pluggable encrypt()/decrypt() API.  (rejected: proposed for Edinburgh in 2018 and failed to obtain consensus; technical issues not fully resolved)

PR #2574, opened 2020-05-28
feat: Implement encryption of vault master key

bnevis-i:

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #1919

- [ ] Yes
- [x] No


internal/security/kdf/methods.go
```diff
@@ -0,0 +1,100 @@
...
+
+const (
+	saltFile   string = "kdf-salt.dat"
+	saltLength        = 32
+)
+
+// file implements a unit testing hook for file IO
+type file interface {
+	io.Reader
+	io.Closer
```
anonymouse64, line 30: this could just be io.ReadCloser no?

internal/security/kdf/methods.go
```diff
@@ -0,0 +1,100 @@
...
+	io.WriterAt
```
anonymouse64, line 31: if you don't need specifically WriteAt (which IMHO you don't), then you could get rid of this whole interface and just use `io.ReadWriteCloser` instead :tada:

internal/security/secretstore/init.go
```diff
@@ -80,6 +83,17 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
 	vaultHost := fmt.Sprintf("%s:%v", configuration.SecretService.Server, configuration.SecretService.Port)
 	intervalDuration := time.Duration(b.vaultInterval) * time.Second
 	vc := secretstoreclient.NewSecretStoreClient(lc, req, vaultScheme, vaultHost)
+	pipedHexReader := pipedhexreader.NewPipedHexReader()
+	kdf := kdf.NewKdf(fileOpener, configuration.SecretService.TokenFolderPath, sha256.New)
```
anonymouse64, line 87: is sha256 here good enough? should we be using something like sha3_384 instead ?

internal/security/secretstore/init.go
```diff
@@ -80,6 +83,17 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
...
+	vmkEncryption := NewVMKEncryption(fileOpener, pipedHexReader, lc, kdf)
+
+	if err := vmkEncryption.LoadIKM(os.Getenv("IKM_HOOK")); err != nil {
```
anonymouse64, line 90: mmm I think this code would be safer when written as

```suggestion
	err := vmkEncryption.LoadIKM(os.Getenv("IKM_HOOK"))
	defer vmkEncryption.WipeIKM()
	if err != nil {
```

this way even if we encounter a late error, if we had sensitive data in memory, we ensure it is wiped.

internal/security/secretstore/init.go
```diff
@@ -109,20 +123,33 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
...
-				}
 				_, err = vc.Unseal(&initResponse)
 				if err == nil {
 					shouldContinue = false
 				}
+				// We need the unencrypted initResponse in order to generate a temporary root token later
+				// Make a copy and save the copy, possibly encrypted
+				encryptedInitResponse := initResponse
+				// Optionally encrypt the vault init response based on whether encryption was enabled
+				if err := vmkEncryption.EncryptInitResponse(&encryptedInitResponse); err != nil {
```
anonymouse64, line 134: I like methods that maybe do nothing to be called `MaybeEncryptInitResponse`. Mild preference to re-name, OTOH the other pattern I've seen would be to have `vmkEncryption.LoadIKM()` return a specific sentinel error value that you can check for and if that is returned then the caller doesn't have hardware encryption support and thus doesn't ever call EncryptInitResponse, and EncryptInitResponse will be a bit simpler and easier to reason about since it doesn't have this "maybe" quality about it.

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,280 @@
...
+	initResp.EncryptedKeys = nil
+	initResp.Nonces = nil
+	return nil
+}
+
+//
+// Internal methods
+//
+
+func (v *VMKEncryption) gcmEncryptKeyshare(keyshare []byte, counter int) ([]byte, []byte, error) {
```
anonymouse64, line 196: this function could use some more comments if only to help verify the pattern implemented here

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,280 @@
...
+	nonce := make([]byte, aesgcm.NonceSize())
+	if _, err := io.ReadFull(rand.Reader, nonce); err != nil {
+		v.logger.Error(fmt.Sprintf("failed to initialize random nonce %s", err.Error()))
+		return nil, nil, err
+	}
+
+	// Encrypt they key share, overwrite the input, and wipe the encrypted copy
+	// (input and output size are the same with AES)
+	ciphertext := aesgcm.Seal(nil, nonce, keyshare, nil)
+	wipeKey(keyshare)
```
anonymouse64, line 228: any reason not do do `defer wipeKey(keyshare)` at the top of this function too?

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,280 @@
...
+	return nil
+}
+
+//
+// Internal methods
+//
+
+func (v *VMKEncryption) gcmEncryptKeyshare(keyshare []byte, counter int) ([]byte, []byte, error) {
+
+	info := fmt.Sprintf("vault%d", counter)
```
anonymouse64, line 198: this needs the same comment as from the accompanying decrypt function, as well as one about this needing to be unique I think, no?

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,280 @@
...
+	plaintext, err := aesgcm.Open(nil, nonce, keyshare, nil)
+	if err != nil {
+		wipeKey(keyshare)
+		if plaintext != nil {
+			wipeKey(plaintext)
+		}
+		v.logger.Error(fmt.Sprintf("failed to decrypt ciphertext: %s", err.Error()))
+		return nil, err
+	}
+	wipeKey(keyshare)
```
anonymouse64, line 268: same comment as above, is there a reason not to always do this when we exit the function with a `defer` at the top ?

hutchic:
I hope this question isn't terribly naive; but could one not have reduced a lot of this boilerplate by relying on the built-in mechanisms vault offers to encrypt / decrypt the vault initialization keys?

specifically: https://www.vaultproject.io/docs/concepts/pgp-gpg-keybase

bnevis-i:
> I hope this question isn't terribly naive; but could one not have reduced a lot of this boilerplate by relying on the built-in mechanisms vault offers to encrypt / decrypt the vault initialization keys?
> 
> specifically: https://www.vaultproject.io/docs/concepts/pgp-gpg-keybase

That is a great question.  Vault in its normal mode of operation is designed to work with human operators -- it requires collusion to get together all the key holders and unseal the vault.  The purpose of the PGP mechanism is that the sys admin can initialize the vault and issue key shares without ever having to see the plaintext of the keys.  For an embedded IoT scenario, it is likely that the device is headless (no keyboard nor mouse).  Even if a device had KVM redirection, one still wouldn't want to hire a staff of keyholds to go around unsealing the Vault in fleets of IoT devices.  To do this in an automated way using the PGP functionality would mean that the PGP private key would need to be on-disk, which is exactly the same risk as having the vault key shares unencrypted on disk: i.e.: the key is close-by.  The whole idea here is to avoid putting unencrypted key material on disk.

internal/security/kdf/methods.go
```diff
@@ -0,0 +1,100 @@
...
+	infoBytes := []byte(info)
+	kdfReader := hkdf.New(kdf.hashConstructor, ikm, salt, infoBytes)
+	key := make([]byte, keyLen)
+	kdfReader.Read(key)
+	return key, nil
+}
+
+// initializeSalt recovers the KDF salt value from a file
+// or installs a new salt
+func (kdf *kdfObject) initializeSalt() ([]byte, error) {
```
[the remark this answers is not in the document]
bnevis-i: Can you think of a way to do this without the os.Stat() call.  os.Stat() is a system call and unfortunately there is not an io.Statable interface.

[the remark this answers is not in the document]
anonymouse64: why is os.Stat difficult to mock? what I would do normally is just define in this file:

```go
var (
    osStat = os.Stat
)

... 
_, err := osStat(saltFile)
...
```

and then another test file, export_test.go:

```go
package methods

func MockOsStat(f func(string) (os.File, error)) (restore func()) {
    old := osStat
    osStat = f 
    return func() {
        osStat = old
    }
}
```

then finally mock it in a test like this:

```go
package methods_test

func (m *methodsSuite)TestMethods(c *C) {
     r := methods.MockOsStat(func(fname string) (os.File, error) {
        c.Assert(fname,Equals,"expected-file-name")
        return nil, nil
     })
     defer r()

     // do the test
}
```

internal/security/kdf/methods.go
bnevis-i: Fixed.

internal/security/kdf/methods.go
bnevis-i: Fixed.

internal/security/kdf/methods.go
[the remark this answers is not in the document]
bnevis-i: That worked. Fixed.

internal/security/kdf/methods.go
```diff
@@ -0,0 +1,100 @@
...
+	salt := make([]byte, saltLength)
+	saltPath := path.Join(kdf.persistencePath, saltFile)
+	saltFileWriter, err := kdf.fileIoPerformer.OpenFileWriter(saltPath, os.O_RDWR|os.O_CREATE, 0600)
```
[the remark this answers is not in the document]
bnevis-i: n/a - read/write paths are now completely separated in rewritten salt function.

internal/security/kdf/methods.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, errors.New("saltFileWriter does not implement required read/write methods")
+	}
+	nbytes, _ := saltFileObj.Read(salt)
+	if nbytes == 0 || nbytes != saltLength {
+		_, err := rand.Read(salt)
+		if err != nil {
+			_ = saltFileObj.Close()
+			return nil, err
+		}
+		nwritten, err := saltFileObj.WriteAt(salt, 0)
```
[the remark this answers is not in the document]
bnevis-i: see rewritten method.

internal/security/kdf/methods.go
```diff
@@ -0,0 +1,100 @@
...
+		if err != nil {
+			_ = saltFileObj.Close()
+			return nil, err
+		}
+		if nwritten != len(salt) {
+			err := errors.New("Failed to write entire contents of salt file; encryption key will likely be unrecoverabe")
```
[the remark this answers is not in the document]
bnevis-i: fixed

internal/security/kdf/methods.go
```diff
@@ -0,0 +1,100 @@
...
+	err = saltFileObj.Close()
+	if err != nil {
+		return nil, err
+	}
+	return salt, nil
+}
+
+//
+// ECCN: 5D002TSU
+//
```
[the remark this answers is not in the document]
bnevis-i: moved.

internal/security/kdf/methods_test.go
```diff
@@ -0,0 +1,141 @@
+//
```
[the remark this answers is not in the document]
bnevis-i: it is now an error for salt to be of unexpected length for both reading and writing.

internal/security/secretstore/init.go
bnevis-i: Yes, SHA256 is sufficient for key deriviation.

internal/security/secretstore/init.go
bnevis-i: rewritten as suggested.

internal/security/secretstore/init.go
bnevis-i: Resolved by "if vmkEncryption.IsEncrypting()" guard around the call and throwing an error inside of the implementation is a key was not supplied.

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+	v.ikm = ikm
+	v.encrypting = true
+	v.logger.Info("Enabled encryption of Vault master key")
+	return nil
+}
+
+// WipeIKM scrubs the input key material from memory
+func (v *VMKEncryption) WipeIKM() {
+	copy(v.ikm, make([]byte, len(v.ikm)))
```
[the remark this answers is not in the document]
bnevis-i: comment added

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,280 @@
...
+			return err
+		}
+
+		keyShare, err := v.gcmDecryptKeyshare(cipherText, nonce, i) // Unwrap using a unique AES key
+		if err != nil {
+			v.logger.Error(fmt.Sprintf("failed to unwrap key %d: %s", i, err.Error()))
+			return err
+		}
+
+		newKeys[i] = hex.EncodeToString(keyShare)
```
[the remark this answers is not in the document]
bnevis-i: Not concerned since data is opaque and the key is summetric.

internal/security/secretstore/vmkencryption.go
bnevis-i: added

internal/security/secretstore/vmkencryption.go
bnevis-i: addressed in previous comment.

internal/security/secretstore/vmkencryption.go
bnevis-i: removed to the top as a defer.

internal/security/secretstore/vmkencryption.go
bnevis-i: technically not needed since this keyshare comes in encrypted, but made same change as for encrypt

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,291 @@
...
+		kdf:            kdf,
+		logger:         logger,
+		encrypting:     false,
+	}
+}
+
+// LoadIKM loads input key material from the specified path
+func (v *VMKEncryption) LoadIKM(ikmBinPath string) error {
+	if ikmBinPath == "" {
+		v.logger.Info("Not encrypting Vault master key")
```
lenny-intel, line 85: This should be returned as an error. i.e. shouldn't get here if not `encrypting Vault master key`.

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,291 @@
...
+		return nil
+	}
+	ikm, err := v.pipedHexReader.ReadHexBytesFromExe(ikmBinPath, []string{})
+	if err != nil {
+		v.logger.Error(fmt.Sprintf("Error reading input key material from IKM_HOOK - encryption not enabled: %s", err.Error()))
```
lenny-intel, line 90: I don't think this module should be logging errors. It should add context to the error and return the error. Otherwise the error is getting logged again by the caller.

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,291 @@
...
+		return err
+	}
+	v.ikm = ikm
+	v.encrypting = true
+	v.logger.Info("Enabled encryption of Vault master key")
```
lenny-intel, line 95: This should be logged by the caller.

internal/security/secretstore/vmkencryption_test.go
```diff
@@ -0,0 +1,111 @@
...
+	initialInitResp := secretstoreclient.InitResponse{
+		Keys:       []string{"aabbcc", "ddeeff"},
+		KeysBase64: []string{"qrvM", "3e7/"},
+	}
+	initResp := initialInitResp
+
+	// Act & Assert
+	vmkEncryption := NewVMKEncryption(fileOpener, pipedHexReader, lc, kdf)
+	err := vmkEncryption.LoadIKM("/bin/myikm")
+	assert.NoError(t, err)
```
lenny-intel, line 63: Use `require.NoError()` to stop test here if fails. Do this in other locations and other tests where appropriate.

internal/security/secretstore/vmkencryption_test.go
```diff
@@ -0,0 +1,111 @@
...
+	// Act / Assert
+	vmkEncryption := NewVMKEncryption(fileOpener, pipedHexReader, lc, kdf)
+	encrypting := vmkEncryption.IsEncrypting()
+	assert.False(t, encrypting)
+
+	vmkEncryption.WipeIKM()
+
+	fileOpener.AssertExpectations(t)
+	pipedHexReader.AssertExpectations(t)
+	kdf.AssertExpectations(t)
```
lenny-intel, line 40: Consider using table drive tests to consolidate this duplicated code.

lenny-intel (changes requested):

anonymouse64:
Just a comment that I do intend to do another pass over this sometime in the next 2 weeks, but if that's too late feel free to move on without me

internal/security/kdf/methods_test.go
```diff
@@ -0,0 +1,279 @@
...
+	. "github.com/edgexfoundry/go-mod-secrets/pkg/token/fileioperformer/mocks"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+const expectedKey = "1060e4e72054653bf46623844033f5ccc9cff596a4a680e074ef4fd06aae60df"
+
+// TestNoErrorKdfCreateSalt tests the golden path for a new salt
+func TestNoErrorKdfCreateSalt(t *testing.T) {
```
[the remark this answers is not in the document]
bnevis-i: It looks duplicated, but each test case runs through a different set of mocks because they test error handling at different points in the logic.

internal/security/secretstore/vmkencryption_test.go
bnevis-i: For same reason as earlier (each test registers different mocks) these tests do not lend themselves well to table-ization.

bnevis-i:
Rebased.  Awaiting final review comments.

internal/security/pipedhexreader/pipedhexreader.go
```diff
@@ -0,0 +1,53 @@
...
+type pipedHexReader struct{}
+
+// NewPipedHexReader creates a new PipedHexReader
+func NewPipedHexReader() PipedHexReader {
+	return &pipedHexReader{}
+}
+
+// ReadHexBytesFromExe see interface.go
+func (phr *pipedHexReader) ReadHexBytesFromExe(executablePath string, args []string) ([]byte, error) {
+	sanitizedExecutable, err := exec.LookPath(executablePath)
```
tonyespy, line 26: Wouldn't requiring a fully qualified path lock things down a bit more? I suppose the only the only thing leaked would be the set of command-line args passed to the IKM application...

tonyespy (changes requested):
Please see comments inline... A couple of global comments:

 * For such a large PR (+1k lines of diff) I would have expected more than a single commit. I'm not going to ask that you break it up at this point, but please consider for future PRs.
  * The commit itself is a single commit summary, at least include a "fixes: 1919" in the body of your commit message, although an actual commit message body with a little more context would be nice.
  * I assume you intend to squash the "fixup!" commits before landing?
  * Can you please add a (conventional commit) scope to your commit summary?

internal/security/kdf/interface.go
```diff
@@ -0,0 +1,15 @@
...
+//
+
+package kdf
+
+// KeyDeriver is the interface that the main program expects
+// for returning a derived key
+type KeyDeriver interface {
+	// DeriveKey returns a byte array that is of keyLen length and
+	// an error if errors where encountered while deriving the key
+	DeriveKey(ikm []byte, keylkeyLen uint, info string) ([]byte, error)
```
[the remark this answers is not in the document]
bnevis-i: > Is the 2nd parameter supposed to be `keyLen` (v.s. `keylkeyLen`)?

Typo fixed.

> Also why type of key is being returned, symmetric, public, private? Do algorithms matter here?

The key is algorithm-less -- it is a bunch of random stuff that could be used as a symmetric key, a password, input to a second KDF, et cetera.  It is for sure not a public/private keypair, however.

> It would also be helpful to document the parameters of DeriveKey

Added

internal/security/kdf/methods.go
```diff
@@ -0,0 +1,131 @@
...
+// Copyright (c) 2020 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// ECCN: 5D002TSU
```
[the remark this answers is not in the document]
bnevis-i:
```
// US Export Control Classification Number (ECCN): 5D002TSU
```

Apache has more user-consumable writeup here: http://www.apache.org/licenses/exports/

internal/security/kdf/methods.go
```diff
@@ -0,0 +1,131 @@
...
+	"io"
+	"os"
+	"path"
+
+	"golang.org/x/crypto/hkdf"
+
+	"github.com/edgexfoundry/go-mod-secrets/pkg/token/fileioperformer"
+)
+
+const (
```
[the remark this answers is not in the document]
bnevis-i: Added documentation.

// The KDF salt adds additional randomness to the KDF input key
// and provides a mechanism to force generation of unique keys
// in the event that the KDF inputKeyMaterial is less than random.

Basically, it is random insurance, and ensures that even reinstallation of EdgeX on the same hardware generates unique keys to that installation.

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,291 @@
...
+DO NOT REMOVE EXCEPT BY PERMISSION OF THE AUTHOR
+
+Vault Master Key Encryption Feature
+
+The purpose of this feature is to provide a secure way to unlock Vault
+without requiring human intervention to supply the Vault Master Key.
+
+This feature requires a sufficiently secure source of randomness in order
+to unlock the Vault. This randomness should have at least 256 bits of entropy
+and be backed by hardware secure storage such as a TPM or TEE or a hardware
```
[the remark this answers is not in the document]
bnevis-i: I agree that this is splitting hairs but have removed the explanatory text so that it doesn't create additional confusion.

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,291 @@
...
+security token.  If using a TPM, the secret should be bound to platform
+configuration registers to attest the system state.
```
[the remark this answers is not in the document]
bnevis-i: > So what you're saying is that when using a TPM, a secret of at least 256 bits of entropy should be sealed to the TPM using a PCR-based policy? I suppose the actual policy (i.e. which PCRs) is left up to the implementer of the IKM?

Yes.

> How is the secret protected from other processes on the system (i.e. if the secret is bound using a PCR-based policy, as long as the PCR values on boot match the PCR values when the secret was sealed, the secret can be read by any application on the system)? Sometimes this is handled by extending one or more secrets after something has been unsealed, but this approach would cause restarts of EdgeX to fail (i.e. the IKM application must always return the same secret given an identical set of command-line options/inputs).

Arguably if one cannot prevent he execution of malicious code on the system this mechanism will do little.  Poisoning the PCR to prevent re-reads would have the above stated effect (only have one chance per boot to unseal Vault).  An implementation might choose to do something in between, such as reading the value out of a TPM, storing it in a kernel keyring, and then poisoning the PCR, and rely on user/group permissions to prevent access from then on.

The scope of the implementation is providing a mechanism to encrypt `resp-ini.json` using a key that is ultimately not stored in plaintext on media.

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,291 @@
...
+		return nil, nil, err
+	}
+
+	nonce := make([]byte, aesgcm.NonceSize())
+	if _, err := rand.Read(nonce); err != nil {
+		v.logger.Error(fmt.Sprintf("failed to initialize random nonce %s", err.Error()))
+		return nil, nil, err
+	}
+
+	// Encrypt they key share (plaintext to be wiped on exit by deferred function)
```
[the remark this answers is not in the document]
bnevis-i: ✔

go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	github.com/pkg/errors v0.8.1
 	github.com/robfig/cron v0.0.0-20180505203441-b41be1df6967
 	github.com/stretchr/testify v1.5.1
+	golang.org/x/crypto v0.0.0-20190701094942-4def268fd1a4
```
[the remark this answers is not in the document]
bnevis-i: > Is there any particular reason the tag here is from last Summer? Is that when you started this implementation? 

In fact, yes, this PR has been in the works for well over a year, been open for public review since March 28, has gone through an internal SDL process, and had its design and implementation looked over by professional cryptographers.
However, I have bumped the version using "go get -u golang.org/x/crypto"

> Also does dependabot grok these versions? If not, how do we know if there have been fixes to the hkdf package being used?

Inspection of https://github.com/golang/crypto will in fact show that there are no release tags ever dropped on that repository.  This is outside of my control.  However, it is generally a good idea with crypto to not update the code casually due to the risk of changing the output.

internal/security/kdf/interface.go
```diff
@@ -0,0 +1,15 @@
...
+//
+
+package kdf
+
+// KeyDeriver is the interface that the main program expects
+// for returning a derived key
+type KeyDeriver interface {
+	// DeriveKey returns a byte array that is of keyLen length and
+	// an error if errors where encountered while deriving the key
+	DeriveKey(inputKeyingMaterial []byte, keyLen uint, info string) ([]byte, error)
```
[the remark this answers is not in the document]
bnevis-i:
```
// KeyDeriver is the interface that the main program expects
// for returning a derived key. inputKeyingMaterial and info
// are inputs to the key deriviation function, an keyLen
// is the desired length of the derived key.
```

According to this page https://blog.golang.org/godoc there is no formal javadoc-like annotion syntax for golang.

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,289 @@
+//
+// Copyright (c) 2020 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// ECCN: 5D002TSU
```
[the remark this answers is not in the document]
bnevis-i: ibid

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,289 @@
...
+without requiring human intervention to supply the Vault Master Key.
+
+This feature requires a sufficiently secure source of randomness in order
+to unlock the Vault. This randomness should have at least 256 bits of entropy
+and be backed by hardware secure storage such as a TPM or TEE or a hardware
+security token.  If using a TPM, the secret should be bound to platform
+configuration registers to attest the system state.
+
+The randomness should be output as a string of hex-encoded octets to standard
+output and the path to the executable should be specified in the
```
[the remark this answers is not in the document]
bnevis-i: I only ever proposed and implement the environment variable method.

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,289 @@
...
+
+	// Check prerequisite (key has been loaded)
+	if !v.encrypting {
+		err := fmt.Errorf("Cannot encrypt init response as key has not been loaded")
+		v.logger.Error(err.Error())
+		return err
+	}
+
+	newKeys := make([]string, len(initResp.Keys))
+	newNonces := make([]string, len(initResp.Keys))
```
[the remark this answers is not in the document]
bnevis-i: This is correct as-is.  The original input has a set of N 'Keys'.  The output is N encrypted keys, and N nonces to initialize the cipher.  There is no initResp.Nonces in the original resp-init.json (as all the keys are stored in plaintext).

internal/security/pipedhexreader/pipedhexreader.go
bnevis-i: LookPath("echo") and LookPath("/bin/echo") both return /bin/echo.   I think the current implementation preserves flexibility for a user that wants to lock it down to a full path.

internal/security/kdf/interface.go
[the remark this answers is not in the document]
bnevis-i: > The commit itself is a single commit summary, at least include a "fixes: 1919" in the body of your commit message, although an actual commit message body with a little more context would be nice.
> I assume you intend to squash the "fixup!" commits before landing?

Both of these are addressed.

bnevis-i:
Request re-review from @lenny-intel and @tonyespy

internal/security/kdf/methods_test.go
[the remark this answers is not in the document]
lenny-intel: Yep, what to mock could be indicated by a test struct property. Just a suggestion to reduce the duplication of code.

internal/security/secretstore/init.go
```diff
@@ -109,19 +129,36 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
...
 				if err := loadInitResponse(lc, fileOpener, configuration.SecretService, &initResponse); err != nil {
 					lc.Error(fmt.Sprintf("unable to load init response: %s", err.Error()))
-					os.Exit(1)
+					os.Exit(1) // can't return false since in anonymous func()
+				}
+				// Optionally decrypt the vault init response based on whether encryption was enabled
+				if vmkEncryption.IsEncrypting() {
+					if err := vmkEncryption.DecryptInitResponse(&initResponse); err != nil {
+						lc.Error(fmt.Sprintf("failed to decrypt key shares for sercret store unsealing: %s", err.Error()))
+						os.Exit(1) // can't return false since in anonymous func()
```
lenny-intel, line 160: anonymous func() can have a `bool` return type and then check it on line 138 and return `false` if it is `false`

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,289 @@
...
+		logger:         logger,
+		encrypting:     false,
+	}
+}
+
+// LoadIKM loads input key material from the specified path
+func (v *VMKEncryption) LoadIKM(ikmBinPath string) error {
+	if ikmBinPath == "" {
+		v.logger.Error("ikmBinPath is required")
+		return nil
```
lenny-intel, line 86: Why logging error but not returning the error?

lenny-intel: Also as stated before IMO the module shouldn't be logging errors. It should always return error to caller.

lenny-intel (changes requested):

internal/security/secretstore/init.go
bnevis-i: can do that.

internal/security/secretstore/vmkencryption.go
bnevis-i: vkmencryption.go will now wrap and return errors.

bnevis-i:
@lenny-intel Please see fixup commit 57d66829982dce96f0a7a7c2bb0d46207d90292f.

bnevis-i:
Rebased.  Awaiting further review.  @tonyespy @lenny-intel

internal/security/secretstore/vmkencryption.go
```diff
@@ -0,0 +1,271 @@
...
+	kdf            kdf.KeyDeriver
+	logger         logger.LoggingClient
+	encrypting     bool
+	ikm            []byte
+}
+
+// NewVMKEncryption - constructor
+func NewVMKEncryption(fileOpener fileioperformer.FileIoPerformer,
+	pipedHexReader pipedhexreader.PipedHexReader,
+	logger logger.LoggingClient,
```
lenny-intel, line 70: Can remove logger now that it isn't used.

lenny-intel (changes requested):

internal/security/secretstore/vmkencryption.go
bnevis-i: Logger removed.

lenny-intel (dismissed):
LGTM

internal/security/secretstore/vmkencryption.go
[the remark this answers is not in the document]
tonyespy: Sure, but this very comment says:

> and the path to the executable should be specified in the configuration.toml of security-secretstore-setup.

internal/security/pipedhexreader/interface.go
```diff
@@ -0,0 +1,15 @@
...
+//
+
+package pipedhexreader
+
+// PipedHexReader is an interface to read hex bytes from
+// standard output stream of an executable into a byte array
+type PipedHexReader interface {
+	// ReadHexBytesFromExe invokes executable with args
+	// and reads hex bytes from stdout and returns an array
+	ReadHexBytesFromExe(executablePath string, args []string) ([]byte, error)
```
tonyespy, line 14: From what I recall, your original proposal had the ability for command-line options to be specified and passed to the IKM binary, however in this PR, while `ReadHexBytesFromExe` includes an args parameter, an empty slice is always used (except for some of the test cases). Why keep this parameter?

tonyespy (commented):
Looks much better, but I still have a couple of comments inline...

internal/security/kdf/interface.go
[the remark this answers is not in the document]
bnevis-i:
```
	// DeriveKey returns a byte array that is of keyLen length and
	// an error if errors where encountered while deriving the key
	// inputKeyingMaterial and info are inputs
	// to the key deriviation function, an keyLen
	// is the desired length of the derived key.
	// inputKeyingMaterial is a secret
	// and info is used to cause the KDF to generate
	// different output keys from the same inputKeyingMaterial.
	// Please see the application notes for RFC 5869
	// https://tools.ietf.org/html/rfc5869#3 for
	// details for details about the key derivation algorithm.
```

This is used to ensure that each shard of the Vault master key is encrypted using a unique (but reproducibly created) key.

internal/security/secretstore/vmkencryption.go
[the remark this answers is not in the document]
bnevis-i: Remedied.  Thank you for pointing that out.  I wanted to be able to late bind the IKM_HOOK into the container without necessarily having to fork security-secretstore-setup in order to do so.

```
The randomness should be output as a string of hex-encoded octets to standard
output and the executable (or path to the executable) should be specified in the
IMK_HOOK environment variable.
```

internal/security/pipedhexreader/interface.go
bnevis-i: Good point.  Removed.

bnevis-i:
Rebased to master with latest comments addressed.  Awaiting review by  @tonyespy

tonyespy (approved):
Thanks for addressing all my comments!

lenny-intel:
@bnevis-i , changing the scope from `(security-services)` to `(security)` in the commit message and PR title will fix the `Semantic Pull Request` error not that the BOT is enabled.

lenny-intel (approved):

lenny-intel:
@bnevis-i , Build failure is known intermittent issue  @tonyespy is looking into. When you fix the commit message it will kick off gain and hopefully make it through

bnevis-i:
Rebased and satisfied the bot.  Need re-approval to merge.

lenny-intel (approved):

PR #2605, opened 2020-06-30
fix: Created timestamp is 0 on message queue

### LINKED ISSUE #2602: created timestamp always 0
# 🐞 Bug Report

### Affected Services
Core Data

### Is this a regression?
Worked in Edinburgh but not Fuji and later

The created timestamp on event/reading is 0.  The created timestamp does get set in the database, but it is not set back into the event/reading and when the event/reading is pushed to app services (via 0MQ or message bus), the created timestamp is always zero.

in a nutshell -- Event param to AddEvent the event is not a pointer (https://github.com/edgexfoundry/edgex-go/blob/ad8ce46ea1b4c1cfb589e4a1128cf4f50d0a06e9/internal/pkg/db/mongo/data.go#L51)
the created and modified are being set here in this call just before save to DB but the call to put it on the bus happens after this call returns -- using the same event that was passed in -- which wont have those fields populated

## 🔥 Exception or Error
no exception or error - just no created timestamp

## 🌍  Your Environment
works the same for all environments

**EdgeX Version:**
Fuji or later

**Anything else relevant?**
See #core slack channel exchange around June 25 for exchange between bug finder and Lenny/Rocky

lindseysimple:
Fix by using the db creation time on message queue for Event.
closes #2602

Signed-off-by: Lindsey Cheng <beckysocute@gmail.com>

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The created timestamp of Event/Reading is zero on the message queue (can be seen on the app services pushed via 0MQ or message bus).
Issue Number: #2602 

## What is the new behavior?
The created timestamp of Event/Reading value should be the same as the created field saved in the database on the message queue, which is the entry creation timestamp in the database.

- [ ] Yes
- [x] No

No.

### DIFF
--- a/internal/core/data/event.go
+++ b/internal/core/data/event.go
@@ -134,6 +134,9 @@ func addNewEvent(
 
 	// Add the event and readings to the database
 	if configuration.Writable.PersistData {
+		if e.Created == 0 {
+			e.Created = db.MakeTimestamp()
+		}
 		id, err := dbClient.AddEvent(e)
 		if err != nil {
 			return "", err
--- a/internal/pkg/db/redis/data.go
+++ b/internal/pkg/db/redis/data.go
@@ -1036,10 +1036,6 @@ func (c *Client) ScrubAllValueDescriptors() error {
 
 // ************************** HELPER FUNCTIONS ***************************
 func addEvent(conn redis.Conn, e correlation.Event) (id string, err error) {
-	if e.Created == 0 {
-		e.Created = db.MakeTimestamp()
-	}
-
 	if e.ID == "" {
 		e.ID = uuid.New().String()
 	}


internal/core/data/event.go
```diff
@@ -139,6 +139,10 @@ func addNewEvent(
 			return "", err
 		}
 		e.ID = id
+		savedEvent, err := dbClient.EventById(id)
```
lenny-intel, line 142: Would it be better to simply set `created` prior to calling AddEvent so we don't have to make a call to read back the Event from the DB?

rsdmike: Alternatively, why don't we have AddEvent return the event it inserted? That's more standard isn't it?

lindseysimple: @lenny-intel, updated the fix based on your suggestion to set the `created` timestamp before calling the `dbClient.AddEvent` function.

@rsdmike , the suggestion sounds great and we'll implement in the v2 API changes. Currently the change is included in https://github.com/edgexfoundry/edgex-go/pull/2610.

internal/core/data/event.go
```diff
@@ -134,6 +134,9 @@ func addNewEvent(
 
 	// Add the event and readings to the database
 	if configuration.Writable.PersistData {
+		if e.Created == 0 {
+			e.Created = db.MakeTimestamp()
```
lenny-intel, line 138: Also need to remove setting it in the DB layer for Redis (so saved value is the value set here). Mongo code is going to be removed, so you can ignore it.

lindseysimple:
@lenny-intel Hi Lenny, I've removed the `created` value check in `addEvent` function of db layer, could you please re-review that? Thanks for your help.

PR #2666, opened 2020-08-21
fix(snap): remove support-logging

tonyespy:
Remove deprecated support-logging service from snap.

## PR Checklist

- [NA] `make test` has completed successfully

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: removes the deprecated support-logging service from the snap

- [x] Yes
- [ ] No

Up til this point, it's not been possible to update from one major/minor version of EdgeX due to the use of sequential epochs in snapcraft.yaml. This policy may be changed with the Hanoi. If so, when someone attempts to refresh from the geneva snap and they have support-logging enabled, the refresh/update should fail to install with an appropriate error message (**note**, this has not been implemented in this PR). This is similar to what would happen if someone were to update from geneva and they had mongo or the old support rules-engine enabled.

No

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -233,23 +233,6 @@ apps:
     daemon: simple
     plugs: [network, network-bind]
     stop-timeout: 10s
-  support-logging:
-    adapter: none
-    after:
-      - redis
-      - security-proxy-setup
-    command: bin/support-logging -confdir $SNAP_DATA/config/support-logging/res --registry
-    command-chain:
-      - bin/security-secret-store-env-var.sh
-    environment:
-      # for support-logging we also want the default persistence to be on a
-      # file so that we can use redis (which isn't supported with support-logging)
-      # in the snap by default
-      Writable_Persistence: 'file'
-      SecretStore_TokenFile: $SNAP_DATA/secrets/edgex-support-logging/secrets-token.json
-    daemon: simple
-    plugs: [network, network-bind]
-    stop-timeout: 10s
   support-notifications:
     adapter: none
     after:
@@ -295,7 +278,6 @@ apps:
       - security-proxy-setup
       - core-data
       - core-metadata
-      - support-logging
     command: bin/device-virtual $CONF_ARG $PROFILE_ARG $REGISTRY_ARG
     daemon: simple
     environment:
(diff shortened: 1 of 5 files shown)


tonyespy:
> > Up til this point, it's not been possible to update from one major/minor version of EdgeX due to the use of sequential epochs in snapcraft.yaml. This policy may be changed with the Hanoi, however if someone attempts to refresh from the geneva snap and they have support-logging enabled, the refresh/update will fail to install. This is similar to what would happen if someone were to update from geneva and they had mongo or the old support rules-engine enabled.
> 
> To be clear, what this PR is doing is removing the service so that it doesn't attempt to be started, so before this PR if someone refreshed from geneva/stable (or latest/stable) to edge, they would get broken because support-logging is broken on edge, but now after this PR if they did that same refresh it wouldn't be broken at least due to support-logging because now support-logging just disappears correct?

I'm not sure what you mean by "support-logging is broken on edge"? We haven't yet updated the epoch for Hanoi, so technically it's possible for someone to refresh from geneva to latest/edge, and it they had support-logging enabled, then it will be gone after the refresh, and no errors will be generated.

When we landed the changes for the legacy rules-engine and mongo, we didn't add logic to the refresh hook to fail if either were being used, because we'd been using epochs to prevent refreshes between releases. As such, I'll adjust this PR to bump the epoch to 4 to prevent refreshes. If we decide to support refreshes from geneva, then we'll need to ensure that the refresh hook as the proper logic to detect current usage of deprecated services.

anonymouse64:
@tonyespy can you clarify what you mean by this in the PR description: 

> however if someone attempts to refresh from the geneva snap and they have support-logging enabled, the refresh/update will fail to install

I'm confused about this statement since in your response just now you also say:

> if they had support-logging enabled, then it will be gone after the refresh, and no errors will be generated.

snap/snapcraft.yaml
```diff
@@ -16,10 +16,9 @@ icon: snap/local/assets/edgex-snap-icon.png
 # Different epochs prevent refreshes between major versions of EdgeX due
 # to continued configuration changes.
 #
-# delhi: 0, edinburgh: 1, fuji: 2, geneva: 3
-epoch: 3
+# delhi: 0, edinburgh: 1, fuji: 2, geneva: 3, hanoi: 4
+epoch: 4
 
-# TODO: add armhf here when that's supported
```
anonymouse64, line 22: why remove this TODO ?

tonyespy: Because we're never going  to support it, and I got tired of seeing this comment everytime I edited snapcraft.yaml. ;)-

anonymouse64 (dismissed):

anonymouse64:
Built and installed on ubuntu 20.04 classic, all services came up and support-logging was gone, so lgtm (meant for this to be a part of the previous review)

lenny-intel (approved):
LGTM

Issue #2667, opened 2020-08-21
Update Core Data API Swagger for new `Tags` field on Event


lenny-intel:
# 📚 Docs or Wiki Bug Report

### Description

 Tags field has been added to the V1 Event model and the V2 Event Model and DTO. Core Data API Swagger files need to be updated for both V1 & V2 to include this new Tags field.

This is dependent on this issue which adds this new field. 
https://github.com/edgexfoundry/go-mod-core-contracts/issues/270

PR #2676, opened 2020-09-01
feat: Add Tags property to Event in V1 & V2 swagger.

lenny-intel:
## PR Checklist

- [ ] `make test` has completed successfully

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes

Swagger not updated for recently add Tags property on Event

Issue Number: #2667

## What is the new behavior?

V1 & V2 Swagger now updated for recently add Tags property on Event

- [ ] Yes
- [x] No

No

### DIFF
--- a/openapi/v1/core-data.yaml
+++ b/openapi/v1/core-data.yaml
@@ -954,6 +954,12 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/reading'
+        tags:
+          title: tags
+          uniqueItems: false
+          type: array
+          items:
+            $ref: '#/components/schemas/tagsKeyValue'
       description: Core device/sensor event
     reading:
       title: reading
@@ -1030,6 +1036,18 @@ components:
           type: string
       description: Core and MetaData value descriptor that describes
         device/sensor data sent and received
+    tagsKeyValue:
+      description: Defines the key/value pair of the Tag
+      type: object
+      properties:
+        key:
+          description: The key to identify the Tag
+          type: string
+          example: "gateway-id"
+        value:
+          description: The value for the Tag
+          type: string
+          example: "HoustonStore-000123"
   requestBodies:
     event:
       content:
--- a/openapi/v1/core-data.yaml
+++ b/openapi/v1/core-data.yaml
@@ -956,10 +956,14 @@ components:
             $ref: '#/components/schemas/reading'
         tags:
           title: tags
+          description: "List of zero or more Tags attached to the Event which give more context to the Event"
           uniqueItems: false
-          type: array
-          items:
-            $ref: '#/components/schemas/tagsKeyValue'
+          type: object
+          example: {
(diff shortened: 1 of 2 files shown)


lenny-intel:
recheck

cloudxxx8:
recheck

openapi/v1/core-data.yaml
```diff
@@ -1030,6 +1036,18 @@ components:
           type: string
       description: Core and MetaData value descriptor that describes
         device/sensor data sent and received
+    tagsKeyValue:
```
cloudxxx8, line 1039: we don't need to define `tagsKeyValue` object, because it would become.
[{"key": "gateway-id", "value": "HoustonStore-000123"}, {"key": "OOO", "value": "XXX"}]

openapi/v1/core-data.yaml
```diff
@@ -954,6 +954,12 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/reading'
+        tags:
+          title: tags
+          uniqueItems: false
+          type: array
```
cloudxxx8, line 960: the type could a pure object without any reference, but the example could be identified clearly.
please see the example of protocol properties
https://github.com/edgexfoundry/edgex-go/blob/9f0651306d868a93c4f6d04f573f0b50f68ed542/openapi/v1/core-metadata.yaml#L2210-L2214

cloudxxx8 (changes requested):

openapi/v1/core-data.yaml
lenny-intel: Yep, yo are correct!! Good catch!

I thought this is required for the data to be un-marshaled correctly into the `map[string]string`, but per my test below it is a simple object with the the keys as properties and the values as the property's values.

I add this to V1's event.go just prior to persisting the Event and then queered the events.
```
	e.Tags = make(map[string]string)
	e.Tags["gateway-id"] = "HoustonStore-000123"
	e.Tags["OOO"] = "YYY"
```
and the resulting JSON is this:
```
    {
        "id": "a1abc227-f17d-4c74-957e-523d57d6ac2a",
        "device": "Random-Boolean-Device",
        "created": 1599066555273,
        "origin": 1599066555271553579,
        "readings": [
            {
                "id": "0230e6ad-35ba-4af1-a724-f80e62a781b1",
                "created": 1599066555273,
                "origin": 1599066555271455776,
                "device": "Random-Boolean-Device",
                "name": "Bool",
                "value": "true",
                "valueType": "Bool"
            }
        ],
        "tags": {
            "OOO": "YYY",
            "gateway-id": "HoustonStore-000123"
        }
```

I will use this as the example. :-)

lenny-intel: Corrected.

cloudxxx8 (approved):
LGTM

Issue #2738, opened 2020-10-01
[V2, core-data] Query not found should not log error message


cloudxxx8:
# 🐞 Bug Report

### Affected Services

 The issue is located in: core-data

### Is this a regression?

 Yes, the previous version in which this bug was not present was: ....

### Description and Minimal Reproduction
Sometime, we use query to confirm whether the object exists, so not found should not be treated as an error to log.

**EdgeX Version:**
master

PR #2739, opened 2020-10-01
Fix(data):query event doesn't print error message

cloudxxx8:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: Fix https://github.com/edgexfoundry/edgex-go/issues/2738

## What is the new behavior?
If the query error is errors.KindEntityDoesNotExist, the error log
should not be printed out because it is not a real error.
In the infrastructure layer, if the event doesn't exist when adding, the
error should not be returned.

- [ ] Yes
- [x] No

### DIFF
--- a/internal/core/data/v2/controller/http/event.go
+++ b/internal/core/data/v2/controller/http/event.go
@@ -11,6 +11,7 @@ import (
 	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/container"
 	"github.com/edgexfoundry/go-mod-bootstrap/di"
 
+	"github.com/edgexfoundry/go-mod-core-contracts/errors"
 	"github.com/edgexfoundry/go-mod-core-contracts/v2"
 	commonDTO "github.com/edgexfoundry/go-mod-core-contracts/v2/dtos/common"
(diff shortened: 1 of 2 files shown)


internal/core/data/v2/controller/http/event.go
```diff
@@ -97,7 +98,9 @@ func (ec *EventController) EventById(w http.ResponseWriter, r *http.Request) {
 	// Get the event
 	e, err := application.EventById(id, ec.dic)
 	if err != nil {
-		lc.Error(err.Error())
+		if errors.Kind(err) != errors.KindEntityDoesNotExist {
```
lenny-intel, line 101: Add comment why need to do this.

lenny-intel (changes requested):

internal/core/data/v2/controller/http/event.go
cloudxxx8: added comment

lenny-intel (approved):
LGTM

Issue #2803, opened 2020-10-16
[V2] GET /deviceservice/all returns inconsistent response when specifying labels or not


judehung:
# 🐞 Bug Report

### Affected Services

 The issue is located in: Metadata

### Is this a regression?

 
No, it's from V2 API which is still under development.

### Description
Based on the current implementation of GET /deviceservice/all, when client specify an out-of-range offset, query with labels will return out of range errors(http status=416), but query without labels will return empty array(http status=200).

We shall make query without labels to also return out of range errors(http status=416) under such case.  

Moreover, Swagger spec indicates that the maximum for limit query string is 50 records, which is too arbitrary and not flexible enough for clients to decide their own limit.  Ideally, the maximum shall refer to `MaxResultCount` as defined in the metadata configuration.  For the minimum of limit, we shall consider to have -1, meaning no limit and return all records.  Our Swagger spec shall be revised to reflect such changes.

**Anything else relevant?**
No

judehung:
Fix merged.  Close this issue.

PR #2805, opened 2020-10-16
fix(metadata): V2 GET /deviceservice/all returns inconsistent response when specifying labels or not

judehung:
## PR Checklist

**If your build fails** due to your commit message not passing the build checks, please review the guidelines here: https://github.com/edgexfoundry/edgex-examples/blob/master/.github/CONTRIBUTING.md.

## Issue Number:
#2803 

## What is the new behavior?
Based on the original implementation of GET /deviceservice/all, when client
specify an out-of-range offset, query with labels will return out of range
errors(http status=416), but query without labels will return empty
array(http status=200).

This commit make query without labels to also return out of range errors
(http status=416) under such case.

Moreover, Swagger spec indicates that the maximum for limit query string
is 50 records, which is too arbitrary and not flexible enough for clients
to decide their own limit. Ideally, the maximum shall refer to MaxResultCount
as defined in the metadata configuration. For the minimum of limit, we shall
consider to have -1, meaning no limit and return all records. Our Swagger
spec is also revised to reflect such changes.

- [ ] Yes
- [x] No

## New Imports
Are there any new imports or modules? If so, what are they used for and why?
No
## Specific Instructions

## Other information

### DIFF
--- a/internal/core/metadata/v2/application/deviceservice.go
+++ b/internal/core/metadata/v2/application/deviceservice.go
@@ -131,9 +131,10 @@ func GetDeviceServices(offset int, limit int, labels []string, ctx context.Conte
 	if err != nil {
 		return deviceServices, errors.NewCommonEdgeXWrapper(err)
 	}
-	for _, ds := range services {
+	deviceServices = make([]dtos.DeviceService, len(services))
+	for i, ds := range services {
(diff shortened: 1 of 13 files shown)


lenny-intel (approved):
LGTM

PR #2807, opened 2020-10-16
fix(snap): update redis to 6.0.8

hareshkainthcanonical:
Updated the source-tag to 6.0.8 to align with docker-compose redis version.

## PR Checklist

**If your build fails** due to your commit message not passing the build checks, please review the guidelines here: https://github.com/edgexfoundry/edgex-examples/blob/master/.github/CONTRIBUTING.md.

Using an older version of redis.

## What is the new behavior?
Updated the source-tag to 6.0.8 - newer stable release of redis.

- [ ] Yes
- [x] No

## New Imports
Are there any new imports or modules? If so, what are they used for and why?

## Specific Instructions
Build snap from source.

## Other information
Ran blackbox tests and all tests passed. Note, you will need to install the latest/edge version of app-service-configurable before running the test suite.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -572,7 +572,7 @@ parts:
 
   redis:
     source: https://github.com/antirez/redis.git
-    source-tag: "5.0.8"
+    source-tag: "6.0.8"
     source-depth: 1
     plugin: make
     make-install-var: PREFIX


tonyespy (changes requested):
Two issues need to be resolved...

  * First your PR is out of sync with master, so it can't be merged right now. Please wait for me to come online on Monday and I'll help you resolve this. That way I can approve and merge as soon as you've updated it too.
  * Your PR comment message is failing the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) check because you've specified an invalid scope (redis). Unfortunately the link in the PR template to the `CONTRIBUTING.md` file is broken (which I just reported on slack). For this commit since you're actually changing the snap packaging, your scope should be "snap" instead of "redis". Also updating a component in the snap really isn't really a feature, so "fix" is probably a better choice here for type (and it's what we've used in the past when bumping part versions). Also no need to mention what files you changed in the commit message, as someone can use `git show` or `git diff` to see those. So suggest the following commit summary:

_fix(snap): update redis to 6.0.8_

tonyespy (approved):
LGTM

Issue #2800, opened 2020-10-15
[V2 data] Implement Core Data PUT /event/pushed V2 API


judehung:
# 🚀 Feature Request

### Relevant Package

 This feature request is for coredata

### Description
 A clear and concise description of the problem or missing capability...
Per https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/default/put_event_pushed, implement the V2 API for 
PUT /event/pushed

judehung:
PR merged.  Close this feature request.

PR #2827, opened 2020-10-20
feat(data): Implement Core Data PUT /event/pushed V2 API

judehung:
## PR Checklist

## Issue Number:
#2800 

## What is the new behavior?
Per https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/default/put_event_pushed,
implement the V2 API for PUT /event/pushed

- [ ] Yes
- [ ] No

## New Imports

- [ ] Yes
- [ ] No

## Specific Instructions
https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/default/put_event_pushed

## Other information


internal/pkg/v2/infrastructure/redis/client.go
```diff
@@ -279,3 +280,29 @@ func (c *Client) AddDevice(d model.Device) (model.Device, errors.EdgeX) {
...
+func (c *Client) UpdateEvent(event model.Event) errors.EdgeX {
+	conn := c.Pool.Get()
+	defer conn.Close()
+
+	e, err := c.EventById(event.Id)
+	if err != nil {
+		return err
+	}
+
+	mergeErr := mergo.Merge(&event, e)
```
lenny-intel, line 294: Why do we need to merge when we loaded the Event by Id, changed a property passed the event to this func and now just need to replace the old event?

internal/pkg/v2/infrastructure/redis/event.go
```diff
@@ -84,6 +83,34 @@ func addEvent(conn redis.Conn, e models.Event) (addedEvent models.Event, edgeXer
...
+	storedKey := fmt.Sprintf("%s:%s", EventsCollection, id)
+	_ = conn.Send(MULTI)
+	_ = conn.Send(UNLINK, storedKey)
+	_ = conn.Send(UNLINK, EventsCollection+":readings:"+storedKey)
+	_ = conn.Send(ZREM, EventsCollection, storedKey)
+	_ = conn.Send(ZREM, EventsCollection+":created", storedKey)
+	_ = conn.Send(ZREM, EventsCollection+":pushed", storedKey)
+	_ = conn.Send(ZREM, EventsCollection+":device:"+e.DeviceName, storedKey)
+
+	res, err := redis.Values(conn.Do(EXEC))
```
lenny-intel, line 102: Does this also delete the events readings? Or should be calling `deleteReading()`?

lenny-intel (changes requested):

internal/pkg/v2/infrastructure/redis/event.go
judehung: Ooops, I shall invoke `deleteReadingById` as well.  Thanks for the catch!

cloudxxx8: since aside from pushed, event and reading are both immutable, we don't need to recreate readings.
it consumes lots of computing resource to update readings for just a pushed if there are more than 100 readings in an event.

internal/pkg/v2/infrastructure/redis/client.go
judehung: Thanks for pointing this out.  My original implementation is based on what V1 API did.  After further consideration, I've revised the codes to only update database records that are associated with `pushed` timestamp.  As Cloud also mentioned in his comment below, the only event update of V2 API is to update the pushed, so I think this new implementation would be better than removing and adding event.

internal/pkg/v2/infrastructure/redis/event.go
judehung: After further consideration, I've revised the codes to only update database records that are associated with `pushed` timestamp.

lenny-intel (dismissed):
LGTM

judehung:
rebased

cloudxxx8:
recheck

cloudxxx8 (approved):
LGTM

Issue #2839, opened 2020-10-22
[V2 data] Implement DELETE /event/scrub V2 API


judehung:
# 🚀 Feature Request

### Relevant Package

 This feature request is for coredata

### Description
 A clear and concise description of the problem or missing capability...
Per https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/default/delete_event_scrub, implement DELETE /event/scrub for V2 API

PR #2840, opened 2020-10-22
feat(data): Implement DELETE /event/scrub V2 API

judehung:
## PR Checklist

## Issue Number:
#2839 

## What is the new behavior?
Implement DELETE /event/scrub for V2 API per
https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/default/delete_event_scrub

As DELETE /event/scrub might deal with high volume of events/readings deletion at redis server, I leverage the idea from v1 API to simply accepted the deletion request and then pass the event/reading Ids to be removed through golang channel to two background goroutines to do the real deletion.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No


internal/core/data/v2/application/event.go
```diff
@@ -135,6 +135,20 @@ func EventCountByDevice(deviceName string, dic *di.Container) (uint32, errors.Ed
 	return count, nil
 }
 
+// The DeleteAllEvents function will be invoked by controller functions
+// and then invokes DeleteAllEvents function in the infrastructure layer to remove
+// all events
+func DeleteAllEvents(dic *di.Container) errors.EdgeX {
```
lenny-intel, line 141: Please rename all these functions to `DeletePushedEvents` be more explicate on what is being deleted.

internal/pkg/v2/infrastructure/redis/event.go
```diff
@@ -130,6 +130,16 @@ func deleteEvent(conn redis.Conn, id string) (edgeXerr errors.EdgeX) {
 	return edgeXerr
 }
 
+func scrubRecordsByKey(conn redis.Conn, key string) error {
```
lenny-intel, line 133: rename to `deleteRecordsByKey` as it is more explicate than `scrub` and follows the names used for the higher level functions that get us to this one.

internal/pkg/v2/infrastructure/redis/client.go
```diff
@@ -281,6 +281,22 @@ func (c *Client) AddDevice(d model.Device) (model.Device, errors.EdgeX) {
...
 }
 
+// Delete all readings and events
+func (c *Client) DeleteAllEvents() errors.EdgeX {
+	conn := c.Pool.Get()
+	defer conn.Close()
+
+	err := scrubRecordsByKey(conn, ReadingsCollection)
+	if err != nil {
+		return errors.NewCommonEdgeX(errors.KindDatabaseError, "scrub all readings failed", err)
```
lenny-intel, line 291: change message to `delete all pushed  readings failed`

internal/pkg/v2/infrastructure/redis/client.go
```diff
@@ -281,6 +281,22 @@ func (c *Client) AddDevice(d model.Device) (model.Device, errors.EdgeX) {
...
+	}
+	err = scrubRecordsByKey(conn, EventsCollection)
+	if err != nil {
+		return errors.NewCommonEdgeX(errors.KindDatabaseError, "scrub all events failed", err)
```
lenny-intel, line 295: change message to `delete all pushed  events failed`

lenny-intel (changes requested):

judehung:
Per swagger API spec, DELETE /event/scrub is to delete all pushed events and readings; however, my original implementation is to delete all events regardless it's pushed or not.  Converted to draft for further code adjustment.

internal/core/data/v2/application/event.go
judehung: Renamed.

internal/pkg/v2/infrastructure/redis/event.go
judehung: Renamed

internal/pkg/v2/infrastructure/redis/client.go
judehung: Refactored

internal/pkg/v2/infrastructure/redis/client.go
judehung: Refactored

judehung:
rebased

internal/core/data/v2/controller/http/event.go
```diff
@@ -215,6 +215,32 @@ func (ec *EventController) EventCountByDevice(w http.ResponseWriter, r *http.Req
...
+	// retrieve all the service injections from bootstrap
+	lc := container.LoggingClientFrom(ec.dic.Get)
+
+	ctx := r.Context()
+	correlationId := correlation.FromContext(ctx)
+
+	var response interface{}
+	var statusCode int
+
+	// Delete all events
```
lenny-intel, line 228: Remove comment. Old `delete all` and now function name is clear what it does.

internal/pkg/v2/infrastructure/redis/client.go
```diff
@@ -35,6 +35,8 @@ func NewClient(config db.Configuration, logger logger.LoggingClient) (*Client, e
 	if err != nil {
 		return nil, errors.NewCommonEdgeX(errors.KindDatabaseError, "redis client creation failed", err)
 	}
+	go dc.AsyncDeleteReadings()
+	go dc.AsyncDeleteEvents()
```
lenny-intel, line 39: This is a core data specific action, so it doesn't belong in the common NewClient code. i.e. What happens when Core Metadata calls this NewClient() func? 

Suggest these calls go in Code Data bootstrap handler once the client is created. These calls could go here:
https://github.com/edgexfoundry/edgex-go/blob/151af978cf25de30735135da97f91b1f15ae3e78/internal/core/data/init.go#L61
Since this handler runs after the V2 DBCleint has been created and is in the DIC.

internal/pkg/v2/infrastructure/redis/event.go
```diff
@@ -25,6 +26,9 @@ const (
 	EventsCollectionReadings   = EventsCollection + ":readings"
 )
 
+// deleteEventsByIdChannel channel used to delete events asynchronously
+var deleteEventsByIdChannel = make(chan []string, 50)
```
lenny-intel, line 30: This should be in Core Data application layer. App layer should request the ids from the DB layer and pass to the channel .

internal/pkg/v2/infrastructure/redis/event.go
```diff
@@ -131,6 +135,101 @@ func deleteEventById(conn redis.Conn, id string) (edgeXerr errors.EdgeX) {
 	return edgeXerr
 }
 
+// AsyncDeleteEvents Handles the deletion of device readings asynchronously. This function is expected to be running
+// in a go-routine and works with the "deleteReadingsByIds" function for better performance.
+func (c *Client) AsyncDeleteEvents() {
```
lenny-intel, line 140: Move to Core Data App layer and have it use the DB Layer to do the deletes.

internal/pkg/v2/infrastructure/redis/event.go
```diff
@@ -131,6 +135,101 @@ func deleteEventById(conn redis.Conn, id string) (edgeXerr errors.EdgeX) {
...
+				} else {
+					c.loggingClient.Debug(fmt.Sprintf("Deleted events successfully. elapsed time: %s", time.Since(startTime)))
+				}
+				conn.Close()
+			}
+		}
+	}
+}
+
+// deleteEventsByIds deletes all readings with given event Ids
```
lenny-intel, line 161: This seems to only delete the events, not readings.

internal/core/data/v2/application/event.go
```diff
@@ -158,6 +158,20 @@ func EventCountByDevice(deviceName string, dic *di.Container) (uint32, errors.Ed
 	return count, nil
 }
 
+// The DeletePushedEvents function will be invoked by controller functions
+// and then invokes DeletePushedEvents function in the infrastructure layer to remove
+// all events that have been pushed (pushed timestamp is greater than zero)
+func DeletePushedEvents(dic *di.Container) errors.EdgeX {
+	dbClient := v2DataContainer.DBClientFrom(dic.Get)
+
+	err := dbClient.DeletePushedEvents()
```
lenny-intel, line 167: This func should have the logic to get the ids of pushed events and send to async channel.

internal/pkg/v2/infrastructure/redis/client.go
```diff
@@ -294,6 +296,20 @@ func (c *Client) AddDevice(d model.Device) (model.Device, errors.EdgeX) {
...
+	conn := c.Pool.Get()
+	defer conn.Close()
+
+	pushedEventIds, readingIds, err := getPushedEventReadingIds(conn)
+	if err != nil {
+		return errors.NewCommonEdgeXWrapper(err)
+	}
+	deleteReadingsByIdChannel <- readingIds
+	deleteEventsByIdChannel <- pushedEventIds
+	return nil
```
lenny-intel, line 310: This logic should be in Core Data's app layer. DB layer expose GetPushedEventReadingIds

internal/pkg/v2/infrastructure/redis/event.go
```diff
@@ -131,6 +135,101 @@ func deleteEventById(conn redis.Conn, id string) (edgeXerr errors.EdgeX) {
...
+
+func deleteRecordsByKey(conn redis.Conn, key string) error {
+	_ = conn.Send(MULTI)
+	s := redisScripts[unlinkZSETMembers]
+	_ = s.Send(conn, key)
+	s = redisScripts[unlinkCollection]
+	_ = s.Send(conn, key)
+	_, err := conn.Do(EXEC)
+	return err
+}
```
lenny-intel, line 231: Is this still needed? I.e. do we have a `Purge` feature?

internal/pkg/v2/infrastructure/redis/reading.go
```diff
@@ -26,6 +26,9 @@ const (
 	ReadingsCollectionName       = ReadingsCollection + ":" + v2.Name
 )
 
+// deleteReadingsByIdChannel channel used to delete readings asynchronously
+var deleteReadingsByIdChannel = make(chan []string, 50)
```
lenny-intel, line 30: Move to Core Data app layer

internal/pkg/v2/infrastructure/redis/reading.go
```diff
@@ -97,6 +100,77 @@ func deleteReadingById(conn redis.Conn, id string) (edgeXerr errors.EdgeX) {
 	return nil
 }
 
+// AsyncDeleteReadings Handles the deletion of device readings asynchronously. This function is expected to be running
+// in a go-routine and works with the "deleteReadingsByIds" function for better performance.
+func (c *Client) AsyncDeleteReadings() {
```
lenny-intel, line 105: Move to Core Data app layer

internal/core/data/v2/infrastructure/interfaces/db.go
```diff
@@ -18,6 +18,7 @@ type DBClient interface {
 	DeleteEventById(id string) errors.EdgeX
 	EventTotalCount() (uint32, errors.EdgeX)
 	EventCountByDevice(deviceName string) (uint32, errors.EdgeX)
+	DeletePushedEvents() errors.EdgeX
```
lenny-intel, line 21: This changes to `GetPushedEventReadingIds` and `DeleteEventsByIds` and `DeleteReadingsByIds`

lenny-intel (changes requested):
Logic and async channels need to move from common Redis code to Core Data application layer.

internal/pkg/v2/infrastructure/redis/event.go
cloudxxx8: @lenny-intel I thought this should remain in the infrastructure layer, because it is Redis specific mechanism.  If we use a SQL database, we can delete them in one SQL command by specifying "Where".
Also, core-metadata and other services has its own DBClient interface, so they won't call this func.  

After Hanoi, I will find a way to refactor the infrastructure layer.  Currently, the DBClient struct here would contain all the function for all the services using database.  It would be better put the common logic here move the service specific logics to their own package.

internal/core/data/v2/controller/http/event.go
judehung: removed

internal/pkg/v2/infrastructure/redis/client.go
judehung: I moved the go routines start up to the very end of init.go.

internal/pkg/v2/infrastructure/redis/event.go
judehung: comments updated.

internal/pkg/v2/infrastructure/redis/event.go
judehung: We don't define a purge feature at this moment, so these two functions could be removed.

internal/pkg/v2/infrastructure/redis/event.go
judehung: Per comments from Cloud below, I sill leave the channel in the infrastructure layer.

internal/pkg/v2/infrastructure/redis/client.go
judehung: Moved to the app layer under DeletePushedEvents func.

internal/pkg/v2/infrastructure/redis/reading.go
judehung: Per comments from Cloud above, I still leave this channel in the infra layer.

internal/pkg/v2/infrastructure/redis/reading.go
judehung: Per comments from Cloud above, I still leave this channel in the infra layer.

internal/core/data/v2/infrastructure/interfaces/db.go
judehung: Updated.

internal/pkg/v2/infrastructure/redis/event.go
lenny-intel: Please remove since not in plan.

internal/core/data/init.go
```diff
@@ -154,5 +152,12 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, wg *sync.WaitGroup, st
 		},
 	})
 
+	if configuration.Databases["Primary"].Type == "redisdb" {
```
lenny-intel, line 155: @cloudxxx8 , Code checks like this is why these functions need to be part of the App layer. The will be needed for other DB implementations.

internal/core/data/v2/application/event.go
```diff
@@ -199,3 +222,49 @@ func EventsByDeviceName(offset int, limit int, name string, dic *di.Container) (
 	}
 	return events, nil
 }
+
+// AsyncDeleteEvents Handles the deletion of device events asynchronously. This function is expected to be running
+// in a go-routine and works with the "DeleteEventsByIds" function for better performance.
+func AsyncDeleteEvents(dic *di.Container, deleteEventsByIdChannel chan []string) {
```
lenny-intel, line 228: channel should be local to this package, initialized by this function before the for loop, not passed in.

Also the ctx.Done needs to be checked in the select to exit when SIGTERM and the Wait group needs to be incremented at the start and decremented when ctx.Done seen. The wg.add(1) need to be called where the `go` statement is used and the a deferred wg.Done() in this func. ctx & wg need to be parameters passed from init(). 
Similar to this:https://github.com/edgexfoundry/edgex-go/blob/70c918de78a758e1281f9e18824a6a08fc9647aa/internal/core/data/init.go#L113

Same for other async func

internal/core/data/v2/application/event.go
```diff
@@ -26,6 +28,12 @@ import (
 	"github.com/google/uuid"
 )
 
+type EventBackgroundScrubber struct {
+	dic                       *di.Container
+	deleteEventsByIdChannel   chan []string
+	deleteReadingsByIdChannel chan []string
```
lenny-intel, line 34: for now, just make these package private variables. I think the App layer should be refactored to be a object with a new method, but that needs to wait.

internal/pkg/v2/infrastructure/redis/event.go
```diff
@@ -25,6 +29,84 @@ const (
...
+				if err != nil {
+					lc.Error(fmt.Sprintf("Deleted events failed.  Err: %s", err.Error()))
+				} else {
+					lc.Debug(fmt.Sprintf("Deleted events successfully. elapsed time: %s", time.Since(startTime)))
+				}
+			}
+		}
+	}
+}
+
```
lenny-intel, line 57: Remove once we have agreement these need to be in App Layer.

internal/pkg/v2/infrastructure/redis/reading.go
```diff
@@ -26,8 +29,85 @@ const (
...
+				err := dbClient.DeleteReadingsByIds(readingIds)
+				if err != nil {
+					lc.Error(fmt.Sprintf("Deleted readings failed.  Err: %s", err.Error()))
+				} else {
+					lc.Debug(fmt.Sprintf("Deleted readings successfully. elapsed time: %s", time.Since(startTime)))
+				}
+			}
+		}
+	}
+}
```
lenny-intel, line 58: Delete once we have agreement this belongs in app layer.

lenny-intel (changes requested):

internal/pkg/v2/infrastructure/redis/event.go
lenny-intel: @cloudxxx8 , see my comments on init.go. This redis specific code points out why these need to be in the App Layer. Redis is currently the only DB for V2.  If we ever implemented a SQL DB we can revisit the best way to deal with the DBClient interface. 

Question, is the async channel the best why to do this with Redis, just because V1 did it that way?
It seems more complicated than needed. Can the Redis layer DeletePushedEvents just have call a one shot go func that does the work rather than having these long running go threads waiting on a channel to do the same async deleting?

internal/pkg/v2/infrastructure/redis/event.go
```diff
@@ -25,6 +29,84 @@ const (
...
+// DeleteEventsByIds deletes all events with given event Ids
+func (c *Client) DeleteEventsByIds(eventIds []string) (edgeXerr errors.EdgeX) {
```
lenny-intel, line 59: @cloudxxx8 , @judehung , After more consideration I think we can do away with the async channels. All they do is provide background processing of the deletes. This can be done in a much simpler way here if we have deletePushedEvents() (like others funcs in this layer) that uses a short running Go func (i.e. returns after complete). 

Here is some pseudo code
```
func deletePushedEvents(conn redis.Conn) edgeXerr errors.EdgeX  {
    get the Ids for events that have been pushed
   if err then 
       return error

   go func(eventIds string) {
       get each event by id, remove the Event's readings and then remove the Event 
       log any errors, but errors don't stop the deleting of other events or readings.
   }
}
```

internal/pkg/v2/infrastructure/redis/event.go
lenny-intel: @cloudxxx8 , @judehung , please focus on my last comment on eliminating the async functions and channels for a short running local Go func.

internal/core/data/v2/application/event.go
lenny-intel: Please ignore an focus on my lest recommendation.

cloudxxx8:
@lenny-intel @jpwhitemn we don't have time to re-write this by today (code freeze), so we would modify it next week for post Hanoi

lenny-intel:
> we don't have time to re-write this by today (code freeze), so we would modify it next week for post Hanoi

Yes. Also please move PR into draft as with any V2 PRs created during code freeze. THX!

judehung:
Converted this PR to draft per Lenny's comment.  Will revise the code post Hanoi.

judehung:
@lenny-intel I understand that we're under code freeze at this moment.  I've revised the implementation per your recommendation.  Please kindly take your time to review my code change.  Thanks!

lenny-intel (dismissed):
Nice job!

judehung:
rebased

judehung:
recheck

cloudxxx8 (changes requested):
please also modify the swagger file because the response code is changed

judehung:
> please also modify the swagger file because the response code is changed

Updated

cloudxxx8 (approved):
LGTM

cloudxxx8:
recheck

PR #2848, opened 2020-10-23
fix(snap): update to use kong deb from bintray

tonyespy:
## PR Checklist

NA

## Issue Number:
NA

## What is the new behavior?
NA

- [ ] Yes
- [X] No

## New Imports

- [ ] Yes
- [X] No

## Specific Instructions
Are there any specific instructions or things that should be known prior to reviewing? No

## Other information
This PR updates the edgexfoundry snap to use the kong debian packages published on bintray vs. building lua, luarocks, kong, and openresty as individual parts (some from source, some downloaded tarballs).

Two other minor changes are included in the PR:

  * filter out man pages included in the snap from the postgresql part
  * strip the date+commit suffix from the snap version

We've verified that kong comes up by default on both amd64 and arm64, and also run the blackbox-tests with security enabled, and found no regressions.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -65,6 +65,8 @@ passthrough:
 layout:
   /etc/luarocks:
     bind: $SNAP/etc/luarocks
+  /usr/local:
+    bind: $SNAP/usr/local
 
 grade: stable
 confinement: strict
@@ -324,7 +326,7 @@ apps:
     plugs: [home, removable-media]
   kong:
     adapter: full
-    command: bin/kong
+    command: usr/local/bin/kong
     command-chain:
       - bin/perl5lib-launch.sh
       - bin/kong-launch.sh
@@ -510,9 +512,7 @@ parts:
       tar -C $SNAPCRAFT_STAGE/go1.15 -xf "$FILE_NAME" --strip-components=1
     prime:
       - "-*"
-    # build after kong to prevent staging go binaries into the snap as they
-    # are large - see the TODO on the kong part
-    after: [go-build-helper, kong]
+    after: [go-build-helper]
 
   consul:
     after: [go115]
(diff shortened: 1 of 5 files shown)


snap/snapcraft.yaml
```diff
@@ -407,7 +409,7 @@ parts:
         PROJECT_VERSION=local-dev
       fi
       GIT_REVISION=$(git rev-parse --short HEAD)
-      snapcraftctl set-version ${PROJECT_VERSION}-$(date +%Y%m%d)+${GIT_REVISION}
+      snapcraftctl set-version ${PROJECT_VERSION}
```
anonymouse64, line 412: perhaps also call out this and the postgresql change in the PR description too? and thanks for making those things specific, separate commits

snap/snapcraft.yaml
```diff
@@ -508,8 +512,8 @@ parts:
       sha256sum -c sha256 | grep OK
       mkdir -p $SNAPCRAFT_STAGE/go1.15
       tar -C $SNAPCRAFT_STAGE/go1.15 -xf "$FILE_NAME" --strip-components=1
-    prime:
-      - "-*"
+    #prime:
+    #  - "-*"
```
anonymouse64, line 516: why this change? Won't this add all the go binaries back into the snap?

anonymouse64 (changes requested):
some code in the PR seems leftover from a debugging session?

lenny-intel (commented):
LGTM, but will let the Snap experts have final approval.
I did validate version of Kong is correct

lenny-intel:
At Tony, failing Semantic bot due to missing `:` after scope. Should be `fix(snap): update to use kong deb from bintray`

tonyespy:
@anonymouse64 @lenny-intel changes addressed, and re-pushed.

lenny-intel (dismissed):

anonymouse64:
Well I built and tested this PR on amd64 on my focal desktop and everything is groovy :wink: but I haven't been able to build it on arm64 yet to see if it's all good on the pi, I will have to do that on Monday, but I think that is a necessary check that should be done before merging since this is supposed to fix arm64.

anonymouse64 (changes requested):
see previous comment

tonyespy:
@anonymouse64 just a heads-up that @siggiskulason ran the blackbox tests with security enabled against this PR on his RPi4, and although there were a couple of assertion errors thought to be timing related, we're confident that this change works fine on arm64.

anonymouse64 (approved):
Tested on my arm64 rpi4 running uc18 and kong worked fine in my testing, so lgtm. :tada: 
thanks for figuring this out

hutchic:
I don't have a lot of knowledge about snaps but this seems sensible and who am I to complain about a ~200 line reduction 😆

tonyespy:
recheck

lenny-intel (approved):

PR #2871, opened 2020-11-05
fix(snap): remove external symlink to openresty

siggiskulason:
The Kong Debian package has a symbolic link to an external path, which the review tools in the snap store do not allow. The openresty symlink to /usr/local/openresty/nginx/sbin/nginx needs therefore to be changed to ../nginx/sbin/nginx

## PR Checklist

NA

## Issue Number:
NA

## What is the new behavior?
NA

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions
No

## Other information

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -675,6 +675,15 @@ parts:
 
       mkdir -p $SNAPCRAFT_PART_INSTALL/config/security-proxy-setup
       cp $SNAPCRAFT_PART_INSTALL/etc/kong/kong.conf.default $SNAPCRAFT_PART_INSTALL/config/security-proxy-setup/kong.conf
+
+      # by default the Kong deb contains an absolute symlink @ /usr/local/openresty/bin/openresty which points to /usr/local/openresty/nginx/sbin/nginx
+      # because the review-tools for the snap store do not currently allow using absolute symlinks that point outside of the snap
+      # (and are not smart enough to realize there is a layout involved in our usage here), we delete the absolute symlink
+      # and re-create it as a relative symlink pointing to ../nginx/sbin/nginx instead
+      cd $SNAPCRAFT_PART_INSTALL/usr/local/openresty/bin
+      rm openresty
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -675,6 +675,12 @@ parts:
 
       mkdir -p $SNAPCRAFT_PART_INSTALL/config/security-proxy-setup
       cp $SNAPCRAFT_PART_INSTALL/etc/kong/kong.conf.default $SNAPCRAFT_PART_INSTALL/config/security-proxy-setup/kong.conf
+
+      # this step resolves "package contains external symlinks: usr/local/openresty/bin/openresty lint-snap-v2_external_symlinks"
```
anonymouse64, line 679: super minor nitpick, but rather than explaining the esoteric error that the review-tools pumps out, could you explain what it is doing to resolve and what the problem is, i.e. something like

```suggestion
      # by default the deb contains an absolute symlink @ /usr/local/foo/bar which points to /foo/bar/something
      # because the review-tools for the snap store do not currently allow using absolute symlinks that point outside of the snap
      # (and are not smart enough to realize there is a layout involved in our usage here), we delete the absolute symlink
      # and re-create it as a relative one instead
```

obviously fix the paths I put there

anonymouse64 (commented):
lgtm, but one small suggestion about documentation

snap/snapcraft.yaml
siggiskulason: Thanks @anonymouse64 - good point. I have changed the comment and re-pushed my changes

anonymouse64 (approved):
thanks for the changes, LGTM

tonyespy (approved):
LGTM too...

tonyespy:
Note, we ran this past @lenny-intel and @jpwhitemn for a code freeze exception, and were given the thumbs-up.  As such, I'm going to go ahead and merge.

Issue #2495, opened 2020-04-14
Configuration of root CA file is inconsistent across services...


tonyespy:
While working on a recent update to the snap packaging (issue #2051), I noticed a few inconsistencies involving the configuration keys, and and sometimes the value of the root CA path across the various configuration files used by core-*, security-*, and support-* services.

One inconsistency is that some services use the top-level configuration key **SecretStore** (core-*, and support-* services), whereas some use **SecretService**.

The next inconsistency is the name of the root CA file path. Some use **CaFilePath** (security-secretstore-setup), some use **CACertPath** (security-proxy-setup), and some use **RootCaCertPath** (core-*, and support-* services).

The final issue is that the values used in the various configuration files differ as well, and thus most of them need to be overridden in the docker-compose files (and the snap). For instance, security-secretstore-setup uses `/tmp/edgex/secrets/ca/ca.pem`, whereas security-proxy-setup uses `res/EdgeXFoundryCA/EdgeXFoundryCA.pem` (the latter is overridden by a service-specific env var override in [docker-compose-nexus-mongo.yml](https://github.com/edgexfoundry/developer-scripts/blob/master/releases/nightly-build/compose-files/docker-compose-nexus-mongo.yml#L250)).  

Also note, in the snap, due to a misconfiguration of security-secrets-setup, the directory `$SNAP_DATA/pki/EdgeXFoundryCA`, isn't properly cleaned up, which means there are two identical root certificates, one named **ca.pem** and one named **EdgeXFoundryCA.pem** that get created (see issue #2547 for details).

At minimum, we should make sure all the default values are correct in the **configuration.toml** files, and if possible at least fix the security-proxy-setup to use **CaFilePath** instead of **CAFilePath** so that the security services are aligned.

This bug is being reported against the master branch.

bnevis-i:
> For instance, security-secretstore-setup uses /tmp/edgex/secrets/ca/ca.pem, whereas security-proxy-setup uses res/EdgeXFoundryCA/EdgeXFoundryCA.pem (the latter is overridden by a service-specific env var override in docker-compose-nexus-mongo.yml). I've always wondered why we have two identical root certificates, one named ca.pem and one named EdgeXFoundryCA.pem that get created?

I've always wondered why we are compiling the configurations files into the containers, instead of injecting the configuration at runtime.  IMO the configuration.toml files don't even belong in the codebase.

mkbhanda:
[...]

PR #2877, opened 2020-11-13
fix(security): fix proxy vault configuration

tonyespy:
This change fixes an issue with security-proxy-setup
which prevents the SecretService.Scheme from being
overridden via environment configuration override.
The existing code hard-coded Scheme to "https", which
this change overrides.

Fixes: 2873

## PR Checklist

If a configuration environment override is created for security-proxy-setup's SecretService.Scheme configuration value, it's currently ignored because security-proxy-setup hard-codes this value to "https".

## What is the new behavior?
security-proxy-setup honors a configuration override for `SecretService.Scheme`

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information
A configuration override of Scheme (set to "http") has been tested in the snap. No changes were made to the configuration.toml, and the override now actually works.

### DIFF
--- a/internal/security/proxy/config/config.go
+++ b/internal/security/proxy/config/config.go
@@ -76,6 +76,7 @@ type KongAclInfo struct {
 }
 
 type SecretServiceInfo struct {
+	Scheme          string
 	Server          string
 	Port            int
 	HealthcheckPath string
@@ -130,11 +131,18 @@ func (c *ConfigurationStruct) UpdateWritableFromRaw(rawWritable interface{}) boo
 func (c *ConfigurationStruct) GetBootstrap() bootstrapConfig.BootstrapConfiguration {
 	//To keep config file for proxy unchanged in Geneva we need to create a temporary SecretStore struct so that bootstrapHandler can use it to create a secret client
 	//The config file may be changed in the future version and SecretStore can be used directly like other core services
+
+	// Note - see issue #2873 for details on why this default value is set here
+	scheme := c.SecretService.Scheme
+	if scheme == "" {
+		scheme = "https"
+	}
+	
 	ss := bootstrapConfig.SecretStoreInfo{
 		Host:                    c.SecretService.Server,
 		Port:                    c.SecretService.Port,
 		Path:                    c.SecretService.CertPath,
-		Protocol:                "https",
+		Protocol:                scheme,
 		RootCaCertPath:          c.SecretService.CACertPath,
 		ServerName:              c.SecretService.Server,
 		Authentication:          vault.AuthenticationInfo{AuthType: "X-Vault-Token"},


internal/security/proxy/config/config.go
```diff
@@ -76,6 +76,7 @@ type KongAclInfo struct {
 }
 
 type SecretServiceInfo struct {
+	Scheme          string
```
lenny-intel, line 79: This should be Protocol to be consistent with every other service

internal/security/proxy/config/config.go
```diff
@@ -134,7 +135,7 @@ func (c *ConfigurationStruct) GetBootstrap() bootstrapConfig.BootstrapConfigurat
 		Host:                    c.SecretService.Server,
 		Port:                    c.SecretService.Port,
 		Path:                    c.SecretService.CertPath,
-		Protocol:                "https",
+		Protocol:                c.SecretService.Scheme,
```
lenny-intel, line 138: What if the config  value is blank? Default to `https` if blank?

lenny-intel (changes requested):
Still don't agree this is a bug , but agree it should be added for the upcoming Ireland work.

internal/security/proxy/config/config.go
tonyespy: `Protocol` is an attribute of `SecretStore`, not `SecretService` and I pointed out this and other inconsistencies in security configuration in issue #2495, which was punted until we could make breaking changes...

internal/security/proxy/config/config.go
tonyespy: Nice catch, yes that would preserve existing behavior.

internal/security/proxy/config/config.go
jim-wang-intel: Should this issue #2495 be non-issue any more if we go to `non-TLS` approach because there is no CA cert to deal with in the code in Ireland release?

tonyespy:
Updated (with a force push) to address @lenny-intel's comment about ensuring that there's a default value set the code still behaves the same way if `SecretService.Scheme` isn't specified in proxy-setup's` configuration.toml` file.

internal/security/proxy/config/config.go
lenny-intel: `Protocol` is the standard name we use. Using `Scheme` creates more inconsistencies.

bnevis-i, line 79:
```suggestion
	Protocol          string
```

internal/security/proxy/config/config.go
```diff
@@ -130,11 +131,18 @@ func (c *ConfigurationStruct) UpdateWritableFromRaw(rawWritable interface{}) boo
 func (c *ConfigurationStruct) GetBootstrap() bootstrapConfig.BootstrapConfiguration {
 	//To keep config file for proxy unchanged in Geneva we need to create a temporary SecretStore struct so that bootstrapHandler can use it to create a secret client
 	//The config file may be changed in the future version and SecretStore can be used directly like other core services
+
+	// Note - see issue #2873 for details on why this default value is set here
+	scheme := c.SecretService.Scheme
+	if scheme == "" {
+		scheme = "https"
+	}
```
bnevis-i, line 139:
```suggestion
	protocol := c.SecretService.Protocol
	if protocol == "" {
		protocol = "https"
	}
```

internal/security/proxy/config/config.go
```diff
@@ -130,11 +131,18 @@ func (c *ConfigurationStruct) UpdateWritableFromRaw(rawWritable interface{}) boo
...
+	
 	ss := bootstrapConfig.SecretStoreInfo{
 		Host:                    c.SecretService.Server,
 		Port:                    c.SecretService.Port,
 		Path:                    c.SecretService.CertPath,
-		Protocol:                "https",
+		Protocol:                scheme,
```
bnevis-i, line 145:
```suggestion
		Protocol:                protocol,
```

bnevis-i (changes requested):

bnevis-i (changes requested):
`internal/security/proxy/config/config.go` needs to be reformatted ala gofmt

internal/security/proxy/config/config.go
```diff
@@ -130,11 +131,18 @@ func (c *ConfigurationStruct) UpdateWritableFromRaw(rawWritable interface{}) boo
 func (c *ConfigurationStruct) GetBootstrap() bootstrapConfig.BootstrapConfiguration {
 	//To keep config file for proxy unchanged in Geneva we need to create a temporary SecretStore struct so that bootstrapHandler can use it to create a secret client
 	//The config file may be changed in the future version and SecretStore can be used directly like other core services
+
+	// Note - see issue #2873 for details on why this default value is set here
+	scheme := c.SecretService.Scheme
+	if scheme == "" {
+		scheme = "https"
```
lenny-intel, line 138: Now that this is going in after Hanoi, do we want the default to be `http`?

bnevis-i: I believe we discussed in the planning meeting that there is generally no broad support for defaults in code.  I think we should take it out, for the reason stated above.

jim-wang-intel:
Hi @tonyespy  this PR has been covered by the efforts of [`Remove Vault TLS`](https://github.com/edgexfoundry/edgex-go/pull/2890/files) PR and hence this PR can be closed I guess.

jim-wang-intel:
Close as this fix had been addressed in [another PR: Remove Vault TLS](https://github.com/edgexfoundry/edgex-go/pull/2890/files).

Issue #2866, opened 2020-11-04
Implement missing "proxy tls" functionality in edgex-secrets-config


bnevis-i:
# 🚀 Feature Request

### Relevant Package
edgex-secrets-config

### Description
The related [ticket to implement edgex-secrets-config](https://github.com/edgexfoundry/edgex-go/pull/2808) left out a major feature to support uploading of a user-supplied TLS certificate in order to make the EdgeX Hanoi code freeze deadline.

This request is to add the missing functionality as described in the edgex-secrets-config man page under "proxy tls".

This request realizes the objective of https://github.com/edgexfoundry/edgex-go/issues/1926 and https://github.com/edgexfoundry/edgex-go/issues/1922

jim-wang-intel:
currently this one depends on PR #2808

PR #2930, opened 2020-12-07
feat(security): implement secrets-config proxy tls

jim-wang-intel:
Add implementation for secrets-config: proxy tls subcommand
It uploads the user-provided Kong TLS certificate/key pair to proxy server.

Closes: #2866

## PR Checklist

- [X ] Tests for the changes have been added (for bug fixes / features)

## What is the new behavior?
Added new implementation of proxy tls into secrets-config to upload the kong tls cert

- [ ] Yes
- [X ] No

## New Imports

- [ ] Yes
- [X ] No

## Specific Instructions

## Other information


internal/security/config/command/proxy/tls/command.go
```diff
@@ -61,9 +73,172 @@ func NewCommand(
 }
 
 func (c *cmd) Execute() (statusCode int, err error) {
-	fmt.Println("TODO: Configure inbound TLS certificate.")
+	fmt.Println("Configure inbound proxy TLS certificate.")
 	fmt.Printf("--incert %s\n", c.certificatePath)
 	fmt.Printf("--inkey %s\n", c.privateKeyPath)
```
lenny-intel, line 78: Use logging client so all messages are consistent.

lenny-intel (changes requested):

internal/security/config/command/proxy/tls/command.go
jim-wang-intel: Good catch, I missed these.

internal/security/config/command/proxy/tls/command.go
```diff
@@ -61,9 +73,172 @@ func NewCommand(
...
+
+	// post the certKeyPair as the new one
+	if err := c.postKongTLSCertificate(certKeyPair); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *cmd) listKongTLSCertificates() ([]map[string]interface{}, error) {
```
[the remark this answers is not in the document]
jim-wang-intel: yea, we can. it is just that the element  `map[string]interface{}` inside the array is very generic though. Everything with array of json structure returns this kind of structure  and thus it is not as bad as we thought about the leaking data structure.

internal/security/config/command/proxy/tls/command.go
bnevis-i: Delete these three lines.  They were placeholders.

jim-wang-intel: ok, will delete them

bnevis-i:
Ok.  The code is uploading the cert but not making it active.

```
 openssl s_client -connect localhost:8443
```

should output the TLS cert that we uploaded.  It is currently returning the default cert.

jim-wang-intel:
> Ok. The code is uploading the cert but not making it active.
> 
> ```
>  openssl s_client -connect localhost:8443
> ```
> 
> should output the TLS cert that we uploaded. It is currently returning the default cert.
> 
> Looks like the SNIs parameter is required and it won't take '*'

Added SNIS association, now the command line: `openssl s_client -servername localhost -connect localhost:8443` responds with the user-provided TLS cert.

jim-wang-intel:
To test this locally on docker-based Kong, do the following steps:
 1. git clone this PR
 1. build codes by `make build`
 1. copy the temporary res/ directory from `security-proxy-setup`'s: 
    ```
        cp -r cmd/security-proxy-setup/res cmd/secrets-config/
    ```
 1. open another terminal and go to developer-scripts master branch and change directory to `compose-builder`
 1. run the edgex docker stack by `make run`
 1. wait for few seconds to let the whole stack up and running
 1. go back to the previous terminal that has `edgex-go`, and change directory to secrets-config:
    ```
      cd cmd/secrets-config/
    ```
 1. run the executable `proxy tls ` subcommand by:
    ```
      ./secrets-config proxy tls --incert <Path_to_Your_Cert_In_Encoded_PEM_Format>yourcert.pem --inkey <Path_to_Your_Cert_Private_Key_In_PEM_Format>yourprivatekey
    ```
 1. You should see no error in the console prompt after  you run that subcommand.
 1. You can also verify the certificate being used against the running edgex docker-stack via curl command or openssl provided by Bryon's:
   ```
      openssl s_client -servername localhost -connect localhost:8443
   ```

bnevis-i:
Please update README.md with new command line option.

internal/security/config/command/proxy/tls/command.go
```diff
@@ -7,26 +7,40 @@
...
-	loggingClient   logger.LoggingClient
-	configuration   *config.ConfigurationStruct
-	certificatePath string
-	privateKeyPath  string
+	loggingClient           logger.LoggingClient
+	client                  internal.HttpCaller
+	configuration           *config.ConfigurationStruct
+	certificatePath         string
+	privateKeyPath          string
+	serveNameIndictionsList string
```
bnevis-i, line 43:
```suggestion
	serverNameIndicatorList string
```

internal/security/config/command/proxy/tls/command.go
```diff
@@ -61,9 +78,208 @@ func NewCommand(
...
+
+	if len(snisList) > 0 {
+		splitSnis := parseAndTrimSpaces(snisList)
+		for _, sni := range splitSnis {
+			if _, exists := uniqueSnisMap[sni]; !exists {
+				uniqueSnisMap[sni] = true
+				snis = append(snis, sni)
+			}
+		}
+	}
```
bnevis-i, line 272: This algorithm is confusing.  There is no need for the existence test.

Suggest:
create the map
iterate builtin sni list and set key of map
iterate user-supplied sni list and set key of map (potential overwrite, which is harmless)
iterate keys of resulting map and build the array

internal/security/config/command/proxy/tls/command.go
```diff
@@ -61,9 +78,208 @@ func NewCommand(
...
+	if err := c.postKongTLSCertificate(certKeyPair); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *cmd) listKongTLSCertificates() ([]string, error) {
+	// list certificates first to see if any already exists
+	certKongURL := strings.Join([]string{c.configuration.KongURL.GetProxyBaseURL(), "certificates"}, "/")
```
bnevis-i, line 139: I have decided that this algorithm is unsafe.  As Kong uses a central "certificates" table for all certificates no matter what their purpose is, I think it is better to enumerate the /snis  API, get the unique list of certificate IDs, and delete those instead.  A blanket deletion of certificates might get certificates that aren't being used to serve TLS.

bnevis-i (changes requested):

internal/security/config/command/proxy/tls/command.go
bnevis-i: Also if the builtin sni list was an array to start with, it could also save some extra code.

bnevis-i: Nevermind the above.  Arrays aren't compatible with the configuration provider.  Perhaps default SNIs should come from configuration.toml as a string vs an array.

jim-wang-intel: currently there is no toml file for this utility. so you suggest creating one? it sounds to me it's kind of overdone just for that single entry

internal/security/config/command/proxy/tls/command.go
jim-wang-intel: ok

internal/security/config/command/proxy/tls/command.go
bnevis-i: Just thinking out loud.  You are probably right.

lenny-intel (changes requested):

internal/security/config/command/proxy/tls/command.go
jim-wang-intel: changed.

internal/security/config/command/proxy/tls/command.go
jim-wang-intel: changed and now it uses `/snis` to figure out the unique certificate ids to be deleted.

bnevis-i (dismissed):
LGTM

internal/security/config/command/proxy/tls/command.go
```diff
@@ -61,9 +83,235 @@ func NewCommand(
...
+	jsonData, err = json.Marshal(parsedResponse["data"])
+	if err != nil {
+		return nil, fmt.Errorf("Failed to json marshal parsed response data: %w", err)
+	}
+
+	// the list snis get API returns an array of snis-certs association
+	var parsedSnisCertData []map[string]interface{}
+	if err := json.NewDecoder(bytes.NewReader(jsonData)).Decode(&parsedSnisCertData); err != nil {
+		return nil, fmt.Errorf("Unable to parse response for parsed SNIS cert data: %w", err)
+	}
```
lenny-intel, line 182: Should only need one Decode, and not marshal and decode again if struct is defined appropriately.

What does the raw JSON response look like?

lenny-intel (changes requested):

internal/security/config/command/proxy/tls/command.go
jim-wang-intel: not a fan to this rule. see the structure in test for the raw JSON response.

jim-wang-intel: Hi @lenny-intel: 

example of the first time response: 
```json
 {"next":null,"data":[]}
```

after post some cert successfully, example of the second time response: 
```json
{"next":null,"data":[{"created_at":1607534629,"id":"646c762c-f86f-4222-a7bb-acb27d108cd2","tags":null,"name":"localhost","certificate":{"id":"6e8dbc72-d175-4e47-8064-9a45634a822b"}},{"created_at":1607534629,"id":"70f4aa8c-94c7-4985-881d-ab915e30c663","tags":null,"name":"kong","certificate":{"id":"6e8dbc72-d175-4e47-8064-9a45634a822b"}}]}
```

there is "next" on the first level and thus `map[string]interface{}`
but there is "certificate" object inside `data` after the first level, so the `certificate`'s id is nested inside the `data`:`certificate` and thus requires the second time of decoding.

We can change the mechanism to do marshal/unmarshal the predefined structures but there is two drawbacks to do this:
 1. The json user-defined structure is "Globally" exposed (for json to marshal/unmarshal to work) which is undesired in this utility.
 1. Inconsistent with other implementation of subcommands: the other existing subcommnads like adduser/deluser, what have you, currently don't implement this way; so using coding/encoding is to make consistent with the implementation of other subcommands.  For the sake of consistence, then will have to change all codes, and this is beyond the scope of this PR.

internal/security/config/command/proxy/tls/command.go
```diff
@@ -61,9 +78,208 @@ func NewCommand(
...
+	// the list certificate get API returns the array of certificates
+	var parsedCertData []map[string]interface{}
+	if err := json.NewDecoder(bytes.NewReader([]byte(outputData))).Decode(&parsedCertData); err != nil {
+		return nil, fmt.Errorf("Unable to parse response for parsed cert data: %w", err)
+	}
+
+	certIDs := make([]string, len(parsedCertData))
+	for i, certMap := range parsedCertData {
+		certIDs[i] = fmt.Sprintf("%s", certMap["id"])
+	}
```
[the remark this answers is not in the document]
jim-wang-intel: changed.

lenny-intel (approved):
LGTM

PR #2942, opened 2020-12-11
refactor: Use new enhanced SecretProvider

lenny-intel:
## PR Checklist

Currently use SecretProvider from go-mod-bootstrap that is only for edgex-go services

## What is the new behavior?
Use new reworked/enhanced SecretProvider from go-mod-bootstrap that is for all services
Includes new InsecureSecrets capability to pull from configuration when Vault is not being used.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information


jim-wang-intel (dismissed):
LGTM

cmd/core-command/Attribution.txt
```diff
@@ -154,6 +154,9 @@ https://github.com/golang/net/blob/master/LICENSE
 gopkg.in/yaml.v2 (Apache 2.0) https://github.com/go-yaml/yaml/
 https://github.com/go-yaml/yaml/blob/v2.2.2/LICENSE
 
+gopkg.in/yaml.v3 (MIT) https://github.com/go-yaml/yaml/
```
cloudxxx8, line 157: I don't see we are using yaml.v3 in go.mod.  Why do you add this section?

cmd/core-command/res/configuration.toml
```diff
@@ -1,5 +1,11 @@
 [Writable]
 LogLevel = 'INFO'
+  [Writable.InsecureSecrets]
+    [Writable.InsecureSecrets.DB]
+    path = "redisdb"
```
cloudxxx8, line 5: other configuration keys are using PascalCase.  Should we also use this format for the secrets config?

internal/core/data/init.go
```diff
@@ -62,14 +64,15 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, wg *sync.WaitGroup, st
...
+		secretProvider := container.SecretProviderFrom(dic.Get)
+		credentials, err := secretProvider.GetSecrets(configuration.Databases["Primary"].Type)
 		if err != nil {
 			lc.Error(fmt.Sprintf("Error getting DB creds for RedisStreams: %s", err.Error()))
 			return false
 		}
 
 		lc.Info("DB Credentials set for using Redis Streams")
-		configuration.MessageQueue.Optional["Password"] = credentials.Password
+		configuration.MessageQueue.Optional["Password"] = credentials[secret.PasswordKey]
```
cloudxxx8, line 75: if we change PasswordKey to PascalCase, it could also be applied to `configuration.MessageQueue.Optional[secret.PasswordKey]` here

internal/security/proxy/main.go
```diff
@@ -23,10 +23,10 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal"
 	"github.com/edgexfoundry/edgex-go/internal/security/proxy/config"
 	"github.com/edgexfoundry/edgex-go/internal/security/proxy/container"
+	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/handlers"
```
cloudxxx8, line 26: this import should be moved to the go-mod-boostrap block

internal/support/notifications/main.go
```diff
@@ -30,13 +30,10 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/pkg/telemetry"
 	notificationsConfig "github.com/edgexfoundry/edgex-go/internal/support/notifications/config"
 	"github.com/edgexfoundry/edgex-go/internal/support/notifications/container"
+	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/handlers"
```
cloudxxx8, line 33: this import should be moved to the go-mod-bootstrap block

internal/support/scheduler/main.go
```diff
@@ -24,13 +24,10 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/pkg/telemetry"
 	"github.com/edgexfoundry/edgex-go/internal/support/scheduler/config"
 	"github.com/edgexfoundry/edgex-go/internal/support/scheduler/container"
+	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/handlers"
```
cloudxxx8, line 27: this import should be moved to the go-mod-bootstrap block

internal/core/data/main.go
```diff
@@ -25,14 +25,11 @@ import (
 	v2DataContainer "github.com/edgexfoundry/edgex-go/internal/core/data/v2/bootstrap/container"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/handlers/database"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/telemetry"
-	"github.com/edgexfoundry/edgex-go/internal/pkg/v2/bootstrap/handlers"
+	v2Handlers "github.com/edgexfoundry/edgex-go/internal/pkg/v2/bootstrap/handlers"
+	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/handlers"
```
cloudxxx8, line 29: these two imports should be moved to the go-mod-bootstrap block

internal/core/metadata/main.go
```diff
@@ -25,14 +25,11 @@ import (
 	v2MetadataContainer "github.com/edgexfoundry/edgex-go/internal/core/metadata/v2/bootstrap/container"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/bootstrap/handlers/database"
 	"github.com/edgexfoundry/edgex-go/internal/pkg/telemetry"
-	"github.com/edgexfoundry/edgex-go/internal/pkg/v2/bootstrap/handlers"
+	v2Handlers "github.com/edgexfoundry/edgex-go/internal/pkg/v2/bootstrap/handlers"
+	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/handlers"
```
cloudxxx8, line 29: these two imports should be moved to the go-mod-bootstrap block

cmd/core-command/Attribution.txt
lenny-intel: It is brought in by testify

cmd/core-command/res/configuration.toml
lenny-intel: No, this path has to match to value used for the DB type, which is all lowercase

cmd/core-command/Attribution.txt
cloudxxx8: do we have to upgrade testify?  or, should we use yaml.v3 and remove yaml.v2 in core-metadata later?

internal/core/data/init.go
lenny-intel: I like where you are going with this, but are dependent of existing code that is creating them with lowercase. Changing that code is out of scope for this PR.
https://github.com/edgexfoundry/edgex-go/blob/master/internal/security/secretstore/password.go#L66

cloudxxx8 (approved):
LGTM

jim-wang-intel (approved):
LGTM other than issues brought by Cloud @

cmd/core-command/Attribution.txt
lenny-intel: testify 1.6.1 is not required, just keeping updated with latest. Note yaml.v3 use in testify doesn't impact our use of yaml.v2 in core-metadata. We should consider upgrading to yaml.v3 in core-metadata as a separate issue.

cmd/core-command/Attribution.txt
```diff
@@ -154,6 +154,9 @@ https://github.com/golang/net/blob/master/LICENSE
 gopkg.in/yaml.v2 (Apache 2.0) https://github.com/go-yaml/yaml/
 https://github.com/go-yaml/yaml/blob/v2.2.2/LICENSE
```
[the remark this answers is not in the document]
lenny-intel: V2 is used by our code. V3 is indirect dependency used by testify. We really should not need to list V3 here since it is only used in the test code, but the script doesn't know that.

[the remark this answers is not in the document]
jim-wang-intel: ok, agree; but it seems to bring up one issue that our testing system for attribution file should skip that based on that reasoning then. maybe worth some discussion in future WG meeting. @lenny-intel

PR #2940, opened 2020-12-11
feat: create optional kong certificate config options

beaufrusetta:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**If your build fails** due to your commit message not passing the build checks, please review the guidelines here: https://github.com/edgexfoundry/edgex-go/blob/master/.github/Contributing.md.

## What is the current behavior?
Values for CertPath, CertFilePath, and KeyFilePath are required in the configuration.toml.

## Issue Number:
#2928 

## What is the new behavior?
The kong config options for CertPath, CertFilePath, and KeyFilePath
in configuration.toml are now optional. By default, those values
are now empty. I also cleaned up a single quote situation in the
token provider configuration.toml.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions
Are there any specific instructions or things that should be known prior to reviewing?

## Other information

PR #2953, opened 2020-12-15
fix(snap): remove kong TLS config overrides

tonyespy:
Due to the recent decision to remove use of TLS on a
single device/node, it was also decided to let Kong
generate its own self-signed certificate and get rid
of the EdgeX CA altogether. A PR landed this weekend
which updated some of the default configuration values
in security-secretstore-setup's configuration.toml file
related to TLS configuration. As the snap still contained
overrides for some of these files, installation of the
snap fail when security-secretstore-setup fails to start.
This change removes the configuration overrides related
to Kong TLS configuration from snapcraft.yaml, which then
allows the snap to be installed.

## PR Checklist

The snap built from master fails to install due to configuration overrides in snapcraft.yaml specifying alternate paths for Kong TLS files.  This regression was introduced by the followingf PR:

https://github.com/edgexfoundry/edgex-go/pull/2940

## Issue Number:
None

## What is the new behavior?
NA

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -164,8 +164,6 @@ apps:
     environment:
       VAULT_INTERVAL: "--vaultInterval=10"
       SecretService_Server: localhost
-      SecretService_CertFilePath: $SNAP_DATA/secrets/edgex-kong/server.crt
-      SecretService_KeyFilePath: $SNAP_DATA/secrets/edgex-kong/server.key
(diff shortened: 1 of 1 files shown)


lenny-intel (approved):
LGTM

Issue #2983, opened 2021-01-04
Allow containers to run as non-root user - write Vault token files as non-root user


lenny-intel:
# 🚀 Feature Request

### Relevant Package

 This feature request is for secret-store-setup

### Description
For the edgex services to run in docker as a non-root user with security enabled, the Vault token files must be write using the same non-root user. 

### Describe the solution you'd like
 If you have a solution in mind, please describe it.
To do this the new `edgex` user must be defined with a specific (TBD) user id and group id and used to run all edgex services. This will result in secret-store-setup creating the token files with the same user that the other services use to read the token files.

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?

PR #3003, opened 2021-01-08
feat(security): secure containers run as non-root

beaufrusetta:
## PR Checklist

Docker containers run as root user.

## Issue Number:
#2983 

## What is the new behavior?
Modified the dockerfiles of various services to create/run as
"edgex_user:edgex_group".

Modified the entrypoint script of security-secretstore-setup to set
the appropriate ownership of the /tmp/edgex/secrets directory that's
mounted to retrieve secrets in various services used throughout.

Removed a security concern from the entrypoint script of
security-secretstore-setup that allowed root level CLI execution access
to anyone that could add parameters via CMD in the dockerfile.

- [ ] Yes
- [X] No

## New Imports

- [ ] Yes
- [X] No

## Specific Instructions

## Other information

### DIFF
--- a/cmd/core-command/Dockerfile
+++ b/cmd/core-command/Dockerfile
@@ -49,5 +49,6 @@ WORKDIR /
 COPY --from=builder /edgex-go/cmd/core-command/Attribution.txt /
 COPY --from=builder /edgex-go/cmd/core-command/core-command /
 COPY --from=builder /edgex-go/cmd/core-command/res/configuration.toml /res/configuration.toml
+
 ENTRYPOINT ["/core-command"]
 CMD ["-cp=consul.http://edgex-core-consul:8500", "--registry", "--confdir=/res"]
--- a/cmd/security-secretstore-setup/Dockerfile
+++ b/cmd/security-secretstore-setup/Dockerfile
@@ -50,9 +50,11 @@ COPY --from=builder /edgex-go/cmd/security-secretstore-setup/res/configuration.t
 COPY --from=builder /edgex-go/cmd/security-file-token-provider/security-file-token-provider .
 COPY --from=builder /edgex-go/cmd/security-secretstore-setup/security-secretstore-setup .
 
-# setup the entry point script
+# Setup the entry point script, create token dir, and assign perms
 COPY --from=builder /edgex-go/cmd/security-secretstore-setup/entrypoint.sh /usr/local/bin/
 RUN chmod +x /usr/local/bin/entrypoint.sh \
-    && ln -s /usr/local/bin/entrypoint.sh /
+    && ln -s /usr/local/bin/entrypoint.sh / \
+    && mkdir -p /vault/config/assets \
+    && chown -Rh 100:1000 /vault/    
 
 ENTRYPOINT ["entrypoint.sh"]
--- a/cmd/security-secretstore-setup/entrypoint.sh
+++ b/cmd/security-secretstore-setup/entrypoint.sh
@@ -24,27 +24,21 @@ if [ -n "${SECRETSTORE_SETUP_DONE_FLAG}" ] && [ -f "${SECRETSTORE_SETUP_DONE_FLA
   rm -f "${SECRETSTORE_SETUP_DONE_FLAG}"
 fi
 
-echo "creating /vault/config/assets"
+echo "Starting vault-worker..."
 
-# create token directory and
-# grant permissions of folders for vault:vault
-mkdir -p /vault/config/assets
-chown -Rh 100:1000 /vault/
-
-echo "starting vault-worker..."
-
-echo "Initializing secret store"
+echo "Initializing secret store..."
 /security-secretstore-setup --vaultInterval=10
-
-echo "Executing custom command: $@"
-"$@"
 
 # write a sentinel file when we're done because consul is not
 # secure and we don't trust it it access to the EdgeX secret store
 if [ -n "${SECRETSTORE_SETUP_DONE_FLAG}" ]; then
+
+    echo "Changing ownership of secrets to edgex_user:edgex_group"
+    chown -R ${EDGEX_USER}:${EDGEX_GROUP} /tmp/edgex/secrets
+
     echo "Signaling secretstore-setup completion"
-    mkdir -p $(dirname "${SECRETSTORE_SETUP_DONE_FLAG}")
-    touch "${SECRETSTORE_SETUP_DONE_FLAG}"
+    mkdir -p $(dirname "${SECRETSTORE_SETUP_DONE_FLAG}") && \
+      touch "${SECRETSTORE_SETUP_DONE_FLAG}"
 fi
 
 echo "Waiting for termination signal"
--- a/cmd/support-scheduler/Dockerfile
+++ b/cmd/support-scheduler/Dockerfile
@@ -47,5 +47,6 @@ EXPOSE $APP_PORT
 COPY --from=builder /edgex-go/cmd/support-scheduler/Attribution.txt /
 COPY --from=builder /edgex-go/cmd/support-scheduler/support-scheduler /
 COPY --from=builder /edgex-go/cmd/support-scheduler/res/configuration.toml /res/configuration.toml
+
 ENTRYPOINT ["/support-scheduler"]
 CMD ["-cp=consul.http://edgex-core-consul:8500", "--registry", "--confdir=/res"]
--- a/cmd/core-metadata/Dockerfile
+++ b/cmd/core-metadata/Dockerfile
@@ -49,5 +49,6 @@ WORKDIR /
 COPY --from=builder /edgex-go/cmd/core-metadata/Attribution.txt /
 COPY --from=builder /edgex-go/cmd/core-metadata/core-metadata /
 COPY --from=builder /edgex-go/cmd/core-metadata/res/configuration.toml /res/configuration.toml
+
 ENTRYPOINT ["/core-metadata"]
 CMD ["-cp=consul.http://edgex-core-consul:8500", "--registry", "--confdir=/res"]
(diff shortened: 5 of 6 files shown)


cmd/core-command/Dockerfile
```diff
@@ -36,6 +36,9 @@ COPY . .
 
 RUN make cmd/core-command/core-command
 
+# Create simulated /etc/passwd file with a single user
+RUN echo "edgex_user:x:2002:2001:Linux User,,,:/home/edgex_user:/sbin/nologin" >> /tmp/passwd
+
```
beaufrusetta, line 41: This will change when I pull in the Alpine changes to Scratch images from @jim-wang-intel - however, for future reference, this is how you force a Scratch based image to run as a specific user_id.

cmd/security-secretstore-setup/entrypoint.sh
```diff
@@ -24,27 +24,21 @@ if [ -n "${SECRETSTORE_SETUP_DONE_FLAG}" ] && [ -f "${SECRETSTORE_SETUP_DONE_FLA
...
-
-echo "starting vault-worker..."
-
-echo "Initializing secret store"
+echo "Initializing secret store..."
 /security-secretstore-setup --vaultInterval=10
 
-echo "Executing custom command: $@"
-"$@"
-
```
beaufrusetta, line 41: This is the security issue I called out in the commit message. This change came in on a commit that @tingyuz and @bnevis-i worked on in Jan 2020. Given that this container is a "run and done" container, we left it with root privileges to appropriately create/distribute secrets. Having root perms to execute whatever command you wanted on the command line is just asking for an injection at some point.

cmd/security-secretstore-setup/Dockerfile
```diff
@@ -36,7 +36,7 @@ RUN make cmd/security-file-token-provider/security-file-token-provider \
 
 FROM alpine:3.12
 
-RUN apk add --update --no-cache ca-certificates dumb-init curl
+RUN apk add --update --no-cache ca-certificates dumb-init curl su-exec
```
lenny-intel: Do we still need `su-exe`?

cmd/security-secretstore-setup/Dockerfile
```diff
@@ -55,4 +55,8 @@ COPY --from=builder /edgex-go/cmd/security-secretstore-setup/entrypoint.sh /usr/
 RUN chmod +x /usr/local/bin/entrypoint.sh \
     && ln -s /usr/local/bin/entrypoint.sh /
 
+# Create token directory and assign appropriate permissions
+RUN mkdir -p /vault/config/assets && \
+    chown -Rh 100:1000 /vault/
```
lenny-intel: Add to comment why `100:1000`

cmd/security-secretstore-setup/Dockerfile
```diff
@@ -55,4 +55,8 @@ COPY --from=builder /edgex-go/cmd/security-secretstore-setup/entrypoint.sh /usr/
...
+
```
jim-wang-intel, line 61: minor suggestion: reduce number of docker image layers by combining this with the above run.

cmd/security-secretstore-setup/entrypoint.sh
```diff
@@ -24,27 +24,21 @@ if [ -n "${SECRETSTORE_SETUP_DONE_FLAG}" ] && [ -f "${SECRETSTORE_SETUP_DONE_FLA
...
 if [ -n "${SECRETSTORE_SETUP_DONE_FLAG}" ]; then
+
+    echo "Changing ownership of secrets to edgex_user:edgex_group"
+    chown -R 2002:2001 /tmp/edgex/secrets
+
     echo "Signaling secretstore-setup completion"
-    mkdir -p $(dirname "${SECRETSTORE_SETUP_DONE_FLAG}")
-    touch "${SECRETSTORE_SETUP_DONE_FLAG}"
+    mkdir -p $(dirname "${SECRETSTORE_SETUP_DONE_FLAG}") && \
+      touch "${SECRETSTORE_SETUP_DONE_FLAG}"
```
jim-wang-intel, line 41: just FYI- this whole block will go away in the `security-bootstrapper`'s PR

jim-wang-intel (commented):
LGTM

cmd/security-secretstore-setup/Dockerfile
beaufrusetta: That actually does run in a single "RUN" command (and thus a single image layer) - the line is just split for readability.

cmd/security-secretstore-setup/entrypoint.sh
beaufrusetta: Ok sweet. I think we need to get that one in first, and then I can rebase this and remove it?

cmd/security-secretstore-setup/Dockerfile
jim-wang-intel: no i meant with the above RUN, not just this one

cmd/security-secretstore-setup/entrypoint.sh
jim-wang-intel: i am fine either way; i can rebase it, too

cmd/security-secretstore-setup/Dockerfile
beaufrusetta: Check it again - I think I get what you are saying. I combined the two RUN commands so it creates 1 image layer for those 4 commands instead of 2 image layers.

jim-wang-intel: cool

cmd/security-secretstore-setup/entrypoint.sh
```diff
@@ -24,27 +24,21 @@ if [ -n "${SECRETSTORE_SETUP_DONE_FLAG}" ] && [ -f "${SECRETSTORE_SETUP_DONE_FLA
...
 
-echo "Executing custom command: $@"
-"$@"
-
 # write a sentinel file when we're done because consul is not
 # secure and we don't trust it it access to the EdgeX secret store
 if [ -n "${SECRETSTORE_SETUP_DONE_FLAG}" ]; then
+
+    echo "Changing ownership of secrets to edgex_user:edgex_group"
+    chown -R 2002:2001 /tmp/edgex/secrets
```
lenny-intel, line 37: Please use environment variables EDGEX_USER and EDGEX_GROUP. In compose builder we'll set them in environment section of vault worker. 
```
environment:
   EDGEX_USER=${EDGEX_USER}
   EDGEX_GROUP=${EDGEX_GROUP}
```

cmd/core-command/Dockerfile
```diff
@@ -49,5 +52,10 @@ WORKDIR /
 COPY --from=builder /edgex-go/cmd/core-command/Attribution.txt /
 COPY --from=builder /edgex-go/cmd/core-command/core-command /
 COPY --from=builder /edgex-go/cmd/core-command/res/configuration.toml /res/configuration.toml
+
+# Copy over /tmp/passwd file with injected "user" entry as required by Docker's "USER <user>" command
+COPY --from=builder /tmp/passwd /etc/passwd
+USER edgex_user
```
lenny-intel, line 58: I would like to try doing this from compose file, rather than the Dockerfiles.
```
user: "${EDGEX_USER}:${EDGEX_GROUP}"
```
define EDGEX_USER and EDGEX_GROUP in .env file and use them through out compose and entry point scripts.
```
EDGEX_USER=2002
EDGEX_GROUP=2001
```

lenny-intel (changes requested):

cmd/security-secretstore-setup/Dockerfile
beaufrusetta: Already removed it and checked it in...this is "outdated"...

cmd/security-secretstore-setup/Dockerfile
beaufrusetta: I didn't create that - that was added before I touched it - I just moved it around...

cmd/core-command/Dockerfile
jim-wang-intel: IMO, for image itself, i think it makes sense to define it as ENV variable for docker run to use; doesn't have to be in compose-builder only but also any docker run command

cmd/security-secretstore-setup/Dockerfile
jim-wang-intel: it is 100 because dockerd volume mount is 100. it is 1000 because that is the first grounp in the container :-)

cmd/core-command/Dockerfile
lenny-intel: I think this will land before Jim's PR, so IMO you should make the change from `scratch` to `alpine`

beaufrusetta:
@lenny-intel must be some kind of a Docker Jedi - he sat back, watched me do all these updates, and then get it working - then he said, "wouldn't it be easier if we just put _**this**_ in the compose?". 

Backed out all my Dockerfile changes. Recent changes are checked in.

@jim-wang-intel @bnevis-i

lenny-intel (approved):
LIKE IT! LGTM

cmd/support-scheduler/Dockerfile
```diff
@@ -47,5 +47,6 @@ EXPOSE $APP_PORT
 COPY --from=builder /edgex-go/cmd/support-scheduler/Attribution.txt /
 COPY --from=builder /edgex-go/cmd/support-scheduler/support-scheduler /
 COPY --from=builder /edgex-go/cmd/support-scheduler/res/configuration.toml /res/configuration.toml
+
 ENTRYPOINT ["/support-scheduler"]
 CMD ["-cp=consul.http://edgex-core-consul:8500", "--registry", "--confdir=/res"]
```
bnevis-i, line 52: Need newline at end of this file.

bnevis-i (approved):

PR #3016, opened 2021-01-12
fix: Add defaults for User and Group to Secret Store Setup

lenny-intel:
## PR Checklist

Since the developer scripts update for non-root user has not yet merged yet the EDGEX_USER and EDGEX_GROUP are never set causing Secret Store Setup to fail before signaling done. This causes Proxy Setup to never run which results in no Kong proxies configured. 

## What is the new behavior?
This fix sets the EDGEX_USER and EDGEX_GROUP to a default value when they are not set.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information

### DIFF
--- a/cmd/security-secretstore-setup/entrypoint.sh
+++ b/cmd/security-secretstore-setup/entrypoint.sh
@@ -32,8 +32,13 @@ echo "Initializing secret store..."
 # write a sentinel file when we're done because consul is not
 # secure and we don't trust it it access to the EdgeX secret store
 if [ -n "${SECRETSTORE_SETUP_DONE_FLAG}" ]; then
+    # default User and Group in case never set
+    if [ -z "${EDGEX_USER}" ]; then
+      EDGEX_USER="2002"
+      EDGEX_GROUP="2001"
+    fi
 
-    echo "Changing ownership of secrets to edgex_user:edgex_group"
+    echo "Changing ownership of secrets to ${EDGEX_USER}:${EDGEX_GROUP}"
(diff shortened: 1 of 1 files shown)


jim-wang-intel (approved):
LGTM

cmd/security-secretstore-setup/entrypoint.sh
```diff
@@ -32,8 +32,13 @@ echo "Initializing secret store..."
...
 if [ -n "${SECRETSTORE_SETUP_DONE_FLAG}" ]; then
+    # default User and Group in case never set
+    if [ -z "${EDGEX_USER}" ]; then
+      EDGEX_USER="2002"
+      EDGEX_GROUP="2001"
+    fi
 
-    echo "Changing ownership of secrets to edgex_user:edgex_group"
+    echo "Changing ownership of secrets to ${EDGEX_USER}:${EDGEX_GROUP}"
     chown -R ${EDGEX_USER}:${EDGEX_GROUP} /tmp/edgex/secrets
```
bnevis-i, line 42: Please while you're at it change to `chown -Rh` here.  Without the `-h` chown will follow symlinks and can corrupt other parts of the OS (like making a symlink to /etc/shadow suddenly owned by edgex-user.

lenny-intel: @bnevis-i , didn't see this comment before merging. Please submit an issue so this doesn't get dropped.

PR #3051, opened 2021-01-19
fix(core-data): Set ApiVersion on Event/Reading DTOs prior to publishing to MessageBus

lenny-intel:
## PR Checklist

ApiVersion is not set in Event/Readings when published to MessageBus since it is not required to be set in the Add Event Request.
It needs to be set since the App Services use it to determine the version of object received and also needs to be included when the Event/Readings are exported.

## What is the new behavior?

ApiVersion is now set in Event/Readings when published to MessageBus

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information

### DIFF
--- a/internal/core/data/v2/application/event.go
+++ b/internal/core/data/v2/application/event.go
@@ -20,6 +20,7 @@ import (
 	"github.com/edgexfoundry/go-mod-core-contracts/v2/clients"
 	"github.com/edgexfoundry/go-mod-core-contracts/v2/errors"
 	"github.com/edgexfoundry/go-mod-core-contracts/v2/v2/dtos"
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/v2/dtos/common"
 	dto "github.com/edgexfoundry/go-mod-core-contracts/v2/v2/dtos/requests"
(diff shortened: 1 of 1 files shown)


lenny-intel:
recheck

cloudxxx8 (approved):
LGTM,
However, I thought we should set the ApiVersion field as required, and create the helper function like NewXXXRequest.  We have NewXXXResponse helper function now.

cloudxxx8:
My team can help handle the changes in contract if you agree with this change

lenny-intel:
> However, I thought we should set the ApiVersion field as required, and create the helper function like NewXXXRequest. We have NewXXXResponse helper function now.

@cloudxxx8 , we didn't make the ApiVersion required since all the endpoints are have the version. Having factory methods doesn't  solve setting it as the user of the API is not forced to you the factory method. We could set the ApiVersion when the request is un-marshaled. But this is only good for JSON, since CBOR is published w/o un-marshaling. We should discuss making **ApiVersion required** in Core WG meeting. @jpwhitemn , another agenda item.

cloudxxx8:
Got it.  I believe we can merge this PR now.

cloudxxx8 (approved):
LGTM

Issue #3100, opened 2021-01-29
Implement alternative to [dockerize] 3rd party lib for security-bootstrapper


jim-wang-intel:
# 🚀 Feature Request

### Relevant Package

 

This feature request is for `security-bootstrapper` to find an alternative (of the existing binary `dockerize`) to utilize a mechanism that can have waiting ability on URI and files.  

### Description
 

Though the current `dockerize` from https://github.com/jwilder/dockerize can fulfill the waiting ability on URIs/files, the CI builds of that encountered a few hiccups due to the across organization boundary of github.com.   Furthermore, the activity of that dockerize repo seems to be extremely low and thus may not be viable in the long run in terms of maintenance reliability perspective.

### Describe the solution you'd like
 
Bring the concept of `dockerize` into house and implement it to become one of subcommands for `security-bootstrapper` to use.

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?

PR #3101, opened 2021-01-29
feat(security): Add waitFor subcommand for security-bootstrapper

jim-wang-intel:
- New subcommand **waitFor** is added and it replaces the functionality of the former `dockerize`
- Add waitForInfo for the toml configuration (`Timeout` and `RetryInterval`)
- Building related to `dockerize` in Dockerfile is removed

Fixes: #3100

## PR Checklist

Currently we are using `dockerize` 3rd party lib to do the waiting for other services.

## What is the new behavior?
Replacing the `dockerize` with internal implemented `waitFor` as part of `security-bootstrapper` subcommand.

- [ ] Yes
- [x ] No

## New Imports

- [ ] Yes
- [x ] No

## Specific Instructions

## Other information
After this merged, need to update the `developer-script` compose files to include newly added env var. overrides used in `waitFor`.
Update PR created: https://github.com/edgexfoundry/developer-scripts/pull/399

### DIFF
--- a/internal/security/bootstrapper/command/waitfor/command.go
+++ b/internal/security/bootstrapper/command/waitfor/command.go
@@ -1,0 +1,251 @@
+/*******************************************************************************
+ * Copyright 2021 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ *******************************************************************************/
+
+package waitfor
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/edgexfoundry/edgex-go/internal/security/bootstrapper/config"
+	"github.com/edgexfoundry/edgex-go/internal/security/bootstrapper/interfaces"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/clients/logger"
+)
+
+const (
+	CommandName string = "waitFor"
+)
+
+type cmd struct {
+	waitGroup     *sync.WaitGroup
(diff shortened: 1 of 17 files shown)


internal/security/bootstrapper/command/waitfor/command.go
```diff
@@ -0,0 +1,251 @@
...
+			Timeout: c.timeout,
+		}
+
+		defer c.waitGroup.Done()
+		for {
+			req, err := http.NewRequest(http.MethodGet, httpURL.String(), nil)
+			if err != nil {
+				c.loggingClient.Infof("Problem with dial: %v. Sleeping %s", err.Error(), c.retryInterval)
+				time.Sleep(c.retryInterval)
+			}
```
lenny-intel, line 234: Should have a `continue` here?

internal/security/bootstrapper/command/waitfor/uri_flags.go
```diff
@@ -0,0 +1,29 @@
...
+
+import "fmt"
+
+type uriFlagsVar []string
+
+func (uris *uriFlagsVar) String() string {
+	return fmt.Sprint(*uris)
+}
+
+func (uris *uriFlagsVar) Set(value string) error {
```
lenny-intel, line 26: Should this be named `Add` rather than `Set`? To me `Set` would overwrite and this is adding.

lenny-intel (changes requested):
Nice job! Couple small things.

internal/security/bootstrapper/command/waitfor/command.go
jim-wang-intel: good catch

internal/security/bootstrapper/command/waitfor/uri_flags.go
jim-wang-intel: No, this Set is the method to set the flag value, part of the `flag.Value` interface from Golang Flag.

jim-wang-intel: See https://golang.org/pkg/flag/#Value

internal/security/bootstrapper/command/waitfor/command.go
jim-wang-intel: fixed

internal/security/bootstrapper/command/waitfor/uri_flags.go
jim-wang-intel: Added comments implementing Golang's flag.Value interface.

lenny-intel: K

lenny-intel (approved):
LGTM

PR #3102, opened 2021-01-30
fix(snap): fix redis snapshots

tonyespy:
## PR Checklist

The edgexfoundry snap doesn't properly generate redis persistence snapshots.

## What is the new behavior?
Snapshots are properly generated.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Other information
This fix has already been applied to master. Also as part of this PR, the snap is also updated to use the latest maintenance release of Redis (6.0.10).

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -84,9 +84,11 @@ apps:
   redis:
     adapter: full
     after: [security-secretstore-setup]
-    command: bin/redis-server --dir $SNAP_DATA/redis $REDIS_SAVE_OPTS
+    command: bin/redis-server $DIR_OPT $SAVE_OPT1 $SAVE_OPT2
     environment:
-      REDIS_SAVE_OPTS: "--save 900 1 --save 300 10"
+      DIR_OPT: "--dir $SNAP_DATA/redis"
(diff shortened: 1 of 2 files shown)


lenny-intel (dismissed):

snap/snapcraft.yaml
```diff
@@ -591,8 +593,8 @@ parts:
       - zip
 
   redis:
-    source: https://github.com/antirez/redis.git
-    source-tag: "6.0.8"
+    source: https://github.com/redis/redis.git
+    source-tag: "6.0.10"
```
lenny-intel, line 597: For Docker, the `hanoi` branch is using `6.0.9`. I assume we want these to be the same?
https://github.com/edgexfoundry/developer-scripts/blob/hanoi/compose-builder/.env#L33

tonyespy: Sure, I can make that change...

lenny-intel (approved):

snap/snapcraft.yaml
tonyespy: Just force pushed and adjusted the commit which bumped the snap to redis 6.0.10 to bump instead to 6.0.9.

Issue #3123, opened 2021-02-03
Remove Writable config section from Security services that don't use Config Provider


lenny-intel:
Many of the security services have the `Writable` section in their configuration structure that just holds the LogLevel. These services do not use the Configuration Provider and thus there isn't a Writable capability.

These service should be refactored to move the LogLevel to the top level in the struct and remove the Writable section.

Implementations of the configuration interface also need adjusting as follows:

- UpdateFromRaw should do nothing as it will never be called
- UpdateWritableFromRaw should do nothing as it will never be called
- EmptyWritablePtr should just return nil, also never called.

Issue #2863, opened 2020-11-04
Redis always started in insecure mode


tonyespy:
The Redis startup sequence has been updated for Hanoi to be compliant with our Secret Creation and Distribution ADR. Instead of the Redis credentials being read from Vault via security-secretstore-read and written to disk, a new service security-bootstrap-redis has been introduced which reads the newly generated credentials and calls into Redis (running in insecure mode) to set the credentials. The credentials are not persisted however which means the next time the EdgeX system is started, the same sequence happens all over again, so basically every time EdgeX starts, there's a small window where Redis is running in insecure mode.  

Version: v1.2.2-dev.104

jim-wang-intel:
Hi Tony @tonyespy : when running in security enable mode, the redis db password has been stored into vault  (See: https://github.com/edgexfoundry/edgex-go/blob/master/internal/security/secretstore/init.go#L335)  as part of `secretstore-setup` process. Inside the `secretstore-setup`, in that `addDBCredential` function, it has that checking logic  whether redis db password is already in vault or not (see https://github.com/edgexfoundry/edgex-go/blob/master/internal/security/secretstore/init.go#L411-L429).  So it will re-use the stored db password from vault when EdgeX restart next time.  Could you please elaborate what this issue more on what you were trying to resolve for?

PR #3147, opened 2021-02-08
feat(security): Remove Writable from security services

jim-wang-intel:
Update for security services on:

- Loglevel is on the top level of ConfigurationStruct
- Remove Writable from Configuration struct
- Make UpdateFromRaw, UpdateWritableFromRaw do nothing and make EmptyWritablePtr just return nil

Fixes: #3123

## PR Checklist

security services have Writable struct but actually it never uses

## What is the new behavior?
Refactor the `LogLevel` inside Writable struct and remove Writable.

- [ ] Yes
- [x ] No

## New Imports

- [ ] Yes
- [x ] No

## Specific Instructions
Depends on PR #2863 as the redis' toml conf would also be removed about the Writable section.

## Other information


cmd/security-proxy-setup/res/configuration.toml
```diff
@@ -16,8 +16,9 @@
 
 # This is a TOML config file for edgexsecurity service.
 
-[Writable]
 LogLevel = "DEBUG"
+
+[ProxySetup]
```
lenny-intel, line 21: Why was this added?

internal/security/proxy/config/config.go
```diff
@@ -33,8 +35,7 @@ type ConfigurationStruct struct {
 	Clients       map[string]bootstrapConfig.ClientInfo
 }
 
-type WritableInfo struct {
-	LogLevel       string
+type ProxySetupInfo struct {
 	RequestTimeout int
```
lenny-intel, line 39: Why does this need to be in a sub struct? Just have it add the top level also.

lenny-intel (changes requested):

cmd/security-proxy-setup/res/configuration.toml
jim-wang-intel: no particular strong reason, i just felt that it can be more structural in this way since potentially you can have many more `RequestTimout` for that naming alone.

lenny-intel: It is redundant since this is `Proxy Setup's` specific configuration. IMO, the extra level isn't needed. We added the extra levels in the bootstrapper config to make the Env override more readable, but that isn't needed here.

jim-wang-intel: fixed.

internal/security/proxy/config/config.go
jim-wang-intel: fixed.

lenny-intel (dismissed):
LGTM

lenny-intel (approved):

Issue #3161, opened 2021-02-09
Update Redis ACLs to use hashed passwords and decouple ACL settings into separate file


andresrinivasan:
# 🚀 Feature Request

### Relevant Package

Redis Security

### Description

After getting a better understanding from @bnevis-i about the security concern and as a follow on to #2863 and #3115, I recommend using ACLs with hashed passwords and putting the ACLs in a separate file from redis.conf.

### Describe the solution you'd like

The Redis 6 ACL implementation supports SHA256 hashed passwords. See https://redis.io/topics/acl#how-passwords-are-stored-internally. For example

```sh
user andre on +@all -@dangerous ~jobs:* #64-CHARACTER-SHA256
```

See https://redis.io/topics/acl#acl-rules to decipher the syntax.

Note from the docs

> However ACL passwords are not really passwords: they are shared secrets between the server and the client, because in that case the password is not an authentication token used by a human being.

Also
> * There are no length limits, the password will just be memorized in some client software, there is no human that need to recall a password in this context.
> * The ACL password does not protect any other thing: it will never be, for instance, the password for some email account.
> * Often when you are able to access the hashed password itself, by having full access to the Redis commands of a given server, or corrupting the system itself, you have already access to what such password is protecting: the Redis instance stability and the data it contains.

Finally, see https://redis.io/topics/acl#using-an-external-acl-file regarding using a separate ACL file.

### Describe alternatives you've considered

N/A

PR #3162, opened 2021-02-10
feat(security): Make separate redis ACL conf file and use hashed pwd in config

jim-wang-intel:
- Separate out ACL rule config into a different config file
- Use hashed password in the ACL config file
- Remove requirepass since it is using hashed pwd now

Fixes: #3161

## PR Checklist

- [x ] Docs have been added / updated (for bug fixes / features)

Currently clear text pwd from Vault can be observed in `redis.conf` file and acl rules are also in the same config file.

## What is the new behavior?
Hashed pwd in config file and ACL rule config in a separate file.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x ] No

## Specific Instructions

## Other information

### DIFF
--- a/internal/security/bootstrapper/helper/redis_conf.go
+++ b/internal/security/bootstrapper/helper/redis_conf.go
@@ -16,8 +16,10 @@
 package helper
 
 import (
+	"bufio"
+	"crypto/sha256"
 	"fmt"
-	"io"
+	"os"
 	"text/template"
 )
 
@@ -55,9 +57,6 @@ import (
 #  allcommands  Alias for +@all. Note that it implies the ability to execute
 #               all the future commands loaded via the modules system.
 #  allkeys      Alias for ~*
-#  ><password>  Add this password to the list of valid password for the user.
-#               For example >mypass will add "mypass" to the list.
-#               This directive clears the "nopass" flag (see later).
 #
 # ACL rules can be specified in any order: for instance you can start with
 # passwords, then flags, or key patterns. However note that the additive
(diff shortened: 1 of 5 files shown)


jim-wang-intel:
recheck

lenny-intel (commented):
LGTM, but will leave final approval to @andresrinivasan

andresrinivasan (commented):
This looks correct. One suggestion is to go with an explicit user other than _default_. If we think forward to each microservice having access to only their keys and therefore each having a unique user, maybe this first user with access to everything should be _admin_ or something self describing.

jim-wang-intel:
> This looks correct. One suggestion is to go with an explicit user other than _default_. If we think forward to each microservice having access to only their keys and therefore each having a unique user, maybe this first user with access to everything should be _admin_ or something self describing.

Hi @andresrinivasan , that is a good suggestion and I see where you are coming from, but that is another enhancement  and that is beyond the scope of this PR.  As with the current master, we are all using `default` as the user to authenticate with Redis database from all the database clients of other EdgeX core services (because we all omitted the user name in the database client).   Currently the credentials are read from Vault and the username we use from Vault is `redis5`, do you suggest to using that as the admin name?

andresrinivasan:
> > This looks correct. One suggestion is to go with an explicit user other than _default_. If we think forward to each microservice having access to only their keys and therefore each having a unique user, maybe this first user with access to everything should be _admin_ or something self describing.
> 
> Hi @andresrinivasan , that is a good suggestion and I see where you are coming from, but that is another enhancement and that is beyond the scope of this PR. As with the current master, we are all using `default` as the user to authenticate with Redis database from all the database clients of other EdgeX core services (because we all omitted the user name in the database client). Currently the credentials are read from Vault and the username we use from Vault is `redis5`, do you suggest to using that as the admin name?

Re: redis5. No, that was intended for a different purpose a couple of iterations ago. Yes, all that code will need to be reworked and is beyond this PR.

andresrinivasan (approved):
Approved @lenny-intel

lenny-intel (approved):

Issue #3155, opened 2021-02-09
[Secure Consul Ph. 1] Implement the token creation of specific Vault token for Consul's ACL bootstrapping


jim-wang-intel:
# 🚀 Feature Request

### Relevant Package

 
This feature request is for `secretstore-setup`

### Description
 
`Secretstore-setup` will create a Vault token for consul secrets engine configuration.  

### Describe the solution you'd like
 
1. Create a new `CreateToken` implementation inside the service `secretstore-setup` and assign the proper ACL attributes for `bootstrapping Consul ACL`  in Vault's config file. 
2. Store this token to a place using a volume mount (temp. file system, tmpfs) to distribute this Vault token to Consul's bootstrapper so that in Consul's bootstrapper can utilitze it to do Consul's ACL bootstrap step.
3.  Need to modify or update the docker's compose-file for this token-sharing volume.

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?

PR #3192, opened 2021-02-24
feat(security): Create a Vault mgmt token for Consul Secrets API Operations

jim-wang-intel:
Part of Phase 1 Securing Consul stories: Create a Vault token to configure consul access

Changes in `secretstore-setup`:
 - add token file writer implementation for creating and writing Vault's Consul secrets admin token
 - add Consul ACL feature flag `"ENABLE_REGISTRY_ACL"`
 - add configuration setting for `ConsulSecretAdminTokenPath` for specifying token file location
 - refactor TokenMaintenance to share code with token file writer's implementation

Closes: #3155

## PR Checklist

N/A

## What is the new behavior?
New Vault token created for special purpose of management Consul Secrets APIs operations policy: configure access / create update roles later on.  This token will be later used by Consul's bootstrapper.

- [ ] Yes
- [x ] No

## New Imports

- [ ] Yes
- [x ] No

## Specific Instructions

## Other information
To verify this locally, one can git clone this PR and build a dev version of `docker_secretstore_setup` and then add the following environment variables to docker-compose file of `secretstore-setup`:
```yaml
 environment:
      ENABLE_REGISTRY_ACL: "true"
```
so that the ACL feature is turned on.  Run `make run dev` and then observe the following docker logs from `edgex-secretstore-setup`:
`docker logs edgex-secretstore-setup | grep "edgex-consul/admin/token"`

```

jim@jim-NUC7i5DNHE:~/go/src/github.com/edgexfoundry/edgex-compose/compose-builder$ docker logs edgex-secretstore-setup | grep "edgex-consul/admin/token"
level=INFO ts=2021-02-24T22:47:57.466345924Z app=edgex-security-secretstore-setup source=tokenfilewriter.go:110 msg="token's written to /tmp/edgex/secrets/edgex-consul/admin/token.json"
jim@jim-NUC7i5DNHE:~/go/src/github.com/edgexfoundry/edgex-compose/compose-builder$ 

```
And also list the file, too:
`$ sudo ls -al /tmp/edgex/secrets/edgex-consul/admin/`
```
[sudo] password for jim: 
total 12
drwx------ 2 2002 2001 4096 Feb 24 12:56 .
drwx------ 3 2002 2001 4096 Feb 24 12:56 ..
-rw------- 1 2002 2001  536 Feb 24 15:47 token.json
```

### DIFF
--- a/cmd/security-secretstore-setup/res/configuration.toml
+++ b/cmd/security-secretstore-setup/res/configuration.toml
@@ -38,6 +38,7 @@ TokenProviderAdminTokenPath = "/run/edgex/secrets/tokenprovider/secrets-token.js
 PasswordProvider = ""
 PasswordProviderArgs = [ ]
 RevokeRootTokens = true
+ConsulSecretsAdminTokenPath = "/tmp/edgex/secrets/edgex-consul/admin/token.json"
 
 [Databases]
   [Databases.admin]
--- a/internal/security/secretstore/constants.go
+++ b/internal/security/secretstore/constants.go
@@ -19,10 +19,8 @@
 package secretstore
 
 const (
-	// KVSecretsEngineMountPoint is the name of the mount point base for Vault's key-value secrets engine
-	KVSecretsEngineMountPoint = "secret"
-	// ConsulSecretEngineMountPoint is the name of the mount point base for Vault's Consul secrets engine
-	ConsulSecretEngineMountPoint = "consul"
+	// this is a feature key to indicate whether to enable Registry Consul's ACL or not
+	RegistryACLFeatureFlag = "ENABLE_REGISTRY_ACL"
 
 	VaultToken             = "X-Vault-Token"
 	TokenCreatorPolicyName = "privileged-token-creator"
--- a/internal/security/secretstore/init.go
+++ b/internal/security/secretstore/init.go
@@ -35,6 +35,7 @@ import (
 	"github.com/edgexfoundry/edgex-go/internal/security/secretstore/config"
 	"github.com/edgexfoundry/edgex-go/internal/security/secretstore/container"
 	"github.com/edgexfoundry/edgex-go/internal/security/secretstore/secretsengine"
+	"github.com/edgexfoundry/edgex-go/internal/security/secretstore/tokenfilewriter"
 
 	bootstrapContainer "github.com/edgexfoundry/go-mod-bootstrap/v2/bootstrap/container"
 	"github.com/edgexfoundry/go-mod-bootstrap/v2/bootstrap/startup"
@@ -223,13 +224,11 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
 	healthOkCh := make(chan struct{})
 	go func() {
 		for {
-			select {
-			case <-ticker.C:
-				if sCode, _ := client.HealthCheck(); sCode == http.StatusOK {
-					close(healthOkCh)
-					ticker.Stop()
-					return
-				}
+			<-ticker.C
+			if sCode, _ := client.HealthCheck(); sCode == http.StatusOK {
+				close(healthOkCh)
+				ticker.Stop()
+				return
 			}
 		}
 	}()
@@ -290,9 +289,10 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
 
 	// If configured to do so, create a token issuing token
 	if secretStoreConfig.TokenProviderAdminTokenPath != "" {
-		revokeIssuingTokenFuc, err := makeTokenIssuingToken(lc, configuration, tokenMaintenance, fileOpener, rootToken)
+		revokeIssuingTokenFuc, err := tokenfilewriter.NewWriter(lc, client, fileOpener).
+			CreateAndWrite(rootToken, secretStoreConfig.TokenProviderAdminTokenPath, tokenMaintenance.CreateTokenIssuingToken)
 		if err != nil {
-			lc.Errorf("failed to create token issuing token %s", err.Error())
+			lc.Errorf("failed to create token issuing token: %s", err.Error())
 			os.Exit(1)
 		}
 		if secretStoreConfig.TokenProviderType == OneShotProvider {
@@ -318,14 +318,14 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
 	}
 
 	// Enable KV secret engine
-	if err := secretsengine.New(KVSecretsEngineMountPoint, secretsengine.KeyValue).
+	if err := secretsengine.New(secretsengine.KVSecretsEngineMountPoint, secretsengine.KeyValue).
 		Enable(&rootToken, lc, client); err != nil {
 		lc.Errorf("failed to enable KV secrets engine: %s", err.Error())
 		os.Exit(1)
 	}
 
 	// Enable Consul secret engine
-	if err := secretsengine.New(ConsulSecretEngineMountPoint, secretsengine.Consul).
+	if err := secretsengine.New(secretsengine.ConsulSecretEngineMountPoint, secretsengine.Consul).
 		Enable(&rootToken, lc, client); err != nil {
 		lc.Errorf("failed to enable Consul secrets engine: %s", err.Error())
 		os.Exit(1)
@@ -422,6 +422,19 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
 		lc.Info("proxy certificate pair upload was skipped because cert secretStore value(s) were blank")
 	}
 
+	// If Registry Consul ACL feature is enabled, then we need to create and save a Vault token to configure
+	// Consul secret engine access, role operations, and managing Consul agent tokens.
+	registryACLEnabled := os.Getenv(RegistryACLFeatureFlag)
+	if registryACLEnabled == "true" {
+		// generate a management token for Consul secrets engine operations:
+		tokenFileWriter := tokenfilewriter.NewWriter(lc, client, fileOpener)
+		if _, err := tokenFileWriter.CreateAndWrite(rootToken, configuration.SecretStore.ConsulSecretsAdminTokenPath,
+			tokenFileWriter.CreateMgmtTokenForConsulSecretsEngine); err != nil {
+			lc.Errorf("failed to create and write the token for Consul secret management: %s", err.Error())
+			os.Exit(1)
+		}
+	}
+
 	lc.Info("Vault init done successfully")
 	return false
 
@@ -467,74 +480,6 @@ func addDBCredential(lc logger.LoggingClient, db string, cred Cred, service stri
 	}
 
 	return err
-}
-
-func makeTokenIssuingToken(
-	lc logger.LoggingClient,
-	configuration *config.ConfigurationStruct,
-	tokenMaintenance *TokenMaintenance,
-	fileOpener fileioperformer.FileIoPerformer,
-	rootToken string) (RevokeFunc, error) {
-
-	configAdminTokenPath := configuration.SecretStore.TokenProviderAdminTokenPath
-	if configAdminTokenPath == "" {
-		err := fmt.Errorf("TokenProviderAdminTokenPath is a required configuration setting")
-		lc.Error(err.Error())
-		return nil, err
-	}
-
-	// Create delegate credential for use by the token provider
-	tokenIssuingToken, revokeIssuingTokenFuc, err := tokenMaintenance.CreateTokenIssuingToken(rootToken)
-	if err != nil {
-		lc.Errorf("failed to create token issuing token %s", err.Error())
-		return nil, err
-	}
-	lc.Info("created token issuing token")
-
-	// Write the token issuing token to disk to pass it to the token provider
-	adminTokenPath, err := filepath.Abs(configAdminTokenPath)
-	if err != nil {
-		lc.Errorf("failed to convert to absolute path %s: %s", configAdminTokenPath, err.Error())
-		revokeIssuingTokenFuc()
-		return nil, err
-	}
-	dirOfAdminToken := filepath.Dir(adminTokenPath)
-	err = fileOpener.MkdirAll(dirOfAdminToken, 0700)
-	if err != nil {
-		lc.Errorf("failed to create tokenpath base dir: %s", err.Error())
-		revokeIssuingTokenFuc()
-		return nil, err
-	}
-	tokenWriter, err := fileOpener.OpenFileWriter(adminTokenPath, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0600)
-	if err != nil {
-		lc.Errorf("failed to create token issuing file %s: %s", adminTokenPath, err.Error())
-		revokeIssuingTokenFuc()
-		return nil, err
-	}
-
-	encoder := json.NewEncoder(tokenWriter)
-	if encoder == nil {
-		err := fmt.Errorf("failed to create token encoder")
-		lc.Error(err.Error())
-		_ = tokenWriter.Close()
-		revokeIssuingTokenFuc()
-		return nil, err
-	}
-
-	if err = encoder.Encode(tokenIssuingToken); err != nil {
-		lc.Errorf("failed to write token issuing token: %s", err.Error())
-		_ = tokenWriter.Close()
-		revokeIssuingTokenFuc()
-		return nil, err
-	}
-
-	if err = tokenWriter.Close(); err != nil {
-		lc.Errorf("failed to close token issuing file: %s", err.Error())
-		revokeIssuingTokenFuc()
-		return nil, err
-	}
-
-	return revokeIssuingTokenFuc, nil
 }
 
 func loadInitResponse(
--- a/internal/security/secretstore/init.go
+++ b/internal/security/secretstore/init.go
@@ -324,13 +324,6 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
 		os.Exit(1)
 	}
 
-	// Enable Consul secret engine
-	if err := secretsengine.New(secretsengine.ConsulSecretEngineMountPoint, secretsengine.Consul).
-		Enable(&rootToken, lc, client); err != nil {
-		lc.Errorf("failed to enable Consul secrets engine: %s", err.Error())
-		os.Exit(1)
-	}
-
 	// credential creation
 	gen := NewPasswordGenerator(lc, secretStoreConfig.PasswordProvider, secretStoreConfig.PasswordProviderArgs)
 	cred := NewCred(httpCaller, rootToken, gen, secretStoreConfig.GetBaseURL(), lc)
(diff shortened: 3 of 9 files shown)


internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
```diff
@@ -0,0 +1,172 @@
...
+	if fileErr = json.NewEncoder(fileWriter).Encode(tokenCreated); fileErr != nil {
+		_ = fileWriter.Close()
+		return nil, fmt.Errorf("failed to write created token: %s", fileErr.Error())
+	}
+
+	if fileErr = fileWriter.Close(); fileErr != nil {
+		return nil, fmt.Errorf("failed to close token file: %s", fileErr.Error())
+	}
+
+	w.logClient.Infof("token's written to %s", tokenFilePath)
```
lenny-intel, line 110: A nit, but should `token's` be `token` since it is writing a single token?

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
```diff
@@ -0,0 +1,172 @@
...
+	return revokeTokenFunc, nil
+}
+
+// CreateMgmtTokenForConsulSecretsEngine creates a new Vault token that
+// allows the Consul bootstrapper to operate on managing Vault's Consul secrets engine related APIs (see reference:
+// https://www.vaultproject.io/api-docs/secret/consul). The created Vault token is meant for serving
+// the purpose of Consul ACL's bootstrapping as part of securing Consul process.
+//
+// Requires a root token to create, and returns data/information containing the token,
+// keeping the token without revoking it and hence always returning nil RevokeFunc;
```
lenny-intel, line 121: Why have `RevokeFunc`  when always returning nil for it? Is it due to function signature? If so add this `why` to this description.

internal/security/secretstore/tokenfilewriter/tokenfilewriter_test.go
```diff
@@ -0,0 +1,189 @@
...
+			}
+		})
+	}
+}
+
+func TestFileWriteErrorForConsulSecretEngine(t *testing.T) {
+	curUser, _ := user.Current()
+	if curUser != nil && curUser.Uid == "0" {
+		// it is root user then we skip this test as root can have permission to write everything
+		t.Skip()
```
lenny-intel, line 149: Should we fail rather than skip so developer is forced to not run the tests as root?

internal/security/secretstore/tokenfilewriter/tokenfilewriter_test.go
```diff
@@ -0,0 +1,189 @@
...
+	// as only the root user can perform this
+	testTokenFileDir := "/root/test"
+	testTokenFilePath := filepath.Join(testTokenFileDir, "mgmt-token.json")
+	fileWriter := NewWriter(lc, secretClient, flOpener)
+	_, err := fileWriter.CreateAndWrite(testRootToken, testTokenFilePath,
+		fileWriter.CreateMgmtTokenForConsulSecretsEngine)
+	require.Error(t, err)
+}
+
+func setup() func() (logger.LoggingClient, fileioperformer.FileIoPerformer) {
```
lenny-intel, line 176: This should be TestMain() which is the standard for this type of setup and doesn't requires a call in each test. `lc` and `flOpener`  become package level globals.

lenny-intel (changes requested):

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
jim-wang-intel: yea, i meant it to be `token is written` i'll change it to spell it out

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
jim-wang-intel: yes, it is due to function signature. I'll add the comments

internal/security/secretstore/tokenfilewriter/tokenfilewriter_test.go
jim-wang-intel: not sure, if CI is running as root then the test will be failing. Is CI running as root user?

internal/security/secretstore/tokenfilewriter/tokenfilewriter_test.go
jim-wang-intel: ok

internal/security/secretstore/tokenfilewriter/tokenfilewriter_test.go
lenny-intel: I doubt it, but @ernestojeda should be able to answer the question.

ernestojeda: Yes, when `make test` runs, it runs inside the context of a docker image and we run as root due to permissions issues with the LF build agents. See: https://github.com/edgexfoundry/edgex-global-pipelines/blob/master/vars/edgeXBuildGoApp.groovy#L181

jim-wang-intel: Thank you, @ernestojeda.  @lenny-intel i'll see this comment as resolved then.

cmd/security-secretstore-setup/res/configuration.toml
```diff
@@ -38,6 +38,7 @@ TokenProviderAdminTokenPath = "/run/edgex/secrets/tokenprovider/secrets-token.js
 PasswordProvider = ""
 PasswordProviderArgs = [ ]
 RevokeRootTokens = true
+ConsulSecretsAdminTokenPath = "/tmp/edgex/secrets/edgex-consul/admin/token.json"
```
bnevis-i, line 41: Don't code the whole path into the executable.  Use existing environment variables.

internal/security/secretstore/constants.go
```diff
@@ -19,10 +19,8 @@
 package secretstore
 
 const (
-	// KVSecretsEngineMountPoint is the name of the mount point base for Vault's key-value secrets engine
-	KVSecretsEngineMountPoint = "secret"
-	// ConsulSecretEngineMountPoint is the name of the mount point base for Vault's Consul secrets engine
-	ConsulSecretEngineMountPoint = "consul"
+	// this is a feature key to indicate whether to enable Registry Consul's ACL or not
+	RegistryACLFeatureFlag = "ENABLE_REGISTRY_ACL"
```
bnevis-i, line 23: Remove. This feature doesn't need to be in secretstore setup.  Secretstore-setup doesn't enable the Consul ACL feature.  We can create the token for configuring the consul secrets provider regardless of whether or not we enable Consul ACLS.

internal/security/secretstore/init.go
```diff
@@ -422,6 +422,19 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
 		lc.Info("proxy certificate pair upload was skipped because cert secretStore value(s) were blank")
 	}
 
+	// If Registry Consul ACL feature is enabled, then we need to create and save a Vault token to configure
+	// Consul secret engine access, role operations, and managing Consul agent tokens.
+	registryACLEnabled := os.Getenv(RegistryACLFeatureFlag)
+	if registryACLEnabled == "true" {
```
bnevis-i, line 428: Do this every time.  It is harmless.

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
```diff
@@ -0,0 +1,172 @@
...
+	if fileErr != nil {
+		return nil, fmt.Errorf("failed to create tokenpath base dir: %s", fileErr.Error())
+	}
+
+	fileWriter, fileErr := w.fileOpener.OpenFileWriter(tokenFileAbsPath, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0600)
+	if fileErr != nil {
+		return nil, fmt.Errorf("failed to create token file %s: %s", tokenFileAbsPath, fileErr.Error())
+	}
+
+	if fileErr = json.NewEncoder(fileWriter).Encode(tokenCreated); fileErr != nil {
```
bnevis-i, line 101: This line can fail if NewEncoder() returns nil.

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
```diff
@@ -0,0 +1,172 @@
...
+		_ = fileWriter.Close()
+		return nil, fmt.Errorf("failed to write created token: %s", fileErr.Error())
+	}
+
+	if fileErr = fileWriter.Close(); fileErr != nil {
```
bnevis-i, line 106: See internal/security/fileprovider/provider.go lines 172-196.  Since we are passing the token to another container, we want to be able to configure file permissions, uid, and gid.

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
bnevis-i: For creating the token creating token used by filetokenprovider, we revoke to token after it is used via the `RevokeFunc`.  I needs to be here for the other use case.

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
```diff
@@ -0,0 +1,172 @@
...
+	if err := w.secretClient.InstallPolicy(rootToken,
+		consulSecretsEngineOpsPolicyName,
+		consulSecretsEngineOpsPolicyDocument); err != nil {
+		return nil, nil, fmt.Errorf("failed to install Consul secrets engine operations policy: %v", err)
+	}
+
+	// setup new token's properties
+	tokenParams := make(map[string]interface{})
+	tokenParams["type"] = "service"
+	tokenParams["display_name"] = "managing with policy name " + consulSecretsEngineOpsPolicyName
```
bnevis-i, line 149: Suggest display_name = "Configuration token for consul secrets engine"

internal/security/secretstore/tokenfilewriter/tokenfilewriter_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+			expectError := test.installPolicyCallErr != nil ||
+				test.createTokenCallErr != nil ||
+				len(test.rootToken) == 0
+
+			if expectError {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+				require.FileExists(t, testTokenFilePath)
```
bnevis-i, line 139: I don't understand.  If we're using a mockable FileOpener, why is an actual file created?

internal/security/secretstore/tokenfilewriter/tokenfilewriter_test.go
```diff
@@ -0,0 +1,189 @@
...
+	fileWriter := NewWriter(lc, secretClient, flOpener)
+	_, err := fileWriter.CreateAndWrite(testRootToken, testTokenFilePath,
+		fileWriter.CreateMgmtTokenForConsulSecretsEngine)
+	require.Error(t, err)
+}
+
+func setup() func() (logger.LoggingClient, fileioperformer.FileIoPerformer) {
+	return func() (logger.LoggingClient, fileioperformer.FileIoPerformer) {
+		lc := logger.MockLogger{}
+		flOpener := fileioperformer.NewDefaultFileIoPerformer()
```
bnevis-i, line 179: Can you pass a mock object here instead of NewDefaultFileIoPerformer so that you can test file reading and writing without actually having to read and write files?

internal/security/secretstore/tokenfilewriter/tokenfilewriter_test.go
bnevis-i: See my other comment about mocking the file creation commands and thus having to avoid this debacle althogether.

bnevis-i (changes requested):

cmd/security-secretstore-setup/res/configuration.toml
jim-wang-intel: don't get it, what existing env?

internal/security/secretstore/constants.go
jim-wang-intel: I disagree, if no ACL enable why need to create this token at all?

internal/security/secretstore/init.go
jim-wang-intel: same comment above, no need to create useless token if no ACL

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
jim-wang-intel: could not think of any case, could you give me a concrete example that json.NewEncoder will fail?

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
jim-wang-intel: i think our go.mod.secret code in Vault's CreateToken() automatically prefaces `token-`  -xxxxxx something or Vault itself does it, so with my example above, the display_name will be `token-managing-with-policy-name-xxxxxxx`,  that is why I keep the display name this way.

internal/security/secretstore/tokenfilewriter/tokenfilewriter_test.go
jim-wang-intel: we are not using mockable fileopener here, concrete file creates

internal/security/secretstore/tokenfilewriter/tokenfilewriter_test.go
jim-wang-intel: i don't see a need here as there are already tests in`NewDefaultFileIoPerformer`  itself doing the mocking test why duplicate tests here?

cmd/security-secretstore-setup/res/configuration.toml
bnevis-i: Sorry, it wasn't an env var.  It was OutputDir, here https://github.com/edgexfoundry/edgex-go/blob/12a188d7528ee379d18cf3e1ac715e90dfef7258/cmd/security-secretstore-setup/res-file-token-provider/configuration.toml#L11

jim-wang-intel: ok

jim-wang-intel: but that is in fileprovider, nothing to with that utility here

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
jim-wang-intel: I don't understand this: the existing code for creating token-issuing token doesn't have this logic at all;  are you suggesting we modify the existing privilege for token issuing token , too???

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
jim-wang-intel: @bnevis-i :   It is impossible for `json.NewEncoder` to return `nil`, please check out the golang source code at https://golang.org/src/encoding/json/stream.go  line 191-194, in which it is always return an instance of `&Encoder{w: w, escapeHTML: true}`.  Thus `json.NewEncoder()` will never return nil.

bnevis-i: Good to know.

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
jim-wang-intel: fixed

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
jim-wang-intel: added comments

internal/security/secretstore/tokenfilewriter/tokenfilewriter_test.go
jim-wang-intel: added TestMain()

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
jim-wang-intel: @bnevis-i  need more discussion on this. I do not quite understand the logic in the line 172-196 you mentioned above there and I thought the file permission already specified by the fileOpenWriter.  We are not using this token to create another token so I don't understand why we need to do that step. Let's have a discussion tomorrow morning if you have time.

bnevis-i: fileOpenWriter only specifies the mode.  If we are running containers as non-root users the file owner and group must match that of the consuming container.

jim-wang-intel: Hi @bnevis-i  if the `secretstore-setup` is running as non-root, then the chown operations won't work, right?  Also as of today, this code in the entrypoint script running as root, https://github.com/edgexfoundry/edgex-go/blob/master/cmd/security-secretstore-setup/entrypoint.sh#L33-L36  would change the ownership to 2002:2001 anyway, so that the ownership change in the provider is a moot point.

internal/security/secretstore/tokenfilewriter/tokenfilewriter_test.go
```diff
@@ -32,17 +32,21 @@ import (
 	"github.com/edgexfoundry/go-mod-secrets/v2/secrets/mocks"
 )
 
-func TestNewTokenFileWriter(t *testing.T) {
-	lc, flOpener := setup()()
+var lc logger.LoggingClient
+var flOpener fileioperformer.FileIoPerformer
+
+func TestMain(m *testing.M) {
+	lc, flOpener = setup()()
```
lenny-intel, line 39: Why not move code from setup() into here? It is just a couple line that can be here.

lenny-intel (changes requested):

cmd/security-secretstore-setup/res/configuration.toml
jim-wang-intel: resolved

internal/security/secretstore/constants.go
jim-wang-intel: resolved, also use this same flag to not install Cosnul's secret engine

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
jim-wang-intel: resolve as we discussed

internal/security/secretstore/tokenfilewriter/tokenfilewriter.go
jim-wang-intel: "token" + "display_name":  **for Consul ACL bootstrap**
resolved

internal/security/secretstore/tokenfilewriter/tokenfilewriter_test.go
jim-wang-intel: ok, i'll move it.

jim-wang-intel: fixed

lenny-intel (approved):
LGTM

Issue #3156, opened 2021-02-09
[Secure Consul Ph. 1] Enable and Bootstrap Consul ACL to configure Vault's Consul secret engine Access


jim-wang-intel:
# 🚀 Feature Request

### Relevant Package

 
This feature request is for `Consul's bootstrapper` or the entrypoint script of consul service.

### Description

Consul to be enabled with persistent agent tokens and a default "allow" policy. Bootstrap Consul ACL and save the Consul's bootstrap management token into file.

### Describe the solution you'd like
 
1. Create Consul's default `allow` policy with `persistent agent token` turned on
2. Bootstrap Consul's ACL and save the returned Consul bootstrap management token into file for later use.
3. Use both of the pre-stored special `Vault token` in the temp. volume mount and `Consul bootstrap management token` to configure the Vault's Consul secret engine access.
4. Add logic to check whether Consul's ACL has been bootstrapped already because it will produce an error if we try to bootstrap Consul's ACL again in the 2nd time or later.

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?

PR #3215, opened 2021-03-02
feat(security): Implementation to set up Consul ACL

jim-wang-intel:
This PR is part of Phase 1 to secure Consul.  Main feature added is to enable Consul's ACL and bootstrap ACL. The following are detailed changes:
 - Add Consul's ACL configuration file with "allow" policy
 - Disable Consul's DNS port
 - Add ACL related configuration toml on security-bootstrapper
 - Add setupRegistryACL subcommand to security-bootstrapper
 - Add checking and retry logic for "non-empty" Consul leader
 - Add implementation for setting up Consul's ACL, including bootstrap ACL and configure Consul secrets access for Vault

Closes: #3156

## PR Checklist

No ACL of Consul as of today.

## What is the new behavior?
Start up Consul with ACL enabled with "allow" policy and also bootstrap its ACL for the first time use.  Once bootstrapping is successfully, we will use Vault mgmt token to configure consul secret access.

- [ ] Yes
- [x ] No

## New Imports

- [ ] Yes
- [x ] No

## Specific Instructions
This implements the majority of consul bootstrap process in golang code and can be running in docker-compose.  The implementation for snap would be in a separate PR. 

## Other information
To verify this locally, one needs to git clone this PR and then add environment override `ENABLE_REGISTRY_ACL=true` into both `secretstore-setup` and `consul` service of docker-compose file:
```yaml
   environment:
      ENABLE_REGISTRY_ACL: "true"
```

Also make sure to make the volume of secrets to have written permission in the Consul container: 
```yaml
volumes:
      - edgex-init:/edgex-init:ro,z
      - /tmp/edgex/secrets/edgex-consul:/tmp/edgex/secrets/edgex-consul:z
```
as the generated Consul will be written into that secrets volume.

### DIFF
--- a/cmd/security-bootstrapper/entrypoint-scripts/consul_wait_install.sh
+++ b/cmd/security-bootstrapper/entrypoint-scripts/consul_wait_install.sh
@@ -93,6 +93,7 @@ if [ "${ENABLE_REGISTRY_ACL}" == "true" ]; then
     echo "$(date) failed to set up Consul ACL"
   fi
   set -e
+  # no need to wait for Consul's port since it is in ready state after all ACL stuff
 else
   echo "$(date) Starting edgex-core-consul with ACL disabled ..."
   docker-entrypoint.sh agent \
@@ -101,6 +102,7 @@ else
     -server \
     -client 0.0.0.0 &
   # wait for the consul port
+  # this waitFor is not necessary in the other ACL case, as it is already in the ready state
   echo "$(date) Executing waitFor on Consul with waiting on its own port \
     tcp://${STAGEGATE_REGISTRY_HOST}:${STAGEGATE_REGISTRY_PORT}"
   /edgex-init/security-bootstrapper --confdir=/edgex-init/res waitFor \
--- a/cmd/security-bootstrapper/entrypoint-scripts/consul_wait_install.sh
+++ b/cmd/security-bootstrapper/entrypoint-scripts/consul_wait_install.sh
@@ -52,7 +52,10 @@ done
 DEFAULT_CONSUL_LOCAL_CONFIG='
 {
     "enable_local_script_checks": true,
-    "disable_update_check": true
+    "disable_update_check": true,
+    "ports": {
+      "dns": -1
+    }
 }
 '
 
@@ -63,15 +66,47 @@ export CONSUL_LOCAL_CONFIG
 
 echo "$(date) CONSUL_LOCAL_CONFIG: ${CONSUL_LOCAL_CONFIG}"
 
-echo "$(date) Starting edgex-consul..."
-exec docker-entrypoint.sh agent -ui -bootstrap -server -client 0.0.0.0 &
+echo "$(date) ENABLE_REGISTRY_ACL = ${ENABLE_REGISTRY_ACL}"
 
-# wait for the consul port
-echo "$(date) Executing waitFor on Consul with waiting on its own port \
-  tcp://${STAGEGATE_REGISTRY_HOST}:${STAGEGATE_REGISTRY_PORT}"
-/edgex-init/security-bootstrapper --confdir=/edgex-init/res waitFor \
-  -uri tcp://"${STAGEGATE_REGISTRY_HOST}":"${STAGEGATE_REGISTRY_PORT}" \
-  -timeout "${STAGEGATE_WAITFOR_TIMEOUT}"
+if [ "${ENABLE_REGISTRY_ACL}" == "true" ]; then
+  echo "$(date) Starting edgex-core-consul with ACL enabled ..."
+  docker-entrypoint.sh agent \
+    -ui \
+    -bootstrap \
+    -server \
+    -config-file=/edgex-init/consul-bootstrapper/config_consul_acl.json \
+    -client 0.0.0.0 &
+  # wait for the secretstore tokens ready as we need the token for bootstrapping
+  echo "$(date) Executing waitFor on Consul with waiting on TokensReadyPort \
+    tcp://${STAGEGATE_SECRETSTORESETUP_HOST}:${STAGEGATE_SECRETSTORESETUP_TOKENS_READYPORT}"
+  /edgex-init/security-bootstrapper --confdir=/edgex-init/res waitFor \
+    -uri tcp://"${STAGEGATE_SECRETSTORESETUP_HOST}":"${STAGEGATE_SECRETSTORESETUP_TOKENS_READYPORT}" \
+    -timeout "${STAGEGATE_WAITFOR_TIMEOUT}"
+
+  # we don't want to exit out the whole Consul process when ACL bootstrapping failed, just that
+  # Consul won't have ACL to be used
+  set +e
+  # call setupRegistryACL bootstrapping command, containing both ACL bootstrapping and re-configure consul access steps
+  /edgex-init/security-bootstrapper --confdir=/edgex-init/res setupRegistryACL
+  setupACL_code=$?
+  if [ "${setupACL_code}" -ne 0 ]; then
+    echo "$(date) failed to set up Consul ACL"
+  fi
+  set -e
+else
+  echo "$(date) Starting edgex-core-consul with ACL disabled ..."
+  docker-entrypoint.sh agent \
+    -ui \
+    -bootstrap \
+    -server \
+    -client 0.0.0.0 &
+  # wait for the consul port
+  echo "$(date) Executing waitFor on Consul with waiting on its own port \
+    tcp://${STAGEGATE_REGISTRY_HOST}:${STAGEGATE_REGISTRY_PORT}"
+  /edgex-init/security-bootstrapper --confdir=/edgex-init/res waitFor \
+    -uri tcp://"${STAGEGATE_REGISTRY_HOST}":"${STAGEGATE_REGISTRY_PORT}" \
+    -timeout "${STAGEGATE_WAITFOR_TIMEOUT}"
+fi
 
 # Signal that Consul is ready for services blocked waiting on Consul
 /edgex-init/security-bootstrapper --confdir=/edgex-init/res listenTcp \
--- a/cmd/security-bootstrapper/res/configuration.toml
+++ b/cmd/security-bootstrapper/res/configuration.toml
@@ -20,6 +20,12 @@ LogLevel = 'INFO'
     Host = "edgex-core-consul"
     Port = 8500
     ReadyPort = 54324
+    [StageGate.Registry.ACL]
+      Protocol = "http"
+      # this is the filepath for the generated Consul management token from ACL bootstrap
+      BootstrapTokenPath = "/tmp/edgex/secrets/edgex-consul/admin/bootstrap_token.json"
+      # this is the filepath for the Vault token created from secretstore-setup
+      SecretsAdminTokenPath = "/tmp/edgex/secrets/edgex-consul/admin/token.json"
   [StageGate.KongDb]
     Host = "kong-db"
     Port = 5432
@@ -27,3 +33,13 @@ LogLevel = 'INFO'
   [StageGate.WaitFor]
     Timeout = "10s"
     RetryInterval = "1s"
+
+# this configuration is just part of the whole go-mod-bootstrap's secret store to have
+# protocol, host, and port of secretstore using in the security-bootstrapper
+# we are not really using the secret store provider from go-mod-bootstrap in the code
+# also this is needed as snap does not have those environments from env-files
+[SecretStore]
+Type = 'vault'
+Protocol = 'http'
+Host = 'edgex-vault'
+Port = 8200
--- a/cmd/security-bootstrapper/res/configuration.toml
+++ b/cmd/security-bootstrapper/res/configuration.toml
@@ -26,6 +26,8 @@ LogLevel = 'INFO'
       BootstrapTokenPath = "/tmp/edgex/secrets/edgex-consul/admin/bootstrap_token.json"
       # this is the filepath for the Vault token created from secretstore-setup
       SecretsAdminTokenPath = "/tmp/edgex/secrets/edgex-consul/admin/token.json"
+      # this is the filepath for the sentinel file to indicate the registry ACL is set up successfully
+      SentinelFilePath = "/edgex-init/consul-bootstrapper/consul_acl_done"
   [StageGate.KongDb]
     Host = "kong-db"
     Port = 5432
@@ -41,5 +43,5 @@ LogLevel = 'INFO'
 [SecretStore]
 Type = 'vault'
 Protocol = 'http'
-Host = 'edgex-vault'
+Host = 'localhost'
 Port = 8200
--- a/internal/security/bootstrapper/command/setupacl/aclbootstrap.go
+++ b/internal/security/bootstrapper/command/setupacl/aclbootstrap.go
@@ -41,9 +41,9 @@ type Policy struct {
 	Name string `json:"Name"`
 }
 
-// this should only be called once per Consul agent
-func (c *cmd) callConsulACLBootstrapAPI() (*BootStrapACLTokenInfo, error) {
-	aclBootstrapURL, err := c.getAPI_URL(consulACLBootstrapAPI)
+// generateBootStrapACLToken should only be called once per Consul agent
+func (c *cmd) generateBootStrapACLToken() (*BootStrapACLTokenInfo, error) {
+	aclBootstrapURL, err := c.getRegistryApiUrl(consulACLBootstrapAPI)
 	if err != nil {
 		return nil, err
 	}
@@ -79,7 +79,7 @@ func (c *cmd) callConsulACLBootstrapAPI() (*BootStrapACLTokenInfo, error) {
 	}
 }
 
-func (c *cmd) saveBootstrapToken(tokenInfoToBeSaved *BootStrapACLTokenInfo) error {
+func (c *cmd) saveBootstrapACLToken(tokenInfoToBeSaved *BootStrapACLTokenInfo) error {
 	// Write the token to the specified file
 	tokenFileAbsPath, err := filepath.Abs(c.configuration.StageGate.Registry.ACL.BootstrapTokenPath)
 	if err != nil {
--- a/internal/security/bootstrapper/command/setupacl/aclbootstrap.go
+++ b/internal/security/bootstrapper/command/setupacl/aclbootstrap.go
@@ -1,0 +1,114 @@
+/*******************************************************************************
+ * Copyright 2021 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ *******************************************************************************/
+
+package setupacl
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+
+	"github.com/edgexfoundry/go-mod-secrets/v2/pkg/token/fileioperformer"
+)
+
+// BootStrapACLTokenInfo is the key portion of the response metadata from consulACLBootstrapAPI
+type BootStrapACLTokenInfo struct {
+	AccessorID string   `json:"AccessorID"`
+	SecretID   string   `json:"SecretID"`
+	Policies   []Policy `json:"Policies"`
+	CreateTime string   `json:"CreateTime"`
+}
+
+// Policy is the metadata for ACL policy
+type Policy struct {
+	ID   string `json:"ID"`
+	Name string `json:"Name"`
+}
+
+// this should only be called once per Consul agent
+func (c *cmd) callConsulACLBootstrapAPI() (*BootStrapACLTokenInfo, error) {
+	aclBootstrapURL, err := c.getAPI_URL(consulACLBootstrapAPI)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequest(http.MethodPut, aclBootstrapURL, http.NoBody)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to prepare request for http URL: %w", err)
+	}
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to send request for http URL: %w", err)
+	}
+
+	defer func() {
+		_ = resp.Body.Close()
+	}()
+
+	responseBody, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to read response body of bootstrap ACL: %w", err)
+	}
+
+	var bootstrapACLToken BootStrapACLTokenInfo
+	switch resp.StatusCode {
+	case http.StatusOK:
+		if err := json.NewDecoder(bytes.NewReader(responseBody)).Decode(&bootstrapACLToken); err != nil {
+			return nil, fmt.Errorf("failed to decode bootstrapACLToken json data: %v", err)
+		}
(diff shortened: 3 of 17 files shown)


internal/security/bootstrapper/command/setupacl/command.go
```diff
@@ -0,0 +1,364 @@
...
+		// because Consul is still on its way to initialize the new ACL system internally
+		// for the details of this issue, see related issue on Consul's Github website:
+		// https://github.com/hashicorp/consul/issues/5218#issuecomment-457212336
+		if !strings.Contains(err.Error(), consulLegacyACLModeError) {
+			// other type of ACL bootstrapping error, cannot continue
+			return interfaces.StatusCodeExitWithError, fmt.Errorf("failed to bootstrap registry's ACL: %v", err)
+		}
+
+		c.loggingClient.Warnf("found Consul still in ACL legacy mode, will retry once again: %v", err)
+		time.Sleep(5 * time.Second)
```
lenny-intel, line 120: 5 seconds seems like a very long time to wait. I would prefer 1 second with 5 retries.

internal/security/bootstrapper/command/setupacl/command.go
```diff
@@ -0,0 +1,364 @@
...
+// Consul's agent tokens later on
+func (c *cmd) Execute() (statusCode int, err error) {
+	c.loggingClient.Infof("Security bootstrapper running %s", CommandName)
+
+	if err := c.waitForNonEmptyConsulLeader(); err != nil {
+		return interfaces.StatusCodeExitWithError, fmt.Errorf("failed to wait for Consul leader: %v", err)
+	}
+
+	var bootstrapTokenInfo *BootStrapACLTokenInfo
+	bootstrapTokenInfo, err = c.callConsulACLBootstrapAPI()
```
lenny-intel, line 107: Variables should not have the type in the name, so go short with simple `token` or ` aclToken` for name.

internal/security/bootstrapper/command/setupacl/command.go
```diff
@@ -0,0 +1,364 @@
...
+		if err != nil {
+			return interfaces.StatusCodeExitWithError, fmt.Errorf("failed to bootstrap registry's ACL: %v", err)
+		}
+	}
+
+	c.loggingClient.Info("successfully bootstrap registry ACL")
+
+	// Save the bootstrap token into the file so that it can be used later on
+	if err := c.saveBootstrapToken(bootstrapTokenInfo); err != nil {
+		return interfaces.StatusCodeExitWithError, fmt.Errorf("failed to save registry's bootstrap token: %v", err)
```
lenny-intel, line 131: `registry's bootstrap token` => `registry's bootstrap ACL token`

internal/security/bootstrapper/command/setupacl/command.go
```diff
@@ -0,0 +1,364 @@
...
+
+	// retrieve the secretstore (Vault) token from the file produced by secretstore-setup
+	secretstoreToken, err := c.getSecretStoreTokenFromFile()
+	if err != nil {
+		return interfaces.StatusCodeExitWithError, fmt.Errorf("failed to retrieve secretstore token: %v", err)
+	}
+
+	c.loggingClient.Info("successfully get secretstore token and configuring the registry access for secretestore")
+
+	// configure Consul access with both Vault mgmt token and consul's bootstrap token
```
lenny-intel, line 142: `Vault mgmt` => `Secret Store`

internal/security/bootstrapper/command/setupacl/command.go
```diff
@@ -0,0 +1,364 @@
...
+
+	return
+}
+
+// GetCommandName returns the name of this command
+func (c *cmd) GetCommandName() string {
+	return CommandName
+}
+
+func (c *cmd) getAPI_URL(api string) (string, error) {
```
lenny-intel, line 155: underscore in names are non-standard. How about `getRegistryApiUrl`

internal/security/bootstrapper/command/setupacl/command.go
```diff
@@ -0,0 +1,364 @@
...
+			return
+		}
+	}()
+}
+
+// callConsulLeaderAPI makes http request call to get the registry Consul leader
+// the response of getting leader call could be an empty leader (represented by "")
+// even if the http status code is 200 when Consul is just booting up and
+// it will take a bit of time to elect the raft leader
+func (c *cmd) callConsulLeaderAPI() error {
```
lenny-intel, line 226: how about `createConsulLeader()` for the name?

internal/security/bootstrapper/command/setupacl/command.go
```diff
@@ -0,0 +1,364 @@
...
+				c.loggingClient.Warnf("error from getting Consul leader API call, will retry it again: %v", err)
+				c.errs <- err
+				continue
+			}
+			// once reached here, we have a leader and terminate this goroutine
+			c.errs <- nil
+			return
+		}
+	}()
+}
```
lenny-intel, line 220: This code seems too complex for a simple retry loop. Why not just the code in line 206-218 w/o the channel for `waitForNonEmptyConsulLeader`

internal/security/bootstrapper/command/setupacl/command.go
```diff
@@ -0,0 +1,364 @@
...
+	if err := json.NewDecoder(tokenJSONFile).Decode(&tokenData); err != nil {
+		return emptyToken, fmt.Errorf("failed to decode secretstore json token file: %v", err)
+	}
+
+	c.loggingClient.Infof("successfully retrieved secretstore management token from %s", trimmedFilePath)
+
+	return tokenData.Authentication.Token, nil
+}
+
+// configureConsulAccess is to enable the Consul config access to Vault via consul/config/access API
```
lenny-intel, line 307: `Vault` => `the SecretStore`

internal/security/bootstrapper/command/setupacl/command.go
```diff
@@ -0,0 +1,364 @@
...
+	_, err := url.Parse(configAccessURL)
+	if err != nil {
+		return fmt.Errorf("failed to parse config Access URL: %v", err)
+	}
+
+	c.loggingClient.Debugf("configAccessURL: %s", configAccessURL)
+
+	type ConfigAccess struct {
+		RegistryAddress string `json:"address"`
+		BootstrapToken  string `json:"token"`
```
lenny-intel, line 321: `BootstrapToken ` => `consulToken` since you have been using `BootstrapToken ` for the ACL token. or better yet `AccessToken`??

internal/security/bootstrapper/interfaces/command.go
```diff
@@ -22,6 +22,8 @@ const (
 	StatusCodeNoOptionSelected = 1
 	// StatusCodeExitWithError is exit code for error case
 	StatusCodeExitWithError = 2
+	// JSONContentType is the content type for JSON based body/payload
+	JSONContentType = "application/json"
```
lenny-intel, line 26: why not use the constant already here: https://github.com/edgexfoundry/go-mod-core-contracts/blob/b3069d58daddec4f0489f199d23d3a6acd3e16a6/clients/constants.go#L79

lenny-intel (changes requested):

internal/security/bootstrapper/command/setupacl/command.go
jim-wang-intel: not really, acl bootstrapping is kind of expensive process, so less tries is better actually

internal/security/bootstrapper/command/setupacl/aclbootstrap.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// Policy is the metadata for ACL policy
+type Policy struct {
+	ID   string `json:"ID"`
+	Name string `json:"Name"`
+}
+
+// this should only be called once per Consul agent
+func (c *cmd) callConsulACLBootstrapAPI() (*BootStrapACLTokenInfo, error) {
```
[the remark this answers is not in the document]
jim-wang-intel: ok

internal/security/bootstrapper/command/setupacl/aclbootstrap.go
```diff
@@ -0,0 +1,114 @@
...
+			return nil, fmt.Errorf("failed to decode bootstrapACLToken json data: %v", err)
+		}
+		return &bootstrapACLToken, nil
+	default:
+		return nil, fmt.Errorf("failed to bootstrap Consul's ACL via URL [%s] and status code= %d: %s", aclBootstrapURL,
+			resp.StatusCode, string(responseBody))
+	}
+}
+
+func (c *cmd) saveBootstrapToken(tokenInfoToBeSaved *BootStrapACLTokenInfo) error {
```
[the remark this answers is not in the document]
jim-wang-intel: ok

internal/security/bootstrapper/command/setupacl/command.go
jim-wang-intel: ok

internal/security/bootstrapper/command/setupacl/command.go
jim-wang-intel: we not "creating" though... this funciton is actually calling consul's get leader api

internal/security/bootstrapper/command/setupacl/command.go
jim-wang-intel: i don't know how to do it just with those lines 206-218 and at the same time with time-out limit. let me know if you can figure out a better solution.

internal/security/bootstrapper/command/setupacl/command.go
jim-wang-intel: it is the same token actually, so i can change it to BootstrapACLToken for field name and this is post body/payload by the way

internal/security/bootstrapper/command/setupacl/command.go
lenny-intel: The result of call the API is the leader is create or set... callXyzApi is not a descriptive name of have the real action is.

internal/security/bootstrapper/command/setupacl/command.go
lenny-intel: Why? The timeout is there with the ticker?? 
Also I just noticed that this would need to return the timeout error after the for loop

internal/security/bootstrapper/command/setupacl/command.go
lenny-intel: Right, I was confused with the SecretStoreToken... The parameter `consulToken` name should be consistent as `bootstrapACLToken` `

lenny-intel (changes requested):

internal/security/bootstrapper/command/setupacl/command.go
jim-wang-intel: time out error is in **line 196-197**, we have already logged all the error during the getting leader process; there is no need for redundant error

jim-wang-intel: also, note that this function is go routine so no return for the function allowed actually

lenny-intel: Yes, a little refactoring is needed. Doesn't need to be a go func.

jim-wang-intel: let me know your solution, i really don't know how to do it gracefully without go func() when there are retries interval and time limit

jim-wang-intel: plus, to avoid data race condition if channels

internal/security/bootstrapper/command/setupacl/command.go
jim-wang-intel: ok

internal/security/bootstrapper/command/setupacl/command.go
jim-wang-intel: thanks for the help, looks great

jim-wang-intel:
> what cleans up the file when the is a restart of all the services?

no need to clean up even when restarting all services since Consul's ACL is in a good bootstrapped ACL state so no need to re-run the setup registry ACL again.  Unless you want to clean out all the volumes.

lenny-intel:
> Unless you want to clean out all the volumes.

Right, so I see an issue here when volumes are cleared, but sentinel file is left behind. Is there a Consul API that will tell us that it is setup already or not?

internal/security/bootstrapper/command/setupacl/command.go
lenny-intel: @bnevis-i , thoughts on this long timeout?

jim-wang-intel:
> > Unless you want to clean out all the volumes.
> 
> Right, so I see an issue here when volumes are cleared, but sentinel file is left behind. Is there a Consul API that will tell us that it is setup already or not?

the sentinel file is on the volume. I am aware of any Consul API can tell us the ACL bootstrap status.

lenny-intel:
> the sentinel file is on the volume

ohhh! Didn't realize that. ;-) Then I am good!

lenny-intel (dismissed):
LGTM, I'll concede on the 5 sec wait... ;-)

jim-wang-intel:
> LGTM, I'll concede on the 5 sec wait... ;-)

Thanks, it rarely happens and I've put the issue link notes.  Since last time I saw, I've never seen it happened again on my box afterwards. So I can not even re-produce, and my guess is probably last time i've had some left-over state for some reason in consul as Consul is stateful machine.

jim-wang-intel:
recheck

internal/security/bootstrapper/command/setupacl/aclbootstrap.go
```diff
@@ -0,0 +1,114 @@
...
+	CreateTime string   `json:"CreateTime"`
+}
+
+// Policy is the metadata for ACL policy
+type Policy struct {
+	ID   string `json:"ID"`
+	Name string `json:"Name"`
+}
+
+// generateBootStrapACLToken should only be called once per Consul agent
```
bnevis-i, line 44: If each Consul agent has its own bootstrap token what does that mean when you have multiple Consul instances in the infrastructure?  Does that mean multiple bootstrap tokens?  Would a bootstrap token on one Consul instance be honored by another Consul instance?

internal/security/bootstrapper/command/setupacl/command.go
bnevis-i: Looks OK to me.  I set up a HA Vault before and it takes quite a number of seconds for leader election, etc.

internal/security/bootstrapper/command/setupacl/command.go
```diff
@@ -0,0 +1,372 @@
...
+			Token string `json:"client_token"`
+		} `json:"auth"`
+	}
+
+	tokenJSONFile, err := os.Open(trimmedFilePath)
+	defer func() { tokenJSONFile.Close() }()
+
+	if err != nil {
+		return emptyToken, fmt.Errorf("failed to open secretstore token file: %v", err)
+	}
```
bnevis-i, line 274: Replace with call to github.com/edgexfoundry/go-mod-secrets/pkg/token/authtokenloader

internal/security/bootstrapper/command/setupacl/command.go
```diff
@@ -0,0 +1,372 @@
...
+
+	c.loggingClient.Infof("successfully retrieved secretstore management token from %s", trimmedFilePath)
+
+	return tokenData.Authentication.Token, nil
+}
+
+// configureConsulAccess is to enable the Consul config access to the SecretStore via consul/config/access API
+// see the reference: https://www.vaultproject.io/api-docs/secret/consul#configure-access
+func (c *cmd) configureConsulAccess(secretStoreToken string, bootstrapACLToken string) error {
```
bnevis-i, line 288: Wouldn't it be better to extend go-mod-secrets SecretStoreClient, which is the administrative interface to Vault, and put this code there instead?

cmd/security-bootstrapper/entrypoint-scripts/consul_wait_install.sh
```diff
@@ -63,15 +66,47 @@ export CONSUL_LOCAL_CONFIG
...
+    -ui \
+    -bootstrap \
+    -server \
+    -client 0.0.0.0 &
+  # wait for the consul port
+  echo "$(date) Executing waitFor on Consul with waiting on its own port \
+    tcp://${STAGEGATE_REGISTRY_HOST}:${STAGEGATE_REGISTRY_PORT}"
+  /edgex-init/security-bootstrapper --confdir=/edgex-init/res waitFor \
+    -uri tcp://"${STAGEGATE_REGISTRY_HOST}":"${STAGEGATE_REGISTRY_PORT}" \
+    -timeout "${STAGEGATE_WAITFOR_TIMEOUT}"
```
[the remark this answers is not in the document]
jim-wang-intel: because in the "if" branch, consul is in ready state after all ACL stuff so we don't need to wait for its port

cmd/security-bootstrapper/res/configuration.toml
```diff
@@ -20,10 +20,26 @@ LogLevel = 'INFO'
...
     RetryInterval = "1s"
+
+# this configuration is just part of the whole go-mod-bootstrap's secret store to have
+# protocol, host, and port of secretstore using in the security-bootstrapper
+# we are not really using the secret store provider from go-mod-bootstrap in the code
+# also this is needed as snap does not have those environments from env-files
+[SecretStore]
+Type = 'vault'
+Protocol = 'http'
+Host = 'edgex-vault'
```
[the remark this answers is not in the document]
jim-wang-intel: we could do that way, too. with this then we don't need env var override in docker-compose file; but we do env override in the snap env.  That said, I am ok with either way.

cmd/security-bootstrapper/entrypoint-scripts/consul_wait_install.sh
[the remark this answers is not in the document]
bnevis-i: Request to add comments to that effect.

cmd/security-bootstrapper/res/configuration.toml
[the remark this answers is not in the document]
bnevis-i: Just asking to be consistent with the over configuration.toml's.

internal/security/bootstrapper/command/setupacl/aclbootstrap.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/edgexfoundry/go-mod-secrets/v2/pkg/token/fileioperformer"
+)
+
+// BootStrapACLTokenInfo is the key portion of the response metadata from consulACLBootstrapAPI
+type BootStrapACLTokenInfo struct {
+	AccessorID string   `json:"AccessorID"`
+	SecretID   string   `json:"SecretID"`
+	Policies   []Policy `json:"Policies"`
+	CreateTime string   `json:"CreateTime"`
+}
```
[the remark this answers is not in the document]
jim-wang-intel: we also need global management policy name later for associate that with the agent token (I don't want to hard-coded the global management policy name with prior knowledge) and I want to save this token for later use; other than that I am fine to remove other fields.

internal/security/bootstrapper/command/setupacl/command.go
jim-wang-intel: yes, good point, I had in my mind to do that refactor later after phase 2 for all Vault's Consul secret engine APIs, making it work first for now is higher priority

internal/security/bootstrapper/command/setupacl/command.go
jim-wang-intel: ok, will need to spend time on trying to understand that piece first in order to fit in here

cmd/security-bootstrapper/entrypoint-scripts/consul_wait_install.sh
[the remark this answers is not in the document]
jim-wang-intel: ok, i'll add the comments

internal/security/bootstrapper/command/setupacl/command.go
bnevis-i: AR: File technical debt issue when this PR is merged.

cmd/security-bootstrapper/res/configuration.toml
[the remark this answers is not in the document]
jim-wang-intel: As of today's master branch, I am seeing there are mixture of both uses on today's configuration toml's. `secretstore-setup` uses `edgex-vault` but `proxy-setup` uses `localhost`.  If want to, we need to discuss this more.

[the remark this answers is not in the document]
bnevis-i: @lenny-intel Any opinion here?  This is a cross-cutting concern.

internal/security/bootstrapper/command/setupacl/command.go
jim-wang-intel: will do

cmd/security-bootstrapper/res/configuration.toml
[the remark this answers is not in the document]
lenny-intel: @jim-wang-intel , @bnevis-i,  We should be consistently using `localhost` so Snap doesn't have to override and a non-docker and non-snap deployment works without overrides.

jim-wang-intel:
Hi @ernestojeda , I am starting seeing the snap pulling core18 from internet error in the CI pipeline and it happens 2-in-a-row now, could you please look into that?

cmd/security-bootstrapper/res/configuration.toml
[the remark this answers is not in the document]
jim-wang-intel: ok

internal/security/bootstrapper/command/setupacl/aclbootstrap.go
jim-wang-intel: @bnevis-i : I googled it and from [Hashicorp's documentation](https://learn.hashicorp.com/tutorials/consul/access-control-setup-production#enable-acls-on-the-agents), the answers to your questions seem to be **yes** on `multiple tokens`: 

In order for ACL configuration to be applied correctly you will need to **apply the same parameters to every server and every client in your datacenter.** 

otherwise, you would think Consul should sync'ed up with one bootstrap token with all cluster automatically and internally, and no need to do it for every server and client in a datacenter.

ernestojeda:
Not much I can do other than re-run the job. I will open a ticket with the LF to see if they can diagnose any network issue.

jim-wang-intel:
> Not much I can do other than re-run the job. I will open a ticket with the LF to see if they can diagnose any network issue.

Ok, thanks for the information.

internal/security/bootstrapper/command/setupacl/aclbootstrap.go
bnevis-i: > In order for ACL configuration to be applied correctly you will need to **apply the same parameters to every server and every client in your datacenter.**

The way I read that was that the config.hcl needed to specify ACL mode on each Consul agent.

Later on it reads "To create the initial bootstrap token, use the acl bootstrap command _on one of the servers_."  (emphasis mine).  So it LOOKS like we're ok.

jim-wang-intel: ok, thanks for the details, good to know.

cmd/security-bootstrapper/entrypoint-scripts/consul_wait_install.sh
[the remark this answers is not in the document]
jim-wang-intel: added

cmd/security-bootstrapper/res/configuration.toml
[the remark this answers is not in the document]
jim-wang-intel: changed `secretstore-setup`, `filetokenprovider` and this one to `localhost`

internal/security/bootstrapper/command/setupacl/aclbootstrap.go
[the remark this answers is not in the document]
jim-wang-intel: trimmed down

internal/security/bootstrapper/command/setupacl/command.go
jim-wang-intel: replaced

bnevis-i (approved):

internal/security/bootstrapper/command/setupacl/command.go
jim-wang-intel: #3227 created

lenny-intel (approved):
LGTM

PR #3228, opened 2021-03-04
refactor: Rework Proxy Setup to use KongRoute struct for configuration

lenny-intel:
## PR Checklist

Proxy Setup uses repurposed ClientInfo struct for the Route configuration which has become an issue once the standards keys changed to service keys. This is because the key name is also used for the route name

## What is the new behavior?
Now Proxy Setup uses existing KongRoute struct that has extra name property for the route name. The Route configuration is properly named `Routes`, no longer `Clients`

- [x ] Yes
- [ ] No

BREAKING CHANGE: Names for Route configuration has changed

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information

### DIFF
--- a/cmd/security-proxy-setup/res/configuration.toml
+++ b/cmd/security-proxy-setup/res/configuration.toml
@@ -54,38 +54,45 @@ RetryWaitPeriod = "1s"
   [SecretStore.Authentication]
   AuthType = "X-Vault-Token"
 
-[Clients]
-  [Clients.CoreData]
+[Routes]
+  [Routes.CoreData]
+  Name = "coredata"
   Protocol = "http"
   Host = "localhost"
   Port = 48080
 	
-  [Clients.Metadata]
+  [Routes.Metadata]
+  Name = "metadata"
   Protocol = "http"
   Host = "localhost"
   Port = 48081
 	
-  [Clients.Command]
+  [Routes.Command]
+  Name = "command"
   Protocol = "http"
   Host = "localhost"
   Port = 48082
 	
-  [Clients.Notifications]
+  [Routes.Notifications]
+  Name = "notifications"
   Protocol = "http"
   Host = "localhost"
   Port = 48060
 
-  [Clients.Scheduler]
+  [Routes.Scheduler]
+  Name = "scheduler"
   Protocol = "http"
   Host = "localhost"
   Port = 48085
 
-  [Clients.RulesEngine]
+  [Routes.RulesEngine]
+  Name = "rulesengine"
   Protocol = "http"
   Host = "localhost"
   Port = 48075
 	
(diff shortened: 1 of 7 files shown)


cmd/security-proxy-setup/res/configuration.toml
```diff
@@ -54,38 +54,45 @@ RetryWaitPeriod = "1s"
   [SecretStore.Authentication]
   AuthType = "X-Vault-Token"
 
-[Clients]
-  [Clients.CoreData]
+[Routes]
+  [Routes.CoreData]
+  Name = "coredata"
```
jim-wang-intel, line 59: could you add a comment on the name is always being lower cases as internally we setup kong route that way?

jim-wang-intel (commented):
overall LGTM

cmd/security-proxy-setup/res/configuration.toml
lenny-intel: Code is still forcing it to lowercase, so case here doesn't matter.

bnevis-i (approved):

cmd/security-proxy-setup/res/configuration.toml
jim-wang-intel: i think it matters. the service name is the key to the map, so the coredata != CoreData in terms of map's key.

jim-wang-intel (changes requested):
map key is case sensitive

cmd/security-proxy-setup/res/configuration.toml
lenny-intel: Lowercased `name` is used to create the proxy on line 149 & 153 of service.go

jim-wang-intel: ok, i am confused by line 254-256, which is using the name directly from config.route. so if one puts mix or big cases in the name of config file then the name to the map will be like that.
```go
	for serviceName, route := range s.configuration.Routes {
		merged[serviceName] = route
	}
```

jim-wang-intel (approved):

cmd/security-proxy-setup/res/configuration.toml
lenny-intel: `serviceName` is just the map key which is not used for the route. `route` now has the Name filed which is used on lines 149 & 153 to create the actual proxy route. We no longer use the map key as part of the `proxy route`.

jim-wang-intel: i got that, it's just that we have this new "name" field from config now and it creates the contention in the map's key with it. admitted it's the same behavior as before in the map, so seeing resolved.

Issue #3255, opened 2021-03-15
[Secure Consul Ph. 2] Integrate the Consul token Header changes from go-mod-bootstrap


jim-wang-intel:
# 🚀 Feature Request

### Relevant Package

 
This feature request is for `edgex-go` all services.

### Description
 
This enhancement maybe just a `go.mod` update to utilize the changes made from https://github.com/edgexfoundry/go-mod-bootstrap/issues/124 so that all EdgeX services will benefit from this change.

### Describe the solution you'd like
 
- Update the go-mod-bootstrap's version number to utilize the changes made from this issue https://github.com/edgexfoundry/go-mod-bootstrap/issues/124 after the PR to close that issue merged into master create a version for that.

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?

jim-wang-intel:
Now this depends on https://github.com/edgexfoundry/go-mod-bootstrap/issues/199

PR #3331, opened 2021-04-01
feat(security): Integrate EdgeX core servcies/app service with Consul tokens

jim-wang-intel:
Changes made:
 - Integrate with the latest `go-mod-bootstrap` and using it to generating Consul tokens per service's Vault token
 - Add `dumb-init` and re-arrange a bit on `system-management` Dockerfile so that it is bootstrappable for using waiting mechanism from security-bootstrapper entrypoint script
 - Adjust EdgeX's Consul token policy so that the services in snap making registry API calls to Consul registry's API on KV  store can access Consul catalog and registry
 - In snap, add the right starting order (eg. `after security-consul-bootstrapper`) for those EdgeX services generating Consul tokens based on service key or role name from security-bootstrapper.

Closes: #3255

## PR Checklist

Currently, Consul tokens generated by `go-mod-bootstrap`'s using secretProvider on the fly via each service token are not hook up yet.

## What is the new behavior?
Now consul tokens are used by EdgeX services via go-mod-bootstrap's token generating and set to consul configuration on the fly for each service.

- [ ] Yes
- [x ] No

## New Imports

- [ ] Yes
- [x ] No

## Specific Instructions
This one should be merged after PR #3336 

## Other information
 - To test this locally, one can git clone this PR and do the snap build and one should see all EdgeX services are running ok.
 - To test this in docker, use compose file using the updated compose files from Docker-compose PR: https://github.com/edgexfoundry/edgex-compose/pull/53

### DIFF
--- a/cmd/security-bootstrapper/res/configuration.toml
+++ b/cmd/security-bootstrapper/res/configuration.toml
@@ -41,6 +41,8 @@ LogLevel = 'INFO'
       # this section contains the list of registry roles for EdgeX services
       # the service keys are the role names
       [StageGate.Registry.ACL.Roles]
+        [StageGate.Registry.ACL.Roles.appservice-rules-engine]
+          Description = "role for application service of rules engine"
         [StageGate.Registry.ACL.Roles.edgex-core-data]
           Description = "role for coredata"
         [StageGate.Registry.ACL.Roles.edgex-core-metadata]
@@ -51,6 +53,8 @@ LogLevel = 'INFO'
           Description = "role for notifications"
         [StageGate.Registry.ACL.Roles.edgex-support-scheduler]
           Description = "role for scheduler"
+        [StageGate.Registry.ACL.Roles.edgex-sys-mgmt-agent]
+          Description = "role for system management agent"
           
   [StageGate.KongDb]
     Host = "kong-db"
--- a/cmd/security-bootstrapper/res/configuration.toml
+++ b/cmd/security-bootstrapper/res/configuration.toml
@@ -41,7 +41,7 @@ LogLevel = 'INFO'
       # this section contains the list of registry roles for EdgeX services
       # the service keys are the role names
       [StageGate.Registry.ACL.Roles]
-        [StageGate.Registry.ACL.Roles.appservice-rules-engine]
+        [StageGate.Registry.ACL.Roles.app-rules-engine]
           Description = "role for application service of rules engine"
         [StageGate.Registry.ACL.Roles.edgex-core-data]
           Description = "role for coredata"
@@ -55,6 +55,10 @@ LogLevel = 'INFO'
           Description = "role for scheduler"
         [StageGate.Registry.ACL.Roles.edgex-sys-mgmt-agent]
           Description = "role for system management agent"
+        [StageGate.Registry.ACL.Roles.device-virtual]
+          Description = "role for device virtual service"
+        [StageGate.Registry.ACL.Roles.device-rest]
+          Description = "role for device rest service"
           
   [StageGate.KongDb]
     Host = "kong-db"
--- a/cmd/sys-mgmt-agent/res/configuration.toml
+++ b/cmd/sys-mgmt-agent/res/configuration.toml
@@ -57,3 +57,19 @@ Type = 'consul'
   Protocol = 'http'
   Host = 'localhost'
   Port = 48085
+
+[SecretStore]
+Type = 'vault'
+Protocol = 'http'
+Host = 'localhost'
+Port = 8200
+Path = '/v1/secret/edgex/sys-mgmt-agent/'
+TokenFile = '/vault/config/assets/resp-init.json'
+RootCaCertPath = ''
+ServerName = ''
+# Number of attempts to retry retrieving secrets before failing to start the service.
+AdditionalRetryAttempts = 10
+# Amount of time to wait before attempting another retry
+RetryWaitPeriod = "1s"
+  [SecretStore.Authentication]
+  AuthType = 'X-Vault-Token'
--- a/cmd/sys-mgmt-agent/res/configuration.toml
+++ b/cmd/sys-mgmt-agent/res/configuration.toml
@@ -26,11 +26,13 @@ MaxResultCount = 50000
 StartupMsg = 'This is the System Management Agent Service'
 Timeout = 20000
 FormatSpecifier = '%(\\d+\\$)?([-#+ 0(\\<]*)?(\\d+)?(\\.\\d+)?([tT])?([a-zA-Z%])'
+ConfigAccessTokenFile = '/tmp/edgex/secrets/edgex-sys-mgmt-agent/consul-token'  #ignored in non-secure mode
 
 [Registry]
 Host = 'localhost'
 Port = 8500
 Type = 'consul'
+AccessTokenFile = '/tmp/edgex/secrets/edgex-sys-mgmt-agent/consul-token'  #ignored in non-secure mode
 
 [Clients]
   [Clients.edgex-support-notifications]
--- a/cmd/sys-mgmt-agent/res/configuration.toml
+++ b/cmd/sys-mgmt-agent/res/configuration.toml
@@ -26,13 +26,11 @@ MaxResultCount = 50000
 StartupMsg = 'This is the System Management Agent Service'
 Timeout = 20000
 FormatSpecifier = '%(\\d+\\$)?([-#+ 0(\\<]*)?(\\d+)?(\\.\\d+)?([tT])?([a-zA-Z%])'
-ConfigAccessTokenFile = '/tmp/edgex/secrets/edgex-sys-mgmt-agent/consul-token'  #ignored in non-secure mode
 
 [Registry]
 Host = 'localhost'
 Port = 8500
 Type = 'consul'
-AccessTokenFile = '/tmp/edgex/secrets/edgex-sys-mgmt-agent/consul-token'  #ignored in non-secure mode
 
 [Clients]
   [Clients.edgex-support-notifications]
--- a/internal/security/bootstrapper/redis/configure.go
+++ b/internal/security/bootstrapper/redis/configure.go
@@ -28,7 +28,6 @@ import (
 
 	"github.com/edgexfoundry/go-mod-bootstrap/v2/bootstrap"
 	"github.com/edgexfoundry/go-mod-bootstrap/v2/bootstrap/flags"
-	"github.com/edgexfoundry/go-mod-bootstrap/v2/bootstrap/handlers"
 	"github.com/edgexfoundry/go-mod-bootstrap/v2/bootstrap/interfaces"
 	"github.com/edgexfoundry/go-mod-bootstrap/v2/bootstrap/startup"
 	"github.com/edgexfoundry/go-mod-bootstrap/v2/di"
@@ -63,8 +62,8 @@ func Configure(ctx context.Context,
 		nil,
 		startupTimer,
 		dic,
+		true,
 		[]interfaces.BootstrapHandler{
-			handlers.SecureProviderBootstrapHandler,
 			redisBootstrapHdl.GetCredentials,
 			redisBootstrapHdl.SetupConfFiles,
 		},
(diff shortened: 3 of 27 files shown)


snap/snapcraft.yaml
```diff
@@ -215,18 +215,23 @@ apps:
       STAGEGATE_REGISTRY_ACL_SECRETSADMINTOKENPATH: $SNAP_DATA/secrets/edgex-consul/admin/token.json
       STAGEGATE_REGISTRY_ACL_SENTINELFILEPATH: $SNAP_DATA/consul/config/consul_acl_done
       STAGEGATE_REGISTRY_ACL_TOKENBASEDIR: $SNAP_DATA/secrets
+      ADD_REGISTRY_ACL_ROLES: edgex-sys-mgmt-agent, appservice-rules-engine
```
bnevis-i, line 218: Why enable these specially vs just making them enabled by default?

bnevis-i (dismissed):

snap/snapcraft.yaml
jim-wang-intel: these two services right now are not running in secure mode by default so that is why we add it in when running in secure mode and would require consul token in secure mode.

lenny-intel: I think these could be added to the default list since they are always present in secure mode.

snap/snapcraft.yaml
```diff
@@ -237,64 +242,84 @@ apps:
...
       - security-bootstrap-redis
+      # have consul token generated first for this service
+      - security-consul-bootstrapper
       - security-proxy-setup
     command: bin/core-metadata -confdir $SNAP_DATA/config/core-metadata/res -cp -r
     command-chain:
       - bin/security-secret-store-env-var.sh
     environment:
       SECRETSTORE_TOKENFILE: $SNAP_DATA/secrets/edgex-core-metadata/secrets-token.json
+      SERVICE_CONFIGACCESSTOKENFILE: $SNAP_DATA/secrets/edgex-core-metadata/consul-token
```
lenny-intel, line 253: Since we have to override here due to the change from `/tmp` to  `$SNAP_DATA` should we do the same in secure docker and have the default in the config files be set to empty? That way in non-secure mode they don't see the funky `Warning`?

If we did take that approach, that `Warning` could become an `Error`

lenny-intel (changes requested):

snap/snapcraft.yaml
jim-wang-intel: we could do that way, too. that could save one environment override

snap/snapcraft.yaml
jim-wang-intel: not sure which way is better: that would increase quite a few env. overrides in secure docker file

snap/snapcraft.yaml
lenny-intel: IMO, it is more consistent with the usage. The environment variable is for services that we don't know have been added. These two are always there.

jim-wang-intel: good point, ok i'll change it to that default list

snap/snapcraft.yaml
lenny-intel: Yea, that is what I was try to avoid, but now that we have to have them in Snap, it would make it consistent across snap and docker. I also felt having it in the default (which is constantly non-secure) is a bit of a code smell. ;-)

jim-wang-intel: oh, man, now we would resurrect all your PRs done in last few days? ;-)

lenny-intel: Yup, I can do ASC and SDK. You can just move forward with these PRs assuming overrides needed.

jim-wang-intel: awesome, thanks Lenny!

jim-wang-intel:
This PR should not merged until we resolve the issue #3335

snap/snapcraft.yaml
```diff
@@ -319,6 +347,8 @@ apps:
     environment:
       BINDING_PUBLISHTOPIC: events
       EDGEX_SECURITY_SECRET_STORE: "false"
+      SERVICE_CONFIGACCESSTOKENFILE: $SNAP_DATA/secrets/appservice-rules-engine/consul-token
+      REGISTRY_ACCESSTOKENFILE: $SNAP_DATA/secrets/appservice-rules-engine/consul-token
```
tonyespy, line 351: These need to be `$SNAP_DATA/secrets/edgex-application-service/consul-token` to ensure that they end up in the same directory as the vault token.

tonyespy (commented):
The changes look good, although I there's one comment inline about the app-service-configurable token paths. Also we probably need to add overrides for device-virtual, as it's included in this snap as well.

snap/snapcraft.yaml
jim-wang-intel: that actually is the service key for using the rule engine's profile, see https://github.com/edgexfoundry/app-service-configurable/blob/master/res/rules-engine/configuration.toml#L36  The "edgex-application-service" folder is actually too generic and it's not being used by application service.  I think we should remove that one but removing that one seems to be out of scope of this PR.

snap/snapcraft.yaml
jim-wang-intel: both `edgex-sys-mgmt-agent` and `appservice-rules-engine` have been added into the default list

jim-wang-intel:
> This PR should not merged until we resolve the issue #3335

No longer depends on that, it now depends on using the version of this https://github.com/edgexfoundry/go-mod-bootstrap/pull/202

snap/snapcraft.yaml
jim-wang-intel: Note:   now will change to clean up these envs and from configuration as it will be no longer needed once https://github.com/edgexfoundry/go-mod-bootstrap/pull/202 is merged in and uses that version of `go-mod-bootstrap`.

internal/security/bootstrapper/redis/configure.go
```diff
@@ -63,8 +62,8 @@ func Configure(ctx context.Context,
 		nil,
 		startupTimer,
 		dic,
+		true,
```
bnevis-i, line 65: Isn't there a way to use a named constant here.  "true" and "false" is so non-self-explanatory.

jim-wang-intel: that is self-explanatory in go-mod-bootstrap's API IMO.

jim-wang-intel: this is calling go-mod-bootstrap's API btw

snap/snapcraft.yaml
jim-wang-intel: those file-based configuration of edgex-go services are removed now, so no longer an issue here.

snap/snapcraft.yaml
jim-wang-intel: those file-based configuration of edgex-go services are removed now, so no longer an issue here.

internal/security/bootstrapper/redis/configure.go
lenny-intel: @bnevis-i , there reason I don't think it is an issue is because this is what my IDE display for me:
image

internal/system/agent/main.go
```diff
@@ -66,6 +66,9 @@ func Main(ctx context.Context, cancel context.CancelFunc, router *mux.Router, re
 		configuration,
 		startupTimer,
 		dic,
+		// would need to change to true and add secretstore config in future once we have Consul access tightened up
+		// and thus required access token
+		false,
```
lenny-intel, line 71: I think this has to be `true` and add the SecretStore config section to the services config. Otherwise it will never be able to use Consul. This PR is all about enabling `all` the services to use secure consul.

snap/snapcraft.yaml
```diff
@@ -237,6 +239,8 @@ apps:
     adapter: none
     after:
       - security-bootstrap-redis
+      # have consul role generated first for this service
```
lenny-intel, line 242: How about  `This generates the consul role for this service before the service starts'. Here and other same comments

cmd/security-bootstrapper/res/configuration.toml
```diff
@@ -41,6 +41,8 @@ LogLevel = 'INFO'
       # this section contains the list of registry roles for EdgeX services
       # the service keys are the role names
       [StageGate.Registry.ACL.Roles]
+        [StageGate.Registry.ACL.Roles.appservice-rules-engine]
+          Description = "role for application service of rules engine"
```
lenny-intel, line 45: Need to also add `appservice-rules-engine` to secret store setup static list so its token is also generated

lenny-intel (changes requested):

internal/system/agent/main.go
jim-wang-intel: ok

cmd/security-bootstrapper/res/configuration.toml
jim-wang-intel: so this is considered the default service i think because we can also add that through ENV override from secretstore-setup

internal/security/bootstrapper/redis/configure.go
bnevis-i: Sweet.

cmd/security-bootstrapper/res/configuration.toml
lenny-intel: Point is if we added it statically here we need to do the same for vault token

internal/system/agent/main.go
jim-wang-intel: changed to true and added secretstore config

snap/snapcraft.yaml
jim-wang-intel: fixed

cmd/security-bootstrapper/res/configuration.toml
jim-wang-intel: added `app-rules-engine` as default

cmd/sys-mgmt-agent/res/configuration.toml
```diff
@@ -57,3 +57,19 @@ Type = 'consul'
...
   Host = 'localhost'
   Port = 48085
+
+[SecretStore]
+Type = 'vault'
+Protocol = 'http'
+Host = 'localhost'
+Port = 8200
+Path = '/v1/secret/edgex/sys-mgmt-agent/'
+TokenFile = '/vault/config/assets/resp-init.json'
```
lenny-intel, line 67: Note that for App Services I have been using the `/tmp/edgex/secrets/... ` location of the token that was created from Secret Store running in docker. I assume this is also overridden in Snap for its specific location Correct?

Something we should consider doing for all the services, but is out of scope for this PR.

internal/system/agent/router.go
```diff
@@ -70,7 +70,8 @@ func loadRestRoutes(r *mux.Router, dic *di.Container) {
 		func(w http.ResponseWriter, r *http.Request) {
 			healthHandler(w, r, bootstrapContainer.LoggingClientFrom(dic.Get), bootstrapContainer.RegistryFrom(dic.Get))
 		}).Methods(http.MethodGet)
-	b.HandleFunc(
+	// temporarily resolves the health check critical issue in Consul
```
lenny-intel, line 73: Make this a `TODO` comment

lenny-intel (changes requested):

cmd/sys-mgmt-agent/res/configuration.toml
jim-wang-intel: yes, `system-agent` has its overriden in snap as well.

jim-wang-intel: the majority of core- services follow the current way as I did above. not sure which way to standardize in future. maybe worth a discussion in next Weekly Security WG or Core WG meeting.  
@jpwhitemn  and @hutchic, thoughts?

lenny-intel: Yup, we should discuss in next Security WG.

 Snap will always need the override since the root of the path is ` $SNAP_DATA`. 
IMO, we should standardize the default config on using `/tmp/edgex/secrets/...` so we can eliminate a bunch of env overrides and allows me to run the service from command-line and it will find the token generated from the docker stack.

internal/system/agent/router.go
jim-wang-intel: fixed

lenny-intel (approved):
LGTM

Issue #3335, opened 2021-04-02
[Secure Consul Ph. 2] Consul tokens generated by Vault mgmt tokens expired after 1 hour


jim-wang-intel:
# 🚀 Feature Request

### Relevant Package

 
This feature request is for `security-bootstrapper` or `security-consul-bootstrapper`

### Description
 
The issue is located in: all EdgeX services and application services using Consul tokens generated by Vault mgmt token through Vault's `/consul/creds/` API are expired after 1 hour and thus all calls to consul will be failed after tokens are expired.  You will see 403 Status Forbidden when that happens.

This is kind of newly emergent bug once we integrated the Consul token with Vault's mgmt token generation. 

Currently, the TTL or period of Vault management token is set to `1h` (1 hour) and hence the consul tokens generated by this Vault token are inherited the TTL behavior from Vault as Vault is managing the life cycle of these tokens including consul tokens generated by it.

We should renew the Vault management token so that the expiration would not happen and thus consul tokens would not be revoked by Vault.

If not doing the renewal of Vault management token, once consul tokens using by EdgeX services expired, we start to see the 403 status code from those services calling to consul shown as follows:
<pre><code>
edgex-core-consul        |     2021-04-02T17:34:15.110Z [ERROR] agent.http: Request error: method=GET url=/v1/kv/edgex/appservices/1.0/AppService-rules-engine/Writable/?index=357&recurse=&wait=1800000ms from=192.168.80.16:38122 error="ACL not found"
edgex-app-service-configurable-rules | level=ERROR ts=2021-04-02T17:34:15.111195933Z app=AppService-rules-engine source=config.go:421 msg="Unexpected response code: 403"
edgex-core-consul        |     2021-04-02T17:34:16.108Z [ERROR] agent.http: Request error: method=GET url=/v1/kv/edgex/core/1.0/edgex-support-scheduler/Writable/?index=178&recurse=&wait=1800000ms from=192.168.80.10:43482 error="ACL not found"
edgex-support-scheduler  | level=ERROR ts=2021-04-02T17:34:16.109513144Z app=edgex-support-scheduler source=config.go:421 msg="Unexpected response code: 403"
edgex-core-consul        |     2021-04-02T17:34:16.259Z [ERROR] agent.http: Request error: method=GET url=/v1/kv/edgex/core/1.0/edgex-core-data/Writable/?index=74&recurse=&wait=1800000ms from=192.168.80.14:32952 error="ACL not found"
[...]

PR #3336, opened 2021-04-03
feat(security): Add renewal mechanism for Vault management token in security Consul bootstrapper

jim-wang-intel:
Changes made:
 - add renewal process in security-bootstrapper's consul entrypoint
 - add renewal process in snap package as a new simple service using script

The renewal process will be kicked off once the Consul ACLs setup is successfully done and is running as long-running process.
The Vault token renewal happens when its TTL is less than 10% of its original period if it is not expired yet.

Closes: #3335

## PR Checklist

No renewal of token

## What is the new behavior?
Renewal for token process added

- [ ] Yes
- [x ] No

## New Imports

- [ ] Yes
- [x ] No

## Specific Instructions

## Other information
Locally tested and observed logging message something like `"successfully renewed"`:
```
......
Mon Apr  5 16:16:22 UTC 2021 try to renew management token with http://edgex-vault:8200/v1/auth/token/renew-self:
Mon Apr  5 16:16:22 UTC 2021   successfully renewed management token
Mon Apr  5 16:16:22 UTC 2021 checking Vault management token's TTL with http://edgex-vault:8200/v1/auth/token/lookup-self ...
Mon Apr  5 16:16:23 UTC 2021 management token will be renewed after 161 seconds
......
```

### DIFF
--- a/cmd/security-bootstrapper/entrypoint-scripts/consul_wait_install.sh
+++ b/cmd/security-bootstrapper/entrypoint-scripts/consul_wait_install.sh
@@ -35,6 +35,50 @@ vault_ready()
   fi
 }
 
+# function to self-renew Consul secrets engine's Vault management token
+# the self-renew occurs after its TTL < 10% of its period
+renewToken()
+{
+    mgmt_token="$1"
+
+    lookup_url="http://${SECRETSTORE_HOST}:${SECRETSTORE_PORT}/v1/auth/token/lookup-self"
+    renew_url="http://${SECRETSTORE_HOST}:${SECRETSTORE_PORT}/v1/auth/token/renew-self"
+
+    # loop forever for token-renewal process
+    while true; do
+        echo "$(date) checking Vault management token's TTL with ${lookup_url} ..."
+        # self lookup token's period and remaining life (ttl)
+        response=$(curl -s -w "__STATUS_CODE__=%{http_code}" -H "Cache-Control: no-cache" \
+            -H "X-Vault-Token:${mgmt_token}" "${lookup_url}")
+        status_code="${response#*__STATUS_CODE__=}"
+        json_data="${response%__STATUS_CODE__=*}"
+        if [ "${status_code}" -eq 200 ]; then
+            # caluclate the amount of time to sleep before renew the token
+            # the self-renew will occur when its TTL < 10% of its period
+            wait_time=$(echo "${json_data}" | jq -r '.data.ttl-.data.period/10')
+            if [ "${wait_time}" -gt 0 ]; then
+                echo "$(date) management token will be renewed after ${wait_time} seconds"
+                sleep "${wait_time}"
+            else
+                # alredy below the threshold: no wait, just renew it if hasn't expired yet
+                echo "$(date) management token's TTL is lower than 10%"
+            fi
+            echo "$(date) try to renew management token with ${renew_url}:"
+            renew_status_code=$(curl -X POST -s -o /dev/null -w "%{http_code}" -H "X-Vault-Token:${mgmt_token}" "${renew_url}")
+            if [ "${renew_status_code}" -eq 200 ]; then
+              echo "$(date)   successfully renewed management token"
+            else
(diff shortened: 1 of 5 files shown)


cmd/security-bootstrapper/entrypoint-scripts/consul_wait_install.sh
```diff
@@ -91,7 +135,22 @@ if [ "${ENABLE_REGISTRY_ACL}" == "true" ]; then
...
   if [ "${setupACL_code}" -ne 0 ]; then
     echo "$(date) failed to set up Consul ACL"
+  else
+    # kick off the token renewal process
+    echo "$(date) launching vault management token renewal process:"
+    if [ -f "${STAGEGATE_REGISTRY_ACL_SECRETSADMINTOKENPATH}" ]; then
+      mgmt_token=$(jq -r '.auth.client_token' "${STAGEGATE_REGISTRY_ACL_SECRETSADMINTOKENPATH}")
+      parsing_token_ret_code=$?
+      if [ "${parsing_token_ret_code}" -eq 0 ]; then
+        renewToken "${mgmt_token}" &
```
lenny-intel, line 145: How is this running it as a different process?

snap/local/runtime-helpers/bin/renew_mgmt_token.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash -e
```
lenny-intel, line 1: Couldn't we reuse this script and call it from the consul wait script. Obviously the Docker file would have to copy it to the folder that is volume mounted.

cmd/security-bootstrapper/entrypoint-scripts/consul_wait_install.sh
jim-wang-intel: the "&" symbol make it run and fork as a background process and thus it is different from the main process we are running in this entry point.

snap/local/runtime-helpers/bin/renew_mgmt_token.sh
jim-wang-intel: never seen done this way before; i thought snap is a separate world by itself. we haven't done sharing a file between a snap and a docker

lenny-intel: This script is duplicate code for what you have in the consul wait script so the Docker file just needs to copy this file to the shared location next to the consul wait script.

cmd/security-bootstrapper/entrypoint-scripts/consul_wait_install.sh
lenny-intel: Ahh,  ok. did see that....

snap/local/runtime-helpers/bin/renew_mgmt_token.sh
jim-wang-intel: they are not exactly the same: 
- the snap is running as systemd service and thus I remove all the date-time logging information from the script eg. $(date)
- the renewToken in snap is not running as background process as it is running as a separate service here.

so we cannot just copy the exact file.

lenny-intel: But, will little refactoring it could be. It is 99% duplicate code.

jim-wang-intel: not sure how you can do refactoring?  remember every line of `echo $(date) .....` is different from the two.

lenny-intel: I am sure you can figure out a way ;-)

jim-wang-intel: it is kind of kludge solution refactoring either way IMO.  I think the better solution is to code this function in a security-bootstrapper subcommand and both snap and docker world can share. I'll rework on that to reduce the duplication concern you have.

jim-wang-intel:
This PR is no longer needed as the latest approach will be using dynamic generation of consul-token on-the-fly through each service's token.

PR #3408, opened 2021-04-28
feat(security): Add new environment variable to seed Secret Store with known secrets

lenny-intel:
## PR Checklist

No ability to add known secrets to services which have secret store added via the ADD_SECRETSTORE_TOKENS env var.

## What is the new behavior?
Secret Store Setup now has ADD_KNOWN_SECRETS environment variable to specify known secrets to add to the specified services` secret stores.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information

### DIFF
--- a/internal/security/secretstore/init.go
+++ b/internal/security/secretstore/init.go
@@ -59,17 +59,17 @@ const (
 	serviceListSeparator = ";"
 )
 
-var validKnownSecrets = map[string]bool{redisSecretName: true}
-
 type Bootstrap struct {
 	insecureSkipVerify bool
 	vaultInterval      int
+	validKnownSecrets  map[string]bool
 }
 
 func NewBootstrap(insecureSkipVerify bool, vaultInterval int) *Bootstrap {
 	return &Bootstrap{
 		insecureSkipVerify: insecureSkipVerify,
 		vaultInterval:      vaultInterval,
+		validKnownSecrets:  map[string]bool{redisSecretName: true},
 	}
 }
 
@@ -336,7 +336,7 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
 		os.Exit(1)
 	}
 
-	knownSecretsToAdd, err := getKnownSecretsToAdd()
+	knownSecretsToAdd, err := b.getKnownSecretsToAdd()
 	if err != nil {
 		lc.Error(err.Error())
 		os.Exit(1)
@@ -468,7 +468,7 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
 
 }
 
-func getKnownSecretsToAdd() (map[string][]string, error) {
+func (b *Bootstrap) getKnownSecretsToAdd() (map[string][]string, error) {
 	// Process the env var for adding known secrets to the specified services' secret stores.
 	// Format of the env var value is:
 	//   "<secretName>[<serviceName>;<serviceName>; ...], <secretName>[<serviceName>;<serviceName>; ...], ..."
@@ -494,7 +494,7 @@ func getKnownSecretsToAdd() (map[string][]string, error) {
 
 		secretName := strings.TrimSpace(secretItems[0])
 
-		_, valid := validKnownSecrets[secretName]
+		_, valid := b.validKnownSecrets[secretName]
 		if !valid {
 			return nil, fmt.Errorf(
 				"invalid specification for %s environment vaiable: '%s' is not a known secret",
--- a/internal/security/secretstore/init.go
+++ b/internal/security/secretstore/init.go
@@ -51,8 +51,12 @@ import (
 )
 
 const (
-	addKnownSecretsEnv = "ADD_KNOWN_SECRETS"
-	redisSecretName    = "redisdb"
+	addKnownSecretsEnv   = "ADD_KNOWN_SECRETS"
+	redisSecretName      = "redisdb"
+	knownSecretSeparator = ","
+	serviceListBegin     = "["
+	serviceListEnd       = "]"
+	serviceListSeparator = ";"
 )
 
 var validKnownSecrets = map[string]bool{redisSecretName: true}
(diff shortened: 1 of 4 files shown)


internal/security/secretstore/init.go
```diff
@@ -49,6 +50,13 @@ import (
 	"github.com/edgexfoundry/go-mod-secrets/v2/secrets"
 )
 
+const (
+	addKnownSecretsEnv = "ADD_KNOWN_SECRETS"
+	redisSecretName    = "redisdb"
+)
+
+var validKnownSecrets = map[string]bool{redisSecretName: true}
```
jim-wang-intel, line 58: please avoid using the global variable, put into calling object instance if you can

bnevis-i (approved):
Pretty comprehensive test suite... but... is it really necessary to put the known secrets into code?

internal/security/secretstore/init.go
```diff
@@ -440,6 +464,61 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
...
+	addKnownSecretsValue := strings.TrimSpace(os.Getenv(addKnownSecretsEnv))
+	if len(addKnownSecretsValue) == 0 {
+		return knownSecretsToAdd, nil
+	}
+
+	serviceNameRegx := regexp.MustCompile(ServiceNameValidationRegx)
+	knownSecrets := strings.Split(addKnownSecretsValue, ",")
+	for _, secretSpec := range knownSecrets {
+		// each secretSpec has format of "<secretName>[<serviceName>;<serviceName>; ...]"
+		secretItems := strings.Split(secretSpec, "[")
```
jim-wang-intel, line 482: please use a named constant for `[` to better readability

internal/security/secretstore/init.go
```diff
@@ -440,6 +464,61 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
...
+		if len(secretItems) != 2 {
+			return nil, fmt.Errorf("invalid specification for %s environment vaiable: Format of value '%s' is invalid. Missing or too many '['", addKnownSecretsEnv, secretSpec)
```
jim-wang-intel, line 484: replace `[` with the named constant

internal/security/secretstore/init.go
```diff
@@ -440,6 +464,61 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
...
+
+		secretName := strings.TrimSpace(secretItems[0])
+
+		_, valid := validKnownSecrets[secretName]
+		if !valid {
+			return nil, fmt.Errorf("invalid specification for %s environment vaiable: '%s' is not a known secret", addKnownSecretsEnv, secretName)
+		}
+
+		serviceNameList := secretItems[1]
+		if !strings.Contains(serviceNameList, "]") {
```
jim-wang-intel, line 495: please use a named constant for `]` to better readability

internal/security/secretstore/init.go
```diff
@@ -440,6 +464,61 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
...
+			return nil, fmt.Errorf("invalid specification for %s environment vaiable: Service list for '%s' missing closing ']'", addKnownSecretsEnv, secretName)
```
jim-wang-intel, line 496: replace `]` with the named constant

internal/security/secretstore/init.go
```diff
@@ -440,6 +464,61 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
...
+		}
+
+		serviceNameList = strings.TrimSpace(strings.Replace(serviceNameList, "]", "", 1))
```
jim-wang-intel, line 499: replace ']' with the named constant

internal/security/secretstore/init.go
```diff
@@ -440,6 +464,61 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
...
+		}
+
+		// This supports listing known secret multiple times.
+		// Same service name listed twice is not an issue since the add logic checks if the secret is already present.
+		existingServices := knownSecretsToAdd[secretName]
+		knownSecretsToAdd[secretName] = append(existingServices, serviceNames...)
+	}
+
+	return knownSecretsToAdd, nil
+}
```
jim-wang-intel, line 520: this logic seems too tedious to parse all of that. would a regular expression to parse out things better?

jim-wang-intel (changes requested):
I prefer not to have a global variable at all

internal/security/secretstore/init.go
lenny-intel: Not, for single occurance.

internal/security/secretstore/init.go
lenny-intel: Not, for single occurrence. Reads just fine.

internal/security/secretstore/init.go
lenny-intel: Ok, not for readability, but it is in multiple locations now with the error message.

internal/security/secretstore/init.go
lenny-intel: If you can come up with a regular expression for this, I am willing to give it a try, but since it is multi level, I think that will be all most impossible

internal/security/secretstore/init.go
lenny-intel: Technically this isn't a global variable since it is only known to the package, but since this service all in one package... it is a basically a global. ;-) 

Moved to the Bootstrap struct

internal/security/secretstore/init.go
lenny-intel: I tried a couple online regex generator with no luck. Need to move on.

jim-wang-intel: @lenny-intel      tried to play in go playground here: https://play.golang.org/p/HIGjo_dnYRS   getting closer. you can try it out there.

lenny-intel: The multi layer dynamic lists make it too complex. The output of what you have has to be parsed and has lost the context. This function work and isn't too tedious. Let move on.

jim-wang-intel: I think it works if we can split by "," first then each of slice shares the same regex pattern that I just have in that go playground

lenny-intel: Yes, it works on good value, but no error feedback on bad values.

jim-wang-intel: Ok, i'll call it more lenient to error but i'm ok with your point on more robust handling here. ;-)

jim-wang-intel (approved):

internal/security/secretstore/init.go
jim-wang-intel: nice! i like this much better

internal/security/secretstore/init_test.go
```diff
@@ -78,6 +79,52 @@ func TestSaveInitResponse(t *testing.T) {
 	fileOpener.AssertExpectations(t)
 }
 
+func TestGetKnownSecretsToAdd(t *testing.T) {
+	defer os.Clearenv()
```
jim-wang-intel, line 83: @lenny-intel  this screwed the other test relies on the env $PATH for "touch" so that is why the other unit test is broken

lenny-intel: Ahh, ok yup. That would do it. Thanks!

jim-wang-intel (approved):
LGTM

Issue #3403, opened 2021-04-27
[Kong Admin API] Snap: secure and enable Kong's admin APIs


jim-wang-intel:
# 🚀 Feature Request

### Relevant Package

 
This feature request is for `snap-building` to secure and enabling Kong's admin APIs in snap.

### Description
 
After `edgex-go`'s [securing Kong's Admin API issue](https://github.com/edgexfoundry/edgex-go/issues/2537) got resolved and [PR merged into master](https://github.com/edgexfoundry/edgex-go/pull/3328), snap also needs to update to reflect this capability.

### Describe the solution you'd like
 
1. Upgrade snap's Kong version (to be done by Tony) to 2.3.x as the current version of Kong in snap doesn't support the plugin template version 2.1
2. Modify the snap builds to utilize [the feature from @beaufrusetta 's PR](https://github.com/edgexfoundry/edgex-go/pull/3328)

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?

PR #3410, opened 2021-04-29
build(snap): Enable secure Kong Admin APIs in snap build

jim-wang-intel:
- Add installation of Kong's Admin plugins template yaml to enable Admin JWT, ACL plugins etc.
- Update the Kong daemon startup shell script to import the plugins into database
- Update the snapcraft yml to integrate the secure Kong's Admin API features

Closes: #3403

## PR Checklist

kong's admin APIs at 8001 are wide open without authentication and ACLs in snap

## What is the new behavior?
Kong's admin APIs are now protected with required JWT authentication with ACLs in snap

- [ ] Yes
- [x ] No

## New Imports

- [ ] Yes
- [x ] No

## Specific Instructions

In Draft PR mode until other dependent PRs for this snap to be working:
- Feature of [securing Kong's Admin APIs from this PR](https://github.com/edgexfoundry/edgex-go/pull/3328)  MERGED
- Kong's binary upgrade to new version 2.3.x PR (https://github.com/edgexfoundry/edgex-go/pull/3415)  MERGED

Now this PR is ready.

## Other information

### DIFF
--- a/snap/hooks/install
+++ b/snap/hooks/install
@@ -36,6 +36,12 @@ for service in security-file-token-provider security-proxy-setup security-secret
     fi
 done
 
+# install the template config yaml file for securing Kong's admin APIs in security-secretstore-setup service
+if [ ! -f "$SNAP_DATA/config/security-secretstore-setup/res/kong-admin-config.template.yml" ]; then
+    cp "$SNAP/config/security-secretstore-setup/res/kong-admin-config.template.yml" \
+        "$SNAP_DATA/config/security-secretstore-setup/res/kong-admin-config.template.yml"
+fi
+
 # handle app-service-configurable's various profiles:
 # 1. ensure all the directories from app-service-configurable exist
 # 2. copy the config files from $SNAP into $SNAP_DATA
--- a/snap/local/runtime-helpers/bin/kong-daemon.sh
+++ b/snap/local/runtime-helpers/bin/kong-daemon.sh
@@ -12,7 +12,7 @@ until "$KONG" migrations bootstrap --conf "$KONG_CONF"; do
 done
 
 # set up Kong's admin API plugins via kong.yml:
-"$KONG" config db_import "$KONGADMIN_CONFIGFILEPATH"
+"$KONG" config db_import "$KONGADMIN_CONFIGFILEPATH" || true
 
 # now start kong normally
 "$KONG" start --conf "$KONG_CONF" --v
--- a/snap/local/runtime-helpers/bin/kong-daemon.sh
+++ b/snap/local/runtime-helpers/bin/kong-daemon.sh
@@ -11,5 +11,8 @@ until "$KONG" migrations bootstrap --conf "$KONG_CONF"; do
     sleep 5
 done
 
+# set up Kong's admin API plugins via kong.yml:
+"$KONG" config db_import "$KONGADMIN_CONFIGFILEPATH"
+
 # now start kong normally
 "$KONG" start --conf "$KONG_CONF" --v
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -656,7 +656,9 @@ parts:
       done
 
       # For securing Kong's admin APIs, we also need to install kong-admin-config's template yml
-      # into secretstore-setup service
+      # into secretstore-setup service because secretstore-setup uses it to create a JWT file that
+      # proxy-setup can later on enable Kong's JWT plugin on admin API route hence secure it with
+      # JWT authentication and authorization
       install -DT "./cmd/security-secretstore-setup/res/kong-admin-config.template.yml" \
         "$SNAPCRAFT_PART_INSTALL/config/security-secretstore-setup/res/kong-admin-config.template.yml"
 
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -124,7 +124,15 @@ apps:
       KONG_PROXY_ERROR_LOG: $SNAP_COMMON/logs/kong-proxy-error.log
       KONG_ADMIN_ERROR_LOG: $SNAP_COMMON/logs/kong-admin-error.log
       KONG_ADMIN_LISTEN: "localhost:8001, localhost:8444 ssl"
-      KONG_STATUS_LISTEN: "0.0.0.0:8100"
+      KONG_STATUS_LISTEN: "localhost:8100"
+      # The DNS order was modified because of an issue with a Windows based host DNS issue.
+      # Keeping the default order would cause the first API request to Kong's admin API much longer time to connect for
+      # taking something minimum of 20 seconds to complete, and then others would complete as expected.
+      # Moreover, changing the DNS order did not impact those non-Windows based systems,
+      # so implementing it across the board seemed pretty low risk.
+      # The work-around that was found in one of the documented issues was to remove the SRV record check
+      # by manipulating the Kong DNS checking order via environment variable. Once we did that,
+      # the prolonged connecting issue went away.
       KONG_DNS_ORDER: "LAST,A,CNAME"
       LC_ALL: C.UTF-8
       LANG: C.UTF-8
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -108,6 +108,8 @@ apps:
     adapter: full
     after:
       - postgres
+      # to ensure kong.yml is ready for kong to process as the file is set up from security-secretstore-setup
+      - security-secretstore-setup
     command: bin/kong-daemon.sh
     command-chain:
       - bin/perl5lib-launch.sh
@@ -115,12 +117,15 @@ apps:
     daemon: forking
     environment:
       KONG_CONF: $SNAP_DATA/config/security-proxy-setup/kong.conf
+      KONGADMIN_CONFIGFILEPATH: $SNAP_DATA/config/kong.yml
       KONG_LOGS_DIR: $SNAP_COMMON/logs
       KONG_PROXY_ACCESS_LOG: $SNAP_COMMON/logs/kong-proxy-access.log
       KONG_ADMIN_ACCESS_LOG: $SNAP_COMMON/logs/kong-admin-access.log
       KONG_PROXY_ERROR_LOG: $SNAP_COMMON/logs/kong-proxy-error.log
       KONG_ADMIN_ERROR_LOG: $SNAP_COMMON/logs/kong-admin-error.log
       KONG_ADMIN_LISTEN: "localhost:8001, localhost:8444 ssl"
+      KONG_STATUS_LISTEN: "0.0.0.0:8100"
+      KONG_DNS_ORDER: "LAST,A,CNAME"
       LC_ALL: C.UTF-8
       LANG: C.UTF-8
     start-timeout: 15m
@@ -170,6 +175,11 @@ apps:
       TOKENFILEPROVIDER_CONFIGFILE: $SNAP_DATA/config/security-file-token-provider/res/token-config.json
       TOKENFILEPROVIDER_OUTPUTDIR: $SNAP_DATA/secrets
 
+      # securing Kong's admin API plugin related environment variables:
+      KONGADMIN_CONFIGTEMPLATEPATH: $SNAP_DATA/config/security-secretstore-setup/res/kong-admin-config.template.yml
+      KONGADMIN_CONFIGFILEPATH: $SNAP_DATA/config/kong.yml
+      KONGADMIN_CONFIGJWTPATH: $SNAP_DATA/secrets/edgex-security-proxy-setup/kong-admin-jwt
+
     start-timeout: 15m
     plugs: [network]
   security-proxy-setup:
@@ -183,6 +193,7 @@ apps:
       INIT_ARG: "--init=true"
       SECRETSTORE_TOKENFILE: $SNAP_DATA/secrets/edgex-security-proxy-setup/secrets-token.json
       ACCESSTOKENFILE: $SNAP_DATA/secrets/consul-acl-token/bootstrap_token.json
+      KONGAUTH_JWTFILE: $SNAP_DATA/secrets/edgex-security-proxy-setup/kong-admin-jwt
     daemon: oneshot
     start-timeout: 15m
     plugs: [network]
@@ -347,6 +358,7 @@ apps:
     environment:
       SECRETSTORE_TOKENFILE: $SNAP_DATA/secrets/edgex-security-proxy-setup/secrets-token.json
       ACCESSTOKENFILE: $SNAP_DATA/secrets/consul-acl-token/bootstrap_token.json
+      KONGAUTH_JWTFILE: $SNAP_DATA/secrets/edgex-security-proxy-setup/kong-admin-jwt
     plugs: [home, removable-media, network]
   secrets-config:
     adapter: none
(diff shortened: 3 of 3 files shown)


snap/hooks/install
```diff
@@ -36,6 +36,12 @@ for service in security-file-token-provider security-proxy-setup security-secret
     fi
 done
 
+# install the template config yaml file for securing Kong's admin APIs in security-secretstore-setup service
+if [ ! -f "$SNAP_DATA/config/security-secretstore-setup/res/kong-admin-config.template.yml" ]; then
+    cp "$SNAP/config/security-secretstore-setup/res/kong-admin-config.template.yml" \
+        "$SNAP_DATA/config/security-secretstore-setup/res/kong-admin-config.template.yml"
+fi
+
```
bnevis-i, line 44: How to update this file if it is being conditionally replaced?

snap/snapcraft.yaml
```diff
@@ -108,19 +108,24 @@ apps:
...
     environment:
       KONG_CONF: $SNAP_DATA/config/security-proxy-setup/kong.conf
+      KONGADMIN_CONFIGFILEPATH: $SNAP_DATA/config/kong.yml
       KONG_LOGS_DIR: $SNAP_COMMON/logs
       KONG_PROXY_ACCESS_LOG: $SNAP_COMMON/logs/kong-proxy-access.log
       KONG_ADMIN_ACCESS_LOG: $SNAP_COMMON/logs/kong-admin-access.log
       KONG_PROXY_ERROR_LOG: $SNAP_COMMON/logs/kong-proxy-error.log
       KONG_ADMIN_ERROR_LOG: $SNAP_COMMON/logs/kong-admin-error.log
       KONG_ADMIN_LISTEN: "localhost:8001, localhost:8444 ssl"
+      KONG_STATUS_LISTEN: "0.0.0.0:8100"
```
bnevis-i, line 127: Can you listen on `127.0.0.1:8100` instead?  Snaps share the host's network stack and this exposes the port to the world.

snap/snapcraft.yaml
```diff
@@ -640,6 +652,11 @@ parts:
 
       done
 
+      # For securing Kong's admin APIs, we also need to install kong-admin-config's template yml
+      # into secretstore-setup service
```
bnevis-i, line 656: Comment doesn't fully explain what is going on.

snap/snapcraft.yaml
```diff
@@ -670,16 +687,20 @@ parts:
...
-          FILE_HASH=450223a011dba0d6eeb6a192be368e7bcd45a5dc063c743ed6f3d37eb95237b8
+          # TODO: this is a temporary fix for the broken file sha from the bintray downloads
+          # there will be another PR to fix this and upgrade the version of Kong
+          FILE_NAME=kong-2.3.3.xenial.amd64.deb
+          FILE_HASH=0909891d483b0894460bb7f7b69381abaf6b4ebf2c7d13c7076a88087f739294
           ;;
         arm64)
           FILE_NAME=kong-2.0.5.xenial.arm64.deb
           FILE_HASH=15aa5fb2eb0c992afa277b09cb03b3b940df9704ab9e8d93327a72c0cb87eb09
           ;;
```
bnevis-i, line 698: Just to point out, arm64 hasn't been fixed.

bnevis-i (changes requested):

snap/hooks/install
jim-wang-intel: not sure what is conditionally replaced? I don't think we have that capability yet.

snap/snapcraft.yaml
jim-wang-intel: i think most of cases would, not sure about the corner case of "remote" configuration for consul which may need to listen not just `localhost`, correct me if I am wrong.

snap/snapcraft.yaml
jim-wang-intel: not sure what you want me to add here: it is just try to copy that template yml to the proper location.

snap/snapcraft.yaml
jim-wang-intel: yea, my local machine is amd64 so I won;'t be able to touch it and test it here- will leave that to @siggiskulason's PR  on upgrading to Kong's version to fix that.

snap/hooks/install
jim-wang-intel: please describe and elaborate the use case you are thinking for "conditionally replaced"

snap/snapcraft.yaml
bnevis-i: It doesn't say WHY secretstore-setup needs this template.

snap/hooks/install
bnevis-i: The `cp` command is skipped if `$SNAP_DATA/config/security-secretstore-setup/res/kong-admin-config.template.yml` already exists.  If it is a template, why not overwrite it every time?

snap/snapcraft.yaml
bnevis-i: Unfortunately, you won't be able to pass the pre-commit hook as it is.

snap/hooks/install
jim-wang-intel: oh, that, i just refer to the same logic above when dealing withe the toml configuration file, too. see line 14 above. just to keep it consistent with that.

snap/snapcraft.yaml
jim-wang-intel: got you, I'll add that, thanks.

snap/snapcraft.yaml
jim-wang-intel: right- i'll rebase his PR once that is there.

snap/snapcraft.yaml
lenny-intel: Kong is our gateway, so it needs to be exposed to the world

lenny-intel (commented):
LGTM, but will leave to @tonyespy / @siggiskulason for final approval.

snap/snapcraft.yaml
jim-wang-intel: Rebased

jim-wang-intel:
rebased to pull in the latest changes about the kong's binary image downloading

snap/snapcraft.yaml
jim-wang-intel: yea, that is what I think it should be as well

snap/snapcraft.yaml
jim-wang-intel: added

tonyespy:
@jim-wang-intel just a quick note that we started seeing failures in our automated github-action based tests with the snap install failing because of security-proxy-setup failing.

I just built your PR and as expected, it resolves this problem.

I'll review the actual code changes later today, but wanted to let you know that I've started looking at. Thanks for your changes!

jim-wang-intel:
> @jim-wang-intel just a quick note that we started seeing failures in our automated github-action based tests with the snap install failing because of security-proxy-setup failing.
> 
> I just built your PR and as expected, it resolves this problem.
> 
> I'll review the actual code changes later today, but wanted to let you know that I've started looking at. Thanks for your changes!

@tonyespy:  Sure, thanks for the notes!

snap/snapcraft.yaml
tonyespy: I'm going to ask the same question as @bnevis-i, why does this port (which seems to be a read-only status API) need to be accessible from a different host? Are we actively using this status API in the new security model for the admin API?

tonyespy (commented):
@jim-wang-intel for the most part this PR looks good, although there are a few comments/questions inline that I'd like to see answers for...

Also I built and installed a snap from this PR and it works as expected, however I did notice one error logged by security-proxy-setup after a reboot that caused the command to fail:

```
Starting Service for snap application edgexfoundry.security-proxy-setup...
INFO app=edgex-security-proxy-setup source=config.go:359 msg="Loaded service configuration from /var/sn\
ap/edgexfound
INFO  app=edgex-security-proxy-setup source=variables.go:357 msg="Variables override of 'SecretStore.To\
kenFile' by e
INFO app=edgex-security-proxy-setup source=variables.go:357 msg="Variables override of 'KongAuth.JWTFi\
le' by enviro
INFO app=edgex-security-proxy-setup source=variables.go:357 msg="Variables override of 'AccessTokenFil\
e' by environ
INFO app=edgex-security-proxy-setup source=secret.go:49 msg="Creating SecretClient"
INFO app=edgex-security-proxy-setup source=secret.go:56 msg="Reading secret store configuration and authentication
INFO  app=edgex-security-proxy-setup source=secret.go:68 msg="Attempting to create secret client"
INFO  app=edgex-security-proxy-setup source=secret.go:73 msg="Created SecretClient"
INFO app=edgex-security-proxy-setup source=config.go:494 msg="Using local configuration from file (3 envVars overri
INFO app=edgex-security-proxy-setup source=secrets.go:322 msg="kick off token renewal with interval: 30m0s"
INFO  app=edgex-security-proxy-setup source=service.go:103 msg="the service on http://127.0.0.1:8100/status is up su
ERROR  app=edgex-security-proxy-setup source=service.go:361 msg="proxy service for scheduler returned status 401"
ERROR app=edgex-security-proxy-setup source=init.go:53 msg="proxy service for scheduler returned status 401"
snap.edgexfoundry.security-proxy-setup.service: Main process exited, code=exited, status=1/FAILURE
snap.edgexfoundry.security-proxy-setup.service: Failed with result 'exit-code'.
Failed to start Service for snap application edgexfoundry.security-proxy-setup.
```

snap/local/runtime-helpers/bin/kong-daemon.sh
```diff
@@ -11,5 +11,8 @@ until "$KONG" migrations bootstrap --conf "$KONG_CONF"; do
     sleep 5
 done
 
+# set up Kong's admin API plugins via kong.yml:
+"$KONG" config db_import "$KONGADMIN_CONFIGFILEPATH"
```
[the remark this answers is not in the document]
jim-wang-intel: it should, but i think i can add `|| true` at the end of this command just in case then

snap/snapcraft.yaml
```diff
@@ -108,19 +108,24 @@ apps:
...
       KONG_CONF: $SNAP_DATA/config/security-proxy-setup/kong.conf
+      KONGADMIN_CONFIGFILEPATH: $SNAP_DATA/config/kong.yml
       KONG_LOGS_DIR: $SNAP_COMMON/logs
       KONG_PROXY_ACCESS_LOG: $SNAP_COMMON/logs/kong-proxy-access.log
       KONG_ADMIN_ACCESS_LOG: $SNAP_COMMON/logs/kong-admin-access.log
       KONG_PROXY_ERROR_LOG: $SNAP_COMMON/logs/kong-proxy-error.log
       KONG_ADMIN_ERROR_LOG: $SNAP_COMMON/logs/kong-admin-error.log
       KONG_ADMIN_LISTEN: "localhost:8001, localhost:8444 ssl"
+      KONG_STATUS_LISTEN: "0.0.0.0:8100"
+      KONG_DNS_ORDER: "LAST,A,CNAME"
```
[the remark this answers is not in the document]
jim-wang-intel: not sure about the order, i just copied whatever has done in the compose file to the snap here.  @beaufrusetta maybe you can help to answer this better?

snap/local/runtime-helpers/bin/kong-daemon.sh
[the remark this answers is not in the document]
beaufrusetta: The "kong config db_import <file>" command does not return a clean exit code even though it successfully imports. It returns like "6" or something like that - the LUA code behind the scenes is where I found that. 

To get around that, so that the script will continue, I added the `|| true` to the end of the command, and it carries through gracefully. If the import fails, it will continue as if it were good...however, if the import does fail, there are going to be failures immediately down the line when proxy-setup attempts to start - so it doesn't concern me that much because if it does a fail/continue/fail because it won't end up in an exposed middle state - it will fail to setup the Admin API completely and proxy-setup won't be able to complete successfully.

snap/snapcraft.yaml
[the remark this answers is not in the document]
beaufrusetta: Port 8100 doesn't need to be exposed externally (that is my fault) - just needs to be available to the local container/snap network in order for proxy-setup to verify that it is up and running so that it can continue setting up services/routes (proxy-setup was previously pinging the admin port to see if it was up, when it should have been using the status port all along). I have not tested binding it to the local Docker network, however, it **_should_** work. We should probably open up a different issue and I can quickly modify/fix that.

The DNS order was modified because of an issue with a Windows based host DNS issue. Keeping the default order would cause the first API request to the admin API to take min of 20 secs to complete, and then others would complete as expected. Over a certain amount of time, that would then reset, and the next API request coming in would exhibit the same behavior. Changing the order did not impact non-Windows based systems, so implementing it across the board seemed pretty low risk.

There are documented cases of this happening in other projects that utilize Kong containers on Windows hosts, and I verified that using my various machines. The work-around that was found in one of the documented issues was to remove the SRV record check by manipulating the Kong DNS checking order via environment variable. Once we did that, the issue went away.

(Thanks to @rsdmike for helping to expose and find a solution for that DNS issue)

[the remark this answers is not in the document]
beaufrusetta: Actually - in the latest version of the `edgex-compose` code, I bound 8100 to the local network through the `ports` directive in the compose file. Setting `KONG_STATUS_LISTEN` is still set to 0.0.0.0:8100, however, Docker compose cuts off the ports access when it creates the container.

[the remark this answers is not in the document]
jim-wang-intel: thanks, @beaufrusetta . I"ll  try to test 8100 port on snap locally here to see if works.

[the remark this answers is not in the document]
tonyespy: I just updated KONG_STATUS_LISTEN to use localhost in the snap and everything comes up just fine.

That said, I'm still seeing the 401 errors after a reboot. The failing service always seems to be different, sometimes its core-data, sometimes -metadata, -scheduler, ... The only consistent is it always seems to be the first that fails. 

I tried adding a 120s sleep before security-proxy-setup starts to rule out a race condition between security-secretstore-setup, and I still get the same failures.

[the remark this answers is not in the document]
jim-wang-intel: > I just updated `KONG_STATUS_LISTEN` to use localhost in the snap and everything comes up just fine.
> 
> That said, I'm still seeing the 401 errors after a reboot. The failing service always seems to be different, sometimes its core-data, sometimes -metadata, -scheduler, ... The only consistent is it always seems to be the first that fails.
> 
> I tried adding a 120s sleep before security-proxy-setup starts to rule out a race condition between security-secretstore-setup, and I still get the same failures.

yea, I also observed the same thing- it is really weird issue in snap. However, when I tried `snap stop` services first right after `snap install`, the 401 error seems gone away. So this is kind of unique in snap as I didn't see anything like this in docker-compose environment.   @tonyespy  any thoughts?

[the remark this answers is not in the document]
hutchic: security wg meeting discussion: we believe this is caused by running [initKongService](https://github.com/edgexfoundry/edgex-taf/issues/403#issuecomment-839338389) on an already initialized Kong

As a way to validate this is the error we could try querying the Kong admin api or run a kong config db_export maybe.

Are we erroniously starting proxy setup?

snap/snapcraft.yaml
jim-wang-intel: changed to bind to localhost only

snap/local/runtime-helpers/bin/kong-daemon.sh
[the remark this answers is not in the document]
jim-wang-intel: added `|| true`

snap/snapcraft.yaml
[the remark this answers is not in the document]
jim-wang-intel: added comments about `KONG_DNS_ORDER` based on @beaufrusetta 's feedback- thanks to @beaufrusetta !

tonyespy (dismissed):
Based upon our discussion at this week's security WG meeting, I'm approving this PR.

We'll continue investigation of the 401 on restart as a part of the effort to ensure robustness of security services across restarts (@lenny-intel do we have an issue to track this work for Ireland).

**Note** - the branch is out-of-sync with master, so needs a rebase before we merge.

beaufrusetta (approved):
LGTM

jim-wang-intel:
recheck

hutchic (approved):
looks sensible

jim-wang-intel:
> looks sensible

Thanks!

PR #3541, opened 2021-06-03
fix: Don't regenerate DB passwords if already exists & tweaks for Kuiper config

lenny-intel:
## PR Checklist

When Secret Store Setup is run it regenerates the DB password and since it is already in Vault it isn't used by Redis.
The new password is used for any new services and written to the Kuiper configuration.

## What is the new behavior?
password is not regenerated is it already exist in Vault.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information

### DIFF
--- a/internal/security/secretstore/init.go
+++ b/internal/security/secretstore/init.go
@@ -60,6 +60,8 @@ const (
 	secretBasePath       = "/v1/secret/edgex"
 )
 
+var errNotFound = errors.New("credential NOT found")
+
 type Bootstrap struct {
 	insecureSkipVerify bool
 	vaultInterval      int
--- a/internal/security/secretstore/init.go
+++ b/internal/security/secretstore/init.go
@@ -57,6 +57,7 @@ const (
 	serviceListBegin     = "["
 	serviceListEnd       = "]"
 	serviceListSeparator = ";"
+	secretBasePath       = "/v1/secret/edgex"
 )
 
 type Bootstrap struct {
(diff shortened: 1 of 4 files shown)


internal/security/secretstore/init.go
```diff
@@ -360,14 +361,26 @@ func (b *Bootstrap) BootstrapHandler(ctx context.Context, _ *sync.WaitGroup, _ s
...
-	redis5Password, err := cred.GeneratePassword(ctx)
+	redis5Pair, err := getDBCredential("security-bootstrapper-redis", cred, "redisdb")
 	if err != nil {
-		lc.Error("failed to generate redis5 password")
-		os.Exit(1)
-	}
-	redis5Pair := UserPasswordPair{
-		User:     "redis5",
-		Password: redis5Password,
+		if err != errNotFound {
```
jim-wang-intel, line 366: this returned error message is actually misleading. we should change the message of it. It is currently used for certificate pair not found or not uploaded.

jim-wang-intel (changes requested):
could you please change the message of errNotFound?  currently the error message is misleading and misused, I think.

internal/security/secretstore/init.go
lenny-intel: This isn't a message. It is used to determine if the `not found error` was returned from Vault. If not found we generate it. Otherwise use it.

bnevis-i (dismissed):

jim-wang-intel (approved):
LGTM

lenny-intel:
recheck

PR #3570, opened 2021-06-15
feat(sma): add 'edgex-' prefix for executor-based API

### LINKED ISSUE #3559: Add code to statically add "edgex-" prefix to service keys for purposes of SMA
# 🚀 Feature Request
Per conversation in #3501, with the change in Docker image names, the service key names no longer match the Docker container names.
It was decided in Core WG of 6/10/21, to just statically prepend `edgex-` to the service key names in the SMA code in order to allow the SMA (which relies on the executor) to refer to the service by its correct container name.
### Relevant Package
SMA

We discussed that in the future this should be handled with more flexibility and without hard-coding the prefix.

hahattan:

## PR Checklist

For API that relies on executor, user needs to refer to the service by its container name:

```
curl http://localhost:58890/api/v2/system/metrics?services=edgex-core-data,edgex-core-metadata
207

...
```

## What is the new behavior?
`edgex-` prefix will be automatically added for default service (those defined in edgex-compose file)
So user can either refer to those services by their service name (e.g. core-data) or container name (e.g. edgex-core-data)

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information

### DIFF
--- a/internal/system/agent/application/executor/metrics.go
+++ b/internal/system/agent/application/executor/metrics.go
@@ -45,7 +45,7 @@ func (m *metrics) Get(_ context.Context, services []string) ([]interface{}, erro
 		go func(serviceName string) {
 			defer wg.Done()
 
-			res, err := m.executor(m.executorPath, serviceName, "metrics")
+			res, err := m.executor(m.executorPath, edgexPrefix+serviceName, "metrics")
 			if err != nil {
 				mu.Lock()
 				responses = append(responses, common.BaseWithMetricsResponse{
--- a/internal/system/agent/application/executor/executor.go
+++ b/internal/system/agent/application/executor/executor.go
@@ -16,6 +16,8 @@ package executor
 
 import "os/exec"
 
+const edgexPrefix = "edgex-"
+
 // CommandExecutor provides the common callout to the configuration-defined executor.
 func CommandExecutor(executorPath, serviceName, operation string) (string, error) {
 	bytes, err := exec.Command(executorPath, serviceName, operation).CombinedOutput()
--- a/internal/system/agent/application/executor/operation.go
+++ b/internal/system/agent/application/executor/operation.go
@@ -46,7 +46,7 @@ func (o *operation) Do(_ context.Context, operations []requests.OperationRequest
 			defer wg.Done()
 
 			o.lc.Debugf("Executing '%s' action on %s", operation.Action, operation.ServiceName)
-			res, err := o.executor(o.executorPath, operation.ServiceName, operation.Action)
+			res, err := o.executor(o.executorPath, edgexPrefix+operation.ServiceName, operation.Action)
 			if err != nil {
 				mu.Lock()
 				responses = append(responses, common.BaseWithServiceNameResponse{
--- a/openapi/v2/system-agent.yaml
+++ b/openapi/v2/system-agent.yaml
@@ -53,6 +53,7 @@ components:
         servieName:
           description: "A field indicates the original requested serviceName"
           type: string
+          example: "core-data"
     BaseWithMetricsResponse:
       allOf:
         - $ref: '#/components/schemas/BaseResponse'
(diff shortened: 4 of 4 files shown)


cloudxxx8 (changes requested):
please also modify the example in the Swagger file

hahattan:
Swagger is using service name not container name right now.

cloudxxx8:
currently, the service name in the example is "string", not a real example.
https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/system-agent/2.x#/default/get_system_config__services_

cloudxxx8 (approved):
LGTM

internal/system/agent/application/executor/metrics.go
```diff
@@ -45,7 +45,7 @@ func (m *metrics) Get(_ context.Context, services []string) ([]interface{}, erro
 		go func(serviceName string) {
 			defer wg.Done()
 
-			res, err := m.executor(m.executorPath, serviceName, "metrics")
+			res, err := m.executor(m.executorPath, addEdgexPrefix(serviceName), "metrics")
```
lenny-intel, line 48: Do we expect these service to just be the `edgex` service or can they be 3rd party service like `redis` and `kong` . If includes 3rd party, we will need to update the docker compose to name all 3rd party containers with `edgex-` prefix. Currently it is inconsistent for 3rd party container names. We have `edgex-redis`, `edgex-core-consul`, `edgex-vault` and then have `kong` and `kong-db`  and then `edgex-app-service-configurable-rules` which doesn't follow the service key approach....

lenny-intel: Added these issues:
https://github.com/edgexfoundry/edgex-compose/issues/126
https://github.com/edgexfoundry/edgex-taf/issues/464

internal/system/agent/application/executor/util.go
```diff
@@ -0,0 +1,26 @@
...
+	switch serviceName {
+	case common.CoreDataServiceKey:
+	case common.CoreCommandServiceKey:
+	case common.CoreMetaDataServiceKey:
+	case common.SupportNotificationsServiceKey:
+	case common.SupportSchedulerServiceKey:
+	case common.SystemManagementAgentServiceKey:
+	case common.SecurityBootstrapperKey:
+	default:
+		return serviceName
```
lenny-intel, line 22: This doesn't solve the inconsistency issue and doesn't work for any device or apps services.

lenny-intel (changes requested):

internal/system/agent/application/executor/metrics.go
hahattan: > Do we expect these service to just be the `edgex` service or can they be 3rd party service like `redis` and `kong` . If includes 3rd party, we will need to update the docker compose to name all 3rd party containers with `edgex-` prefix. Currently it is inconsistent for 3rd party container names. We have `edgex-redis`, `edgex-core-consul`, `edgex-vault` and then have `kong` and `kong-db` and then `edgex-app-service-configurable-rules` which doesn't follow the service key approach....

Yep we need to make a decision here (or maybe discuss in coreWG?) 
Although I kinda want this to be a more general API (you can operate 3rd party services, or even custom one), considering the naming inconsistent it seems impossible.
I'll be in favor to make this edgex service only. For device services we can directly add `edgex-` prefix. We might need to update app service and security services in either core-contracts or edgex-compose.

lenny-intel: > considering the naming inconsistent it seems impossible.

Since we have total control in the compose files we can fix the naming inconsistency.

lenny-intel: @hahattan, container/hostnames have been updated so all services running in docker have the `edgex-`, so your code can assume to just add the prefix to any service name it receives.

internal/system/agent/application/executor/util.go
lenny-intel: Just add the prefix no matter what

hahattan: updated.

lenny-intel (approved):
LGTM

PR #3617, opened 2021-07-11
fix(notifications): Remove warning from 32-bit build.

mutterc:
Building the core services on 32-bit ARM, I got a warning "Constant
4294967295 is bigger than 'int'", which stopped the build (my environment
may be too pedantic). This resolves it.

Signed-off-by: Corey Mutter <corey@friendlyandhelpful.com>

## PR Checklist

Did not open an issue for this, found it locally.

- [ ] Yes
- [ X ] No

## New Imports

- [ ] Yes
- [ X ] No

## Specific Instructions
I don't know if this is the best fix - maybe it would be better to change the function's prototype to take unsigned int?
New to both EdgeX and Golang.
Don't know how one would update tests or docs for this.

## Other information


cloudxxx8 (approved):
LGTM

PR #3652, opened 2021-08-06
revcover panic when http status is zero.(may be a tcp connection cann…

wangshihui:
fix(all): http response cannot be completed

revcover panic when http status is zero.(may cased by  tcp connection cannot be established)

http.go is a util tool  for all core module.

PR #3662, opened 2021-08-09
fix(all): http response cannot be completed

wangshihui:
revcover panic when http status is zero.(may cased by tcp connection cannot be established)

http.go a tool for all core module.

Signed-off-by: shihui.wang <shihui.wang@thundersoft.com>

## PR Checklist

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information


wangshihui:
fix the pre pr issue
#3652 (comment)

cloudxxx8:
by the way, we fixed the possibility to cause the status code being zero in the following PR:
https://github.com/edgexfoundry/go-mod-core-contracts/pull/646/files

cloudxxx8 (approved):
LGTM

cloudxxx8:
Please open an issue to describe the problem before PR creation next time.  Thanks.

wangshihui:
> by the way, we fixed the possibility to cause the status code being zero in the following PR:
> https://github.com/edgexfoundry/go-mod-core-contracts/pull/646/files
this is a better solution

> Please open an issue to describe the problem before PR creation next time. Thanks.
ok got it

cloudxxx8 (changes requested):
Since there is a specific status code range, we should just use `if` statement to check `statusCode` instead of recover() func.
Also, I thought 500 internal server error is better than 503 service unavailable, because the error is unkown.

```
if statusCode < 100 || statusCode > 900 {
  w.WriteHeader(http.StatusInternalServerError)
}
```

wangshihui:
> Since there is a specific status code range, we should just use `if` statement to check `statusCode` instead of recover() func.
> Also, I thought 500 internal server error is better than 503 service unavailable, because the error is unkown.
> 
> ```
> if statusCode < 100 || statusCode > 900 {
>   w.WriteHeader(http.StatusInternalServerError)
> }
> ```
ok  push new commit as above.

BTW
As this logic is from `http.checkWriteHeaderCode` a private func. so didnot pick  it at the first commit.  
and this PR https://github.com/edgexfoundry/go-mod-core-contracts/pull/646/files  makes sure the statusCode  is a correct http status code. 
so this PR is only  as a guarantee  in case of unkonw use case.

cloudxxx8 (approved):
LGTM

PR #3678, opened 2021-08-20
feat(security): Feature to enable CORS headers for API gateway

bnevis-i:
Signed-off-by: Bryon Nevis <bryon.nevis@intel.com>

## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**If your build fails** due to your commit message not passing the build checks, please review the guidelines here: https://github.com/edgexfoundry/edgex-go/blob/master/.github/Contributing.md.

## What is the current behavior?
Current Kong proxy does not return CORS headers to clients.

## Issue Number:
Closes #1913

## What is the new behavior?
New configuration has been added to allow settings of a limited number of CORS headers for the EdgeX routes.

## Does this PR introduce a breaking change?

- [ ] Yes
- [X] No

## New Imports

- [ ] Yes
- [X] No

## Specific Instructions
To test, modify `cmd/security-proxy-setup/res/configuration.toml` and set `EnableCORS` to `true`.  Customize other headers as needed.  Rebuild security-proxy-setup docker container (make docker_security_proxy_setup).  Restart framework.

Test with a command like
```
curl -ki -X OPTIONS -H "Origin: http://localhost" -H "Access-Control-Request-Method: POST" -H "Access-Control-Request-Headers: Content-Type" https://localhost:8443/core-data/api/v2/ping
```

Expected response:
```
HTTP/2 200 
date: Fri, 20 Aug 2021 02:18:02 GMT
vary: Origin
access-control-allow-origin: https://localhost
access-control-allow-headers: Content-Type
access-control-allow-methods: GET,POST,PUT,DELETE
access-control-max-age: 3600
content-length: 0
x-kong-response-latency: 1
server: kong/2.5.0
```

## Other information

Issue #1913, opened 2019-10-15
Enable CORS access to EdgeX microservices


magallardo:
Hello,

# 🚀 Feature Request

### Relevant Package
This feature request is for the API Gateway

### Description
The Edgex components are exposed through REST APIs and those services are usually called from a UI running on a browser.  Most browser prevent calls to services at a different origin (CORS) thus making it difficult the interactions with services like Edgex.
There are basically 2 ways to solve the CORS issue.  The first one is on the client side (UI application) by using a proxy server.  Unfortunately this solution is static and need to know before hand the back end servers.
The second option is to enable the backend server to allow CORS.

### Describe the solution you'd like
Currently I have been bypassing this issue by using a proxy server at the client site.  However, this solution is not working due to the nature of IoT projects and solutions.  The IoT solutions tend to be very dynamic and in my specific example, the number of instances where the Edgex framework is deployed is dynamic and can change.  That makes the proxy option irrelevant as that solution is static.

So, the preferred solution will be to allow the user to enable/disable CORS as desired on the API Gateway.  That will allow calling the API from UI clients without requiring to use a proxy.

### Describe alternatives you've considered
As described above, I am already using a proxy solution but it only works for a static number of Edgex installations.

Thanks in advance,
Marcelo

tingyuz:
For future reference, the current implementation of API gateway is based on KONG. From the document it seems KONG has limited support to enable CORS/cross-origin resource sharing. 

https://docs.konghq.com/hub/kong-inc/cors/

tingyuz:
A quick verification/band aid style fix is to run Curl command against KONG.  Here is the command that may work: 
```
curl -X POST http://localhost:8001/plugins --data "name=cors" --data "config.origins=*"  --data "config.methods=GET"  --data "config.methods=POST" --data "config.headers=Accept" --data "config.headers=Accept-Version" --data "config.headers=Content-Length" --data "config.headers=Content-MD5" --data "config.headers=Content-Type" --data "config.headers=Date" --data "config.headers=X-Auth-Token" --data "config.exposed_headers=X-Auth-Token" --data "config.credentials=true" --data "config.max_age=3600"
```

AJ-N:
If anyone else is looking for a solution to this, simply using `*` for origins did not work for me. Only when restricted to specific origins (with proper `http` or `https` AND port) did it work (i.e. `config.origins=https://localhost:3000`)

[...]

PR #3711, opened 2021-09-16
feat(security): Revert CORS changes for API Gateway

bnevis-i:
Reverts #3678 due to strategy change.  See #1913.

It was decided to move CORS functionality to the
microservice level so that it can be applied both
in the security-enabled as well as security-disabled
states.

This functionality was never implemented in Ireland,
thus implementing and reverting it in Jakarta is
not a breaking change.

@magallardo


lenny-intel (approved):
LGTM

Issue #3565, opened 2021-06-11
Feature request: add a "make lint" target


bnevis-i:
# 🚀 Feature Request

### Relevant Package
Global

### Description

Suggest adding a "make lint" target to the top-level Makefile to point out code quality and readability issues.

This will help enforce common coding practices in the project.

```
golangci-lint run
golangci-lint run -E bodyclose
golangci-lint run -E cyclop
golangci-lint run -E dupl
golangci-lint run -E gochecknoglobals
golangci-lint run -E gocognit
golangci-lint run -E misspell
golangci-lint run -E nilerr
golangci-lint run -E unparam
golangci-lint run -E unconvert
```

Also trial enable gosec and disposition results.

Enable the following linters:
- [x] deadcode
- [x] errcheck
- [x] gosimple
- [x] ineffassign
- [x] staticcheck
- [x] unused
- [x] varcheck
- [x] gosec

bnevis-i:
Also have suggestion from Colin to look at <https://github.com/github/super-linter>

PR #3735, opened 2021-09-29
style: Enable staticcheck linter and associated recommendations

bnevis-i:
See #3565

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?)  Not adding functionality
- [ ] I have fully tested (add details below) this the new feature or bug fix (if not, why?)  Make build test is sufficient
- [ ] I have opened a PR for the related docs change (if not, why?)  Not applicable

## Testing Instructions
make and make test

### DIFF
--- a/internal/pkg/correlation/middleware.go
+++ b/internal/pkg/correlation/middleware.go
@@ -18,12 +18,12 @@ func ManageHeader(next http.Handler) http.Handler {
 			correlationID = uuid.New().String()
 		}
 		// lint:ignore SA1029 legacy
-		// nolint:staticcheck // SA1004 legacy
+		// nolint:staticcheck // See golangci-lint #741
 		ctx := context.WithValue(r.Context(), common.CorrelationHeader, correlationID)
 
 		contentType := r.Header.Get(common.ContentType)
 		// lint:ignore SA1029 legacy
-		// nolint:staticcheck // SA1004 legacy
+		// nolint:staticcheck // See golangci-lint #741
 		ctx = context.WithValue(ctx, common.ContentType, contentType)
 
 		r = r.WithContext(ctx)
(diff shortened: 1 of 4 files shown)


internal/pkg/correlation/middleware.go
```diff
@@ -17,9 +17,13 @@ func ManageHeader(next http.Handler) http.Handler {
 		if correlationID == "" {
 			correlationID = uuid.New().String()
 		}
+		// lint:ignore SA1029 legacy
+		// nolint:staticcheck // SA1004 legacy
```
lenny-intel, line 21: Intersting that it doen't think you should use a key of type string when the example in GoDoc use string. :-)
https://pkg.go.dev/context#example-WithValue
But they do create a new type to obscure that it is a string. Humm, bet this keeps the linter happy.

SA1004 doesn't make sense here: https://staticcheck.io/docs/checks#SA1004
Does `nolint:staticcheck` disable the `staticcheck` for this line? if so, do we really need `lint:ignore SA1029 legacy` also

bnevis-i: It just disables staticcheck for this line

I fixed this to read //nolint:staticcheck // See golangci-lint #741

Its a known issue https://github.com/golangci/golangci-lint/issues/741

lenny-intel (approved):
LGTM

jim-wang-intel (approved):
LGTM

internal/pkg/correlation/middleware.go
bnevis-i: > Intersting that it doen't think you should use a key of type string when the example in GoDoc use string. :-) https://pkg.go.dev/context#example-WithValue But they do create a new type to obscure that it is a string. Humm, bet this keeps the linter happy.

It's more than that.  If you have a web server with a very deep HTTP pipeline, you can get collisions when people stuff things like "id" = "1234" into the context because "id" for one module is not the same as "id" for another module.  By wrapping it in a typed struct, you mostly eliminate the possibility for the collision because the type is going to be scoped by the package so even if the field is the same, across packages the full type name of the field will be different.

bnevis-i: I think it's ok if we just grandfather this one in.  We only have two.

lenny-intel (approved):

Issue #3712, opened 2021-09-16
Consul cannot be freely switched between secure and insecure modes


bnevis-i:
# 🐞 Bug Report

### Affected Services
EdgeX Registry service
EdgeX Configuration service

### Is this a regression?
Depends on your perspective.

### Description and Minimal Reproduction
In Ireland release, we hardened access to Consul, the component that provides Registry and Configuration services to EdgeX microservices.  It used to be possible to start EdgeX with security disabled or security enabled, depending on the user's choice.  

Now, when booting EdgeX in secure mode, Consul ACL support is enabled.  It has been determined through experimentation that once ACL support has been enabled, it is not possible to restart the framework in insecure mode without breaking client's access to the registry and configuration.

cc: @tonyespy please edit description to add more detail.

bnevis-i:
@jim-wang-intel @lenny-intel 

I am having difficulty reproducing this issue.  Test case:

- Start the framework in secure mode (docker)
- Run docker exec -ti edgex-core-consul consul kv get edgex/appservices/2.0/app-rules-engine/SecretStore/Protocol (403 error)
- Edit docker-compose files, make /edgex-init writable, change acl enabled mode to false, docker stop edge-core-consul, make run (dev) to restart services
- Run docker exec -ti edgex-core-consul consul kv get edgex/appservices/2.0/app-rules-engine/SecretStore/Protocol (result: http)

Clearly, Consul is respecting the ACL enabled flag and it CAN be changed.

jim-wang-intel:
@bnevis-i yes, i couldn't re-produce it in Docker compose environment either.  I think it is only having problem for @tonyespy while running in snap environment or settings.

tonyespy:
As requested by the Security WG, I’ve tested editing the `consul_acl.json` file (found in `$SNAP_DATA/consul/config`) to set `acl.enabled=false` to see if this will disable ACL checking in Consul (after a restart), and unfortunately the answer seems to be no.

Here are the steps I’ve tried (`SNAP_DATA=/var/snap/edgexfoundry/current`):

- `sudo snap install edgexfoundry --channel=2.0 `
- `sudo snap set edgexfoundry security-secret-store=off` (this halts kong, postgres, and vault, and sets the env var `EDGEX_SECURITY_SECRET_STORE=false` for all of the services (except for device-virtual and app-service-configurable which aren’t configured in `snapcraft.yaml` to use the required command chain script which handles this var).
[...]

PR #3791, opened 2021-11-03
fix(snap): defer service startup till after configure hook runs

tonyespy:
This PR modifies the way the snap install and configure hooks when the snap in installed in order to ensure that snap configuration provided at install time is processed before services are started. This is working around an issue with snapd where snap configuration at install time is not made available to the snap until the configure hook is called, which is after the services have already been started.

So, this PR disables all of the services in the install hook, and then in the configure hook, the services are started after configuration has been processed.

## PR Checklist

- [*] I have added unit tests for the new feature or bug fix (if not, why?)
  - units tests to be added before this is taken out of draft status 
- [NA] I have opened a PR for the related docs change (if not, why?)
  - This fix relates to the install behavior of the snap, and as such should be transparent to the installer. 

## Testing Instructions
 - Install the snap built from this PR and verify that all of the expected services start
 - Confirm that device-virtual can be enabled (`$ sudo snap set edgexfoundry device-virtual=on`)
 - Confirm that kuiper can be enabled (`$ sudo snap set edgexfoundry kuiper=on`); ensure that both kuiper and app-service-configurable are started
 - Confirm that support-notifications and support-scheduler can be started (e.g. `snap set... support-scheduler=on`)
 - Confirm that security-proxy can be disabled (`$sudo snap set security-proxy=off`)

Additional test instructions will be added to this PR as it progresses. The more advanced cases require a custom gadget snap which provides default configuration. This is currently the only way to disable secret-store usage. If possible, this PR will be extended to support dynamically disabling secret store usage, but this is currently blocked on issue #3712.

Also note, this has been tested using [our in-progress checkbox provider for Jakarta](https://code.launchpad.net/~mengyiw/checkbox-provider-edgex/+git/checkbox-provider-edgex/+merge/409927) and all of the tests pass.

### DIFF
--- a/snap/local/hooks/cmd/configure/configure.go
+++ b/snap/local/hooks/cmd/configure/configure.go
@@ -20,7 +20,10 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
+	"strings"
 
 	hooks "github.com/canonical/edgex-snap-hooks/v2"
 )
@@ -44,8 +47,47 @@ func getKuiperServices() []string {
 	return []string{hooks.ServiceAppCfg, hooks.ServiceKuiper}
 }
 
+// getProxyServices returns the list of services which implement
+// the API Gateway. Note this list *excludes* Consul and the
+// Secret Store services.
 func getProxyServices() []string {
 	return []string{"postgres", "kong-daemon", "security-proxy-setup"}
+}
+
+// getSecretStoreServices returns the list of services which implement
+// the Secret Store and related dependencies (i.e. the services that
+// secure redis and consul which are tightly bound to the secret store
+// being enabled).
+func getSecretStoreServices() []string {
+	return []string{"security-secretstore-setup", "vault",
+		"security-consul-bootstrapper", "security-bootstrapper-redis"}
+}
+
+// getAllServices returns the entire list of snap services.
+func getAllServices() []string {
+	return []string{"consul", "redis", "core-data", "core-metadata",
+	"core-command",	"security-secretstore-setup", "security-proxy-setup",
+	"security-bootstrapper-redis", "security-consul-bootstrapper",
+	"kong-daemon", "postgres", "vault"}
+}
+
+func getEdgeXServices() []string {
+	return []string{"core-data", "core-metadata", "core-command",
+			"device-virtual", "support-notifications",
+			"support-scheduler", "sys-mgmt-agent"}
+}
+
+// getRequiredServices returns the minimum list of required
+// snap services for a working EdgeX instance.
+func getRequiredServices() []string {
+	return []string{"consul", "redis", "core-metadata"}
+}
+
+// getOptionalServices returns the minimum list of optional
+// EdgeX services.
+func getOptionalServices() []string {
+	return []string{"core-command", "core-data", "support-notifications",
+			"support-scheduler","sys-mgmt-agent"}
 }
 
 // handleSingleService starts or stops a service based on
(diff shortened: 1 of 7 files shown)


tonyespy:
FYI, just force pushed a minor update to address a minor gofmt error which was tripping up the CI build.

siggiskulason (commented):
Looks good so far. I followed the instructions and could start the services, stop the proxy and start it up again and all the services restarted correctly.

anonymouse64:
Just a general comment that the change to remove the device-virtual service seems to be also deleted but it isn't mentioned in the commit messages or the PR description.

snap/local/hooks/cmd/configure/configure.go
```diff
@@ -86,146 +109,262 @@ func serviceType(name string) int {
 	}
 }
 
+func appendServices(startServices []string, newServices []string) []string {
+	for _, s := range newServices {
+		startServices = append(startServices, s)
+	}
+	return startServices
```
anonymouse64, line 116: this can also just be simplified to one append:

```suggestion
	return append(startServices, newServices...)
```

snap/local/hooks/cmd/configure/configure.go
```diff
@@ -86,146 +109,262 @@ func serviceType(name string) int {
...
+// - start/stop any tightly coupled services (e.g. if the secret-store
 //   is disabled, the proxy also has to come down) if required
 // - start/stop the service itself if required
 //
 // NOTE - at this time, this function does *not* restart a service based
 // on env configuration changes. If changes are made after a service has
 // been started, the service must be restarted manually.
 //
-func handleAllServices() error {
+func handleAllServices(install bool) (error, []string) {
```
anonymouse64, line 142: a comment about the install parameter would be nice

snap/local/hooks/cmd/configure/configure.go
```diff
@@ -86,146 +109,262 @@ func serviceType(name string) int {
...
+
+func startEnabled(startServices []string) error {
+	var cmd *exec.Cmd
+	var startServicesCmd = []string{
+		"start",
+		"--enable",
+		// NOTE -- see addtl comments about *always* required services.
+		// These should really be removed from the set of configurable
+		// services (see hooks.Services) as if they were actually disabled
+		// the syste would no longer work. If/when they're removed, then
```
anonymouse64, line 351:
```suggestion
		// the system would no longer work. If/when they're removed, then
```

snap/local/hooks/cmd/configure/configure.go
```diff
@@ -86,146 +109,262 @@ func serviceType(name string) int {
...
+				if err := cli.SetConfig("kuiper", ""); err != nil {
+					return err, startServices
 				}
-				if err = handleSingleService("core-metadata", ON); err != nil {
-					return err
+				fallthrough
+			case ON:
+				serviceList = getKuiperServices()
+				if install {
+					startServices = appendServices(startServices, serviceList)
```
anonymouse64, line 196: this helper can go away completely if you do

```suggestion
					startServices = append(startServices, serviceList...)
```

snap/local/hooks/cmd/configure/configure.go
```diff
@@ -86,146 +109,262 @@ func serviceType(name string) int {
...
-			// - vault (because security-proxy-setup will access/store secrets in vault)
-			// - security-secretstore-setup
-			// so if we are turning the security-api-gateway on, then turn
-			// those services on too
-			if status == ON {
-				if err = handleSingleService("vault", ON); err != nil {
-					return err
+				serviceList = getProxyServices()
+				if install {
+					startServices = appendServices(startServices, serviceList)
```
anonymouse64, line 236: same

snap/local/hooks/cmd/configure/configure.go
anonymouse64: I guess even better you could do

```suggestion
					startServices = append(startServices, getKuiperServices()...)
```

snap/local/hooks/cmd/install/install.go
```diff
@@ -389,22 +426,48 @@ func main() {
...
+		// all of the service options set to "install", and leave configuration
+		// that actually happened as a result of provided config (e.g. from the
+		// gadget) alone
+		if err = cli.SetConfig(s, "install"); err != nil {
+			hooks.Error(fmt.Sprintf("edgexfoundry:install setting service: %s to 'on'; %v", s, err))
+			os.Exit(1)
 		}
 	}
+
+	if err = cli.SetConfig("install", "true"); err != nil {
```
anonymouse64, line 469: it does seem inconsistent to me in hindsight that snapd let's you set config items in the install hook (before the configure hook runs), but it doesn't let you read them before the configure hook runs :-|

snap/local/hooks/cmd/install/install.go
```diff
@@ -323,7 +348,19 @@ func installProxy() error {
 }
 
 func installRedis() error {
-	if err := os.MkdirAll(hooks.SnapData+"/redis", 0755); err != nil {
+	setupScriptPath, err := exec.LookPath("install-setup-redis.sh")
```
anonymouse64, line 351: minor nitpick, but this whole file could be implemented in Go with less lines of code than running the script here:

```go
snapData := os.Getenv("SNAP_DATA")
fileName := filepath.Join(snapData,"/redis/conf/redis.conf")
if _, err := os.Stat(filepath.Join(snapData,"redis")); err != nil {
    // dir doesn't exist
    if err := os.MkdirAll(filepath.Dir(fileName),0755); err != nil {
        return err
    }
    if err := ioutil.WriteFile(fileName, nil, 0644); err != nil {
         return err
    }
}
```

with better comments of course

snap/local/hooks/cmd/configure/configure.go
```diff
@@ -250,8 +391,33 @@ func main() {
...
+	if err != nil {
+		fmt.Println(fmt.Sprintf("edgexfoundry:configure: reading 'install': %v", err))
+		os.Exit(1)
+	}
+
+	if install == "true" {
+		installMode = true
+	}
+
+	if err, startServices = handleAllServices(installMode); err != nil {
```
anonymouse64, line 404: a comment that handleAllServices will dynamically stop/start services if installMode is false would be nice

anonymouse64: otherwise it seems that handleAllServices just returns the set of services which should be started, but really it only returns the list of services that should be started when we are in install mode

snap/local/hooks/cmd/configure/configure.go
```diff
@@ -86,146 +109,262 @@ func serviceType(name string) int {
...
+	secretStoreActive := true
 
 	// grab and log the current service configuration
 	for _, s := range hooks.Services {
 		var envJSON string
 
 		status, err := cli.Config(s)
 		if err != nil {
-			return err
+			return err, startServices
```
anonymouse64, line 153: you can just do `return err, nil` everywhere, that is more Go idiomatic

snap/local/hooks/cmd/configure/configure.go
anonymouse64: ah nevermind I see you do also need serviceList too, so ignore the second recommendation

anonymouse64 (commented):
some comments, overall lgtm

tonyespy:
> Just a general comment that the change to remove the device-virtual service seems to be also deleted but it isn't mentioned in the commit messages or the PR description.

@anonymouse64 First, thanks for review!

Re: your device-virtual comment, if you look again, you'll notice that what I removed was an actual case statement for device-virtual. That's because the original case for device-virtual in `handleAllServices` had additional logic to re-enable core-data and core-metadata when device-virtual is enabled.  In the Ireland (2.0) release of EdgeX, device services now publish events direct to the message bus by default, so core-data is no longer a hard-dependency. Regarding core-metadata, if it's not running the whole system falls apart, so we should really remove the ability to disable it, but I decided to leave that for another day. Since there are no dependencies to handle, device-virtual is simply handed like any other service without dependencies (i.e. by the default case).

farshidtz:
You can upgrade the edgex-snap-hooks to use the newly added function. See https://github.com/canonical/edgex-snap-hooks/releases/tag/v2.1.0

```bash
go get github.com/canonical/edgex-snap-hooks/v2@v2.1.0
```

I've explained [here](https://github.com/canonical/edgex-snap-hooks/pull/18#issuecomment-962109016) that the [EnabledServices](https://pkg.go.dev/github.com/canonical/edgex-snap-hooks/v2#CtlCli.EnabledServices) function is not useful when called in the install hook.

tonyespy:
I just force-pushed a rebase which uses the new edgex-snap-hooks version @farshidtz recently published. This uses the new `UnsetConfig` and `StartMultiple` methods. I've re-run the checkbox tests against the snap, and all continue to pass.

tonyespy:
Just force-pushed a rebase to update `go.mod/go.sum` to edgex-snap-hooks 2.1.1 to pickup @siggiskulason's fix for TLS configuration.

tonyespy:
Just pushed a new commit which enables the Secret Store to be disabled at runtime. Thanks again @farshidtz for figuring out issue #3712.

tonyespy:
Closing as this PR has now been superseded by PR #3813.

PR #3794, opened 2021-11-08
docs(snap): Update proxy configuration information in README.md

siggiskulason:
- changed wording in API Gateway section
- removed "$" for consistency and updated examples

## PR Checklist

- [ ] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break) -N/A
- [ ] I am not introducing a new dependency (add notes below if you are) N/A
- [ ] I have added unit tests for the new feature or bug fix (if not, why?) N/A
- [ ] I have fully tested (add details below) this the new feature or bug fix (if not, why?) N/A
- [ ] I have opened a PR for the related docs change (if not, why?) N/A
  <link to docs PR>


snap/README.md
```diff
@@ -260,11 +248,30 @@ sudo snap set edgexfoundry env.security-proxy.user=user01,USER_ID,ES256
...
+```
+
+3. Finally, you need to generate a token using the user ID which you specified:
 
 ```bash
-$ edgexfoundry.secrets-config proxy jwt --algorithm ES256 --private_key private.pem --id USER_ID --expiration=1h
-TOKEN="copy manually from output of secrets-config command"
+# get token
+TOKEN=`edgexfoundry.secrets-config proxy jwt --algorithm ES256 --private_key private.pem --id USER_ID --expiration=1h`
+
```
MonicaisHer, line 274: Maybe we can add "echo $TOKEN" so that the user can save it for further usage. For example, the UI's secure mode needs users to input this token. Please see an example [here](https://github.com/edgexfoundry/edgex-ui-go/tree/main/snap#using-the-edgex-ui-snap).

snap/README.md
```diff
@@ -290,32 +297,23 @@ JWT_SIGNATURE=`echo -n "$JWT_HEADER.$JWT_PAYLOAD" | openssl dgst -sha256 -binary
...
-Additional users can be added by repeatedly calling the secrets-config command as above. Only one user can however be set at any time when using the snap configure hook, so
-the current user must first be removed by setting the snap configuration settings to an empty string, before setting the values again:
-
-```bash
-$ sudo snap set edgexfoundry env.security-proxy.user=""
-$ sudo snap set edgexfoundry env.security-proxy.public-key=""
-$ sudo snap set edgexfoundry env.security-proxy.user=user02,USER_ID2,ES256
-$ sudo snap set edgexfoundry env.security-proxy.public-key="$(cat public.pem)"
-```
+Additional users can be added by repeatedly calling the secrets-config command as above.
```
MonicaisHer, line 311: I suggest this could be more clear: Additional users can be added by repeatedly step 2 and 3 calling the secrets-config command as above.

snap/README.md
```diff
@@ -290,32 +297,23 @@ JWT_SIGNATURE=`echo -n "$JWT_HEADER.$JWT_PAYLOAD" | openssl dgst -sha256 -binary
...
+Additional users can be added by repeatedly calling the secrets-config command as above.  
   
 
 #### API Gateway TLS certificate setup
 
-By default Kong is configured with an EdgeX signed TLS certificate. Client validation of this certificate requires the root CA certificate from the EdgeX instance. This file
-(`ca.pem`) can be copied from directory `$SNAP_DATA/secrets/ca`.
-
-It is also possible to install your own TLS certificate to be used by the gateway. The steps to do so are as follows:
+By default Kong is configured with a selfsigned TLS certificate. It is also possible to install your own TLS certificate to be used by the gateway. The steps to do so are as follows:
```
MonicaisHer, line 316: Is that possible to include this TLS certificate's file path?

MonicaisHer (dismissed):

snap/README.md
farshidtz: +1
I think a note would be useful to remind the user that the token should be kept for future reference. The user can decide to print it to standard output and handle it externally or redirect it to a file.

If it makes sense:
"Keep this token in a safe place for future reuse as the same token cannot be regenerated or recovered using the `secret-config` CLI."

snap/README.md
farshidtz: Removing the line could also be a way forward as the user following those steps is expected to understand them and repeat what is necessary.

snap/README.md
farshidtz: ~selfsigned~ self-signed  TLS certificate

snap/README.md
siggiskulason: updated

snap/README.md
siggiskulason: removed

snap/README.md
siggiskulason: yes, added that

siggiskulason:
@MonicaisHer , @farshidtz thanks for the review. I've updated the PR, including your suggestions and clarifying the examples.

snap/README.md
```diff
@@ -260,11 +248,33 @@ sudo snap set edgexfoundry env.security-proxy.user=user01,USER_ID,ES256
 sudo snap set edgexfoundry env.security-proxy.public-key="$(cat public.pem)"
 ```
 
-The second step is then to generate a token using the user ID which you specified:
+Alternatively, you can create the user using the secrets-config command. You need to provide the following:
```
farshidtz, line 251: This is not only an alternative, but the only way to create multiple users. As mentioned, the previous commands only allow setting a single user.

I suggest:
~Alternatively, you can create the user using~To create multiple users, use the secrets-config command. You need to provide the following:

snap/README.md
```diff
@@ -260,11 +248,33 @@ sudo snap set edgexfoundry env.security-proxy.user=user01,USER_ID,ES256
...
+create the JWT token. If you don't specify it,
+then an autogenerated one will be output by the secrets-config command
+```bash
+
+# get API Gateway/Kong token
+JWT_FILE=/var/snap/edgexfoundry/current/secrets/security-proxy-setup/kong-admin-jwt
+JWT=`sudo cat ${JWT_FILE}`
+
+# use secrets-config to add user
+$ edgexfoundry.secrets-config proxy adduser --token-type jwt --user user01 --algorithm ES256 --public_key public.pem --id USER_ID --jwt ${JWT}
```
farshidtz, line 266: Remove the `$` from the beginning for consistency.

snap/README.md
```diff
@@ -290,72 +300,64 @@ JWT_SIGNATURE=`echo -n "$JWT_HEADER.$JWT_PAYLOAD" | openssl dgst -sha256 -binary
...
-
-Additional users can be added by repeatedly calling the secrets-config command as above. Only one user can however be set at any time when using the snap configure hook, so
-the current user must first be removed by setting the snap configuration settings to an empty string, before setting the values again:
-
-```bash
-$ sudo snap set edgexfoundry env.security-proxy.user=""
-$ sudo snap set edgexfoundry env.security-proxy.public-key=""
-$ sudo snap set edgexfoundry env.security-proxy.user=user02,USER_ID2,ES256
-$ sudo snap set edgexfoundry env.security-proxy.public-key="$(cat public.pem)"
+$ curl -k -X GET https://localhost:8443/core-data/api/v2/ping? -H "Authorization: Bearer $TOKEN"
```
farshidtz, line 311: Remove the `$` character for consistency with other commands in this section.

farshidtz (changes requested):

farshidtz:
> @MonicaisHer , @farshidtz thanks for the review. I've updated the PR, including your suggestions and clarifying the examples.

Thanks. I added some inline comments for minor modifications.

I feel there are additional changes compared to before, but there is no way to know due to the lost history. In future, please avoid amending the commit unless you made a mistake which you need to go back and fix. Having separate commits helps the reviewers to quickly see the differences and possible new additions. Moreover, it helps to detect changes that have been committed by mistake.

In the end, we can still squash and merge to have a single commit on the main branch.

snap/README.md
```diff
@@ -227,30 +226,19 @@ not addressable from another system). In order to make a service accessible remo
...
-`secrets-config` command, or by using `snap set` commands.
+Before the API Gateway can be used, a user and group must be created and a JWT access token generated.
 
-The first step is to add a user. You need to create a public/private keypair, which can be done with
+1. The first step is to create a public/private keypair for the new user, which can be done with
 
 ```bash
 # Create private key:
-$ openssl ecparam -genkey -name prime256v1 -noout -out private.pem
+openssl ecparam -genkey -name prime256v1 -noout -out private.pem
```
MonicaisHer, line 235: openssl ecparam -genkey -name prime256v1 -noout -out private.pem
ecparam: Can't open "private.pem" for writing, Permission denied

(I think it need to be used with `sudo`)

snap/README.md
```diff
@@ -227,30 +226,19 @@ not addressable from another system). In order to make a service accessible remo
...
 
 # Create public key:
-$ openssl ec -in private.pem -pubout -out public.pem
+openssl ec -in private.pem -pubout -out public.pem
```
MonicaisHer, line 238: openssl ec -in private.pem -pubout -out public.pem
Can't open private.pem for reading, Permission denied
140544944436544:error:0200100D:system library:fopen:Permission denied:../crypto/bio/bss_file.c:69:fopen('private.pem','r')
140544944436544:error:2006D002:BIO routines:BIO_new_file:system lib:../crypto/bio/bss_file.c:78:

(I think it need to be used with `sudo`)

snap/README.md
```diff
@@ -260,11 +248,33 @@ sudo snap set edgexfoundry env.security-proxy.user=user01,USER_ID,ES256
...
+3. Finally, you need to generate a token using the user ID which you specified:
 
 ```bash
-$ edgexfoundry.secrets-config proxy jwt --algorithm ES256 --private_key private.pem --id USER_ID --expiration=1h
-TOKEN="copy manually from output of secrets-config command"
+# get token
+TOKEN=`edgexfoundry.secrets-config proxy jwt --algorithm ES256 --private_key private.pem --id USER_ID --expiration=1h`
+
+# Keep this token in a safe place for future reuse as the same token cannot be regenerated or recovered using the secret-config CLI
+echo $TOKEN
```
MonicaisHer, line 276: echo $TOKEN
level=ERROR ts=2021-11-11T12:11:24.430840404Z app=secrets-config source=bootstraphandler.go:76 msg="Could read private key from file private.pem: open private.pem: permission denied"

sudo echo $TOKEN
level=ERROR ts=2021-11-11T12:11:24.430840404Z app=secrets-config source=bootstraphandler.go:76 msg="Could read private key from file private.pem: open private.pem: permission denied"

snap/README.md
MonicaisHer: curl -k -X GET https://localhost:8443/core-data/api/v2/ping? -H "Authorization: Bearer $TOKEN"
{"message":"Bad token; invalid JSON"}

(I think there should be a blank after "Authorization")
curl -k -X GET https://localhost:8443/core-data/api/v2/ping? -H "Authorization : Bearer $TOKEN"
{"message":"Unauthorized"}

Do you have any idea why it is Unauthorized? Thanks!

MonicaisHer (changes requested):
@siggiskulason Thank you for your update! I have re-review this using the latest snap(reversion 3327). It shows errors in steps 1, 3, 4. These steps used work fine for me with the previous snap(reversion 3324). Please see the inline comments.

snap/README.md
siggiskulason: modified

snap/README.md
siggiskulason: modified.

The reason you get Unauthorized is because of the blank - it should not be added after "Authorization". However, the reason you get "Bad token" is probably because the token is invalid. Did you follow the steps to add a user and create a token?

snap/README.md
siggiskulason: sudo should not be needed - but you need to run these commands in a working directory in your home directory that you have write access to... so you were probably running the command somewhere else?

snap/README.md
siggiskulason: As above... you should run this command somewhere in a local directory that you have write access to. You shouldn't need to use sudo.

snap/README.md
siggiskulason: modified

> > @MonicaisHer , @farshidtz thanks for the review. I've updated the PR, including your suggestions and clarifying the examples.
> 
> Thanks. I added some inline comments for minor modifications.
> 
> I feel there are additional changes compared to before, but there is no way to know due to the lost history. In future, please avoid amending the commit unless you made a mistake which you need to go back and fix. Having separate commits helps the reviewers to quickly see the differences and possible new additions. Moreover, it helps to detect changes that have been committed by mistake.
> 
> In the end, we can still squash and merge to have a single commit on the main branch.

Thanks, I've added a new commit with suggested changes.

siggiskulason:
> > @MonicaisHer , @farshidtz thanks for the review. I've updated the PR, including your suggestions and clarifying the examples.
> 
> Thanks. I added some inline comments for minor modifications.
> 
> I feel there are additional changes compared to before, but there is no way to know due to the lost history. In future, please avoid amending the commit unless you made a mistake which you need to go back and fix. Having separate commits helps the reviewers to quickly see the differences and possible new additions. Moreover, it helps to detect changes that have been committed by mistake.
> 
> In the end, we can still squash and merge to have a single commit on the main branch.

 

Thanks, I've added a new commit with suggested changes.

farshidtz (approved):
Thanks @siggiskulason. Looks good.
I suggest squash+merge.

snap/README.md
MonicaisHer: Thanks! I was in the checkbox directory, which already contains used key pairs. sudo should not be needed.

snap/README.md
MonicaisHer: Same as above. I was in the checkbox directory with previous key pairs.

snap/README.md
MonicaisHer: Yes, I added a user and created a token. After changing working directory, now it works. 

I got: `"message":"An invalid response was received from the upstream server"`. Does this mean self-signed TLS  is implemented?

MonicaisHer (approved):

tonyespy (approved):
LGTM

PR #3813, opened 2021-11-12
fix(snap): defer service startup till after configure hook runs

tonyespy:
This PR modifies the way the snap install and configure hooks when the snap in installed in order to ensure that snap configuration provided at install time is processed before services are started. This is working around an issue with snapd where snap configuration at install time is not made available to the snap until the configure hook is called, which is after the services have already been started.

So, this PR disables all of the services in the install hook, and then in the configure hook, the services are started after configuration has been processed.

**Note** - this PR supersedes PR #3791.

## PR Checklist

- [*] I have added unit tests for the new feature or bug fix (if not, why?)
  - units tests to be added before this is taken out of draft status 
- [NA] I have opened a PR for the related docs change (if not, why?)
  - This fix relates to the install behavior of the snap, and as such should be transparent to the installer. 

## Testing Instructions
 - Install the snap built from this PR and verify that all of the expected services start
 - Confirm that device-virtual can be enabled (`$ sudo snap set edgexfoundry device-virtual=on`)
 - Confirm that kuiper can be enabled (`$ sudo snap set edgexfoundry kuiper=on`); ensure that both kuiper and app-service-configurable are started
 - Confirm that support-notifications and support-scheduler can be started (e.g. `snap set... support-scheduler=on`)
 - Confirm that security-proxy can be disabled (`$sudo snap set security-proxy=off`)

Additional test instructions will be added to this PR as it progresses. The more advanced cases require a custom gadget snap which provides default configuration. This is currently the only way to disable secret-store usage. If possible, this PR will be extended to support dynamically disabling secret store usage, but this is currently blocked on issue #3712.

Also note, this has been tested using [our in-progress checkbox provider for Jakarta](https://code.launchpad.net/~mengyiw/checkbox-provider-edgex/+git/checkbox-provider-edgex/+merge/409927) and all of the tests pass.

### DIFF
--- a/snap/README.md
+++ b/snap/README.md
@@ -205,6 +205,9 @@ Thus the following command will disable both:
 $ sudo snap set edgexfoundry security-secret-store=off
 ```
 
+*NOTE* - disabling security after the snap is installed is a convenience for developers. The snap will not allow the Secret Store to be
+re-enabled. The only way to re-enable the Secret Store is to re-install the snap.
+
 #### API Gateway
 Kong is used for access control to the EdgeX services from external systems and is referred to as the API Gateway. 
 
--- a/snap/local/hooks/cmd/configure/configure.go
+++ b/snap/local/hooks/cmd/configure/configure.go
@@ -20,7 +20,10 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
+	"strings"
 
 	hooks "github.com/canonical/edgex-snap-hooks/v2"
 )
@@ -35,26 +38,73 @@ const ( // iota is reset to 0
 )
 
 const (
-	INSTALL = "install"
-	ON      = "on"
-	OFF     = "off"
-	UNSET   = ""
+	ON    = "on"
+	OFF   = "off"
+	UNSET = ""
 )
 
-func getDefaultServices() []string {
-	return []string{"consul", "redis", "core-data",
-		"core-metadata", "core-command",
-		"security-secretstore-setup", "security-proxy-setup",
-		"security-bootstrapper-redis", "security-consul-bootstrapper",
-		"kong-daemon", "postgres", "vault"}
-}
-
+// getKuiperServices returns the list of services used for
+// EdgeX rules-engine processing.
 func getKuiperServices() []string {
 	return []string{hooks.ServiceAppCfg, hooks.ServiceKuiper}
 }
 
+// getProxyServices returns the list of services which implement
+// the API Gateway. Note this list *excludes* Consul and the
+// Secret Store services.
 func getProxyServices() []string {
 	return []string{"postgres", "kong-daemon", "security-proxy-setup"}
+}
+
+// getSecretStoreServices returns the list of services which implement
+// the Secret Store and related dependencies (i.e. the services that
+// secure redis and consul which are tightly bound to the secret store
+// being enabled).
+func getSecretStoreServices() []string {
+	return []string{"security-secretstore-setup", "vault",
+		"security-consul-bootstrapper", "security-bootstrapper-redis"}
+}
+
+// getEdgeXRefServices returns the list of EdgeX reference services in the snap
+// (excludes all non-EdgeX runtime dependencies, and security-*-setup jobs).
+func getEdgeXRefServices() []string {
+	return []string{"core-data", "core-metadata", "core-command",
+		"device-virtual", "support-notifications",
+		"support-scheduler", "sys-mgmt-agent"}
+}
+
+// getRequiredServices returns the minimum list of required
+// snap services for a working EdgeX instance.
+func getRequiredServices() []string {
+	return []string{"consul", "redis", "core-metadata"}
+}
+
+// getCoreDefaultServices returns the list of core services
+// that are started by default (in addition to the required
+// services)
+func getCoreDefaultServices() []string {
+	return []string{"core-command", "core-data"}
+}
+
+// getOptServices returns the list of optional EdgeX services
+// (i.e. disabled by default).
+//
+// Note:
+// - sys-mgmt-agent isn't included because as of Ireland
+//   it's considered deprecated.
+// - kuiper isn't included because it's not yet possible
+//   to provide kuiper configuration via content interface
+func getOptServices() []string {
+	return []string{"support-notifications", "support-scheduler", "device-virtual"}
+}
+
+func isDisableAllowed(s string) error {
+	for _, v := range getRequiredServices() {
+		if s == v {
+			return fmt.Errorf("edgexfoundry:configure: can't disable required service: %s", s)
+		}
+	}
+	return nil
 }
 
 // handleSingleService starts or stops a service based on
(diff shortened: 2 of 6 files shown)


snap/local/hooks/cmd/configure/configure.go
```diff
@@ -86,22 +164,156 @@ func serviceType(name string) int {
...
+		}
+	}
+	return nil
+}
+
 // handleAllServices iterates through all of the services in the
 // edgexfoundry snap and:
 //
-// - queries the config option associated with the service state (on|off)
+// - queries the config option associated with the service state (on|off|install|'')
```
farshidtz, line 299: I believe `install` is no longer one of the possible states.

tonyespy:
I just force-pushed again resolving the following comments from @farshidtz:

 - changed config options `install=true` to `install-mode: defer-startup`
 - rm `INSTALL` const
 - collapsed a few case statements to remove usage of `fallthrough`
 - removed 'install' from the comment describing the possible values for a service config option

tonyespy:
One more forced push which updates the defer-startup logic, after testing with some pre-configured services.

siggiskulason:
Running the current TAF tests against this snap succeeds - we get the same results as with the latest/edge snap. The TAF tests do test functionality including
- installing the snaps and starting services
- enabling and starting  services including device-virtual, app-service-configurable, support-notifications and support-scheduler

farshidtz:
Formatting issues caught by the CI can be resolved:

- away from the IDE:
```
$ cd edgex-go/snap/local/hooks/
$ go fmt ./...
cmd/install/install.go
cmd/configure/configure.go
```

- if IDE is open in `snap/local/hooks`, simply:
```bash
go fmt ./...
```

- using the more flexible `gofmt` (if you have it):
```bash
gofmt -w snap/local/hooks
```

snap/local/hooks/cmd/install/install.go
```diff
@@ -389,22 +410,19 @@ func main() {
...
-	// just like the configure hook, this code needs to iterate over every service
-	// and setup .env files for each if required...
-	for _, v := range hooks.Services {
-		serviceCfg := hooks.EnvConfig + "." + v
-		envJSON, err := cli.Config(serviceCfg)
-		if err != nil {
-			hooks.Error(fmt.Sprintf("edgexfoundry:install failed to read service %s configuration - %v", v, err))
+	// Stop and disable all *enabled* services as they will be
+	// re-enabled in the configure hook if the config option
+	// 'install=true'.
```
farshidtz, line 415: > Stop and disable all *enabled* services as they will be
 re-enabled in the configure hook ~if the config option
 'install=true'.~

snap/local/hooks/cmd/install/install.go
```diff
@@ -389,22 +410,19 @@ func main() {
...
 
-		if envJSON != "" {
-			hooks.Debug(fmt.Sprintf("edgexfoundry:install: service envJSON: %s", envJSON))
-			if err := hooks.HandleEdgeXConfig(v, envJSON, nil); err != nil {
-				hooks.Error(fmt.Sprintf("edgexfoundry:install failed to process service %s configuration - %v", v, err))
-				os.Exit(1)
-			}
-		}
+	if err = cli.SetConfig("install-mode", "defer-startup"); err != nil {
+		hooks.Error(fmt.Sprintf("edgexfoundry:install setting 'install=true'; %v", err))
```
farshidtz, line 425: > edgexfoundry:install setting ~'install=true'~ config to defer startup

snap/README.md
```diff
@@ -205,6 +205,9 @@ Thus the following command will disable both:
 $ sudo snap set edgexfoundry security-secret-store=off
 ```
 
+*NOTE* - disabling security after the snap is installed is a convenience for developers. The snap will not allow the Secret Store to be
+re-enabled. The only way to re-enabled the Secret Store is to re-install the snap.
```
siggiskulason, line 209: "The only way to re-enable**d**" -> "The only way to re-enable"

snap/local/hooks/cmd/configure/configure.go
```diff
@@ -127,113 +342,200 @@ func handleAllServices() error {
...
+			case UNSET:
+				// this is the default status of all services if no
+				// configuration has been specified; no-op
+				continue
+			default:
+				return fmt.Errorf("edgexfoundry:configure: invalid value for security-secret-store: %s", status)
 			}
+
 		default:
 			hooks.Debug("edgexfoundry:configure: other service")
```
farshidtz, line 424: Not from this PR. But I'd add here the name of that other service:
`hooks.Debug("edgexfoundry:configure: other service: "+s)`

snap/local/hooks/cmd/configure/configure.go
```diff
@@ -86,22 +157,165 @@ func serviceType(name string) int {
...
+//
 //
 // NOTE - at this time, this function does *not* restart a service based
 // on env configuration changes. If changes are made after a service has
 // been started, the service must be restarted manually.
 //
-func handleAllServices() error {
+func handleAllServices(deferStartup bool) error {
+	var serviceList []string
+	secretStoreActive := true
```
siggiskulason, line 318: a couple of lines below, in line 324 we have

```
		status, err := cli.Config(s)
		if err != nil {
			return err
		}
```
If the user doesn't use lower case then we fail when processing the status:

```
sudo snap set edgexfoundry device-virtual=ON
error: cannot perform the following tasks:
- Run configure hook of "edgexfoundry" snap (run hook "configure": exit status 1)
```

Should we change this to be case-agnostic? ( by adding something like the following) 


```
		status, err := cli.Config(s)
		if err != nil {
			return err
		}
		status, err := strings.ToLower(status)
```

snap/local/hooks/cmd/configure/configure.go
```diff
@@ -20,26 +20,92 @@ package main
...
+func getRequiredServices() []string {
+	return []string{"consul", "redis", "core-metadata"}
+}
+
+// getCoreDefaultServices returns the list of core services
+// that are started by default (in addition to the required
+// services)
+func getCoreDefaultServices() []string {
+	return []string{"core-command", "core-data"}
+}
```
farshidtz, line 86: Just my personal preference, feel free to skip:
- Use global variables instead of getter functions. This is a "main" and small package, so accidental modifications to those slices from some function are not likely.
- Add line breaks between each line. This will improve readability and more importantly allow changes to different parts from independent PRs.

snap/local/runtime-helpers/bin/start-consul.sh
```diff
@@ -15,16 +15,21 @@ cat > "$SNAP_DATA/consul/config/consul_default.json" <<EOF
...
 {
-    "acl": {
+      "acl": {
       "enabled": true,
       "default_policy": "deny",
       "enable_token_persistence": true
     }
 }
 EOF
+fi
```
farshidtz, line 32: Can be done in another PR: Since this PR makes the services start after configuration, it makes sense to move similar configuration logic to the configure hook. Removing of the acl file already does happen in the configure hook: https://github.com/edgexfoundry/edgex-go/blob/1839698a2dde2b913a8fb2ca4dd0c55fce4f3cab/snap/local/hooks/cmd/configure/configure.go#L189-L195

farshidtz (approved):
I have tested the changes and they seem to work as expected. The code looks good, I only have some minor inline comment.

For the sake of simplifying further reviews, it would be great if you add future changes in new commits, rather than amending old ones.

siggiskulason (dismissed):
Testing: Passes the current baseline functional API TAF tests.
Code review: I had a couple of minor inline comments.

anonymouse64 (commented):
some comments overall lgtm though note I didn't have time to test anything out on my machine, I trust others have done so as they commented

tonyespy:
> For the sake of simplifying further reviews, it would be great if you add future changes in new commits, rather than amending old ones.

@farshidtz given the time frame involved, I'm going to politely decline this request and continue to use forced push(es) to land my remaining changes. I'll be very specific in comments if/when I do this. We also covered this in our team sync, and I suggested we use one of our future meetings to discuss the topic.

tonyespy:
Force pushed two changes:

 - apply `gofmt`
 - rebased on `origin/main`
 -

snap/snapcraft.yaml
```diff
@@ -325,8 +327,6 @@ apps:
     environment:
       SECRETSTORE_TOKENFILE: $SNAP_DATA/secrets/support-notifications/secrets-token.json
     daemon: simple
-    passthrough:
-      install-mode: disable
```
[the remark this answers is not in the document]
tonyespy: correct

snap/snapcraft.yaml
```diff
@@ -395,10 +390,9 @@ apps:
       -confdir $SNAP_DATA/config/app-service-configurable/res
       -profile rules-engine
     command-chain:
+      - bin/security-secret-store-env-var.sh
       - bin/service-config-overrides.sh
```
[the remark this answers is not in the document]
tonyespy: Noted on our internal project board

snap/local/hooks/cmd/configure/configure.go
tonyespy: Line breaks are another thing we should discuss.  I've always used blank lines freely in C/C++/Java/... Over time, I've been adding less and less of them to my Go code, as that most Go developers seems to follow this practice.

snap/local/runtime-helpers/bin/start-consul.sh
tonyespy: @farshidtz can you open an an issue on our internal board for this?

snap/local/hooks/cmd/configure/configure.go
```diff
@@ -20,26 +20,92 @@ package main
...
+// secure redis and consul which are tightly bound to the secret store
+// being enabled).
+func getSecretStoreServices() []string {
+	return []string{"security-secretstore-setup", "vault",
+		"security-consul-bootstrapper", "security-bootstrapper-redis"}
+}
+
+// getEdgeXServices returns the list of EdgeX services in the snap
+func getEdgeXServices() []string {
+	return []string{"core-data", "core-metadata", "core-command",
```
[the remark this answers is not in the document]
tonyespy: I used getEdgeXRefServices and added a comment to the godoc clarifying the purpose of this list.

snap/local/hooks/cmd/configure/configure.go
```diff
@@ -127,113 +342,200 @@ func handleAllServices() error {
...
+				serviceList = getProxyServices()
+			case UNSET:
+				// this is the default status of all services if no
+				// configuration has been specified; no-op
+				continue
+			default:
+				return fmt.Errorf("edgexfoundry:configure: invalid value for security-proxy: %s", status)
 			}
+
 		case secStoreService:
```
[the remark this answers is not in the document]
tonyespy: Comments added to the code, and issue created on our internal project board.

tonyespy:
Force pushed what's hopefully my last changes:

 - fixed a typo in snap/README
 - updated incorrect comments
 - minor tweaks to a few log statements
 - renamed configure hook `getEdgeXServices` to `getEdgeXRefServices` and added clarifying comments
 - replaced a `switch` statement in `disableSecretStoreAndRestart`
 - made `serviceList` a loop var in `handleAllServices`
 - replace two `append(serviceList, ...)` statements with `[]string{ s }`

lenny-intel (approved):

siggiskulason (approved):
LGTM, thanks for the updates

anonymouse64 (approved):
lgtm, thanks for the changes

PR #3826, opened 2021-11-24
fix(snap): deploy secrets token into separate mount points

farshidtz:
This change is mostly a refactoring to make the edgex-secretstore-tokens interface connection compatible with the deferred startup feature: https://github.com/edgexfoundry/edgex-go/pull/3813. But since the change is already releases and led to an issue, I would call this a fix.

Fixes #3827

Signed-off-by: Farshid Tavakolizadeh <farshid.tavakolizadeh@canonical.com>

**If your build fails**  due to your commit message not passing the build checks, please review the guidelines here: https://github.com/edgexfoundry/edgex-go/blob/main/.github/Contributing.md

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [x] I am not introducing a new dependency (add notes below if you are)
- [ ] I have added unit tests for the new feature or bug fix (if not, why?)
- [x] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
- [ ] I have opened a PR for the related docs change (if not, why?)
  <link to docs PR>

#### Install from the branched release to test the auto-connection (device snap before):
```
# remove snaps
$ sudo snap remove edgex-device-camera edgexfoundry

# install device-camera
$ sudo snap install edgex-device-camera --edge
edgex-device-camera (edge) 2.1.0 from Canonical✓ installed

# install edgexfoundy from the branch
$ sudo snap install edgexfoundry --channel=edge/secrets-connect-fix
edgexfoundry (edge/secrets-connect-fix) 2.2.0-dev.1 from Canonical✓ installed

# check the connection, it should be connected automatically:
$ sudo snap connections edgexfoundry
Interface                         Plug                                  Slot                                         Notes
content[edgex-secretstore-token]  edgexfoundry:edgex-secretstore-token  edgex-device-camera:edgex-secretstore-token  -

# ensure that token is copied
$ sudo ls /var/snap/edgex-device-camera/current/device-camera/secrets-token.json
/var/snap/edgex-device-camera/current/device-camera/secrets-token.json
```

Optional (remove, disconnect, re-connect):
```
$ sudo rm /var/snap/edgex-device-camera/current/device-camera/secrets-token.json
$ sudo snap disconnect edgexfoundry:edgex-secretstore-token edgex-device-camera:edgex-secretstore-token
[...]

PR #3848, opened 2021-12-17
build(snap): update edgexfoundry snap base to core20

siggiskulason:
This commit includes a number of updates to migrate the base to core20.

- postgres: updating postgres to 12.9 from 10.19. This update includes changes to the
  post-refresh hook to handle the three cases of a) new install, b) refresh from
  a postgres-10 version and c) refresh from a postgres-12 version

- snapcraft-preload: updates to use latest version and fixed cmake setup

- redis: make-install-var replaced with make-parameters

- kong-daemon: setpriv package replaced with util-linux

## PR Checklist

- [x] I have added unit tests for the new feature or bug fix (if not, why?) See https://github.com/canonical/edgex-checkbox-provider/pull/21
- [ ] I have opened a PR for the related docs change (if not, why?) N/A
  <link to docs PR>

## Testing Instructions

Test by building the edgexfoundry snap and running a) checkbox tests and b) TAF tests to confirm that there are no regressions.

Test the refresh by doing
1: new install
```
sudo snap remove --purge edgexfoundry
sudo snap install ./edgexfoundry_2.2.0-dev.13_amd64.snap --dangerous
```
2: refresh from Ireland
```
sudo snap remove --purge edgexfoundry
sudo snap install edgexfoundry --channel=2.0/stable
sudo snap install ./edgexfoundry_2.2.0-dev.13_amd64.snap --dangerous
```
Expected result: This will fail, as the epoch has been updated to 6. We will provide a later build with a 6* epoch, which will contain the pre-refresh hook code required to upgrade.

3: refresh from current build
```
sudo snap remove --purge edgexfoundry
sudo snap install ./edgexfoundry_2.2.0-dev.13_amd64.snap --dangerous
sudo snap install ./edgexfoundry_2.2.0-dev.13_amd64.snap --dangerous
```


siggiskulason:
When I run the tests on this, then it passes all checkbox tests and it was passing all TAF tests but is now failing the TAF tests in the latest run and there seem to be some issues with the install hook and starting up postgres.. I will need to debug that and will push an update.

snap/snapcraft.yaml
```diff
@@ -637,7 +655,8 @@ parts:
     plugin: make
     # NOTE: make-install-var is deprecated
     # See https://snapcraft.io/docs/make-plugin
```
anonymouse64, line 657: This NOTE can be deleted

snap/local/runtime-helpers/bin/install-setup-postgres.sh
```diff
@@ -1,73 +1,114 @@
...
+# and also from the post-refresh hook. It handles these three scenerios:
+# - new install: create config file and database
+# - refresh from 2.0/stable or older - a snap using postgres 10
+# - refresh from a 2.1/stable or newer snap using postgres 12
+
+# in postgres 12 we are using these directories
+SNAP_DATA_CONFIG_POSTGRES_DIR="$SNAP_DATA/config/postgres"
+SNAP_DATA_CONFIG_POSTGRES_CONFIGFILE="$SNAP_DATA_CONFIG_POSTGRES_DIR/postgresql.conf"
+SNAP_DATA_POSTGRES_DIR="$SNAP_DATA/postgresql"
+SNAP_DATA_POSTGRES_DB_DIR="$SNAP_DATA_POSTGRES_DIR/data"
```
anonymouse64, line 13: I wonder if we should include the postgres version number in the directory path here, if we need to start supporting more versions of postgres when i.e. we move to `base: core22` in X years

snap/local/runtime-helpers/bin/install-setup-postgres.sh
```diff
@@ -1,73 +1,114 @@
...
-# modify postgres authentication config to use 'md5' (password)
-"$SNAP/bin/drop-snap-daemon.sh" sed -i -e "s@trust@md5@g" "$SNAP_DATA/postgresql/10/main/pg_hba.conf"
+if [ -d "$SNAP_DATA_POSTGRES_DB_DIR" ]; then
+  logger "edgexfoundry:install-setup-postgres.sh: refresh - using existing data directory"
+else
+  # set up the postgres data directory - generate a new one if we are upgrading from postgres 10
+
+  if [ -d "$SNAP_DATA_OLD_POSTGRES_DB_DIR" ]; then
+    logger "edgexfoundry:install-setup-postgres.sh: refresh - removing existing Postgres 10 data directory"
+    rm -rf $SNAP_DATA_OLD_POSTGRES_DIR
```
anonymouse64, line 46: should we do some sort of db migration rather than just remove the version 10 data and create brand new version 12 data? That feels slightly wrong to me, though I do realize that db migration is probably complicated to do in shell?

anonymouse64 (commented):
seems reasonable to me for the most part, but I do wonder if it's okay for us to just delete Postgres 10 database on refresh and create a fresh version 12 database...

I have not tested the changes locally

snap/snapcraft.yaml
```diff
@@ -569,6 +569,36 @@ parts:
...
+          # FILE_HASH=2be6414cdce1540c022acda76da55ef6bbd51c537dc2e3d4020652e72daec62d
+          ;;
+        arm64)
+          FILE_NAME=deck_${VERSION}_linux_arm64.tar.gz
+          # FILE_HASH=d4be8f68f0275701bcde71d130a997f5a32b47f91d5fb8a7eae7db666325a33b
+          ;;
+      esac
+      # download the archive and verify the checksum
+      # NOTE: the archive only includes the binary distribution,
+      # with no config or license files
```
anonymouse64, line 593: probably this means we need to manually add these ourselves

also did the project fully switch to using declarative config for Kong? I had not heard of deck before and just saw it here with your PR :-)

siggiskulason: No, the project hasn't (yet), but we can still use deck - it gets the configuration from the underlying Postgres database - and we can then import the configuration back in, which will populate the postgres 12 database. That saves us having to backup/restore the Postgres database itself - which is somewhat more complicated. I have a couple of more things to do before this is done - hence why it's still in draft :-)

snap/local/runtime-helpers/bin/install-setup-postgres.sh
```diff
@@ -1,73 +1,8 @@
...
-        exit 1
-    fi
-done
-
-# stop postgres again in case the security services should be turned off
-snapctl stop "$SNAP_NAME.postgres"
-
-# modify postgres authentication config to use 'md5' (password)
-"$SNAP/bin/drop-snap-daemon.sh" sed -i -e "s@trust@md5@g" "$SNAP_DATA/postgresql/10/main/pg_hba.conf"
+postgres_install
```
farshidtz, line 8: If you do add support for passing the function name as argument (as explained in the previous comment), you would no longer need this script. The install hook can directly call `kong-postgres-setup.sh postgres_install` in [here](https://github.com/siggiskulason/edgex-go/blob/78dc300c83601bfebc69cf8d2d04a9db17ae32de/snap/local/hooks/cmd/install/install.go#L188).

farshidtz: If you decided to remove this script, please add the existing above note to the other script to mention that install hooks calls it.

snap/local/runtime-helpers/bin/kong-postgres-setup.sh
```diff
@@ -0,0 +1,207 @@
+#!/bin/sh
+
+# this script is called from the install hook to set up postgres
+# and also from the post-refresh hook. It handles these three scenerios:
+# - new install: create config file and database
+# - refresh from 2.0/stable or older - a snap using postgres 10
+# - refresh from a 2.1/stable or newer snap using postgres 12
```
farshidtz, line 7: Please use versions instead of channels, otherwise this comment would quickly become outdated. The 2.1/stable channel currently points to a version which uses postgres 10 but that may change in future. We also have other channels which the given scenarios apply to: latest/stable, latest/beta, 2.x/beta, 2.1/edge.

Perhaps it is better to leave the channels and versions out of this comment and just mention the postgres versions. We'll have to use epochs to control which channels gets this update and that is beyond the scope of this script.

snap/local/runtime-helpers/bin/kong-postgres-setup.sh
```diff
@@ -0,0 +1,207 @@
...
+      logger "edgexfoundry:postgres: refresh - loading saved kong data"
+      cd $SNAP_COMMON
+      $SNAP/bin/deck diff > "diff.txt"
+      $SNAP/bin/deck sync | logger
+    fi
+}
+
+postgres_install() {
+     logger "edgexfoundry:postgres: postgres_install"
+
```
farshidtz, line 150: Please add comment to mention that this is called from the install hook as well as the post refresh function.

snap/snapcraft.yaml
siggiskulason: I ended up changing this The `deck` approach did end up having some issues so I went back to implementing this by dumping the postgres 10 database to a .sql file in the pre-refresh hook and then importing that back into a new postgres 12 database in the post-refresh hook.

Briefly, the main reason `deck` doesn't really work for us is:
- for `deck` to run and restore backed up data, Kong needs to be already running.
- I did try having a hook to run after the one shot security-proxy-setup service runs,  to add a call to “ $SNAP/bin/deck sync”
- That did in theory work, running after a new Kong installation has been set up on a new postgres 12 database, and restored the kong configuration
- However, the security-proxy-setup oneshot service has by that point created a new  kong configuration  with a new JWT token. When I run deck sync, the old configuration overrides that and the old JWT token is added to kong, whcih will cause issues.

So doing this with postgres directly in the pre-refresh and post-refresh hooks was a simpler and easier option.

snap/local/runtime-helpers/bin/install-setup-postgres.sh
siggiskulason: Updated this to include the version numbers

farshidtz:
We'll need to increment the epoch from `5` to `6` in this PR to prevent upgrades from existing deployments that have Postgres 10 installed but no way to backup the data in preperation for the Postgres 12 restore. This is to avoid data loss resulted from replacing Postgres 10 with 12; see https://www.postgresql.org/docs/12/upgrading.html

In future, another snap would be created with epoch `6*` to act as a bridge, based on the code before this PR, but including only the backup feature. That snap would be released to the `latest` track to provide a path for 2.0|2.1|2.2 to 2.2.x upgrades. Installations from the LTS 2.1 track will not be affected by this release.

For more details on how the epoch works: https://snapcraft.io/docs/snap-epochs

snap/hooks/connect-plug-edgex-secretstore-token
```diff
@@ -18,4 +18,4 @@
 # Note: Since v2.1.0, the security-secretstore-setup is started by the configure
 # hook instead of automatically after the installation.
 
-$SNAP/bin/security-secretstore-post-setup.sh "connect-plug-edgex-secretstore-token"
+$SNAP/snap/command-chain/snapcraft-runner $SNAP/bin/security-secretstore-post-setup.sh "connect-plug-edgex-secretstore-token"
```
farshidtz, line 21: I don't understand the need for this change. This PR doesn't modify `security-secretstore-post-setup.sh` and there isn't anything there which require the environment variables exported by `snapcraft-runner`.

siggiskulason: The reason was that I removed the following from snapcraft.yaml:

```
    connect-plug-edgex-secretstore-token:
      command-chain:
       - snap/command-chain/snapcraft-runner
```

so the command-chain was already calling the snapcraft-runner. If it's not needed, then we can just remove it.

farshidtz: Right. It was originally required for the `snapctl` command, which got removed as part of #3826. 

If I remember correctly, you had to add it back after removing the passthrough, because of some missing postgres library paths in one of your earlier commits. That is not the case anymore so we can go ahead and revert it.

snap/local/runtime-helpers/bin/kong-postgres-setup.sh
```diff
@@ -0,0 +1,281 @@
...
+CONFIG_KONG_PASSWORD_FILE="$SNAP_DATA/config/postgres/kongpw"
+
+ETC_POSTGRES_10_CONFIGFILE_DIR="$SNAP_DATA/etc/postgresql/10/main"
+
+POSTGRES_CONFIGFILENAME="postgresql.conf"
+
+POSTGRES_12_DB_DIR="$SNAP_DATA/postgresql/12/main" 
+
+# if we use the postgres binaries in /bin, then --devmode install fails.
+PSQL_BIN_PATH="$SNAP/usr/lib/postgresql/12/bin/"
```
farshidtz, line 22: Would be great to maintain a single naming convention. E.g.:
```suggestion
KONG_CONFIG_PASSWORD_DIR="$SNAP_DATA/config/postgres"
KONG_CONFIG_PASSWORD_FILE="$SNAP_DATA/config/postgres/kongpw"

POSTGRES_10_ETC_CONFIG_DIR="$SNAP_DATA/etc/postgresql/10/main"

POSTGRES_CONFIG_FILENAME="postgresql.conf"

POSTGRES_12_DB_DIR="$SNAP_DATA/postgresql/12/main" 

# if we use the postgres binaries in /bin, then --devmode install fails.
PSQL_BIN_PATH="$SNAP/usr/lib/postgresql/12/bin/"
```

snap/local/runtime-helpers/bin/kong-postgres-setup.sh
```diff
@@ -0,0 +1,281 @@
...
+    postgres_init_postgres_database
+
+    # and the config file
+    postgres_create_configuration_file
+
+    # start postgres up and wait a bit for it so we can create the database and user for kong
+    logger "edgex-kong-postgres-setup: install - starting postgres"
+    snapctl start "$SNAP_NAME.postgres"
+ 
+    postgres_create_kong_db
```
farshidtz, line 214: The waiting now happens in another function.
```suggestion
    # start postgres and create the database and user for kong
    logger "edgex-kong-postgres-setup: install - starting postgres"
    snapctl start "$SNAP_NAME.postgres"
 
    postgres_create_kong_db
```

farshidtz (changes requested):
Thanks a lot @siggiskulason for the updates. It looks very good overall. 

I've added a few suggestions inline.

snap/snapcraft.yaml
```diff
@@ -53,14 +53,11 @@ passthrough:
       environment:
         LC_ALL: C.UTF-8
         LANG: C.UTF-8
-        PGHOST: $SNAP_COMMON/sockets
         SNAPCRAFT_PRELOAD_REDIRECT_ONLY_SHM: 1
```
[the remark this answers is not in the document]
siggiskulason: yes, because they are not only for the kong-postgres part. Without them we get errors such as
```
- Run install hook of "edgexfoundry" snap if present (run hook "install": 
-----
mkdir: cannot create directory '/var/snap/edgexfoundry': Read-only file system
mkdir: cannot create directory '/var/snap/edgexfoundry': Read-only file system
mkdir: cannot create directory '/var/snap/edgexfoundry': Read-only file system
mkdir: cannot create directory '/var/snap/edgexfoundry': Read-only file system
```

farshidtz (approved):

PR #3856, opened 2022-01-07
fix(snap): Apply proxy's runtime config options after startup

### LINKED ISSUE #3863: Unable to seed runtime security-proxy config options via a gadget snap
# 🐞 Bug Report

The issue is located in: edgexfoundry snap, security proxy setup

Affected runtime securiy proxy options:
* env.security-proxy.user
* env.security-proxy.public-key
* env.security-proxy.tls-certificate
* env.security-proxy.tls-private-key
* env.security-proxy.tls-sni

Unaffected securiy proxy options:
* env.security-proxy.add-proxy-route

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 2.0.0

The edgexfoundry snap accepts three types of configuration options to
1. switch services on/off
2. set environment variables to be written to env files and read by services during statup (environment override config options)
3. take user input and call certain internal commands (runtime config options)

The changes introduced by #3813 modified the installation to start services after applying configuration options. This has affected the 3rd type of configuration options that rely on initial setup of services. Note that this is only a problem when the configuration options are set prior to the startup, currently only possible via a Gadget snap, or by manually modifying the install hook.

## 🔥 Exception or Error
E.g. error in journal:
```
edgexfoundry:configure: error handling services: failed to write file /var/snap/edgexfoundry/x3/secrets/security-proxy-setup/jwt-user-public-key.pem - open /var/snap/edgexfoundry/x3/secrets/security-proxy-setup/jwt-user-public-key.pem: no such file or directory)
```

## 🌍  Your Environment
**Deployment Environment:**
Snap

**EdgeX Version [**REQUIRED**]:**
2.1.0

**Anything else relevant?**

farshidtz:
Fixes #3863, see the issue for details.

~This PR changes the configure hook to skip configuration options of security-proxy in defer-startup mode. To apply those options, the configure hook is called again after security-proxy-setup completes.~

This PR changes the configure hook to skip **runtime** configuration options of security-proxy in **defer-startup** mode. Those config options should be applied after security-proxy-setup completes.

To achieve that:
a) The config option handling function is changed to ignore runtime config options of security-proxy in defer-startup install mode: https://github.com/canonical/edgex-snap-hooks/pull/25
b) The `configure` executable is extended with an `options` subcommand which takes a service name as follows:
```
./configure options --service=security-proxy
```
The above command is called in a `post-stop` script after the security-proxy setup. 

Prior to this change, the `configure` executable was called only as the snap configure hook. Future work may focus on file re-structuring of various executables under `snap` directory to add better visiblity into various helper executables.

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?) - gadget testing is WIP
  <link to docs PR> - not required

## Testing Instructions

The [checkbox tests](https://github.com/canonical/edgex-checkbox-provider/blob/b7b12b1a471067e7bf7d59d9e1016b34f6910213/data/latest/test-security-services-proxy-certs.sh) have coverage for the affected config options. Those tests were passing even before this fix because the config sets in those tests are executed after service startup.

To test the seeding, we should pass defaults from a Gadget snap:
```
defaults: 
  # edgexfoundry
  AZGf0KNnh8aqdkbGATNuRuxnt1GNRKkV:
    # Services that should auto start
    device-virtual: "on"
    support-scheduler: "on" 
    support-notifications: "on"
    
    env.security-proxy:
        # set a default user
        user: "admin,1,ES256"
        public-key: |
          -----BEGIN PUBLIC KEY-----
          MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAETuwPTr5tL/NxiTkUWh2r93i6yaQf
          XiFV8kNjemjuTpOTc6yZu+zgNyt1pXpHmPmtIZ8X4eE0QGCIjlnfAz0dQA==
          -----END PUBLIC KEY-----
```
The OS must boot without error and the config options must be applied.

To seed this snap into an Ubuntu Core image to test the above, use the build from this branch under channel: `latest/edge/fix-snap-proxy-config`

### DIFF
--- a/snap/local/runtime-helpers/bin/security-proxy-post-setup.sh
+++ b/snap/local/runtime-helpers/bin/security-proxy-post-setup.sh
@@ -1,0 +1,15 @@
+#!/bin/bash -e
+
+# This script is called as a post-stop-command when
+# security-proxy-setup oneshot service stops.
+#
+# Several config options depend on resources that only exist after proxy is 
+# setup. This scripts re-runs the configure hook after the deferred startup 
+# of security-proxy-setup to apply such configurations.
+
+logger "edgex-secretstore-proxy:post-setup: calling configure hook"
+
+# add bin directory to have e.g. secret-config in PATH
+export PATH="$SNAP/bin:$PATH"
+
+$SNAP/snap/hooks/configure
--- a/snap/local/runtime-helpers/bin/security-proxy-post-setup.sh
+++ b/snap/local/runtime-helpers/bin/security-proxy-post-setup.sh
@@ -2,14 +2,12 @@
 
 # This script is called as a post-stop-command when
 # security-proxy-setup oneshot service stops.
-#
-# Several config options depend on resources that only exist after proxy is 
-# setup. This scripts re-runs the configure hook after the deferred startup 
-# of security-proxy-setup to apply such configurations.
 
-logger "edgex-secretstore-proxy:post-setup: calling configure hook"
+logger "edgexfoundry:security-proxy-post-setup"
 
 # add bin directory to have e.g. secret-config in PATH
 export PATH="$SNAP/bin:$PATH"
 
-$SNAP/snap/hooks/configure
+# Several config options depend on resources that only exist after proxy 
+# setup. This re-applies the config options logic after deferred startup:
+$SNAP/snap/hooks/configure options --service=security-proxy
--- a/snap/local/hooks/cmd/configure/configure.go
+++ b/snap/local/hooks/cmd/configure/configure.go
@@ -28,8 +28,6 @@ import (
 	hooks "github.com/canonical/edgex-snap-hooks/v2"
 )
 
-var cli *hooks.CtlCli = hooks.NewSnapCtl()
-
 const ( // iota is reset to 0
 	kuiperService = iota
 	secProxyService
@@ -310,7 +308,6 @@ func handleAllServices(deferStartup bool) error {
 
 	// grab and log the current service configuration
 	for _, s := range hooks.Services {
-		var envJSON string
 		var serviceList []string
 
 		status, err := cli.Config(s)
@@ -320,23 +317,9 @@ func handleAllServices(deferStartup bool) error {
 
 		hooks.Debug(fmt.Sprintf("edgexfoundry:configure: service: %s status: %s", s, status))
 
-		if deferStartup && s == "security-proxy" {
-			hooks.Info("edgexfoundry:configure: defer-startup mode. Skip security-proxy config handling to after the startup")
-			continue
-		}
-
-		serviceCfg := hooks.EnvConfig + "." + s
-		envJSON, err = cli.Config(serviceCfg)
+		err = applyConfigOptions(s)
 		if err != nil {
-			err = fmt.Errorf("edgexfoundry:configure failed to read service %s configuration - %v", s, err)
-			return err
-		}
-
-		if envJSON != "" {
-			hooks.Debug(fmt.Sprintf("edgexfoundry:configure: service: %s envJSON: %s", s, envJSON))
-			if err := hooks.HandleEdgeXConfig(s, envJSON, nil); err != nil {
-				return err
-			}
+			return fmt.Errorf("failed to apply config options for %s: %v", s, err)
 		}
 
 		// if deferStartup is set, don't start/stop services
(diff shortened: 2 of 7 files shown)


lenny-intel:
@farshidtz , please amend you commit with a it `signed-off` to fix the DCO issue and you also must complete the PR check list. Provide reason for any not checked.

bnevis-i (approved):
LGTM.  Should present the feature at the next security WG meeting.

farshidtz:
Thanks for the reviews. I've amended the commit to add the signature.

Please note that the PR is a draft. I will continue testing and update the description before making it ready.

tonyespy (commented):
Nicely done, code changes look good to me. That said, I'd like a quick review from @anonymouse64 to ensure that there are no issues with re-running the configure hook from a script like this...

anonymouse64:
As mentioned in our meeting, there are no issues from snapd's perspective with running the configure hook script/program, since you're really just running some code here. Snapd doesn't care that the code being executed was also executed for the configure hook at all.

That being said, I didn't test this PR so I can't say whether it works, but the intention of the feature is very cool and I'm happy to see it come to fruition :smile:

snap/local/runtime-helpers/bin/security-proxy-post-setup.sh
```diff
@@ -0,0 +1,15 @@
...
+# Several config options depend on resources that only exist after proxy is 
+# setup. This scripts re-runs the configure hook after the deferred startup 
+# of security-proxy-setup to apply such configurations.
+
+logger "edgex-secretstore-proxy:post-setup: calling configure hook"
+
+# add bin directory to have e.g. secret-config in PATH
+export PATH="$SNAP/bin:$PATH"
+
+$SNAP/snap/hooks/configure
```
stolowski, line 15: As discussed on mattermost, I think this is problematic. If this is triggered during installation, it's going to be racing with "seed" change and may cause conflict error ("seed" change in progress...).

The conflict arises because your configure hook code tries to start services with snapctl and since it is triggered "externally" (in this case by systemd) snapd will not consider this to be part of the seeding (install) change.

farshidtz:
Thanks a lot for all the reviews, in particular to @anonymouse64 and @stolowski for rigorously testing the changes.

I've changed the logic to apply specific service-specific configurations after proxy setup, instead of calling the configure executable to perform a full config (as expected from the configure hook). We'll have to do some refactoring in future to better separate the "configure hook" from "configuring a service" without replicating code.

The PR description has been updated with details. This PR remains in draft until https://github.com/canonical/edgex-snap-hooks/pull/25 is merged and included.

siggiskulason (commented):
I've reviewed the code with no comments - it  looks fine to me. I'm proceeding to run checkbox/taf tests and then create the gadget and image.
- TAF functional tests: 5 failures (482 pass). The failures are not related to this PR and latest/edge results are identical.
- TAF integration tests: All pass

- Created a Ubuntu Core image as per instructions in Google Doc, with gadget-yaml specifying the public key for a new user. Then created a JWT token by copying the private key to the image with ​`scp -P8022 private.pem [username]@localhost:`
​, then ssh-ing onto  the box and doing

- `TOKEN=edgexfoundry.secrets-config proxy jwt --algorithm ES256 --private_key /var/snap/edgexfoundry/current/private.pem --id 1 --expiration=1h` (with "1") being the user ID used in the gadget.yaml
- `edgexfoundry.curl -k -X GET https://localhost:8443/core-data/api/v2/ping? -H "Authorization: Bearer $TOKEN"`
(Note that I could obviously have run edgexfoundry.secrets-config proxy jwt on my desktop instead to generate the token)

and this worked fine.

siggiskulason (approved):
LGTM - Approved - with tests passing as per my comment.

farshidtz:
Thanks a lot @siggiskulason for the review and testing.

> * Then created a JWT token by copying the private key to the image with ​`scp -P8022 private.pem [username]@localhost:`
>   ​, then ssh-ing onto  the box and doing
> * `TOKEN=edgexfoundry.secrets-config proxy jwt --algorithm ES256 --private_key /var/snap/edgexfoundry/current/private.pem --id 1 --expiration=1h` (with "1") being the user ID used in the gadget.yaml
> (Note that I could obviously have run edgexfoundry.secrets-config proxy jwt on my desktop instead to generate the token)

Or you could generate the token on your machine using openssl as instructed here (3. alternative approach): https://github.com/edgexfoundry/edgex-go/tree/main/snap#jwt-tokens

PR #3864, opened 2022-01-19
ci: Go 1.17 related changes

ernestojeda:
Changes related to Go 1.17 and cache optimizations

### DIFF
--- a/cmd/support-scheduler/Dockerfile
+++ b/cmd/support-scheduler/Dockerfile
@@ -16,7 +16,7 @@
 # 
 #  ----------------------------------------------------------------------------------
 
-ARG BUILDER_BASE=golang:1.16-alpine3.14
+ARG BUILDER_BASE=golang:1.17-alpine3.15
 FROM ${BUILDER_BASE} AS builder
 
 WORKDIR /edgex-go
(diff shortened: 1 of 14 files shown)


cmd/support-scheduler/Dockerfile
```diff
@@ -16,7 +16,7 @@
 # 
 #  ----------------------------------------------------------------------------------
 
-ARG BUILDER_BASE=golang:1.16-alpine3.14
+ARG BUILDER_BASE=golang:1.17-alpine3.15
```
lenny-intel, line 19: Other changes missing?

lenny-intel (changes requested):
Also need to update go version used by snap here: https://github.com/edgexfoundry/edgex-go/blob/b0e5e643282aa8d6ca4ca25954737fd68a04a1f7/snap/snapcraft.yaml#L584

lenny-intel (approved):
LGTM

lenny-intel (approved):

snap/snapcraft.yaml
```diff
@@ -581,7 +581,7 @@ parts:
     # see comment for static-packages part about specifying a source part here
     source: snap/local/build-helpers
     build-snaps:
-      - go/1.16/stable
+      - go/1.17/stable
```
farshidtz, line 584: This doesn't look right.

farshidtz (changes requested):

snap/snapcraft.yaml
farshidtz: I was too slow. It is already fixed..

ernestojeda: yeah, had a bad find/replace. 🥺

lenny-intel (approved):
LGTM

PR #3888, opened 2022-02-25
feat(snap): add secretstore token for edgex-ekuiper snap

MonicaisHer:
This PR adds edgex-ekuiper as an [add-on service](https://docs.edgexfoundry.org/2.2/security/Ch-Configuring-Add-On-Services/) to `secretStoreTokens` and `secretStoreKnownSecrets` (for redisdb) lists.

The Vault token is required for the standalone deployment of eKuiper with snap, where a script queries the Redis credentials after authenticating with Vault. The snap receives the vault token using the `edgex-secretstore-token` content interface or via other means.

Signed-off-by: Mengyi Wang <mengyi.wang@canonical.com>

## PR Checklist

  <link to docs PR>

### DIFF
--- a/snap/local/hooks/cmd/install/install.go
+++ b/snap/local/hooks/cmd/install/install.go
@@ -56,6 +56,7 @@ var secretStoreTokens = []string{
 	"device-grove",
 	"device-uart",
 	"device-rfid-llrp",
+	"edgex-ekuiper",
 }
 
 var secretStoreKnownSecrets = []string{
@@ -79,6 +80,7 @@ var secretStoreKnownSecrets = []string{
 	"redisdb[device-grove]",
 	"redisdb[device-uart]",
 	"redisdb[device-rfid-llrp]",
+	"redisdb[edgex-ekuiper]",
 }
 
 // services w/configuration that needs to be copied


snap/snapcraft.yaml
```diff
@@ -67,7 +67,6 @@ plugs:
   # the service-specific directory under $SNAP_DATA/secrets.
   edgex-secretstore-token:
     interface: content
-    content: edgex-secretstore-token
```
farshidtz, line 70: Please revert this so that the PR only updates the necessary files.

farshidtz (changes requested):
This is a feature addition. Please change the PR title prefix.

Also, the commits need to have semantic prefixes.

snap/snapcraft.yaml
MonicaisHer: Thanks, reverted.

lenny-intel:
eKuiper doesn't use `secretStoreTokens` and `secretStoreKnownSecrets`, we inject the Redis password into their config. This PR is not needed.

farshidtz:
> eKuiper doesn't use `secretStoreTokens` and `secretStoreKnownSecrets`, we inject the Redis password into their config. This PR is not needed.

@lenny-intel, the embedded eKuiper doesn't. This is for a standalone ekuiper snap. I've just updated the PR description to:
> This PR adds edgex-ekuiper as an [add-on service](https://docs.edgexfoundry.org/2.2/security/Ch-Configuring-Add-On-Services/) to `secretStoreTokens` and `secretStoreKnownSecrets` (for redisdb) lists.
>
>The Vault token is required for the standalone deployment of eKuiper with snap, where a script queries the Redis credentials after authenticating with Vault. The snap receives the vault token using the `edgex-secretstore-token` content interface or via other means.

The standalone edgex-ekuiper snap is [work in progress](https://github.com/MonicaisHer/edgex-ekuiper-snap). But we are almost sure that [using the vault token to query the redis credentials](https://github.com/MonicaisHer/edgex-ekuiper-snap/blob/main/snap/local/runtime-helpers/bin/redis-token-setup.sh) is the cleanest way forward. This makes the edgex-ekuiper's server configurations (mostly unrelated to security) decoupled from an internal security component of EdgeX. Moreover, this makes edgex-ekuiper's interaction with the platform similar to other services.

I think we could eventually have a component which does this both for snapped and dockerized ekuiper. For example an edgex-ekuiper-setup app that takes a vault token and config file templates, queries the redis credentials, and outputs the config files. What do you think?

lenny-intel (approved):
LGTM

farshidtz (approved):

Issue #3612, opened 2021-07-08
Utilize MaxRequestSize configuration in each microservice


cloudxxx8:
# 🚀 Feature Request

### Relevant Package

 This feature request is for all services

### Description
MaxRequestSize configuration is not currently used. 
It defines the maximum size of http request body in bytes.
Thus, we should implement this feature for all microservices.

jinlinGuan:
Currently the MaxRequestSize is set to 0 (unlimited) by default. Do we need a reasonable default limit?
Also, do we need an upper limit?

cloudxxx8:
I don't think we need an upper limit because there is no upper limit in other configuration and 0 (unlimited) is the valid value for the backward compatible.
Let's change the default value to 10240 KB (10 MB) for now.

farshidtz:
Making the max request size configurable is a backward compatible change as long as its default value remains the same.

Reducing the default max size from unlimited to a limited range will break existing applications that do pass larger bodies.

This is a helpful reference:
https://docs.microsoft.com/en-us/dotnet/core/compatibility/#behavioral-changes

> ❌ DISALLOWED: Decreasing the range of accepted values for a property or parameter
>
> ❌ DISALLOWED: Changing the default value of a property, field, or parameter

cloudxxx8:
ok, let's keep the same default value 0

jpwhitemn:
work already completed per PR above

cloudxxx8:
we need to wait for https://github.com/edgexfoundry/go-mod-bootstrap/pull/321 is applied to the go.mod

PR #3891, opened 2022-03-03
feat(data): Make MaxEventSize a service configuration setting

### LINKED ISSUE #3237: feat(data): Make MaxEventSize a service configuration setting
To avoid large data causing unexpected memory exhaustion when decoding the large payload (CBOR or JSON), maxEventSize was introduced as a limit 25MB for handling CBOR payload in edgex-go.  More details could be found at https://github.com/edgexfoundry/edgex-go/issues/2439

@lenny-intel made a good point in https://github.com/edgexfoundry/edgex-go/pull/3202#discussion_r589521540_, we should consider to make such limit configurable, so that users could adjust the event size value per system requirements.  Open this issue to track the status of such change.

jinlinGuan:
close #3237 
25000KB (25MB) by default

Signed-off-by: Ginny Guan <ginny@iotechsys.com>

## PR Checklist

- [ ] I have opened a PR for the related docs change (if not, why?) will open once this PR approved
  <link to docs PR>

## Testing Instructions

Build and run Core Data locally with rest of EdgeX non-secure services running in Docker, and 
- set maxEventSize to -1  
- set maxEventSize to 0
- set maxEventSize to 1 and payload size over 1000 bytes 

 POST an event
e.g. 
```
curl -ki http://localhost:59880/api/v2/event/Random-Integer-Device/Random-Integer-Device/Int8 -X POST -d '{  "apiVersion": "v2",
  "event": {
    "apiVersion": "v2",
    "deviceName": "Random-Integer-Device",
    "profileName": "Random-Integer-Device",
    "sourceName": "Int8",
    "id": "d5471d59-2810-419a-8744-18eb8fad3d65",
    "origin": 1602168089665565200, 
    "readings": [
      {
        "deviceName": "Random-Integer-Device",
        "resourceName": "Int8",
        "profileName": "Random-Integer-Device",
        "origin": 1602168089665565200,
        "valueType": "Int8",
        "value": "12"
      }
    ]
  }
}'
```

For message bus,
e.g. `curl -ki http://localhost:59900/api/v2/device/name/Random-Binary-Device/Binary\?ds-pushevent\=yes`


lenny-intel:
@jinlinGuan , please remember to complete the PR template when you take this PR out of draft.

cmd/core-data/res/configuration.toml
```diff
@@ -1,3 +1,5 @@
+MaxEventSize = 25000  # Defines the maximum event size in kilobytes
```
lenny-intel, line 1: Why are we adding a new setting when we already have Service.MaxRequestSize, which is currently not used. Are you trying to implement this issue?
https://github.com/edgexfoundry/edgex-go/issues/3612#issuecomment-1057731152

cloudxxx8: @lenny-intel we had the same conversation before.  please see https://github.com/edgexfoundry/edgex-go/issues/3237#issuecomment-926760249

cloudxxx8:
Let's hold this PR and discuss the design here https://github.com/edgexfoundry/edgex-go/issues/3237#issuecomment-1060334694

internal/io/event.go
```diff
@@ -17,14 +16,16 @@ import (
 // a reasonable limit appropriate for handling CBOR payload in edgex-go.  More details could be found at
 // https://github.com/edgexfoundry/edgex-go/issues/2439
 
-func ReadAddEventRequestInBytes(w http.ResponseWriter, reader io.ReadCloser, size int64) ([]byte, errors.EdgeX) {
-	// use MaxBytesReader for limiting the size of incoming request bodies to avoid unexpected memory exhaustion
-	bytes, err := io.ReadAll(http.MaxBytesReader(w, reader, size))
+func CheckPayloadSize(reader io.Reader, size int64) ([]byte, errors.EdgeX) {
```
cloudxxx8, line 19: change the first parameter to `payload []byte`
rename the second parameter to `sizeLimit`
only return `errors.EdgeX`

internal/io/event.go
```diff
@@ -17,14 +16,16 @@ import (
...
+	// Treat 0 as unlimit size
+	payload, err := io.ReadAll(reader)
```
cloudxxx8, line 21: move io.ReadAll to http controller because message bus doesn't need this operation

cloudxxx8 (changes requested):

internal/io/event.go
jinlinGuan: updated

internal/io/event.go
jinlinGuan: updated

internal/io/event.go
```diff
@@ -1,28 +1,26 @@
...
 
-func ReadAddEventRequestInBytes(reader io.Reader) ([]byte, errors.EdgeX) {
-	// use LimitReader with defaultMaxEventSize to avoid unexpected memory exhaustion
-	bytes, err := io.ReadAll(io.LimitReader(reader, defaultMaxEventSize))
-	if err != nil {
-		return nil, errors.NewCommonEdgeX(errors.KindIOError, "AddEventRequest I/O reading failed", err)
+func CheckPayloadSize(payload []byte, sizeLimit int64) errors.EdgeX {
+	// Treat 0 as unlimit size
+	if sizeLimit < 0 {
+		return errors.NewCommonEdgeX(errors.KindIOError, "AddEventRequest I/O reading failed", nil)
```
cloudxxx8, line 21: use kind bad request error, and modify the error message to the proper content, such as:
`"sizeLimit cannot be lower than 0, current sizeLimit: %d", sizeLimit`

internal/io/event.go
```diff
@@ -1,28 +1,26 @@
...
 import (
-	"io"
+	"fmt"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/v2/errors"
 )
 
 // To avoid large data causing unexpected memory exhaustion when decoding CBOR payload, defaultMaxEventSize was introduced as
 // a reasonable limit appropriate for handling CBOR payload in edgex-go.  More details could be found at
 // https://github.com/edgexfoundry/edgex-go/issues/2439
```
cloudxxx8, line 16: remove this comment, because this function is not for CBOR only.

internal/io/event.go
```diff
@@ -1,28 +1,26 @@
 //
-// Copyright (C) 2021 IOTech Ltd
+// Copyright (C) 2021-2022 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
 package io
```
cloudxxx8, line 6: rename and move this file to internal/pkg/utils/payload.go

cloudxxx8 (changes requested):

internal/io/event.go
jinlinGuan: updated

internal/io/event.go
jinlinGuan: removed

internal/io/event.go
jinlinGuan: renamed and moved

cloudxxx8 (approved):
LGTM

internal/io/reader.go
```diff
@@ -79,3 +79,12 @@ func (yamlDtoReader) Read(reader io.Reader, v interface{}) errors.EdgeX {
 
 	return nil
 }
+
+func ReadAll(reader io.Reader) ([]byte, errors.EdgeX) {
+	bytes, err := io.ReadAll(reader)
+	if err != nil {
+		return nil, errors.NewCommonEdgeX(errors.KindIOError, "AddEventRequest I/O reading failed", nil)
```
cloudxxx8, line 86: modify the error message to general purpose

cloudxxx8 (changes requested):

internal/io/reader.go
jinlinGuan: moved this to `internal/core/data/controller/http/event.go`

cloudxxx8 (dismissed):
LGTM

internal/core/data/controller/http/event_test.go
```diff
@@ -266,6 +266,70 @@ func TestAddEvent(t *testing.T) {
...
+	tests := []struct {
+		Name               string
+		MaxEventSize       int64
+		RequestContentType string
+		errorExpected      bool
+		ExpectedStatusCode int
+	}{
+		{"Valid - AddEventRequest JSON with valid event size", 25000, common.ContentTypeJSON, false, http.StatusCreated},
+		{"Valid - AddEventRequest JSON with valid event size", 0, common.ContentTypeJSON, false, http.StatusCreated},
+		{"Invalid - AddEventRequest JSON with negative evnet size", -1, common.ContentTypeJSON, true, http.StatusBadRequest},
```
lenny-intel, line 284: How about another error test where `MaxEventSize=10`  to tests the actual capability we are adding here.

lenny-intel (changes requested):

lenny-intel (changes requested):

internal/pkg/utils/payload_test.go
```diff
@@ -0,0 +1,35 @@
...
+func TestCheckPayloadSize(t *testing.T) {
+	payload := make([]byte, 10)
+	tests := []struct {
+		name            string
+		sizeLimit       int64
+		errorExpected   bool
+		expectedErrKind errors.ErrKind
+	}{
+		{"Valid Size", int64(len(payload)), false, ""},
+		{"Invalid Size", int64(len(payload) - 1), true, errors.KindLimitExceeded},
```
cloudxxx8, line 20: add the invalid case for the payload size is larger than the default 25kb

internal/core/data/controller/messaging/subscriber.go
```diff
@@ -1,17 +1,19 @@
 //
-// Copyright (C) 2021 IOTech Ltd
+// Copyright (C) 2021-2022 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
-package application
+package messaging
```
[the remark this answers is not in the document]
cloudxxx8: I opened https://github.com/edgexfoundry/edgex-go/issues/3902 for this change

internal/core/data/controller/http/event.go
```diff
@@ -71,8 +73,15 @@ func (ec *EventController) AddEvent(w http.ResponseWriter, r *http.Request) {
...
 
 	var addEventReqDTO requestDTO.AddEventRequest
+	var err errors.EdgeX
+
+	dataBytes, readErr := io.ReadAll(r.Body)
+	if readErr != nil {
+		err = errors.NewCommonEdgeX(errors.KindIOError, "AddEventRequest I/O reading failed", nil)
+	} else {
+		err = utils.CheckPayloadSize(dataBytes, config.MaxEventSize*1000)
+	}
```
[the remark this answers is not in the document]
cloudxxx8: r.ContentLength is a better option, but why do we limit the check for CBOR only?

internal/core/data/controller/messaging/subscriber.go
```diff
@@ -57,6 +59,12 @@ func SubscribeEvents(ctx context.Context, dic *di.Container) errors.EdgeX {
 			case msgEnvelope := <-messages:
 				lc.Debugf("Event received on message queue. Topic: %s, Correlation-id: %s ", messageBusInfo.SubscribeTopic, msgEnvelope.CorrelationID)
 				event := &requests.AddEventRequest{}
+				maxEventSize := dataContainer.ConfigurationFrom(dic.Get).MaxEventSize
+				edgeXerr := utils.CheckPayloadSize(msgEnvelope.Payload, maxEventSize*1000)
+				if edgeXerr != nil {
+					lc.Errorf("event size exceed MaxEventSize(%d KB)", maxEventSize)
+					break
+				}
```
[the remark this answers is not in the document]
cloudxxx8: After this check, the `unmarshalPayload` can be saved, and the large payload won't be forward to another messaging topic.
For JSON, it's more important, the exceeded payload won't be stored in the database.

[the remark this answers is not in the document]
lenny-intel: @cloudxxx8 , the issue https://github.com/edgexfoundry/edgex-go/issues/3237 was written in the context of CBOR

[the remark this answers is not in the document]
lenny-intel: So` why` comment is that `decoding the large payload may cause memory so checking before decoding`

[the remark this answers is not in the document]
cloudxxx8: I just updated its description

internal/pkg/utils/payload_test.go
jinlinGuan: Do you mean we use 25kb here?
https://github.com/edgexfoundry/edgex-go/blob/dc6b798737b202c23945748eed2b99c1986bd439/internal/pkg/utils/payload_test.go#L12

internal/core/data/controller/http/event.go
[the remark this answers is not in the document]
cloudxxx8: @lenny-intel Although the request usually contains ContentLength, it's not a required field.
Should we depend on it?  It might be unknown.
https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.4

> 	// ContentLength records the length of the associated content.
> 	// The value -1 indicates that the length is unknown.
> 	// Values >= 0 indicate that the given number of bytes may
> 	// be read from Body.
> 	//
> 	// For client requests, a value of 0 with a non-nil Body is
> 	// also treated as unknown.
> 	ContentLength [int64](https://pkg.go.dev/builtin#int64)

I suggest to check ContentLength first, and keep the same `utils.CheckPayloadSize(dataBytes, config.MaxEventSize*1000)` step

internal/core/data/controller/messaging/subscriber.go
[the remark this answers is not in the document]
jinlinGuan: opened the PR https://github.com/edgexfoundry/edgex-go/pull/3903.

internal/core/data/controller/http/event.go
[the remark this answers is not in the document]
lenny-intel: Yes, check first, if -1 use `utils.CheckPayloadSize`

lenny-intel (commented):
Looks good once address @cloudxxx8 comment on ContentLength

cloudxxx8 (approved):
LGTM

lenny-intel (approved):

PR #3913, opened 2022-03-18
fix(metadata): ignore validation result if device service unavailable

### LINKED ISSUE #3909: Log and ignore the validation error if the Device Service is not running when creating and updating Device
# 🐞 Bug Report

 The issue is located in: core-metadata

### Is this a regression?
 Yes, the previous version in which this bug was not present,

Device creation and updates are blocked if the Device Service is not really running.
This is caused by https://github.com/edgexfoundry/edgex-go/pull/3900

For the backward compatibility, just log the error and continue the creation and update in this situation

hahattan:
fix #3909 

## PR Checklist

  <link to docs PR>

## Testing Instructions

1. Run core-metadata from this branch
2. run any device-service and then stop them
3. add a new device under that device-service and verify request succeed

### DIFF
--- a/internal/core/metadata/application/notify.go
+++ b/internal/core/metadata/application/notify.go
@@ -74,7 +74,13 @@ func validateDeviceCallback(ctx context.Context, dic *di.Container, device dtos.
 	request := requests.NewAddDeviceRequest(device)
 	_, err = deviceServiceCallbackClient.ValidateDeviceCallback(ctx, request)
 	if err != nil {
-		return errors.NewCommonEdgeX(errors.KindServerError, "device validation failed", err)
+		// TODO: reconsider the validity in v3
+		// allow this case for the backward-compatability in v2
+		if err.Code() == http.StatusServiceUnavailable {
+			lc.Warnf("Skipping device validation for device %s (device service %s unavailable)", device.Name, device.ServiceName)
+		} else {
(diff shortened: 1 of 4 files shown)


internal/core/metadata/application/notify.go
```diff
@@ -74,7 +74,13 @@ func validateDeviceCallback(ctx context.Context, dic *di.Container, device dtos.
 	request := requests.NewAddDeviceRequest(device)
 	_, err = deviceServiceCallbackClient.ValidateDeviceCallback(ctx, request)
 	if err != nil {
-		return errors.NewCommonEdgeX(errors.KindServerError, "device validation failed", err)
+		// TODO: reconsider the validity in v3
+		// allow this case for the backward-compatability in v2
+		if err.Code() == http.StatusServiceUnavailable {
+			lc.Infof("Skipping device validation for device %s (%s unavailable)", device.Name, device.ServiceName)
```
cloudxxx8, line 80: use Warnf

cloudxxx8 (changes requested):

internal/core/metadata/application/notify.go
cloudxxx8: "Skipping device validation for device %s due to **device service** %s unavailable"

cloudxxx8 (approved):
LGTM

PR #3915, opened 2022-03-18
fix(metadata): add 404 case for not-yet-upgraded device-service

hahattan:

## PR Checklist

  <link to docs PR>

## Testing Instructions

1. run core-metadata from this branch
2. run device-service which is not yet upgraded sdk to v2.2.0-dev.16
3. verify there's no 404 error and device can be successfully added to metadata

### DIFF
--- a/internal/core/metadata/application/notify.go
+++ b/internal/core/metadata/application/notify.go
@@ -78,6 +78,8 @@ func validateDeviceCallback(ctx context.Context, dic *di.Container, device dtos.
 		// allow this case for the backward-compatability in v2
 		if err.Code() == http.StatusServiceUnavailable {
 			lc.Warnf("Skipping device validation for device %s (device service %s unavailable)", device.Name, device.ServiceName)
+		} else if err.Code() == http.StatusNotFound {
+			lc.Warnf("Skipping device validation for device %s (device service %s < v2.2)", device.Name, device.ServiceName)
 		} else {
 			return errors.NewCommonEdgeX(errors.KindServerError, "device validation failed", err)
 		}


internal/core/metadata/application/notify.go
```diff
@@ -78,6 +78,8 @@ func validateDeviceCallback(ctx context.Context, dic *di.Container, device dtos.
 		// allow this case for the backward-compatability in v2
 		if err.Code() == http.StatusServiceUnavailable {
 			lc.Warnf("Skipping device validation for device %s (device service %s unavailable)", device.Name, device.ServiceName)
+		} else if err.Code() == http.StatusNotFound {
+			lc.Warnf("Skipping device validation for device %s (device service %s not upgraded)", device.Name, device.ServiceName)
```
farshidtz, line 82: The error message should be meaningful out of the context of this PR:
```suggestion
			lc.Warnf("Skipping device validation for device %s (device service %s <= v2.1)", device.Name, device.ServiceName)
```

farshidtz (changes requested):
Thanks. I can confirm that it solves the issue.

internal/core/metadata/application/notify.go
cloudxxx8: it should be `<= v2.2`

cloudxxx8: `the device service validation API is available in v2.2 and later version`

farshidtz: I suggested `<= v2.1` because several `v2.2.0-dev` versions do support it and those are smaller than v2.2. See https://semver.org/#spec-item-11

So, `<= v2.2` is not correct, both `< v2.2` and `<= v2.1` are inaccurate.

"the device service validation API is available in v2.2 and later version" sounds good to me.

```
lc.Warnf("Skipping device validation for device %s (device service %s): validation is available in v2.2 and later", 
    device.Name, device.ServiceName)
```

farshidtz (approved):
Thanks. The error message may be confusing until v2.2 is released (see previous comments), but that's okay.

PR #3945, opened 2022-04-06
fix(all): return 416 code when count is zero and offset > count

### LINKED ISSUE #3940: Query device all with out of range offset will return 200 when no device existed
# 🐞 Bug Report

 The issue is located in: metadata, command

### Is this a regression? No

Prerequisite: No device is created
Send API
```
GET Method
 http://localhost/api/v2/device/all?offset=8
```

## 🔥 Exception or Error
<pre><code>

Status code: 200
Should be return 416 status code
</code></pre>

## 🌍  Your Environment
**Deployment Environment:**

**EdgeX Version [**REQUIRED**]:** Main

**Anything else relevant?**

judehung:
The original implementation will return nil result set and nil error when count=0.  With such implementation, when query multiple entities and there is no entity in the DB, specify a out-of-range offset param will result in incorrect 200 status code rather than expected 416 status code.  

This commit fixes this issue by moving the check of offset > count with higher precedence over the check of count = 0, so that the APIs can return 416 status code when offset > count && count =0.

fixes #3940 

Signed-off-by: Jude Hung <jude@iotechsys.com>

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?) N/A, this PR is fixing the infrastructure layer logic.
- [ ] I have opened a PR for the related docs change (if not, why?) N/A, this PR is fixing the internal error determination.

## Testing Instructions

1. Run core-metadata from this branch without any devices or device profiles registered
2. Invoke one of the following APIs:
    ```
    GET metadata:59881/api/v2/device/all?offset=8
    GET metadata:59881/api/v2/deviceprofile/all?offset=8
    GET command:59882/api/v2/device/all?offset=8
    ```
3. Verify that response status code 416 is returned

### DIFF
--- a/internal/pkg/infrastructure/redis/queries.go
+++ b/internal/pkg/infrastructure/redis/queries.go
@@ -71,10 +71,11 @@ func getObjectsBySomeRange(conn redis.Conn, command string, key string, offset i
 		end = limit
 	}
 	count, _ := redis.Int(conn.Do(ZCOUNT, key, InfiniteMin, InfiniteMax))
-	if count == 0 { // return nil slice when there is no records in the DB
-		return nil, nil
-	} else if count > 0 && start > count { // return RangeNotSatisfiable error when start is out of range
+	if start > count { // return RangeNotSatisfiable error when start is out of range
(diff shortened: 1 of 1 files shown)


cloudxxx8 (approved):
LGTM

Issue #3901, opened 2022-03-14
Create new Core Data metric for Events received


lenny-intel:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]
Core Data

### Description [**REQUIRED**]
Create service metric that counts Events received. 

### Describe the solution you'd like
Use the new Metrics Manager define in this issue to create and store the metric :https://github.com/edgexfoundry/go-mod-bootstrap/issues/307 

### Describe alternatives you've considered
none

PR #3965, opened 2022-04-13
fix: Add missing packages for local docker builds

jim-wang-intel:
After CGO go build flag is introduced for core services built, several Dockerfiles of core-services are missing some base build packages; hence this fix.

Fixes: #3901

## PR Checklist

  <link to docs PR>

## Testing Instructions

One can just try any make docker targets that are using CGO , like `make docker_core_metadata` on the console command line.  It should be built without any build error.

### DIFF
--- a/cmd/core-command/Dockerfile
+++ b/cmd/core-command/Dockerfile
@@ -26,7 +26,7 @@ WORKDIR /edgex-go
 
 RUN sed -e 's/dl-cdn[.]alpinelinux.org/nl.alpinelinux.org/g' -i~ /etc/apk/repositories
 
-RUN apk add --update --no-cache make git
+RUN apk add --update --no-cache make git zeromq-dev libsodium-dev pkgconfig build-base
 
 COPY go.mod vendor* ./
 RUN [ ! -d "vendor" ] && go mod download all || echo "skipping..."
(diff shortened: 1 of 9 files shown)


cmd/core-command/Dockerfile
```diff
@@ -26,7 +26,7 @@ WORKDIR /edgex-go
 
 RUN sed -e 's/dl-cdn[.]alpinelinux.org/nl.alpinelinux.org/g' -i~ /etc/apk/repositories
 
-RUN apk add --update --no-cache make git
+RUN apk add --update --no-cache make git zeromq-dev libsodium-dev pkgconfig build-base
```
lenny-intel, line 29: It is more than just the packages in the build stage. Also need ZMQ in the resulting image in each of these docker files.

https://github.com/edgexfoundry/edgex-go/blob/main/cmd/core-data/Dockerfile#L52

bnevis-i:
For what it's worth, what's been done so far fixes the build.

cmd/core-command/Dockerfile
jim-wang-intel: added

bnevis-i (approved):

PR #3976, opened 2022-04-18
fix(metadata): not trigger DS callback if only lastConnected is updated

### LINKED ISSUE #3641: Update lastConnected should not trigger the Device Service callback
### Affected Services

 The issue is located in: core-metadata

### Is this a regression?
No

When lastConnected timestamp is updated, it also triggers the Device Service callback, which is unnecessary.

hahattan:
fix #3641

## PR Checklist

- [ ] I have opened a PR for the related docs change (if not, why?) n/a
  <link to docs PR>

## Testing Instructions

1. run metadata from this branch
2. run device-virtual with `DEBUG` log level
3. verify PATCH `/api/v2/deviceprofile` will not trigger DS callback if only updating `lastConnected` field

### DIFF
--- a/internal/core/metadata/application/deviceservice.go
+++ b/internal/core/metadata/application/deviceservice.go
@@ -1,5 +1,5 @@
 //
-// Copyright (C) 2020-2021 IOTech Ltd
+// Copyright (C) 2020-2022 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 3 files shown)


internal/core/metadata/application/deviceservice.go
```diff
@@ -77,6 +77,11 @@ func PatchDeviceService(dto dtos.UpdateDeviceService, ctx context.Context, dic *
 		"DeviceService patched on DB successfully. Correlation-ID: %s ",
 		correlation.FromContext(ctx),
 	)
+	// Don't invoke callback if only lastConnected field is updated
+	if dto.LastConnected != nil && dto.Description == nil && dto.LastReported == nil &&
+		dto.BaseAddress == nil && dto.AdminState == nil && len(dto.Labels) == 0 {
```
cloudxxx8, line 82: It's a good idea to not invoke callback if only lastConnected field is updated.
However, this if statement is hard to maintain if we add more fields into the Device struct.
Please write a helper function and use the reflect package to replace this if statement.
https://pkg.go.dev/reflect
Use a for loop to go through all the fields in this DTO and check the logic you designed.

cloudxxx8 (changes requested):

cloudxxx8 (approved):
LGTM

lenny-intel (approved):
LGTM

PR #3924, opened 2022-03-24
feat(snap): use updated environment variable injection

siggiskulason:
Signed-off-by: Siggi Skulason <siggi.skulason@canonical.com>

This PR updates edgex-go to use the new environment variable injection. 

```
apps.<app>.config.<my.env.var> -> setting env var MY_ENV_VAR for an app (e.g. apps.core-data.config.service.port=1000)
config.<my.env.var> -> setting env variable for all apps (e.g. config.service.host=localhost)
```
 

**If your build fails**  due to your commit message not passing the build checks, please review the guidelines here: https://github.com/edgexfoundry/edgex-go/blob/main/.github/Contributing.md

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [x] I am not introducing a new dependency (add notes below if you are)
- [ ] I have added unit tests for the new feature or bug fix (if not, why?) N/A - they are in the imported module
- [X] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
- [ ] I have opened a PR for the related docs change (if not, why?) N/A
  <link to docs PR>

## Testing Instructions

Do
```
sudo snap set edgexfoundry env.core-data.service.host=localhost
sudo snap set edgexfoundry apps.core-data.config.service.port=8080
sudo snap set edgexfoundry config.debug=true
```

1) Confirm that all .env files have DEBUG=true
2) Confirm that the core-data .env file also has both the SERVICE_HOST and SERVICE_PORT settings

```
$ more /var/snap/edgexfoundry/current/config/core-data/res/core-data.env 
export SERVICE_HOST=localhost
export DEBUG=true
export SERVICE_PORT=8080
$ more /var/snap/edgexfoundry/current/config/core-command/res/core-command.env 
export DEBUG=true


```

PR #3986, opened 2022-04-22
feat(snap): Add support for environment variable injection

farshidtz:
Supersceding #3924

This PR updates the snap to support environment variable injection as part of introducing the following scheme:
```
apps.<app>.<type>.<key>
```
where:
- `<app>` is the name of the app (service, executable)
- `<type>` is the type of option with respect to the app
- `<key>` is key for the option. It could contain a path to set a value inside an object, e.g. `x.y=z` sets `{"x": {"y": "z"}}`.

This change effectively makes options prefixed with `env.` deprecated. Such options will continue to work but be removed as of EdgeX 3.

The reasons for the schema change is as follows:
- The existing options depended on hardcoded mappings from keys to environment variables that was extremely hard to maintain. The new feature allows passing any environment variable dynamically.
- The environment variable passing was mixed with secrets-config executions causing various technical challanges and usability issues. With this change, the app configurations (via env files) use the `apps.<app>.config.` and `config.` prefixes exclusively.
- This makes it possible to extend the options for other settings (beyond config and secrets-config proxy), e.g. for controlling default service startup.

#### Environment Variables
The environment variable injection key uses the the following format:
- `apps.<app>.config.<env-var>`: setting app-specific env var ENV_VAR (e.g. `apps.core-data.config.service-port=1000`).
- `config.<env-var>`: setting global env vars by omitting the `apps.<app>` prefix (e.g. `config.service-host=localhost`, [`config.writable-loglevel=DEBUG`](https://docs.edgexfoundry.org/2.2/microservices/configuration/CommonEnvironmentVariables/#configuration-overrides))

Other `<env-var>` mapping examples:
- `edgex-startup-duration`: [`EDGEX_STARTUP_DURATION`](https://docs.edgexfoundry.org/2.2/microservices/configuration/CommonEnvironmentVariables/#edgex_startup_duration)
- `add-secretstore-tokens`: [`ADD_SECRETSTORE_TOKENS`](https://docs.edgexfoundry.org/2.2/security/Ch-Configuring-Add-On-Services/#configure-the-services-secret-store-to-use).

> The above design is restricted by snapd which supports only keys with lower-case letters, hypens, and dots. A dot is to define a parent-child relationship (`x.y=z` = `x={"y": "z"}`).

#### Secrets Proxy Options
New [secrets-config proxy](https://docs.edgexfoundry.org/2.2/security/secrets-config-proxy/) options:
- `apps.secrets-config.proxy.tls.cert` equivalent to `env.security-proxy.tls-certificate`
- `apps.secrets-config.proxy.tls.key` equivalent to `env.security-proxy.tls-private-key`
- `apps.secrets-config.proxy.tls.snis` equivalent to `env.security-proxy.tls-sni`
- `apps.secrets-config.proxy.admin.public-key` equivalent to `env.security-proxy.public-key` and `env.security-proxy.user=admin,1,ES256`

For usability purposes, the username, userid, algorithm of the single user have been hardcoded to admin, 1, ES256. This should be sufficient since we don't allow setting multiple users using snap options. The algorithm is set to ES256 rather than RS256, because it is more secure. If needed, it can be made configurable in future using keys `apps.secrets-config.proxy.admin.{user|id|algorithm}`.

#### Usage
**The new scheme is allowed only after opting-in by setting `config-enabled=true`.** It will be enabled by default as of EdgeX 3. 

Setting `config-enabled=true` will:
1. unset existing `env.` options and ignore future sets. 
2. convert:
    - `env.security-secret-store.add-secretstore-tokens` -> `apps.security-secretstore-setup.config.add-secretstore-tokens`
    - `env.security-secret-store.add-known-secrets` -> `apps.security-secretstore-setup.config.add-known-secrets`
    - `env.security-bootstrapper.add-registry-acl-roles` -> `apps.security-bootstrapper.config.add-registry-acl-roles`

---

## PR Checklist

- [x] I have added unit tests for the new feature or bug fix (if not, why?) - [unit tests](https://github.com/canonical/edgex-snap-hooks/blob/main/options/options_test.go), [smoke tests](https://github.com/canonical/edgex-snap-testing/pull/53)
- [ ] I have opened a PR for the related docs change (if not, why?) - snap/README and docs will be updated in separate, future PRs 
  <link to docs PR>

## Testing Instructions

Install
```
snap install edgexfoundry --channel=edge/pr-3986
```

Run
```
sudo snap set edgexfoundry apps.core-data.config.service-port=1234
```
It should be rejected because `config-enabled` is not set.

Run
```
sudo snap set edgexfoundry config-enabled=true
sudo snap set edgexfoundry apps.core-data.config.service-port=8080
sudo snap set edgexfoundry config.service-startupmsg="testing env injection"
```

1) Confirm that all .env files have SERVICE_STARTUPMSG="testing env injection"
2) Confirm that the core-data .env file also has both the SERVICE_STARTUPMSG and SERVICE_PORT settings

```
$ cat /var/snap/edgexfoundry/current/config/core-data/res/core-data.env 
export SERVICE_STARTUPMSG=testing env injection
export SERVICE_PORT=8080
$ cat /var/snap/edgexfoundry/current/config/core-command/res/core-command.env 
export SERVICE_STARTUPMSG=testing env injection
```

### DIFF
--- a/snap/local/hooks/cmd/configure/options.go
+++ b/snap/local/hooks/cmd/configure/options.go
@@ -22,6 +22,7 @@ import (
 	"os"
 
 	hooks "github.com/canonical/edgex-snap-hooks/v2"
+	app_options "github.com/canonical/edgex-snap-hooks/v2/options"
 )
 
 func applyConfigOptions(service string) error {
@@ -60,8 +61,59 @@ func options() {
 
 	hooks.Info("edgexfoundry:configure-options: handling config options for a single service: " + *service)
 
+	// process the EdgeX >=2.2 snap options
+	err = app_options.ProcessAppCustomOptions(*service)
+	if err != nil {
+		hooks.Error(fmt.Sprintf("edgexfoundry:configure-options: could not process custom options: %v", err))
+		os.Exit(1)
+	}
+
+	// process the legacy snap options
 	if err := applyConfigOptions(*service); err != nil {
 		hooks.Error(fmt.Sprintf("edgexfoundry:configure-options: error handling config options for %s: %v", *service, err))
 		os.Exit(1)
+	}
+}
+
+func processAppOptions() {
+	err := app_options.ProcessAppConfig(
+		"core-data",
+		"core-metadata",
+		"core-command",
+		"support-notifications",
+		"support-scheduler",
+		"app-service-configurable",
+		"device-virtual",
+		"security-secret-store",
+		"security-secretstore-setup",
+		"security-proxy-setup",
+		"security-bootstrapper",
+		"sys-mgmgt-agent",
+	)
+	if err != nil {
+		hooks.Error(fmt.Sprintf("edgexfoundry:configure could not process config options: %v", err))
+		os.Exit(1)
+	}
+
+	// After installation, the configure hook initiates the deferred startup of services,
+	// 	processes snap options and exits. The actual services startup happens only
+	// 	after the configure hook exits.
+	//
+	// The following options should not be processed within the configure hook during
+	//	the initial installation (install-mode=defer-startup). They should be processed
+	//	only on follow-up calls to the configure hook (i.e. when snap set/unset is called)
+	installMode, err := hooks.NewSnapCtl().Config("install-mode") // this set in the install hook
+	if err != nil {
+		hooks.Error(fmt.Sprintf("edgexfoundry:configure failed to read 'install-mode': %s", err))
+		os.Exit(1)
+	}
+	if installMode != "defer-startup" {
+		err = app_options.ProcessAppCustomOptions(
+			"secrets-config", // also processed in security-proxy-post-setup.sh
+		)
+		if err != nil {
+			hooks.Error(fmt.Sprintf("edgexfoundry:configure: could not process custom options: %v", err))
+			os.Exit(1)
+		}
 	}
 }
--- a/snap/local/hooks/go.mod
+++ b/snap/local/hooks/go.mod
@@ -1,5 +1,5 @@
 module github.com/canonical/edgex-go/hooks
 
-require github.com/canonical/edgex-snap-hooks/v2 v2.2.0-beta
+require github.com/canonical/edgex-snap-hooks/v2 v2.2.0
 
 go 1.17
--- a/snap/local/hooks/go.mod
+++ b/snap/local/hooks/go.mod
@@ -1,5 +1,6 @@
 module github.com/canonical/edgex-go/hooks
 
-require github.com/canonical/edgex-snap-hooks/v2 v2.2.0-beta.2
+require github.com/canonical/edgex-snap-hooks/v2 v2.2.0-beta.3
+
 
 go 1.17
--- a/snap/local/hooks/go.mod
+++ b/snap/local/hooks/go.mod
@@ -1,6 +1,8 @@
 module github.com/canonical/edgex-go/hooks
 
-require github.com/canonical/edgex-snap-hooks/v2 v2.2.0-beta.4
+require github.com/canonical/edgex-snap-hooks/v2 v2.2.0-beta.5
 
+// replace github.com/canonical/edgex-snap-hooks/v2 => ./edgex-snap-hooks
+// replace github.com/canonical/edgex-snap-hooks/v2 => github.com/farshidtz/edgex-snap-hooks/v2 d43ccc771100d663099c8ca8e3974d78076b2058
 
 go 1.17
--- a/snap/local/hooks/go.mod
+++ b/snap/local/hooks/go.mod
@@ -1,5 +1,5 @@
 module github.com/canonical/edgex-go/hooks
 
-require github.com/canonical/edgex-snap-hooks/v2 v2.1.3
+require github.com/canonical/edgex-snap-hooks/v2 v2.2.0-beta
 
 go 1.17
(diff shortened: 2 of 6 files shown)


MonicaisHer (approved):
@farshidtz Thank you. It works as expected. It also passed the checkbox test [security-services-proxy-certs](https://github.com/canonical/edgex-checkbox-provider/blob/master/data/latest/test-security-services-proxy-certs.sh) locally with new env vars options.

snap/local/hooks/cmd/configure/options.go
```diff
@@ -60,8 +61,59 @@ func options() {
...
+		"core-command",
+		"support-notifications",
+		"support-scheduler",
+		"app-service-configurable",
+		"device-virtual",
+		"security-secret-store",
+		"security-secretstore-setup",
+		"security-proxy-setup",
+		"security-bootstrapper",
+		"sys-mgmgt-agent",
```
tonyespy, line 91: I'd suggest dropping this as the SMA is considered deprecated and will be removed completely in EdgeX 3.0.

snap/local/hooks/go.mod
```diff
@@ -1,5 +1,8 @@
 module github.com/canonical/edgex-go/hooks
 
-require github.com/canonical/edgex-snap-hooks/v2 v2.1.3
+require github.com/canonical/edgex-snap-hooks/v2 v2.2.0-beta.5
```
tonyespy, line 3: Shouldn't this `require` be referencing a released version?

snap/local/runtime-helpers/bin/security-proxy-post-setup.sh
```diff
@@ -11,3 +11,6 @@ export PATH="$SNAP/bin:$PATH"
 # Several config options depend on resources that only exist after proxy 
 # setup. This re-applies the config options logic after deferred startup:
 $SNAP/snap/hooks/configure options --service=security-proxy
+
+# Process the EdgeX >=2.2 snap options
+$SNAP/snap/hooks/configure options --service=secrets-config
```
tonyespy, line 16: I don't love the fact we use the service name `secrets-config` here, as its not really a service, it's an application. Is there a reason you changed this from `security-proxy`?

tonyespy (commented):
Overall this looks good, testing works as expected. That said, there are a couple of comments in-line that should be reviewed before approval.

Also curious why updates to snap/README weren't included (has the doc moved elsewhere)?

snap/local/hooks/go.mod
farshidtz: We'd be doing more beta testing and add a stable version before the release. I believe other 1st party EdgeX dependencies will also get passed the code freeze with dev versions and switch to stable after a week or so. At least that's what happened in the previous release.

snap/local/runtime-helpers/bin/security-proxy-post-setup.sh
farshidtz: The new options are grouped per app and here, we are calling the processing of options for the secrets-config app. It should have been `--app=secrets-config` for accuracy but that'd involve changes to the parts that are being deprecated. I will add an issue to take care of this as part of future refactoring. But for now I would leave it as is because this is an internal CLI typically not seen or used by any user.

snap/local/hooks/cmd/configure/options.go
farshidtz: Combining the new options with the old ones is disallowed. As a result, it will not be possible for someone to configure SMA using the old options and everything else using the new ones. The only way to respect the deprecation policy is to keep this service functional (and configurable) until EdgeX 3.0 release, at which we can drop that too.

farshidtz:
> Overall this looks good, testing works as expected. That said, there are a couple of comments in-line that should be reviewed before approval.

Thanks a lot for the much needed code review. I've replied to your comments.

> Also curious why updates to snap/README weren't included (has the doc moved elsewhere)?

We will be updating the snap/README and docs in the next two weeks, before the release. I'll update the PR description to make that clear.

snap/local/runtime-helpers/bin/security-proxy-post-setup.sh
tonyespy: That doesn't really make sense as you're not configuring the secrets-config application, you're configuring the proxy, secrets-config is a helper application.

snap/local/hooks/cmd/configure/options.go
tonyespy: I wasn't suggesting combining the options, I was just suggesting that you don't support the SMA at all with the new scheme. And afaik there's _nothing_ in the deprecation policy which states that new features must support deprecated services or applications, that would seem to defeat the purpose of deprecation IMHOP.

farshidtz: As explained, this is necessary to allow using the new scheme with applications while still allowing the configuration of SMA. This isn't a new feature for SMA, but a new feature for the snap as a whole and that should not make the SMA incompatible with everything else when using the new scheme.

snap/local/runtime-helpers/bin/security-proxy-post-setup.sh
farshidtz: If at some point the proxy supports provisioning of users and TLS certificates via environment variables, then we can use the `apps.secrets-proxy.config` namespace to set those. In the meantime and in the scope of the new scheme, a user has to rely on the secrets-config app to perform the operations during runtime by setting e.g. `snap set edgexfoundry apps.secrets-config.proxy.admin.public-key=xxx`. Note that there is no `config` in the key because this isn't a configuration override.

The app in the suggested `--app=<app>` CLI argument is equivalent to `apps.<app>` namespace for snap options. In other words, as said, we are calling the processing of options for the secrets-config app. For now, only proxy is supported so we don't need to specify that in the CLI.

PR #4009, opened 2022-05-07
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,7 @@ updates:
       - dependency-name: "github.com/gomodule/redigo"
         # For github.com/gomodule/redigo, ignore version v2.0.0
         versions: ["v2.0.0"]
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule: 
+      interval: daily


farshidtz (commented):
I don't think this is needed. All three Github workflows in this repository follow major versions of the actions, rather than a minor or patch. As a result, the workflows will use latest patches and minor automatically on each run without us having to change anything.

The version is either the git tag, hash, or branch. The publishers that add a tag or branch for the major releases do that for the very same purpose of allowing workflows to use the latest minor/patch.

More info:
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses
- https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-tags-for-release-management

naveensrinivasan:
Sounds good. Thanks

Issue #4024, opened 2022-05-13
Build optimization for spiffe/spire security containers


bnevis-i:
# 🚀 Feature Request

The SPIFFE server, spiffe config, and spiffe agent containers must built the spire components from source due to lack of an Alpine compatible precompiled binary image.

- Needs to work on x86_64 alpine
- Needs to work on arm64 alpine

According to https://github.com/spiffe/spire/issues/2913#issuecomment-1125180358
it is possible to base an image off of a ghcr.io repository that has statically-linked binaries instead.

```
$ docker create --name temp ghcr.io/spiffe/spire-server:1.2.3
$ docker cp temp:/opt/spire/bin/spire-server .
$ docker rm test
$ ldd spire-server 
	not a dynamic executable
$ ./spire-server 
Usage: spire-server [--version] [--help] <command> [<args>]

Available commands are:
    agent          
    bundle         
    entry          
    federation     
    healthcheck    Determines server health status
    jwt            
    run            Runs the server
    token          
    validate       Validates a SPIRE server configuration file
    x509      
```

This will significantly improve build times.

Ideally, should be able to write an entrypoint script that allows the upstream container to be used directly.

bnevis-i:
Committed in Levski planning 5/19/2022

PR #4040, opened 2022-05-25
ci: update spire Dockerfile's to allow conditional compilation

ernestojeda:
The underlying [Go base image](https://github.com/edgexfoundry/ci-build-images/blob/golang-1.17/Dockerfile#L8-L23) that Jenkins is using to build the Dockerfiles now contains the spire-server and spire-agent binaries. This PR adds a new check into the spire Dockerfile's that first checks to see if the spire-server/agent doesn't exist before compiling the binary. For local builds, nothing will change since the base image is just the base golang image.

amd64 Docker build time ~ 2m, previously ~ 6m
arm64 Docker build time ~ 15m, previously ~ 50m

Fixes: #4024 

## PR Checklist

  <link to docs PR>

### DIFF
--- a/cmd/security-spire-agent/Dockerfile
+++ b/cmd/security-spire-agent/Dockerfile
@@ -33,12 +33,17 @@ COPY . .
 ARG SPIRE_RELEASE=1.2.1
 
 # build spire from the source in order to be compatible with arch arm64 as well
-RUN mkdir -p spire-build
+# in CI the BUILDER_BASE will already contain a compiled spire-server/agent
+# so we check to see if the binary is already in the image before compilation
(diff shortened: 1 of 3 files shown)


bnevis-i (commented):
Since the COPY command is looking at /usr/local/bin/spiffe*, I think it would make more sense to use test -f or test -x to look for the file in /usr/local rather than searching the path.

ernestojeda:
> Since the COPY command is looking at /usr/local/bin/spiffe*, I think it would make more sense to use test -f or test -x to look for the file in /usr/local rather than searching the path.

OK, I will make this change.

bnevis-i (approved):

PR #4063, opened 2022-06-28
fix(data): fix deleting numerous events HTTP request timeout by goroutines

jinlinGuan:
Always returns 202 Accepted immediately

Signed-off-by: Ginny Guan <ginny@iotechsys.com>

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?) NA, only manually test
  <link to docs PR>

## Testing Instructions

Steps to reproduce:
1. start edgex services by non-security compose file
2. generate around 10000 events
3. delete events by device name or age
4. return 503 HTTP request timeout

Steps to test the fix:
1. build core-data from this branch and run with other edgex services
2. generate around 10000 events
3. delete events by device name or age
4. return 202 Accepted immediately


lenny-intel (approved):
LGTM

farshidtz:
Does this still check if the device existed before accepting the request?

Ideally, there should be a 404 if the device with that name doesn't exist.

PR #4100, opened 2022-07-21
fix: Fixed Missing error check

drkfmorton:

## PR Checklist

  <link to docs PR>


lenny-intel:
@drkfmorton , please complete the PR checklist. It is not optional.

internal/security/proxy/service.go
```diff
@@ -319,6 +319,10 @@ func (s *Service) postCert(cp bootstrapConfig.CertKeyPair) *CertError {
 	}
 	tokens := []string{s.configuration.KongURL.GetProxyBaseURL(), CertificatesPath}
 	req, err := http.NewRequest(http.MethodPost, strings.Join(tokens, "/"), strings.NewReader(string(data)))
+	if err != nil {
+		s.loggingClient.Error(err.Error())
+		return &CertError{err.Error(), InternalError}
```
lenny-intel, line 324: Please add context to the error, like err = fmt.Errorf( "failed to upload Kong certificate... : %v", err)

lenny-intel (changes requested):

bnevis-i (dismissed):
LGTM assuming unit test passes.

drkfmorton:
Coverage of change:
image

internal/security/proxy/service_test.go
```diff
@@ -87,6 +87,39 @@ func TestPostCertExists(t *testing.T) {
 	}
 }
 
+func TestPostCertHttpError(t *testing.T) {
+	fileName := "./testdata/configuration.toml"
+	contents, err := os.ReadFile(fileName)
+	if err != nil {
+		t.Errorf("could not load configuration file (%s): %s", fileName, err.Error())
+		return
+	}
```
lenny-intel, line 96: We use the Testify module's `require` and `assert` functions to streamline our unit tests. `require` terminates the test if it fails. `assert` continues on failure. Use `require` when the rest of the test is invalid if it fails.
```suggestion
	require.NoError(t, err)
```

internal/security/proxy/service_test.go
```diff
@@ -87,6 +87,39 @@ func TestPostCertExists(t *testing.T) {
...
+
+	configuration := &config.ConfigurationStruct{}
+	err = toml.Unmarshal(contents, configuration)
+	if err != nil {
+		t.Errorf("unable to parse configuration file (%s): %s", fileName, err.Error())
+		return
+	}
```
lenny-intel, line 103:
```suggestion
	require.NoError(t, err)
```

internal/security/proxy/service_test.go
```diff
@@ -87,6 +87,39 @@ func TestPostCertExists(t *testing.T) {
...
+	configuration.KongURL = config.KongUrlInfo{
+		Server:          "{}",
+		ApplicationPort: -1,
+	}
+
+	mockLogger := logger.MockLogger{}
+	service := NewService(NewRequestor(true, 10, "", mockLogger), mockLogger, configuration)
+	mockCertPair := bootstrapConfig.CertKeyPair{Cert: "test-certificate", Key: "test-private-key"}
+	e := service.postCert(mockCertPair)
+	if e.reason != CertExisting {
```
lenny-intel, line 118:
```suggestion
        require.Error(t,e)
	if e.reason != CertExisting {
```

lenny-intel (changes requested):

internal/security/proxy/service_test.go
drkfmorton: Updated

internal/security/proxy/service_test.go
drkfmorton: Updated

internal/security/proxy/service_test.go
drkfmorton: Thanks for your feedback. Updated

internal/security/proxy/service.go
bnevis-i: Yes, it is required because of golang convention.  The issue is not the lack of an error check.  The issue is that by convention, when you return an object and an error, when there is an error, you return nil as the object, so that you don't accidentially operate on an uninitialized object.  Thus, req.Header.Add could panic(), because in the original code, err is not checked until after req.Header is accessed.

bnevis-i: The error check on line 327 (after req.Header.Add) is not needed if the error is checked before req.header.add, because req.header.add never fails.

bnevis-i (approved):

lenny-intel (approved):
LGTM

PR #4101, opened 2022-07-21
feat: Implement Device System Events

### LINKED ISSUE #4099: Core Metadata publish Device System Events
# 🚀 Feature Request

 This feature request is for Core Metadata

Implement publishing of Device System Events per System Events ADR (Link TBD)

See ADR

See ADR

lenny-intel:
Also, refactored to use common Messaging Bootstrap Handler

closes #4099

Implements [related ADR](https://docs.edgexfoundry.org/2.3/design/adr/0024-system-events) 

Dependent on the following PRs:
- https://github.com/edgexfoundry/go-mod-core-contracts/pull/752
- https://github.com/edgexfoundry/go-mod-bootstrap/pull/360

Signed-off-by: Leonard Goodell <leonard.goodell@intel.com>

## PR Checklist

  link TBD

## Testing Instructions
Run edgex stack with Device Virtual
Stop Core Metadata container
Build and run this branch locally (require branches from PR referenced above) with DEBUG enabled
Using curl or postman delete, add and remove Device Virtual devices (can't use EdgeX UI from docker since core-metadata is running locally)
Verify log show debug message that System Event was published


internal/core/metadata/application/device.go
```diff
@@ -237,3 +265,37 @@ func DevicesByProfileName(offset int, limit int, profileName string, dic *di.Con
 	}
 	return devices, totalCount, nil
 }
+
+var noMessagingClientError = goErrors.New("unable to publish Device System Event: MessageBus Client not available")
+
+func publishDeviceSystemEvent(action string, d models.Device, lc logger.LoggingClient, dic *di.Container) error {
```
cloudxxx8, line 271: use `errors.CommonEdgeX` instead of normal error

internal/core/metadata/application/device.go
```diff
@@ -237,3 +265,37 @@ func DevicesByProfileName(offset int, limit int, profileName string, dic *di.Con
...
+		systemEvent.Source,
+		systemEvent.Type,
+		systemEvent.Action,
+		systemEvent.Owner,
+		device.ProfileName)
+
+	payload, _ := json.Marshal(systemEvent)
+
+	envelope := types.NewMessageEnvelope(payload, context.Background())
+	envelope.CorrelationID = uuid.NewString()
```
cloudxxx8, line 291: the correlation id should retrieve from Context, and it will be the same one as the changing device request for debugging

internal/core/metadata/application/device.go
```diff
@@ -56,12 +62,22 @@ func AddDevice(d models.Device, ctx context.Context, dic *di.Container) (id stri
 		addedDevice.Id,
 		correlation.FromContext(ctx),
 	)
-	go addDeviceCallback(ctx, dic, dtos.FromDeviceModelToDTO(d))
+
+	device := dtos.FromDeviceModelToDTO(d)
+	go addDeviceCallback(ctx, dic, device)
+
+	if err := publishDeviceSystemEvent(common.DeviceSystemEventActionAdd, d, lc, dic); err != nil {
```
cloudxxx8, line 69: why not using async call?

cloudxxx8 (changes requested):

internal/core/metadata/application/device.go
cloudxxx8: we don't handle the error and we usually use async call to publish data to message bus.
if the connection to the message bus is blocking due to network issue or something, the Device Add/Update/Delete will be blocked, too.
that's why we use `go addDeviceCallback(ctx, dic, device)` to prevent any unexpected situation.

lenny-intel: OK

lenny-intel: Added go func's

internal/core/metadata/application/device.go
```diff
@@ -165,6 +189,14 @@ func PatchDevice(dto dtos.UpdateDevice, ctx context.Context, dic *di.Container)
 		go updateDeviceCallback(ctx, dic, oldServiceName, device)
 	}
 	go updateDeviceCallback(ctx, dic, device.ServiceName, device)
+
+	go func() {
+		if err := publishDeviceSystemEvent(common.DeviceSystemEventActionUpdate, device, ctx, lc, dic); err != nil {
+			// Don't fail request if couldn't publish, just log it. Error message already has context.
```
cloudxxx8, line 195: since it's asnyc, we might be able to remove this comment

internal/core/metadata/application/device.go
```diff
@@ -165,6 +189,14 @@ func PatchDevice(dto dtos.UpdateDevice, ctx context.Context, dic *di.Container)
...
+			lc.Error(err.Error())
```
cloudxxx8, line 196: log could add more context here:
`lc.Errorf("issue happens when publishing the device update system event: %v", err)`

internal/core/metadata/application/device.go
```diff
@@ -75,6 +91,14 @@ func DeleteDeviceByName(name string, ctx context.Context, dic *di.Container) err
 		return errors.NewCommonEdgeXWrapper(err)
 	}
 	go deleteDeviceCallback(ctx, dic, device)
+
+	go func() {
+		if err := publishDeviceSystemEvent(common.DeviceSystemEventActionDelete, device, ctx, lc, dic); err != nil {
+			// Don't fail request if couldn't publish, just log it. Error message already has context.
+			lc.Error(err.Error())
```
cloudxxx8, line 98: log could add more context here:
`lc.Errorf("issue happens when publishing the device add system event: %v", err)`

internal/core/metadata/application/device.go
```diff
@@ -237,3 +269,40 @@ func DevicesByProfileName(offset int, limit int, profileName string, dic *di.Con
...
+
+var noMessagingClientError = errors.NewCommonEdgeXWrapper(goErrors.New("unable to publish Device System Event: MessageBus Client not available"))
+
+func publishDeviceSystemEvent(action string, d models.Device, ctx context.Context, lc logger.LoggingClient, dic *di.Container) errors.EdgeX {
+	device := dtos.FromDeviceModelToDTO(d)
+	systemEvent := dtos.NewSystemEvent(common.DeviceSystemEventType, action, common.CoreMetaDataServiceKey, device.ServiceName, nil, device)
+
+	messagingClient := bootstrapContainer.MessagingClientFrom(dic.Get)
+	if messagingClient == nil {
+		return &noMessagingClientError
```
cloudxxx8, line 281: you might need to create the error in this way:
declare the error with a normal error in line 273:
`var noMessagingClientError = goErrors.New("unable to publish Device System Event: MessageBus Client not available")`

and wrap the error for return here
`return errors.NewCommonEdgeXWrapper(noMessagingClientError)`

In this way, the errors.CommonEdgeX can record which line in which file cause this error in the debug message.
Users can still use `errors.As` to determine whether the error contains `noMessagingClientError`

cloudxxx8 (changes requested):

internal/core/metadata/application/device.go
lenny-intel: That is already added in the function. I have it returning error in order to validate in unit tests. Think it is better to just have the function log the error.

internal/core/metadata/application/device.go
lenny-intel: Ok, sound good.

internal/core/metadata/application/device.go
lenny-intel: I have reworked the error handling so this is no longer needed.

internal/core/metadata/application/device.go
lenny-intel: No longer needed.

internal/core/metadata/application/device.go
lenny-intel: No loner returning error, just logging it.

internal/core/metadata/application/device.go
lenny-intel: No longer needed

cloudxxx8 (dismissed):
LGTM

farshidtz (changes requested):
The snap fails to start because security-secretstore-setup exits with non-zero code. The CI logs currently don't get the logs when the snap fails to even start (we are working on improving this).

I downloaded the build from Github (see for example [here](https://github.com/edgexfoundry/edgex-go/actions/runs/2750254562); scroll down) and installed manually to check the logs:
```
$ snap install --dangerous ./edgexfoundry_2.3.0-dev.30_amd64.snap
Run service command "start" for services ["consul" "core-command" "core-data" "core-metadata" "kong-daemon" "postgres" "redis" "security-bootstrapper-redis" "security-consul-bootstrapper" "security-proxy-setup" "security-secretstore-setup" "vault"] of snap "edgexfoundry" (systemctl command [start snap.edgexfoundry.security-secretstore-setup.service] failed with exit status 1: Job for snap.edgexfoundry.security-secretstore-setup.service failed because the control process exited with error code.

$ sudo journalctl -n 1000 | grep edgexfoundry | grep ERROR
level=ERROR ts=2022-07-28T08:14:23.57437856Z app=security-secretstore-setup source=init.go:329 msg="token provider failed: /snap/edgexfoundry/x1/bin/security-file-token-provider terminated with non-zero exit code 1"
```
Unfortunately, security-secretstore-setup is not printing the error message. Unrelated to this PR, it looks like standard output and standard error aren't captured by EdgeX logger [here](https://github.com/edgexfoundry/edgex-go/blob/062af8d3cd30019d44081b1021c6648b4b920d9b/internal/security/secretstore/tokenprovider.go#L86-L95). [This code](https://github.com/edgexfoundry/edgex-go/blob/062af8d3cd30019d44081b1021c6648b4b920d9b/internal/security/secretstore/execrunner.go#L36-L41) sets to use os output streams but not sure if this is ever called.

I could not figure out why the changes in this PR are causing this failure.

Full logs: [edgexfoundry_2.3.0-dev.30.logs.txt](https://github.com/edgexfoundry/edgex-go/files/9207796/edgexfoundry_2.3.0-dev.30.logs.txt)

lenny-intel:
@farshidtz , @jim-wang-intel create this PR to test which go-mod and which version has triggered this. We have narrowed it down to go-mod-bootstrap v2.2.1-dev.9
https://github.com/edgexfoundry/edgex-go/pull/4105

I am taking a look at the commit for that to see how it could be impacting this.

lenny-intel:
@farshidtz , that change to go-mod-bootstrap was to make sure the exit code it set properly when bootstrapping fails. Before it always returned with 0 exit code. 

I suspect that the issue with Secret Store setup in the test was always there, be now has the proper exit code.

lenny-intel:
@farshidtz , found the culprit !

The file token provider is incorrectly returning `false` in its bootstrap hander which now causes it to exit with non-zero exit code.

We'll get a fix in for that which will un-block this PR.

lenny-intel:
@jim-wang-intel , @farshidtz , Rebased this PR to pull in fixes from the other PR.

lenny-intel:
Put in draft to it doesn't get merged until this PR is merged otherwise TAF tests will fail.
https://github.com/edgexfoundry/edgex-compose/pull/261#pullrequestreview-1055010670

jim-wang-intel (approved):

PR #4119, opened 2022-08-08
feat(metadata): implement Unit of Measure (UoM) ADR

hahattan:

## PR Checklist

  https://github.com/edgexfoundry/edgex-docs/pull/834

## Testing Instructions

1. Run metadata from this branch with `WRITABLE_UOM_VALIDATION: "true"`
2. add/update device porfile and compare the result with `res/uom.toml`

Also,
I updated the [sample](https://docs.edgexfoundry.org/2.2/design/adr/core/0022-UoM/#specifying-the-units-of-measure) uom TOML in ADR from
```toml
[Uom]
    Source="reference to source for all UoM if not specified below"
    [Uom.temperature]
    Source="www.weather.com"
    Values="C,F,K"
    [Uom.weights]
    Source="www.usa.gov/federal-agencies/weights-and-measures-division"
    Values="lbs,ounces,kilos,grams"
```
to
```toml
[Uom]
Source="reference to source for all UoM if not specified below"
  [Uom.Units]
    [Uom.Units.temperature]
    Source="www.weather.com"
    Values=["C","F","K"]
    [Uom.Units.weights]
    Source="www.usa.gov/federal-agencies/weights-and-measures-division"
    Values=["lbs","ounces","kilos","grams"]
```

so that it can be correctly unmarshaled


cloudxxx8:
@hahattan we merged a PR which should resolved the snap test issue.  please rebase it.

internal/core/metadata/application/deviceprofile.go
```diff
@@ -50,6 +59,15 @@ func UpdateDeviceProfile(d models.DeviceProfile, ctx context.Context, dic *di.Co
...
 	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
 
+	if container.ConfigurationFrom(dic.Get).Writable.UoM.Validation {
+		uom := container.UnitsOfMeasureFrom(dic.Get)
+		for _, dr := range d.DeviceResources {
+			if ok := uom.Validate(dr.Properties.Units); !ok {
+				return errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("DeviceResource %s units %s is invalid", dr.Name, dr.Properties.Units), nil)
+			}
+		}
+	}
```
lenny-intel, line 69: Create helper function to reduce duplicated code

internal/core/metadata/uom/bootstrap.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/edgexfoundry/edgex-go/internal/core/metadata/container"
+)
+
+func BootstrapHandler(_ context.Context, _ *sync.WaitGroup, _ startup.Timer, dic *di.Container) bool {
+	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
+	config := container.ConfigurationFrom(dic.Get)
+	filepath := config.UoM.UoMFile
+
+	contents, err := ioutil.ReadFile(filepath)
```
lenny-intel, line 26: Not backwards compatible with old V2 configuration which will not have this field set. Should check for empty string and create empty `UnitsOfMeasureImpl` when it is empty. `Validation` will also be missing and default to false.

snap/snapcraft.yaml
```diff
@@ -626,6 +626,10 @@ parts:
           install -DT "./cmd/$service/$service" "$SNAPCRAFT_PART_INSTALL/bin/$service"
 
           case "$service" in
+          "core-metadata")
+              install -DT "./cmd/core-metadata/res/configuration.toml" "$SNAPCRAFT_PART_INSTALL/config/core-metadata/res/configuration.toml"
```
lenny-intel, line 630: No needed as this is done at the bottom of the `for/case` for all the services

lenny-intel (changes requested):

snap/snapcraft.yaml
hahattan: Isn't the bottom one acts as default case if no match is found ?

internal/core/metadata/application/deviceprofile.go
```diff
@@ -50,6 +58,14 @@ func UpdateDeviceProfile(d models.DeviceProfile, ctx context.Context, dic *di.Co
 	dbClient := container.DBClientFrom(dic.Get)
 	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
 
+	if container.ConfigurationFrom(dic.Get).Writable.UoM.Validation {
+		uom := container.UnitsOfMeasureFrom(dic.Get)
+		err := deviceProfileUoMValidation(d, uom)
+		if err != nil {
+			return errors.NewCommonEdgeXWrapper(err)
+		}
+	}
```
lenny-intel, line 67: This code is also duplicated. Please include in helper function

internal/core/metadata/uom/bootstrap.go
```diff
@@ -0,0 +1,61 @@
...
+	// backward compatability for using older 2.x configuration
+	// TODO: Remove in EdgeX 3.0
+	if filepath == "" {
+		dic.Update(di.ServiceConstructorMap{
+			container.UnitsOfMeasureInterfaceName: func(get di.Get) interface{} {
+				return uomImpl
+			},
+		})
+
+		lc.Warn("UoM.UoMFile field not set in configuration file, unit of measure validation will always fail")
```
lenny-intel, line 36: Why `unit of measure validation will always fail` if `uomImpl` is empty? 
```suggestion
		lc.Warn("UoM.UoMFile field not set in configuration file, unit of measure validation is disabled")
```

internal/core/metadata/uom/uom.go
```diff
@@ -0,0 +1,39 @@
...
+type Unit struct {
+	Source string
+	Values []string
+}
+
+func (u *UnitsOfMeasureImpl) Validate(unit string) bool {
+	if u == nil {
+		return false
+	}
+	if unit == "" || len(u.Uom.Units) == 0 {
```
lenny-intel, line 26: IMO, If no UOM.Units specified at all, then nothing to validate against, so should always pass.

```suggestion
	if len(u.Uom.Units) == 0 {
```

snap/snapcraft.yaml
lenny-intel: Ahh yep, because you added this case that will not be hit.

lenny-intel (changes requested):
@hahattan , please complete PR check list with ink to PR for the Doc updates for this new capability. THX!

internal/core/metadata/uom/bootstrap.go
hahattan: > Should check for empty string and create empty UnitsOfMeasureImpl when it is empty. Validation will also be missing and default to false.

Based on previous review comment, I think default to false means fail the validation .
If  we want to disable this should we return true instead ?

internal/core/metadata/uom/uom.go
hahattan: I think we should still keep the first check otherwise validation will fail if user doesn't provide the `units` property ?

internal/core/metadata/uom/bootstrap.go
lenny-intel: Correct, that is why I commented on the actual validation to return true if the UOM list is empty.

internal/core/metadata/uom/uom.go
lenny-intel: Ahh , yep. I missed the OR part of the conditional.

lenny-intel (approved):
LGTM

lenny-intel:
@hahattan , FYI I have escalated the snap testing failure to @farshidtz

hahattan:
document PR updated: https://github.com/edgexfoundry/edgex-docs/pull/834

cloudxxx8 (approved):
LGTM

snap/snapcraft.yaml
farshidtz: This is good as it copies also `uom.toml` to the snap package (readonly, [`SNAP`](https://snapcraft.io/docs/environment-variables) env var). But we also need to add code to copy it over to the snap's writable area ([`SNAP_DATA`](https://snapcraft.io/docs/environment-variables) env var) on install time. The given path `UOM_UOMFILE: $SNAP_DATA/config/core-metadata/res/uom.toml` expects it to be under the writable area.

Currently only `configuration.toml` files get copied:
https://github.com/edgexfoundry/edgex-go/blob/e3ca38a5709b6fd134cd3d7a24fe413a337b80c3/snap/local/hooks/cmd/install/install.go#L118-L143

This is really an unnecessary complication in how the snap is packaged. Maybe we should refactor and simply copy the whole `res` directory.

For now, I'll add a commit to copy `uom.toml` from `SNAP` to `SNAP_DATA`.

snap/snapcraft.yaml
```diff
@@ -626,6 +627,10 @@ parts:
...
+              install -DT "./cmd/core-metadata/res/uom.toml" "$SNAPCRAFT_PART_INSTALL/config/core-metadata/res/uom.toml"
+          ;;
```
farshidtz, line 633: This is good as it copies also `uom.toml` to the snap package (readonly, [`SNAP`](https://snapcraft.io/docs/environment-variables) env var). But we also need to add code to copy it over to the snap's writable area ([`SNAP_DATA`](https://snapcraft.io/docs/environment-variables) env var) on install time. The given path `UOM_UOMFILE: $SNAP_DATA/config/core-metadata/res/uom.toml` expects it to be under the writable area.

Currently only `configuration.toml` files get copied:
https://github.com/edgexfoundry/edgex-go/blob/e3ca38a5709b6fd134cd3d7a24fe413a337b80c3/snap/local/hooks/cmd/install/install.go#L118-L143

This is really an unnecessary complication in how the snap is packaged. Maybe we should refactor and simply copy the whole `res` directory.

For now, a quick fix would be to update the referenced code to copy `uom.toml` from `SNAP` to `SNAP_DATA` for core-metadata. I'll work on it and commit to the same PR (since the permission has been granted by the author during PR submission.)

farshidtz (changes requested):

snap/snapcraft.yaml
hahattan: sorry I wasn't aware you're working on it.
I've added a quick fix: https://github.com/edgexfoundry/edgex-go/compare/4c4951218dd81a254464f99d357ca8617bed2473..3aaec565c9b77e4c31d71bd8720703868403a2fd

farshidtz: Thanks for the fix. I was wondering why I couldn't push. I was testing locally so it took some time.

You force-pushed so it isn't clear what exactly you changed. But the install.go part looks good.

farshidtz (approved):

PR #4130, opened 2022-08-23
Update core-data.yaml

SaltySooda:
**GET operations cannot have a requestBody.**

found this error,when I import core-data.yaml to [https://editor.swagger.io/](url)


cloudxxx8:
please open an issue to write down the problem before opening the PR
for the PR, the commit message need to comply with the community standard
https://wiki.edgexfoundry.org/display/FA/Committing+Code+Guidelines#CommittingCodeGuidelines-CommitMessages

also, for this PR, we do implement the GET API with the payload.  it is workable technically.

Issue #4108, opened 2022-07-29
"ACL not found" messages logged as errors


farshidtz:
# 🐞 Bug Report

### Affected Services [**REQUIRED**]

 

There are many "ACL not found" ERRORs in the logs followed by INFO messages about creating them. Are these really meant to be errors? I believe these should be reduced to WARN. 

### Is this a regression?
Yes, the previous version in which this bug was not present was: v2.2

### Description and Minimal Reproduction [**REQUIRED**]

```
sudo snap install edgexfoundry --channel=latest/edge
snap logs -n=1000 edgexfoundry | grep --ignore-case "error"
```

## 🔥 Exception or Error
```
edgexfoundry.consul[156574]: 2022-07-29T17:17:49.335+0200 [ERROR] agent.http: Request error: method=GET url=/v1/acl/policy/name/acl_policy_for_app-functional-tests from=127.0.0.1:52794 error="ACL not found"
edgexfoundry.security-consul-bootstrapper[157086]: level=INFO ts=2022-07-29T15:17:49.346843916Z app=security-bootstrapper source=aclpolicies.go:142 msg="successfully created a new agent policy with name acl_policy_for_app-functional-tests"
edgexfoundry.security-consul-bootstrapper[157086]: level=INFO ts=2022-07-29T15:17:49.348177934Z app=security-bootstrapper source=aclroles.go:129 msg="successfully created a role [app-functional-tests] for secretstore"

edgexfoundry.consul[156574]: 2022-07-29T17:17:49.348+0200 [ERROR] agent.http: Request error: method=GET url=/v1/acl/policy/name/acl_policy_for_device-grove from=127.0.0.1:52794 error="ACL not found"
edgexfoundry.security-consul-bootstrapper[157086]: level=INFO ts=2022-07-29T15:17:49.359151534Z app=security-bootstrapper source=aclpolicies.go:142 msg="successfully created a new agent policy with name acl_policy_for_device-grove"
 edgexfoundry.security-consul-bootstrapper[157086]: level=INFO ts=2022-07-29T15:17:49.360910761Z app=security-bootstrapper source=aclroles.go:129 msg="successfully created a role [device-grove] for secretstore"
...
```

## 🌍  Your Environment
**Deployment Environment:**
Snap

**EdgeX Version [**REQUIRED**]:**
2.3.0-dev.31

**Anything else relevant?**

jim-wang-intel:
[...]

PR #4164, opened 2022-09-21
 fix: added consul policy list check

ItsRico:
Added a check in getPolicyByName to see whether the policy exists in the policy list before attempting to get the policy. This prevents erroneous or unclear logging.

closes: #4108

Signed-off-by: Rico Chavez-Lopez <rchavezlopez@ucdavis.edu>

## PR Checklist

- [x ] I have added unit tests for the new feature or bug fix (if not, why?)
  <link to docs PR>

## Testing Instructions

- aclpolicies unit test
- creating policy not in consul policy list

### DIFF
--- a/internal/security/bootstrapper/command/setupacl/aclpolicies.go
+++ b/internal/security/bootstrapper/command/setupacl/aclpolicies.go
@@ -66,6 +66,7 @@ const (
 	edgeXServicePolicyName = "edgex-service-policy"
 
 	consulCreatePolicyAPI     = "/v1/acl/policy"
+	consulPolicyListAPI       = "/v1/acl/policies"
 	consulReadPolicyByNameAPI = "/v1/acl/policy/name/%s"
 
 	aclNotFoundMessage = "ACL not found"
@@ -104,6 +105,10 @@ const (
 	// edgeXServicePolicyName is the name of the management policy for edgex
 	edgeXManagementPolicyName = "edgex-management-policy"
 )
+
+type PolicyListResponse []struct {
+	Name string `json:"Name"`
+}
 
 // getOrCreateRegistryPolicy retrieves or creates a new policy
 // it inserts a new policy if the policy name does not exist and returns a policy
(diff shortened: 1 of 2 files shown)


internal/security/bootstrapper/command/setupacl/aclpolicies.go
```diff
@@ -184,6 +189,53 @@ func (c *cmd) getOrCreateRegistryPolicy(tokenID, policyName, policyRules string)
...
+	}
+
+	switch policyListResp.StatusCode {
+	case http.StatusOK:
+		for _, policy := range policyList {
+			if policyName == policy.Name {
+				isExistingPolicy = true
+				break
+			}
+		}
```
jim-wang-intel, line 226: If we already found policy here, we can just return policy? why we need to find policy by name again below? (IMO, kind of redundant logic below)

jim-wang-intel (changes requested):
should refactor to remove the find by name logic if already found

internal/security/bootstrapper/command/setupacl/aclpolicies.go
bnevis-i: No, it's not.  The list API only returns the names of the policies.  Still need another call to actually get the policy (i.e. it is just chaining off to the original code).

jim-wang-intel: that is not true: see this: https://www.consul.io/api-docs/acl/policies#list-policies

bnevis-i: @jim-wang-intel   I understand, I think, what you were saying.  We only really need the name and the ID, not the text of the policy.  I think I like the solution as presented because it addresses the bug as stated (error in consul log due to get of non-existent policy.)  Since this is bootstrapping logic, it is not performance critical that we get the answer we need in one or two REST API calls.  I think we should implement the simplest and most readable fix.  We can refactor later if we want.

jim-wang-intel: yep, it sounds ok

jim-wang-intel (approved):
LGTM

bnevis-i (approved):

PR #4185, opened 2022-10-06
feat(command): refine north-south messaging configuration

### LINKED ISSUE #4180: Use RequireMessageBus config instead of MessageQueue.Required
# 🚀 Feature Request

 This feature request is for core-command

Currently, core-command uses MessageQueue.Required to control whether to enable the messaging API for both internal and external command requests, and the default value is false.
https://github.com/edgexfoundry/edgex-go/blob/17acb4d22b6a38b20e3dd1bbaacb8af86852ef24/cmd/core-command/res/configuration.toml#L47-L48

This causes some issues:

1. The config name is too general to understand its feature. We should follow the config in core-metadata
https://github.com/edgexfoundry/edgex-go/blob/39fcbed88f67e201ee240975c941ad828856b121/cmd/core-metadata/res/configuration.toml#L1-L4
2. The default value should be `true` for internal messaging APIs.
3. The user might not need the external messaging, so the config should be separated and the external messaging default should be disabled.

### LINKED ISSUE #4183: Refine the default messaging topics in core-command
# 🚀 Feature Request

 This feature request is for core-command

The current default topics settings are as follows:
https://github.com/edgexfoundry/edgex-go/blob/17acb4d22b6a38b20e3dd1bbaacb8af86852ef24/cmd/core-command/res/configuration.toml#L56-L62

These are accordance with ADR.
However, they don't follow the general edgex topic naming.  Just like event messaging, there are two kinds of topics, published by device services and core service.  Thus, I suggest to use the following topics:

```
    DeviceRequestTopicPrefix = "edgex/device/command/request"    # for publishing requests to the device service; <device-service>/<device-name>/<command-name>/<method> will be added to this publish topic prefix
    DeviceResponseTopic = "edgex/device/command/response/#"      # for subscribing to device service responses
    CommandRequestTopic = "edgex/core/command/request/#"     # for subscribing to internal command requests
    CommandResponseTopicPrefix = "edgex/core/command/response"     # for publishing responses back to internal service /<device-name>/<command-name>/<method> will be added to this publish topic prefix
    QueryRequestTopic = "edgex/core/commandquery/request"   # for subscribing to internal command query requests
    QueryResponseTopic = "edgex/core/commandquery/response"   # for publishing reponsses back to internal service
```

In addtion, the external topics should be just examples, and they should not be conflicted with the internal topics in case the user use the internal message broker to run some tests.
https://github.com/edgexfoundry/edgex-go/blob/17acb4d22b6a38b20e3dd1bbaacb8af86852ef24/cmd/core-command/res/configuration.toml#L92-L96

Thus, I suggest changing them to:
```
    CommandRequestTopic = "external/edgex/command/request/#"           # for subscribing to 3rd party command requests
    CommandResponseTopicPrefix = "external/edgex/command/response"     # for publishing responses back to 3rd party systems /<device-name>/<command-name>/<method> will be added to this publish topic prefix
    QueryRequestTopic = "external/edgex/commandquery/request/#"        # for subscribing to 3rd party command query request
    QueryResponseTopic = "external/edgex/commandquery/response"        # for publishing responses back to 3rd party systems
```

hahattan:

fix #4180 
fix #4183 

## PR Checklist

  <link to docs PR>

### DIFF
--- a/internal/core/command/main.go
+++ b/internal/core/command/main.go
@@ -90,12 +90,10 @@ func Main(ctx context.Context, cancel context.CancelFunc, router *mux.Router) {
 func MessageBusBootstrapHandler(ctx context.Context, wg *sync.WaitGroup, startupTimer startup.Timer, dic *di.Container) bool {
 	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
 	configuration := container.ConfigurationFrom(dic.Get)
-	if configuration.MessageQueue.Required {
-		router := messaging.NewMessagingRouter()
+	router := messaging.NewMessagingRouter()
+
+	if configuration.RequireMessageBus {
 		if !handlers.MessagingBootstrapHandler(ctx, wg, startupTimer, dic) {
-			return false
-		}
-		if !handlers.NewExternalMQTT(messaging.OnConnectHandler(router, dic)).BootstrapHandler(ctx, wg, startupTimer, dic) {
 			return false
 		}
 		if err := messaging.SubscribeCommandRequests(ctx, router, dic); err != nil {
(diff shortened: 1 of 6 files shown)


cloudxxx8 (dismissed):
LGTM

internal/core/command/main.go
```diff
@@ -111,6 +109,11 @@ func MessageBusBootstrapHandler(ctx context.Context, wg *sync.WaitGroup, startup
 			return false
 		}
 	}
+	if configuration.MessageQueue.External.Enabled {
```
lenny-intel, line 112: I think this check should be within `if configuration.RequireMessageBus` scope since receiving external message doesn't make sense if we are not connected to the internal message bus to send them to the device services

lenny-intel (changes requested):

lenny-intel (approved):

PR #4277, opened 2023-01-12
refactor(snap)!: Replace deprecated snap options with app options

farshidtz:
- Remove support for config and proxy config with `env.` prefix in favor of app config options
- Replace option to disable security from `security-secret-store=off` to `security=false`
- Replace `<app>=on|off` option with `apps.<app>.autostart=true|false` and `autostart=true|false` (global) options
- Refactor deferred startup to use autostart logic for starting services after configuration
- Refactor security disabling logic to make use of autostart and `config.edgex-security-secret-store` in instead.
  - Remove obsolete script for starting services with disabled security.
  - Add a new service that runs to apply secrets-config options.
- Rename helper-go's options subcommand from `service` to `app` since it can handle all apps, incl. CLI apps.

## PR Checklist

  <link to docs PR>


MonicaisHer (approved):
@farshidtz Thank you for the changes, looks good!

PR #4300, opened 2023-01-23
fix: Tweaks to Core Common Config Docker and entrypoint script for secure mode

lenny-intel:
Signed-off-by: Leonard Goodell <leonard.goodell@intel.com>

## PR Checklist

- [ ] I have opened a PR for the related docs change (if not, why?) **N/A**
  <link to docs PR>

## Testing Instructions
See follow on edgex-compose PR

### DIFF
--- a/cmd/core-common-config-bootstrapper/Dockerfile
+++ b/cmd/core-common-config-bootstrapper/Dockerfile
@@ -48,5 +48,6 @@ COPY --from=builder /edgex-go/cmd/core-common-config-bootstrapper/entrypoint.sh 
 RUN chmod 755 /usr/local/bin/entrypoint.sh \
     && ln -s /usr/local/bin/entrypoint.sh /
 
-ENTRYPOINT ["entrypoint.sh"]
-CMD ["/core-common-config-bootstrapper", "-cp=consul.http://edgex-core-consul:8500"]
+# Due to security mode compose file overriding overriding entry point script, we must specify this service entrypoint \
+# script in CMD
+CMD ["entrypoint.sh"]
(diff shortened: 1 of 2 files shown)


jim-wang-intel (commented):
IMO, this way loses the ability to override the extra cmd paramters only without specifying the original entrypoint.sh. e.g. you could just specify command: ["/core-common-cofnig", "param1", "param2",...etc]  without touching entrypioint.

lenny-intel:
> IMO, this way loses the ability to override the extra cmd paramters only without specifying the original entrypoint.sh. e.g. you could just specify command: ["/core-common-coifing", "param1", "param2",...etc] without touching entrypioint.

Nope, that doesn't work. We need the local entry point script to run from the bootstrap entrypoint script . Only other way would be to create a new security bootstrap entrypoint script specify for this service like we do for proxy setup. IMO this is cleaner than creating another one off security script. This service doesn't need flexibility to override parameters as it has a very narrow purpose.

bnevis-i (commented):
The way we have done this in all the other Dockerfiles is that the executable is in ENTRYPOINT (in this case, entrypoint.sh), and the arguments are in CMD.

lenny-intel:
> The way we have done this in all the other Dockerfiles is that the executable is in ENTRYPOINT (in this case, entrypoint.sh), and the arguments are in CMD.

Yes that is the standard way, but this case we have two entry point scripts that need to be run and is the original approach we used and works in non-secure...

I think I found a way to override the `command` in the  secure compose where is calls the local entrypoint and passes the command..
```
    command:
    - /entrypoint.sh
    - /core-common-config-bootstrapper
    - -cp=consul.http://edgex-core-consul:8500
```
This is working but was hand edited...
Will close this PR once verify I can get that to be generated properly.

lenny-intel:
Yep, that worked! Closing this PR

PR #4306, opened 2023-01-24
feat: Push common config to config provider

ejlee3:
Signed-off-by: Elizabeth J Lee <elizabeth.j.lee@intel.com>

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?) - NA - added functionality to main
- [ ] I have opened a PR for the related docs change (if not, why?) - docs will need to be added for common config when feature is complete
  <link to docs PR>

## Testing Instructions
1. run `make common-config` to ensure the binary builds
3. run in non-secure mode `export EDGEX_SECURITY_SECRET_STORE=false`
4. run edgex core services from compose builder `make run no-secty`
5. `./core-common-config-bootstrapper -cp=consul.http://localhost:8500 -cf=configuration.yaml`
6. check that the config is loaded into Consul [http://localhost:8500/ui/dc1/kv/edgex/v3/core-common-config-bootstrapper/](http://localhost:8500/ui/dc1/kv/edgex/v3/core-common-config-bootstrapper/)
7. bring down the services `make down`
8. build the docker images `make docker`
9. run in secure mode `export EDGEX_SECURITY_SECRET_STORE=true`
10. use compose builder and run `make run dev`
11. run the core common config bootstrapper from `cmd/core-common-config-bootstrapper` by running `sudo ./core-common-config-bootstrapper -cp=consul.http://localhost:8500 -cf=configuration.yaml`
12. verify that the config is loaded into Consul

### DIFF
--- a/internal/core/common_config/main.go
+++ b/internal/core/common_config/main.go
@@ -25,7 +25,6 @@ import (
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/startup"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/di"
 	"github.com/edgexfoundry/go-mod-configuration/v3/configuration"
-	"github.com/edgexfoundry/go-mod-configuration/v3/pkg/types"
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/clients/logger"
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/common"
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/models"
@@ -74,11 +73,9 @@ func Main(ctx context.Context, cancel context.CancelFunc) {
 
 	lc.Info("Secret Provider created")
 
-	var getAccessToken types.GetAccessTokenCallback
-
 	// need to use in-line function to set the callback type for getAccessToken used in CreateProviderClient to allow
 	// access to the config provider in secure mode
-	getAccessToken = func() (string, error) {
+	getAccessToken := func() (string, error) {
 		accessToken, err := secretProvider.GetAccessToken("consul", common.CoreCommonConfigServiceKey)
 		if err != nil {
 			return "", fmt.Errorf("failed to get Configuration Provider access token: %s", err.Error())
@@ -195,7 +192,7 @@ func buildKeyValues(data map[string]interface{}, kv map[string]interface{}, orig
 	key := origKey
 	for k, v := range data {
 		if len(key) == 0 {
-			key = fmt.Sprintf("%s", k)
+			key = fmt.Sprint(k)
 		} else {
 			key = fmt.Sprintf("%s/%s", key, k)
 		}
--- a/internal/core/common_config/main.go
+++ b/internal/core/common_config/main.go
@@ -16,19 +16,29 @@ package common_config
 
 import (
 	"context"
+	"fmt"
+	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/config"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/container"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/environment"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/flags"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/secret"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/startup"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/di"
+	"github.com/edgexfoundry/go-mod-configuration/v3/configuration"
+	"github.com/edgexfoundry/go-mod-configuration/v3/pkg/types"
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/clients/logger"
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/common"
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/models"
+	"gopkg.in/yaml.v3"
 	"os"
 	"os/signal"
+	"sort"
 	"sync"
 	"syscall"
+)
+
+const (
+	commonConfigDone = "isCommonConfigReady"
 )
 
 func Main(ctx context.Context, cancel context.CancelFunc) {
@@ -56,7 +66,6 @@ func Main(ctx context.Context, cancel context.CancelFunc) {
 			return lc
 		},
 	})
-
 	secretProvider, err := secret.NewSecretProvider(nil, environment.NewVariables(lc), ctx, startupTimer, dic, common.CoreCommonConfigServiceKey)
 	if err != nil {
 		lc.Errorf("failed to create Secret Provider: %v", err)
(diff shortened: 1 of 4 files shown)


internal/core/common_config/main.go
```diff
@@ -56,23 +66,62 @@ func Main(ctx context.Context, cancel context.CancelFunc) {
...
+				return "", fmt.Errorf("failed to get Configuration Provider access token: %s", err.Error())
+			}
+			return accessToken, err
+		}
+		_, err = getAccessToken()
+		if err != nil {
+			lc.Errorf("failed to get Access Token for config provider: %s", err.Error())
+			os.Exit(1)
+		}
+	}
```
jim-wang-intel, line 92: probably it's a good idea to add to handle else- case here. we don't wanna continue in this case....

internal/core/common_config/main.go
```diff
@@ -99,3 +148,74 @@ func translateInterruptToCancel(ctx context.Context, wg *sync.WaitGroup, cancel
...
+
+	for k := range kv {
+		keys = append(keys, k)
+	}
+
+	sort.Strings(keys)
+
+	for _, k := range keys {
+		v := kv[k]
+		// Push key/value into Consul rather than print
```
lenny-intel, line 184: Remove comment. Obvious from code and `rather than print` is out of context.

internal/core/common_config/main.go
```diff
@@ -99,3 +148,74 @@ func translateInterruptToCancel(ctx context.Context, wg *sync.WaitGroup, cancel
...
+
+	// push done flag to config client
+	err = configClient.PutConfigurationValue(commonConfigDone, []byte("true"))
+	if err != nil {
+		return fmt.Errorf("failed to push %s on completion: %s", commonConfigDone, err.Error())
+	}
+	return nil
+}
+
+// buildKeyValues is a helper function to parse the configuration yaml file
```
lenny-intel, line 199:
```suggestion
// buildKeyValues is a helper function to parse the configuration yaml file contents
```

internal/core/common_config/main.go
```diff
@@ -56,23 +66,62 @@ func Main(ctx context.Context, cancel context.CancelFunc) {
...
+		}
+		_, err = getAccessToken()
+		if err != nil {
+			lc.Errorf("failed to get Access Token for config provider: %s", err.Error())
+			os.Exit(1)
+		}
+	}
 
-	_, err = secretProvider.GetAccessToken("consul", common.CoreCommonConfigServiceKey)
+	lc.Info("Got Config Provider Access Token")
```
jim-wang-intel, line 94: move this logging inside the `if secretProvider != nil`  block once you successfully got token (i.e. after line 91)

lenny-intel (changes requested):

internal/core/common_config/main.go
```diff
@@ -99,3 +148,74 @@ func translateInterruptToCancel(ctx context.Context, wg *sync.WaitGroup, cancel
...
 }
+
+func loadYaml(lc logger.LoggingClient, yamlFile string, configClient configuration.Client) error {
+	// push not done flag to configClient
+	err := configClient.PutConfigurationValue(commonConfigDone, []byte("false"))
+	if err != nil {
+		return fmt.Errorf("failed to push %s on startup: %s", commonConfigDone, err.Error())
+	}
+	lc.Infof("reading %s", yamlFile)
+	contents, err := os.ReadFile(yamlFile)
```
jim-wang-intel, line 159: prefer to normalize this input yamlFile path into the absolute path form before opens it

jim-wang-intel (changes requested):

internal/core/common_config/main.go
```diff
@@ -56,23 +66,62 @@ func Main(ctx context.Context, cancel context.CancelFunc) {
...
-
 	secretProvider, err := secret.NewSecretProvider(nil, environment.NewVariables(lc), ctx, startupTimer, dic, common.CoreCommonConfigServiceKey)
 	if err != nil {
 		lc.Errorf("failed to create Secret Provider: %v", err)
 		os.Exit(1)
 	}
 
 	lc.Info("Secret Provider created")
+	var accessToken string
+	var getAccessToken types.GetAccessTokenCallback
```
lenny-intel, line 77: We don't need to worry about the callback.

internal/core/common_config/main.go
lenny-intel: No we don't do that unless the user uses the in the flags. This is consistent with how bootstrap does it for all the other services.

jim-wang-intel: if the end user specifies the input yamlfile location at free will: it could be something like `../../res/common-configu-file/`  

if that path is kind of built-in then probably it is ok.

internal/core/common_config/main.go
ejlee3: this if statement is not used anymore, so I think this is addressed.

internal/core/common_config/main.go
ejlee3: If we move this to inside the callback, it gets printed twice because we make the callback right after to do the error handling.

internal/core/common_config/main.go
```diff
@@ -76,22 +76,22 @@ func Main(ctx context.Context, cancel context.CancelFunc) {
...
+		return accessToken, err
+	}
+	_, err = getAccessToken()
+	if err != nil {
+		lc.Errorf("failed to get Access Token for config provider: %s", err.Error())
+		os.Exit(1)
 	}
 
-	lc.Info("Got Config Provider Access Token")
+	lc.Infof("Got Config Provider Access Token with length %d", len(accessToken))
```
lenny-intel, line 94: Remove, We don't have the access token at this point anymore. Delayed and taken can of by CreateProviderClient

internal/core/common_config/main.go
```diff
@@ -76,22 +76,22 @@ func Main(ctx context.Context, cancel context.CancelFunc) {
...
-			lc.Errorf("failed to get Access Token for config provider: %s", err.Error())
-			os.Exit(1)
+			return "", fmt.Errorf("failed to get Configuration Provider access token: %s", err.Error())
 		}
+		return accessToken, err
+	}
+	_, err = getAccessToken()
+	if err != nil {
+		lc.Errorf("failed to get Access Token for config provider: %s", err.Error())
+		os.Exit(1)
```
lenny-intel, line 91: I think this can be removed also since CreateProviderClient  is calling it.

internal/core/common_config/main.go
jim-wang-intel: never mind, since `if secretProvider != nil` block is gone, this becomes irrelevant

lenny-intel (changes requested):

internal/core/common_config/main.go
lenny-intel: Actually, move this up into the in-line function where we do have the token.

internal/core/common_config/main.go
ejlee3: ok. didn't know if we needed this error handling or not.

internal/core/common_config/main.go
```diff
@@ -56,23 +66,62 @@ func Main(ctx context.Context, cancel context.CancelFunc) {
...
 
 	lc.Info("Secret Provider created")
+	var accessToken string
+	var getAccessToken types.GetAccessTokenCallback
 
-	_, err = secretProvider.GetAccessToken("consul", common.CoreCommonConfigServiceKey)
+	// need to use in-line function to set the callback type for getAccessToken used in CreateProviderClient to allow
+	// access to the config provider in secure mode
+	getAccessToken = func() (string, error) {
+		accessToken, err = secretProvider.GetAccessToken("consul", common.CoreCommonConfigServiceKey)
```
lenny-intel, line 82:
```suggestion
		accessToken, err := secretProvider.GetAccessToken("consul", common.CoreCommonConfigServiceKey)
```

internal/core/common_config/main.go
lenny-intel: We must support relative paths.

lenny-intel (approved):
LGTM

internal/core/common_config/main.go
jim-wang-intel: yes, absolutely. what i was suggesting is to show the absolute path in logging and before opens it. as it could just find the error in the path via 
```go
  yamlFileAbs, err := filepath.Abs(yamlFile)
```
before the os opens it.  

I am ok not a big deal issue and just suggestion.

lenny-intel: > if that path is kind of built-in then probably it is ok.

Yes, common function in bootstrap.

jim-wang-intel (dismissed):
LGTM

lenny-intel (changes requested):
Need to add the `-cf=configuration.yaml` to `CMD` in the Dockerfile

lenny-intel (approved):
LGTM

jim-wang-intel (approved):

PR #4327, opened 2023-02-06
feat: Vault tokens are now identity-based instead of anonymous

bnevis-i:
This is an internal refactoring of security-secretstore-setup, security-file-token-provider, and security-spiffe-token-provider to generate Vault tokens that are based on a Vault identity instead of being anonymous tokens with an attached ACL.

The consequence of this refactoring is that the token issuing components of EdgeX run with fewer required privileges than before, and it is possible to ask Vault for a verifiable JWT-based identity assertion of an EdgeX service.

This change should be completely transparent to all existing EdgeX services.

## PR Checklist

- [X] I have added unit tests for the new feature or bug fix (if not, why?)
- [ ] I have opened a PR for the related docs change (if not, why?) n/a
  <link to docs PR>

## Testing Instructions

```
# in edge-go
make clean_docker_base docker_base docker
# in compose-builder
make run dev ds-virtual
make get-token
# Go to localhost:4000 with the above token and make sure everything runs as usual
```


cmd/security-spiffe-token-provider/res/configuration.toml
```diff
@@ -38,6 +38,17 @@ Port = 6379
 Timeout = 5000
 Type = "redisdb"
 
+
+[TokenConfig]
+#PrivilegedTokenPath = "UNUSED"
+#ConfigFile = "UNUSED"
+#OutputDir = "UNUSED"
+#OutputFilename = "UNUSED"
```
jim-wang-intel, line 46: if we remove this, shouldn't it become breaking changes in terms of update pov?

internal/security/common/tokenpolicy_test.go
```diff
@@ -13,7 +13,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0'
 //
-package fileprovider
+package common
```
jim-wang-intel, line 16: update license header for year 2023

internal/security/fileprovider/provider.go
```diff
@@ -124,37 +131,29 @@ func (p *fileTokenProvider) Run() error {
...
-		if serviceConfig.CustomTokenParameters != nil {
-			// Custom token parameters override the defaults
-			createTokenParameters = mergeMaps(createTokenParameters, serviceConfig.CustomTokenParameters)
-		}
+		// Generate a random password
 
-		// Set a meta property that consuming serices can use to automatically scope secret queries
-		createTokenParameters["meta"] = map[string]interface{}{
-			"edgex-service-name": serviceName,
+		randomPassword, err := credentialGenerator.Generate(context.TODO())
```
jim-wang-intel, line 136: what is `context.TODO()`?

bnevis-i:
```
TODO returns a non-nil, empty Context. Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter)
```

The original implementation of the credential generator used a key derivation scheme that derived multiple passwords from the same base secret.  We later changed it to just do random bits in base64.  Context is no longer used, but left in the API to avoid having to modify a million places in the code.   Per GO BKM, never pass a nil Context.

cmd/security-spiffe-token-provider/res/configuration.toml
bnevis-i: What this is actually doing is reusing the `configuration.toml` snippet from `security-file-token-provider` except that we don't need all of the parameters, just some of them.  I'm just making it clear that we don't use these parameters in `spiffe-file-token-provider`.

internal/security/fileprovider/provider.go
jim-wang-intel: Thanks for details explanation 👍

cmd/security-spiffe-token-provider/res/configuration.toml
jim-wang-intel: ok

internal/security/common/tokenpolicy_test.go
bnevis-i: Oh no!  Just pushed a commit to update copyright dates.

bnevis-i:
Resolved merge conflict in go.mod.

internal/security/common/usermanager.go
```diff
@@ -0,0 +1,137 @@
...
+		err = m.secretStoreClient.DeleteIdentity(m.privilegedToken, username)
+		if err != nil {
+			return err
+		}
+	}
+
+	err = m.secretStoreClient.DeleteUser(m.privilegedToken, m.userPassMountPoint, username)
+	if err != nil {
+		return err
+	}
```
jim-wang-intel, line 134: should we make this transactional or atomic? if deleting user somehow failed then you lost identity for that user isn't it?

bnevis-i: I think its ok.  The ACL is attached to the identity so if you delete the identity the user is really gone.  You could attempt to log in as the user, but it would error out as it isn't bound to anything.  Moreover, if you recreated the user it would overwrite the identity and/or login, so a partial deletion would take up space but that's about it.

jim-wang-intel: ok, good to know that creating user would get a new identity.

internal/security/fileprovider/provider_test.go
```diff
@@ -157,12 +157,15 @@ func TestNoDefaultsCustomPolicy(t *testing.T) {
 	mockAuthTokenLoader.On("Load", privilegedTokenPath).Return("fake-priv-token", nil)
 
 	expectedService1Policy := `{"path":{"secret/non/standard/location/*":{"capabilities":["list","read"]}}}`
-	expectedService1Parameters := makeMetaServiceName("myservice")
+	//expectedService1Parameters := makeMetaServiceName("myservice")
```
jim-wang-intel, line 160: remove the dead code

jim-wang-intel (approved):

Issue #4173, opened 2022-09-29
Collect and publish SPDX SBOM data for EdgeX releases


bnevis-i:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]
All packages

### Description [**REQUIRED**]
In order to help companies comply with US Executive Order (EO) 14028, Improving the Nation’s Cybersecurity (May 12, 2021), the EdgeX Foundry project should generate SBOM artifacts in the SPBX SBOM format.

References:
- https://www.whitehouse.gov/wp-content/uploads/2022/09/M-22-18.pdf
- https://www.whitehouse.gov/briefing-room/presidential-actions/2021/05/12/executive-order-onimproving-the-nations-cybersecurity/
- https://csrc.nist.gov/Projects/ssdf

This repo has a good index of existing tools:
- https://github.com/awesomeSBOM/awesome-sbom

PR #4339, opened 2023-02-09
feat: generate spdx sbom and add Makefile target for sbom generation

jim-wang-intel:
Closes: #4173

## PR Checklist

  <link to docs PR>

## Testing Instructions

- git clone this branch
- run `make sbom` and it should run successfully and produce sbom under the directory sbom\

## New Dependency Instructions (If applicable)

spdx docker container used for generating spdx-sbom:     https://github.com/opensbom-generator/spdx-sbom-generator


bnevis-i:
Pretty cool!  @lenny-intel do you think we should be checking in the resulting SBOM?

sbom/bom-go-mod.spdx
```diff
@@ -0,0 +1,1299 @@
+SPDXVersion: SPDX-2.2
```
ernestojeda, line 1: Is this something we want to store in Git? This seems more like a build artifact that can be stored in Jenkins or Nexus. I would suggest adding the `sbom` directory to `.gitignore`

Makefile
```diff
@@ -344,3 +346,8 @@ docker_security_spiffe_token_provider: docker_base
 
 vendor:
 	$(GO) mod vendor
+
+sbom:
+	docker run -it --rm \
+		-v "$(CUR_DIR):/edgex-go" -v "$(CUR_DIR)/sbom:/sbom" \
```
ernestojeda, line 352: I think you can use `$$PWD` rather than create a separate variable:

```suggestion
		-v "$$PWD:/edgex-go" -v "$$PWD/sbom:/sbom" \
```

```Makefile
CUR_DIR=$(shell pwd)

foo:
        echo $$PWD
        echo $(CUR_DIR)
```

```sh
$ make foo
echo $PWD
/home/eojeda/make-test
echo /home/eojeda/make-test
/home/eojeda/make-test
```

ernestojeda (commented):
See comments

jim-wang-intel:
> Pretty cool! @lenny-intel do you think we should be checking in the resulting SBOM?

Thanks and good question, @bnevis-i ; depending on what is the satisfying requirement for complying that description about `In order to help companies comply with US Executive Order (EO) 14028, Improving the Nation’s Cybersecurity (May 12, 2021)`. i.e if the artifacts are required in github repo then the answer is yes or just having a way to generate SBOM in spdx format.

Makefile
jim-wang-intel: Hi @ernestojeda thanks for the suggestion, but that doesn't seem working for me. when you use $$PWD it becomes $PWD literally instead of the value of `$PWD`

jim-wang-intel: never mind, i got it wrong. sorry. I'll use your suggestion 👍

sbom/bom-go-mod.spdx
jim-wang-intel: good question. same as @bnevis-i brought it up in https://github.com/edgexfoundry/edgex-go/pull/4339#issuecomment-1424604298 .    @bnevis-i  Probably we can discuss and decide this in the security WG meeting.

Makefile
jim-wang-intel: updated.

bnevis-i (approved):
LGTM

sbom/bom-go-mod.spdx
jim-wang-intel: this file is removed and add it into .gitignore

PR #4345, opened 2023-02-13
feat(data): update validateEvent topic check

hahattan:
event topic from device service will be:
`<prefix>/<service-name>/<profile-name>/<device-name>/<source-name>`

corresponding to edgexfoundry/device-sdk-go#1261

## PR Checklist

  <link to docs PR>

### DIFF
--- a/internal/core/data/controller/messaging/subscriber.go
+++ b/internal/core/data/controller/messaging/subscriber.go
@@ -1,5 +1,5 @@
 //
-// Copyright (C) 2021-2022 IOTech Ltd
+// Copyright (C) 2021-2023 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
@@ -12,7 +12,7 @@ import (
 	"strings"
 
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/config"
-	cbor "github.com/fxamacker/cbor/v2"
+	"github.com/fxamacker/cbor/v2"
 
 	"github.com/edgexfoundry/edgex-go/internal/core/data/application"
 	dataContainer "github.com/edgexfoundry/edgex-go/internal/core/data/container"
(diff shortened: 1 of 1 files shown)


hahattan:
@lenny-intel according to document app-service expects event topic to have the same suffix:
https://docs.edgexfoundry.org/3.0/microservices/general/messagebus/#multi-level-topics-and-wildcards

However if event is added by REST API, coredata will not know the service name:
https://github.com/edgexfoundry/edgex-go/blob/main/internal/core/data/application/event.go#L84

Do we allow this inconsistency ? 
`edgex/events/device/<service>/<profile>/<device>/<source>` vs `edgex/events/core/<profile>/<device>/<source>`

or maybe we update the event topic published by coredata to `edgex/events/core/core-data/<profile>/<device>/<source>` ?

lenny-intel (approved):
LGTM

lenny-intel:
recheck

lenny-intel:
> However if event is added by REST API, coredata does not know the device service name:

@hahattan , the Add Event REST API needs to be update to require the `DeviceService`

hahattan:
> > However if event is added by REST API, coredata does not know the device service name:
> 
> @hahattan , the Add Event REST API needs to be updated to include the `DeviceServiceName`

@lenny-intel REST endpoint only includes device name, profile name and source name.
Do we have any other way to retrieve device service name in this case ?

cloudxxx8:
@hahattan after the discussion in the Core WG meeting on 15th-Feb-2023, we should modify the REST API to fit this change
[/event/{profileName}/{deviceName}/{sourceName}](https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.3.0#/default/post_event__profileName___deviceName___sourceName_)
it will become `/event/{serviceName}/{profileName}/{deviceName}/{sourceName}`

After this API change, you can get the device service name from REST API

lenny-intel:
@hahattan , @cloudxxx8 , FYI may PR here https://github.com/edgexfoundry/edgex-go/pull/4348 took care of this except for updating the swagger. 

@cherrycl , FYI for TAF tests

internal/core/data/controller/messaging/subscriber.go
```diff
@@ -114,11 +114,11 @@ func unmarshalPayload(envelope types.MessageEnvelope, target interface{}) error
 }
 
 func validateEvent(messageTopic string, e dtos.Event) errors.EdgeX {
-	// Parse messageTopic by the pattern `edgex/events/<device-profile-name>/<device-name>/<source-name>`
+	// Parse messageTopic by the pattern `edgex/events/device/<device-service-name>/<device-profile-name>/<device-name>/<source-name>`
 	fields := strings.Split(messageTopic, "/")
 
 	// assumes a non-empty base topic with /profileName/deviceName/sourceName appended by publisher
-	if len(fields) < 4 {
+	if len(fields) < 5 {
```
hahattan, line 121: @lenny-intel Do you think we still need to check this ? I didn't see the change in #4348

cloudxxx8: I believe we need this check, and update the comment in line 120
however, why doesn't it `if len(fields) < 6`?

PR #4388, opened 2023-02-23
refactor!: Re-work handling of secrets

vli11:

## PR Checklist

  <link to docs PR>

## Testing Instructions
make test
make notifications
postman: http://localhost:59880/api/v2/secret
view the log with no error

### DIFF
--- a/cmd/core-common-config-bootstrapper/res/configuration.yaml
+++ b/cmd/core-common-config-bootstrapper/res/configuration.yaml
@@ -2,8 +2,8 @@ all-services:
   Writable:
     InsecureSecrets:
       DB:
-        path: "redisdb"
-        Secrets:
+        SecretName: "redisdb"
+        SecretData:
           username: ""
           password: ""
     
--- a/cmd/core-common-config-bootstrapper/res/configuration.yaml
+++ b/cmd/core-common-config-bootstrapper/res/configuration.yaml
@@ -2,6 +2,10 @@ all-services:
   Writable:
     InsecureSecrets:
       DB:
+        path: "redisdb"
+        Secrets:
+          username: ""
+          password: ""
         SecretName: "redisdb"
         SecretData:
           username: ""
--- a/internal/security/spiffetokenprovider/init.go
+++ b/internal/security/spiffetokenprovider/init.go
@@ -373,7 +373,7 @@ func (b *Bootstrap) seedKnownSecrets(ctx context.Context, lc logger.LoggingClien
 		Type:           ssConfig.Type,
 		Host:           ssConfig.Host,
 		Port:           ssConfig.Port,
-		BasePath:       secretBasePath, // make sure path is like /v1/edgex/secrets/ in global area
+		BasePath:       secretBasePath, // make sure path is like /v1/edgex/secrets in global area
 		SecretsFile:    ssConfig.SecretsFile,
 		Protocol:       ssConfig.Protocol,
 		Namespace:      ssConfig.Namespace,
@@ -392,12 +392,12 @@ func (b *Bootstrap) seedKnownSecrets(ctx context.Context, lc logger.LoggingClien
 	}
 
 	// copy known secrets redisdb from redis-bootstrapper to the requested service with serviceKey
-	secrets, err := secretClient.GetSecret(fmt.Sprintf("/%s/%s", edgexRedisBootstrapperServiceKey, redisSecretName))
+	secrets, err := secretClient.GetSecret(fmt.Sprintf("%s/%s", edgexRedisBootstrapperServiceKey, redisSecretName))
 	if err != nil {
 		return fmt.Errorf("found error on getting secrets: %v", err)
 	}
 
-	err = secretClient.StoreSecret(fmt.Sprintf("/%s/%s", serviceKey, redisSecretName), secrets)
+	err = secretClient.StoreSecret(fmt.Sprintf("%s/%s", serviceKey, redisSecretName), secrets)
 	if err != nil {
 		return fmt.Errorf("found error on storing secrets: %v", err)
 	}
--- a/internal/security/spiffetokenprovider/init.go
+++ b/internal/security/spiffetokenprovider/init.go
@@ -373,7 +373,7 @@ func (b *Bootstrap) seedKnownSecrets(ctx context.Context, lc logger.LoggingClien
 		Type:           ssConfig.Type,
 		Host:           ssConfig.Host,
 		Port:           ssConfig.Port,
-		Path:           secretBasePath, // make sure path is like /v1/edgex/secrets/ in global area
+		BasePath:       secretBasePath, // make sure path is like /v1/edgex/secrets/ in global area
 		SecretsFile:    ssConfig.SecretsFile,
 		Protocol:       ssConfig.Protocol,
 		Namespace:      ssConfig.Namespace,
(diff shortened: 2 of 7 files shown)


lenny-intel (dismissed):
LGTM

internal/security/spiffetokenprovider/init.go
```diff
@@ -392,12 +392,12 @@ func (b *Bootstrap) seedKnownSecrets(ctx context.Context, lc logger.LoggingClien
...
 
 	// copy known secrets redisdb from redis-bootstrapper to the requested service with serviceKey
-	secrets, err := secretClient.GetSecrets(fmt.Sprintf("/%s/%s", edgexRedisBootstrapperServiceKey, redisSecretName))
+	secrets, err := secretClient.GetSecret(fmt.Sprintf("/%s/%s", edgexRedisBootstrapperServiceKey, redisSecretName))
 	if err != nil {
 		return fmt.Errorf("found error on getting secrets: %v", err)
 	}
 
-	err = secretClient.StoreSecrets(fmt.Sprintf("/%s/%s", serviceKey, redisSecretName), secrets)
+	err = secretClient.StoreSecret(fmt.Sprintf("/%s/%s", serviceKey, redisSecretName), secrets)
```
lenny-intel, line 400: Need to test these changes

jim-wang-intel (approved):

cmd/core-common-config-bootstrapper/res/configuration.yaml
```diff
@@ -2,8 +2,8 @@ all-services:
   Writable:
     InsecureSecrets:
       DB:
-        path: "redisdb"
-        Secrets:
+        SecretName: "redisdb"
+        SecretData:
           username: ""
           password: ""
```
lenny-intel, line 8: Need to temporary support old and new until Bootstrap changes propagated to all services.
New will be ignored for those services and old will be ignored for updated services.
```
      DB:
        path: "redisdb"
        Secrets:
          username: ""
          password: ""
        SecretName: "redisdb"
        SecretData:
          username: ""
          password: ""
```

lenny-intel (dismissed):
LGTM

internal/security/spiffetokenprovider/init.go
```diff
@@ -392,12 +392,12 @@ func (b *Bootstrap) seedKnownSecrets(ctx context.Context, lc logger.LoggingClien
 	}
 
 	// copy known secrets redisdb from redis-bootstrapper to the requested service with serviceKey
-	secrets, err := secretClient.GetSecrets(fmt.Sprintf("/%s/%s", edgexRedisBootstrapperServiceKey, redisSecretName))
+	secrets, err := secretClient.GetSecret(fmt.Sprintf("/%s/%s", edgexRedisBootstrapperServiceKey, redisSecretName))
```
jim-wang-intel, line 395: we might need a way to get secrets from other service rather than itself: in this case, spiffetokenprovider is the service to retrieve secrets from redis-bootstrapper service. so this may not work

jim-wang-intel: @bnevis-i what's your thought to fix this?

bnevis-i: Are you sure its broken?  Did we actually change the way it works?

bnevis-i: If worse comes to worse, secretstore-setup has an alternative method in https://github.com/edgexfoundry/edgex-go/blob/main/internal/security/secretstore/password.go that it uses to seed known secrets.

lenny-intel: > we might need a way to get secrets from other service rather than itself: in this case, spiffetokenprovider is the service to retrieve secrets from redis-bootstrapper service. so this may not work

@jim-wang-intel , @bnevis-i , The method now is expecting a name not a path, so the leading slash needs to be removed. The name can be a compound name as this is building with the service key and the secret name i.e. `device-virtual/redisdb`. This is how spiffetokenprovider  is dealing with "others" service.

lenny-intel, line 395:
```suggestion
	secrets, err := secretClient.GetSecret(fmt.Sprintf("%s/%s", edgexRedisBootstrapperServiceKey, redisSecretName))
```

internal/security/spiffetokenprovider/init.go
lenny-intel, line 400:
```suggestion
	err = secretClient.StoreSecret(fmt.Sprintf("%s/%s", serviceKey, redisSecretName), secrets)
```

internal/security/spiffetokenprovider/init.go
jim-wang-intel: > Are you sure it's broken? Did we actually change the way it works?

so, based on the go-mod-secrets implementation: https://github.com/edgexfoundry/go-mod-secrets/blob/main/pkg/types/config.go#L49-L51  the base path from the config would be spiffetokenprovider's base path when the secret name is passed in as argument for GetSecret() to provider.  However, in this corner case, we wanna copy from redis-bootstrapper's secret not itself....

jim-wang-intel: the assumption here is the base path now seems tightly coupled with the service's config itself, doesn't it?

lenny-intel: > the assumption here is the base path now seems tightly coupled with the service's config itself, doesn't it?

Yes, for normal services, but you will see the this service set the base path w/o the service-key so that it can accomplish the task of seeding other services Secret Store

vli11: I think we are good now, I was just running it with Jim, print out the secretConfig, here is the log from spiffetokenprovider, you can see from the log, device virtual path seems to be correct too

level=INFO ts=2023-02-23T21:45:59.821882475Z app=security-spiffe-token-provider source=request.go:96 msg="successfully made request to Read Entity"

level=INFO ts=2023-02-23T21:45:59.887763882Z app=security-spiffe-token-provider source=request.go:96 msg="successfully made request to Create or update user"

level=INFO ts=2023-02-23T21:45:59.888737489Z app=security-spiffe-token-provider source=request.go:96 msg="successfully made request to List auth methods"

level=INFO ts=2023-02-23T21:45:59.890679092Z app=security-spiffe-token-provider source=request.go:96 msg="successfully made request to Create OIDC JWT role"

level=INFO ts=2023-02-23T21:45:59.955047821Z app=security-spiffe-token-provider source=request.go:96 msg="successfully made request to login user (service)"

level=DEBUG ts=2023-02-23T21:45:59.955072329Z app=security-spiffe-token-provider source=init.go:275 msg="seeding the known secrets if any..."

level=INFO ts=2023-02-23T21:45:59.955093093Z app=security-spiffe-token-provider source=init.go:387 msg="spiffe secret config: {vault edgex-vault 8200 /v1/secret/edgex  http    {X-Vault-Token hvs.CAESIKZ-ICEH-v_YyCJwRhPp0tObwH12YjNk7iz0fSf-PYlqGh4KHGh2cy5ROURFYjJwck5aN2FQR0M4ZkVSQ24wNkg} {false   0   }}"

level=DEBUG ts=2023-02-23T21:45:59.956767349Z app=security-spiffe-token-provider source=secrets.go:357 msg="Using Secrets URL of `http://edgex-vault:8200/v1/secret/edgex/security-bootstrapper-redis/redisdb`"

level=INFO ts=2023-02-23T21:45:59.956850539Z app=security-spiffe-token-provider source=secrets.go:277 msg="kick off token renewal with interval: 30m0s"

level=DEBUG ts=2023-02-23T21:45:59.957612364Z app=security-spiffe-token-provider source=secrets.go:424 msg="Using Secrets URL of `http://edgex-vault:8200/v1/secret/edgex/device-virtual/redisdb`"

level=INFO ts=2023-02-23T21:45:59.959867217Z app=security-spiffe-token-provider source=init.go:301 msg="successfully got raw token"

jim-wang-intel: yea, that makes sense now :-)

jim-wang-intel (approved):

lenny-intel (approved):
LGTM

bnevis-i (approved):
Smoke tested under docker, seems to work fine.

PR #4433, opened 2023-03-10
refactor(security)!: Change to camelCase args for secrets-config proxy tls

bnevis-i:
BREAKING CHANGE: To be consistent with other secrets-config sub-commands, the tls command will now accept camelCase command line arguments.  Also fixes issue with -h option resulting in an error in addition to printing help.

## PR Checklist

- [X] I have added unit tests for the new feature or bug fix (if not, why?)
  <link to docs PR>

## Testing Instructions
For help, use secrets-config (command) (subcommand) -h
tls subcommand works as usual, with exception of breaking change to argument casing.

### DIFF
--- a/cmd/secrets-config/README.md
+++ b/cmd/secrets-config/README.md
@@ -22,95 +22,89 @@ Proxy configuration commands (listed below) require access to the secret store m
 
     Points to directory containing a configuration.toml file.
 
+!!! edgey "EdgeX 3.0"
+    The `--confdir` command line option is replaced by `--configDir` in EdgeX 3.0.
+
 # SUBCOMMANDS
 
   * **tls**
 
-    Configure inbound TLS certificate. This command will provision the TLS secrets into the secret store and re-deploy them to Kong. Requires additional arguments:
+    Configure inbound TLS certificate.
+    This command will replace the default TLS certificate created with EdgeX is started for the first time.
+    Requires additional arguments:
 
-    * **--incert** _/path/to/certchain_ (required)
+    * **--inCert** _/path/to/certchain_ (required)
 
       Path to TLS leaf certificate (PEM-encoded x.509) (the file extension is arbitrary).
       If intermediate certificates are required to chain to a certificate authority,
       these should also be included.
       The root certificate authority should not be included.
 
-    * **--inkey** _/path/to/private\_key_ (required)
+    * **--inKey** _/path/to/private\_key_ (required)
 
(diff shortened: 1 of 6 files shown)


cmd/secrets-config/README.md
```diff
@@ -22,95 +22,61 @@ Proxy configuration commands (listed below) require access to the secret store m
...
-      Optional user-specified &quot;key&quot; used for linkage with an incoming JWT via Kong&#39;s config.key\_claim\_name setting (defaults to &quot;iss&quot; field).
-      See
-      [Kong documentation for JWT plugin](https://docs.konghq.com/hub/kong-inc/jwt/#craft-a-jwt-with-publicprivate-keys-rs256-or-es256)
-      for an example of how this parameter is used.
+    ```shell
+    password=password-from-above
 
-    Upon completion, for token-type == "jwt", the command outputs the autogenerated _key_ for the **id** command above.
-    This value must be used during later construction of the JWT.
+    vault_token=$(curl -ks "http://localhost:8200/v1/auth/userpass/login/${username}" -d "{\"password\":\"${password}\"}" | jq -r '.auth.client_token')
```
jim-wang-intel, line 73: what happens if this fails, the developer can re-run or restart the service?

bnevis-i: Vault is always running.  Additional runs of secrets-config adduser will set a new password on the account at which point need to go through the login/generate jwt process again.

jim-wang-intel: could you add what you said into docs to help developers in this case? thanks!

bnevis-i: Added that as well as other documentation.

jim-wang-intel (approved):
LGTM

bnevis-i:
> Thanks. Looks good.
> 
> Any reason we have the documentation both in code and [docs](https://docs.edgexfoundry.org/2.3/security/secrets-config-proxy/)? It is hard to maintain and prone to going out of sync.

Which would you rather I keep?  Would it be OK to just put a README.md to point to the docs, or should I delete it entirely?

bnevis-i:
> > Which would you rather I keep? Would it be OK to just put a README.md to point to the docs, or should I delete it entirely?
> 
> I think the README.md can go entirely in favor of the docs. Keeping a link from README.md to docs isn't necessary IMO as it also adds maintenance burden for the versioning of the URL.

Ok.  I'll merge this since I need to copy it over to docs, then follow up with a remove later.

PR #4436, opened 2023-03-11
fix(security): Add back missing configuration to seed message-bus secrets

bnevis-i:
When starting EdgeX with the mqtt-bus option, security-secretstore-setup needs information as to which services need the message-bus credential seeded.

## PR Checklist

  <link to docs PR>

## Testing Instructions
In edgex-compose, start services with `make run dev ds-mqtt mqtt-bus` option.  Without this fix, most of the services will fail.


lenny-intel (approved):
LGTM

PR #4449, opened 2023-03-15
feat: Add env override capability for common configuration

### LINKED ISSUE #4448: [Common Config Bootstrapper] Apply env overrides to values only when loaded from file
As decide in 3/15/2023 Architects meeting, Config Provider will be the System of Record for services' values. 
This means that env overrides are only applied when configuration is loaded from file, including the common config file.
They are not applied to values loaded from Config Provider.
The overridden values are pushed into the Config Provider the first time the service starts or when the -o flag is used.

lenny-intel:
closes #4448

## PR Checklist

  TBD

## Testing Instructions
Run non-secure EdgeX stack
Build core-common-config from this branch
Set the following
- export ALL_SERVICES_WRITABLE_TELEMETRY_INTERVAL=50s
- export APP_SERVICES_WRITABLE_TELEMETRY_METRICS_MESSAGESRECEIVED=true
- export DEVICE_SERVICES_WRITABLE_TELEMETRY_METRICS_EVENTSSENT=true
 
run common config from command line
- `./core-common-config-bootstrapper -cp -cf configuration.yaml -o`

verify consul has the appropriate overridden values
verify logs contain the following:
```
"Variables override of 'app-services/Writable/Telemetry/Metrics/MessagesReceived' by environment variable: APP_SERVICES_WRITABLE_TELEMETRY_METRICS_MESSAGESRECEIVED=true"
"Variables override of 'all-services/Writable/Telemetry/Interval' by environment variable: ALL_SERVICES_WRITABLE_TELEMETRY_INTERVAL=50s"
"Variables override of 'device-services/Writable/Telemetry/Metrics/EventsSent' by environment variable: DEVICE_SERVICES_WRITABLE_TELEMETRY_METRICS_EVENTSSENT=false"
"Common configuration loaded from file with 3 overrides applied"
"Common configuration has been pushed to into Configuration Provider with overrides applied"
```

### DIFF
--- a/internal/core/common_config/main.go
+++ b/internal/core/common_config/main.go
@@ -118,10 +118,10 @@ func Main(ctx context.Context, cancel context.CancelFunc) {
 		os.Exit(1)
 	}
 
-	lc.Infof("Common configuration exists in Config Provider is %v and overwrite flag is %v", hasConfig, f.OverwriteConfig())
-
 	// load the yaml file and push it using the config client
 	if !hasConfig || f.OverwriteConfig() {
+		lc.Info("Pushing common configuration. It doesn't exists or overwrite flag is set")
+
 		yamlFile := config.GetConfigLocation(lc, f)
 		err = pushConfiguration(lc, yamlFile, configClient)
 		if err != nil {
@@ -129,7 +129,7 @@ func Main(ctx context.Context, cancel context.CancelFunc) {
 			os.Exit(1)
 		}
 	} else {
-		lc.Infof("Skipped pushing common configuration")
+		lc.Info("Skipped pushing common configuration. It already exists and overwrite flag not set")
 	}
 
 	lc.Info("Core Common Config exiting")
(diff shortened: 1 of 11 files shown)


internal/core/common_config/main.go
```diff
@@ -117,17 +118,18 @@ func Main(ctx context.Context, cancel context.CancelFunc) {
 		os.Exit(1)
 	}
 
-	lc.Infof("configuration exists: %v", hasConfig)
+	lc.Infof("Common configuration exists in Config Provider is %v and overwrite flag is %v", hasConfig, f.OverwriteConfig())
```
ejlee3, line 121: can we clarify this log statement? It is a bit confusing on first read - I needed to look at the code to understand what we are logging here...

jim-wang-intel (dismissed):

ejlee3 (dismissed):
LGTM

jim-wang-intel (approved):

farshidtz:
As expected by #4448, the snap tests have started failing since this change because the overrides no longer apply on top of configurations coming from config provider: `Common configuration loaded from the Configuration Provider. No overrides applied`

Need to change the snap tests to no longer test that.

@MonicaisHer

PR #4456, opened 2023-03-16
refactor!: Change configuration file format to YAML

lenny-intel:
BREAKING CHANGE: Configuration files are now in YAML format, Default file name is now configuration.yaml

closes # 4452

## PR Checklist

  TBD

## Testing Instructions
run `make docker`
from compose builder run `make run delayed-start dev`
Verify all core/support/security services bootstrap successfully and contain the following log message:
```
msg="Loaded private configuration from ./res/configuration.yaml"
```


cmd/security-secretstore-setup/res/configuration.yaml
```diff
@@ -0,0 +1,84 @@
...
+  CaFilePath: ""
+  CertFilePath: ""
+  KeyFilePath: ""
+  TokenFolderPath: /vault/config/assets
+  TokenFile: resp-init.json
+  VaultSecretShares: 5
+  VaultSecretThreshold: 3
+  TokenProvider: /security-file-token-provider
+  TokenProviderArgs:
+    - "-configDir"
```
jim-wang-intel, line 33: interesting, is it because it starts with - for configDir so we need to put it into double quotes?

lenny-intel: Yep, seems so

jim-wang-intel (dismissed):
LGTM

vli11 (dismissed):

internal/core/metadata/uom/bootstrap.go
```diff
@@ -10,12 +10,11 @@ import (
 	"os"
 	"sync"
 
+	"github.com/edgexfoundry/edgex-go/internal/core/metadata/container"
 	bootstrapContainer "github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/container"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/startup"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/di"
 	"github.com/pelletier/go-toml"
```
marcpfuller, line 17: Since we are switching to yaml, do we still need `github.com/pelletier/go-toml` ?

lenny-intel: Yes, for now. Still used for some processing , which needs to be refactored  later.

marcpfuller (approved):
LGTM

bnevis-i:
In my testing, few of the containers stay running for more than a minute.  overrides such as REGISTRY_HOST are not working:

```
level=INFO ts=2023-03-17T20:03:54.838229874Z app=core-data source=registry.go:85 msg="Using Registry (consul) from http://localhost:8500"
level=WARN ts=2023-03-17T20:03:54.838420073Z app=core-data source=registry.go:118 msg="registry is not available"
```

bnevis-i (approved):
Works for me with the following fix:

https://github.com/edgexfoundry/edgex-compose/pull/353

cmd/core-data/res/configuration.yaml
```diff
@@ -0,0 +1,27 @@
+MaxEventSize: 25000 # Defines the maximum event size in kilobytes
```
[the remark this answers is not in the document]
lenny-intel: Just tested and any case works except all upper case.

cmd/support-notifications/res/configuration.yaml
```diff
@@ -0,0 +1,39 @@
...
+        username: username@mail.example.com
+        password: ''
+Service:
+  Host: localhost
+  Port: 59860
+  StartupMsg: "This is the Support Notifications Microservice"
+Smtp:
+  Host: smtp.gmail.com
+  Port: 587
+  Sender: jdoe@gmail.com
```
bnevis-i, line 18: Can we **not** put in `jdoe@gmail.com` here?

bnevis-i (changes requested):
I definitely agree with Farshid's observation that the camelCasing with the lowercase first letter is the preferred style for YAML.

I definitely think we shouldn't hard code jdoe@gmail.com into our scripts!

Though I would approve the PR with it in there, I also don't think we should be suggesting smtp.gmail.com either.  I have my printer set up to scan to email via GMail, and no matter how many times I click "allow insecure logins" google turns it back on randomly and my scan to email feature is broken.  Suggesting that Gmail is even close to a supportable outgoing SMTP server is just plain misleading.

lenny-intel:
> I definitely agree with Farshid's observation that the camelCasing with the lowercase first letter is the preferred style for YAML.

I googled preferred style for YAML and found there isn't one and found it is dependent on the project. I.e. python projects with use lower case hyphenated names. This is why compose files uses lower case hyphenated names. ;-)

cmd/support-notifications/res/configuration.yaml
lenny-intel: @bnevis-i , how's this: https://github.com/edgexfoundry/edgex-go/pull/4456/commits/fd18b926304cc8dcd1f8fd5870834444f3f0aebc

bnevis-i: Yes.

bnevis-i (approved):

Issue #4507, opened 2023-04-11
Failed to start support services with delayed-start mode


cherrycl:
# 🐞 Bug Report

### Affected Services [**REQUIRED**]

 The issue is located in: support services

### Is this a regression?

 Yes, the previous version in which this bug was not present was: ....

### Description and Minimal Reproduction [**REQUIRED**]
Deploy EdgeX with attached compose file which enabled delayed-start mode for support service, device-virtual and device-modbus.
[docker-compose.yml.zip](https://github.com/edgexfoundry/edgex-go/files/11197731/docker-compose.yml.zip)

## 🔥 Exception or Error
device-virtual and device-modbus started successfully with delayed-start mode, but support services not
Support services logs just stayed on the below and no next.
<pre><code>
level=INFO ts=2023-04-11T07:25:52.733587527Z app=support-notifications source=secret.go:65 msg="Creating SecretClient"
level=INFO ts=2023-04-11T07:25:52.736665941Z app=support-notifications source=variables.go:457 msg="Variables override of 'SecretStore/RuntimeTokenProvider/Enabled' by environment variable: SECRETSTORE_RUNTIMETOKENPROVIDER_ENABLED=true"
level=INFO ts=2023-04-11T07:25:52.737948554Z app=support-notifications source=variables.go:457 msg="Variables override of 'SecretStore/Host' by environment variable: SECRETSTORE_HOST=edgex-vault"
level=INFO ts=2023-04-11T07:25:52.737998133Z app=support-notifications source=variables.go:457 msg="Variables override of 'SecretStore/RuntimeTokenProvider/Host' by environment variable: SECRETSTORE_RUNTIMETOKENPROVIDER_HOST=edgex-security-spiffe-token-provider"
level=INFO ts=2023-04-11T07:25:52.738077843Z app=support-notifications source=secret.go:159 msg="SecretStore information created with 3 overrides applied"
level=INFO ts=2023-04-11T07:25:52.739677589Z app=support-notifications source=secret.go:75 msg="Reading secret store configuration and authentication token"
level=INFO ts=2023-04-11T07:25:52.739717746Z app=support-notifications source=secret.go:196 msg="runtime token provider enabled"
level=INFO ts=2023-04-11T07:25:52.739737279Z app=support-notifications source=methods.go:138 msg="using Unix Domain Socket at unix:///tmp/edgex/secrets/spiffe/public/api.sock"
</code></pre>

## 🌍  Your Environment
[...]

PR #4509, opened 2023-04-11
fix: remove unnecessary service name tweaks for support services

jim-wang-intel:
Since the docker service keys change in docker-compose file for support services (support-notifications from notifications and support-scheduler from scheduler),
  the tweaks in spire configure script for such behavior is no longer necessary and thus removed it

This should fix part of the problem in issue #4507 

## PR Checklist

  <link to docs PR>

## Testing Instructions

see detailed instructions in that PR: https://github.com/edgexfoundry/edgex-compose/pull/365

### DIFF
--- a/cmd/security-spire-config/seed_builtin_entries.sh
+++ b/cmd/security-spire-config/seed_builtin_entries.sh
@@ -1,7 +1,7 @@
 #!/bin/sh -x
 #
 #  ----------------------------------------------------------------------------------
-#  Copyright (c) 2022 Intel Corporation
+#  Copyright (c) 2022-2023 Intel Corporation
 #
 #  Licensed under the Apache License, Version 2.0 (the "License");
 #  you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


bnevis-i (approved):
LGTM

cmd/security-spire-config/seed_builtin_entries.sh
```diff
@@ -34,8 +34,6 @@ for dockerservice in security-spiffe-token-provider support-notifications suppor
     # Temporary workaround because service name in dockerfile is not consistent with service key.
     # TAF scripts depend on legacy docker-compose service name. Fix in EdgeX 3.0.
     service=`echo -n ${dockerservice} | sed -e 's/app-service-/app-/'`
```
lenny-intel, line 36: Also can remove this tweak not that all services App Services use `app-`

jim-wang-intel: I think this is beyond scope of this issue related PR: if we have application services that are not started with `app-service-` in docker-compose service names then it is a different issue we are dealing with here for support services.

lenny-intel: You can easily change the scope of this PR to be `remove unnecessary service name tweaks` so to reduce the need to do another PR.

jim-wang-intel: @lenny-intel   this PR is for issue #4507

lenny-intel (approved):
LGTM

PR #4557, opened 2023-05-08
fix: Remove Devices Service exists Check from Add/Update Provision Watcher

lenny-intel:
The service name in Provision Watcher is the base name. It may not exist if each device service(s) are running with the -i/--instance flag

## PR Checklist

  <link to docs PR>

## Testing Instructions
Run device simple with the `-i 1`. 
Verify `Simple-Provision-Watcher` is created even though the device service `device-simple` doesn't exist.
Update the `Simple-Provision-Watcher`
Verify no error.


cloudxxx8 (approved):
LGTM

PR #4569, opened 2023-05-16
feat: add dockerfile and script to perform fuzzing test on all swagger files and individual

vli11:

## PR Checklist

  <link to docs PR>

## Testing Instructions
to build: `docker build -f Dockerfile.fuzz -t fuzz-edgex-go:latest .`
to run: `docker run --net host -v $(pwd)/fuzz_results:/fuzz_results fuzz-edgex-go:latest`


Dockerfile.fuzz
```diff
@@ -0,0 +1,16 @@
+FROM mcr.microsoft.com/dotnet/sdk:6.0-alpine as builder
```
jim-wang-intel, line 1: why could not we just use Restler Fuzzer docker image from Microsoft https://hub.docker.com/_/microsoft-restlerfuzzer-restler instead?

fuzzing_docker.sh
```diff
@@ -0,0 +1,22 @@
...
+
+EDGEX_PROJECT_NAME=${1}
+echo $EDGEX_PROJECT_NAME
+SWAGGER_FILE_PATH=${2}
+echo $SWAGGER_FILE_PATH
+SWAGGER_FILE_NAME=${3}
+echo $SWAGGER_FILE_NAME
+
+echo "--compile from swagger file"
+./restler_bin/restler/Restler compile --api_spec /$SWAGGER_FILE_PATH/$SWAGGER_FILE_NAME
```
jim-wang-intel, line 11:
```suggestion
./restler_bin/restler/Restler compile --api_spec /"$SWAGGER_FILE_PATH"/"$SWAGGER_FILE_NAME"
```

fuzzing_docker.sh
```diff
@@ -0,0 +1,22 @@
...
+
+echo "--test the grammar"
+./restler_bin/restler/Restler test --grammar_file ./Compile/grammar.py --dictionary_file ./Compile/dict.json --settings ./Compile/engine_settings.json --no_ssl
+
+# assuming edgex service is already running on host
+echo "--run fuzz-lean"
+./restler_bin/restler/Restler fuzz-lean --grammar_file ./Compile/grammar.py --dictionary_file ./Compile/dict.json --settings ./Compile/engine_settings.json --no_ssl
+
+echo "--copy result logs into $EDGEX_PROJECT_NAME"
+mkdir -p /fuzz_result/$EDGEX_PROJECT_NAME
```
jim-wang-intel, line 21:
```suggestion
mkdir -p /fuzz_result/"$EDGEX_PROJECT_NAME"
```

fuzzing_docker.sh
```diff
@@ -0,0 +1,22 @@
...
+cp -r ./Test/RestlerResults/ /fuzz_results/$EDGEX_PROJECT_NAME/
```
jim-wang-intel, line 22:
```suggestion
cp -r ./Test/RestlerResults/ /fuzz_results/"$EDGEX_PROJECT_NAME"/
```

Dockerfile.fuzz
vli11: it looks like microsoft does not update the restlerfuzzer image as often as the github repo, I can update that later when they have updated one. Currently the dockerfile is based on the latest dockerfile from restler repo.

Dockerfile.fuzz
```diff
@@ -0,0 +1,22 @@
...
+
+RUN apk add --no-cache python3 py3-pip bash coreutils
+
+RUN wget https://github.com/microsoft/restler-fuzzer/archive/refs/tags/v9.2.2.tar.gz
```
bnevis-i, line 5: Sonarqube is flagging this line:

https://sonarcloud.io/project/security_hotspots?pullRequest=4569&id=edgexfoundry_edgex-go&tab=how_to_fix

bnevis-i (changes requested):

lenny-intel:
@vli11 , please add make targets for building and running .

- docker-fuzz
   - `docker build -f Dockerfile.fuzz -t fuzz-edgex-go:latest .`
- fuxx-test 
   - `docker run --net host -v $(pwd)/fuzz_results:/fuzz_results fuzz-edgex-go:latest`

Dockerfile.fuzz
```diff
@@ -0,0 +1,22 @@
...
+
+ENTRYPOINT ["/restler-fuzzer/fuzzing.sh"]
+# CMD of "all" will do fuzz-lean test for core-commmand, core-data, core-metadata, support-notifications, and support-scheduler, basically all services under openapi/v3/ directory
+# This can take more than 20 minutes to finish
+CMD ["all"]
+# CMD core-data below only do fuzz-lean for core-data
+# This can take about 2 minutes to finish
+# You may swap out core-data to other server to perform fuzz-lean
+# Comment above CMD of "all" and uncomment below CMD to fuzz core-data only
+# CMD ["core-data", "/restler-fuzzer/openapi/core-data.yaml"]
```
lenny-intel, line 22: I am not wild about `all` being the default. IMO, the `CMD` should not be set in Dockerfile and expected to be passed in when run. This way we have better control and don't accidently run ALL.

fuzzing_docker.sh
```diff
@@ -0,0 +1,43 @@
...
+    do
+        projectname=$(basename "$swagger" .yaml)
+        echo "$projectname"
+        echo "$swagger"
+        if [[ "$projectname" != *"."* ]]
+        then
+            runFuzzLeanPerSwagger $projectname $swagger
+        fi
+    done
+else
```
lenny-intel, line 39: Check for `EDGEX_PROJECT_NAME` == "" if CMD not set and error out.

lenny-intel (changes requested):

Dockerfile.fuzz
```diff
@@ -0,0 +1,22 @@
+FROM mcr.microsoft.com/dotnet/sdk:6.0-alpine as builder
+
+RUN apk add --no-cache python3 py3-pip bash coreutils
+
+RUN wget -q "https://github.com/microsoft/restler-fuzzer/archive/refs/tags/v9.2.2.tar.gz"
+RUN tar -xvf v9.2.2.tar.gz && mv restler-fuzzer-9.2.2 restler-fuzzer
```
jim-wang-intel, line 6: combine these two runs and delete gz file after mv

jim-wang-intel (changes requested):

Dockerfile.fuzz
bnevis-i: Actually, it is more efficient to just pipe the wget output into tar directly, and rename the resulting directory.

`wget -q -O - (url.tar.gz) | tar xz && mv`

Dockerfile.fuzz
```diff
@@ -0,0 +1,22 @@
...
+
+RUN cd restler-fuzzer; mkdir -p restler_bin
```
bnevis-i, line 8:
```suggestion
RUN mkdir -p /restler-fuzzer/restler_bin
```

Can combine into the previous RUN command as well.

Dockerfile.fuzz
jim-wang-intel: that's great

Dockerfile.fuzz
jim-wang-intel, line 1: add copyright header

Dockerfile.fuzz
```diff
@@ -0,0 +1,45 @@
...
+RUN apk add --no-cache python3 py3-pip bash coreutils
+
+RUN wget -q "https://github.com/microsoft/restler-fuzzer/archive/refs/tags/v9.2.2.tar.gz" && \
+    tar -xvf v9.2.2.tar.gz && \
+    mv restler-fuzzer-9.2.2 restler-fuzzer && \
+    rm v9.2.2.tar.gz && \
+    cd restler-fuzzer && \
+    mkdir -p restler_bin
+
+WORKDIR /restler-fuzzer
```
bnevis-i, line 29:
```suggestion
WORKDIR /restler-fuzzer
RUN wget -q -O - "https://github.com/microsoft/restler-fuzzer/archive/refs/tags/v9.2.2.tar.gz" | \
    tar xvf - --strip-components 1 && \
    mkdir -p restler_bin
```

fuzzing_docker.sh
```diff
@@ -0,0 +1,73 @@
...
+    mkdir -p /fuzz_results/"$1"
+    cp -r ./Test/ /fuzz_results/"$1"/
+}
+
+if [ "$EDGEX_PROJECT_NAME" == "" ]
+then
+    echo "Please provide a valid project name."
+    usage
+    exit 1
+elif [ "$EDGEX_PROJECT_NAME" == "all" ]
```
jim-wang-intel, line 55:
```suggestion
if [ "$EDGEX_PROJECT_NAME" == "" ]
then
    echo "Please provide a valid project name."
    usage
    exit 1
fi
if [ "$EDGEX_PROJECT_NAME" == "all" ]
```
since there is forcing exit 1, we don't need elif

Makefile
```diff
@@ -372,4 +372,4 @@ docker-fuzz:
 	docker build -f Dockerfile.fuzz -t fuzz-edgex-go:latest .
 
 fuzz-test:
-	docker run --net host -v $(pwd)/fuzz_results:/fuzz_results fuzz-edgex-go:latest
+	docker run --net host -v "$$PWD/fuzz_results:/fuzz_results" fuzz-edgex-go:latest
```
jim-wang-intel, line 375:
```suggestion
	docker run --rm -v "$$PWD/fuzz_results:/fuzz_results" fuzz-edgex-go:latest
```
avoid using net host. and add --rm for removing container once it is done

jim-wang-intel (changes requested):

Makefile
vli11: `--net host` is needed. In order to run fuzz-test correctly, user will need to run edgex-compose on the host (ex: `make run no-secty ds-virtual` in edgex-compose/compose-builder/), and then container fuzz test can test against host's edgex running services.

jim-wang-intel: edgeX has its own network, join the same network then

jim-wang-intel: the host net has everything..so it is kind of bad to have it into the container you  are running unless it is system thing like device that you cannot avoid

vli11: @lenny-intel @bnevis-i I see @jim-wang-intel 's point here that `--net host` will give fuzz-test container all the running service network, this may not be an ideal way and we only want to limit only edgex-go service into fuzz-test container, so Jim suggested to use something like `--net edgex_edgex-network`, but this is not working for fuzz-test, because fuzz-test only fire the api call according to swagger file as below
```
servers:
  - url: http://localhost:59882/api/v3
    description: URL for local development and testing
    ```
what would you suggest to do? should we change swagger file url link and use `--net edgex_edgex-network` or we can live with `--net host` for now?

Comment (bnevis-i) [reply Makefile]: I'm OK with it.  This is a test/validation tool, not production code.

**Review by bnevis-i — APPROVED** [review]

File: Makefile (hunk 12 above)
Comment (jim-wang-intel) [reply Makefile]: ok, thanks.

File: fuzzing_docker.sh [hunk 13]
```diff
@@ -0,0 +1,73 @@
...
+SWAGGER_FILE_PATH="/restler-fuzzer/openapi"
+
+usage()
+{
+  echo "Usage:"
+  echo "./fuzzing_docker.sh <EDGEX_PROJECT_NAME> <SWAGGER_FILE_NAME_PATH>"
+  echo
+  echo "<EDGEX_PROJECT_NAME> is required, options: all|core-data|core-command|core-metadata|support-notifications|support-scheduler"
+  echo "<SWAGGER_FILE_NAME_PATH> is required for NOT \"all\" EDGEX_PROJECT_NAME, it is the path and filename of a project swagger file"
+  exit 1
```
jim-wang-intel, line 31: i prefer exit code 0 when showing the usage or help.

vli11: I copy that exit code from compose builder's shell script, so I keep same convention to return 1 here.

jim-wang-intel: ook

bnevis-i (approved):
Re-approved.

jim-wang-intel (approved):
LGTM, except `make test` failed

Makefile
lenny-intel: Agreed, since it is test/validation tool, not production.

Dockerfile.fuzz
```diff
@@ -0,0 +1,41 @@
...
+    tar xz --strip-components 1 && \
+    mkdir -p restler_bin
+
+RUN python3 ./build-restler.py --dest_dir ./restler_bin/
+
+COPY fuzzing_docker.sh /restler-fuzzer/fuzzing.sh
+COPY /openapi/v3/* /restler-fuzzer/openapi/
+
+ENTRYPOINT ["/restler-fuzzer/fuzzing.sh"]
+# CMD core-data below only do fuzz-lean for core-data
```
lenny-intel, line 33:
```suggestion
# CMD  [ "<service-name>", "<swagger-path>"] must be specified as CMD to `docker run`
# Example: CMD core-data below only do fuzz-lean for core-data
```

Dockerfile.fuzz
```diff
@@ -0,0 +1,41 @@
...
+# This can take about 2 minutes to finish
+# You may swap out core-data to other server to perform fuzz-lean
+CMD ["core-data", "/restler-fuzzer/openapi/core-data.yaml"]
+# CMD of "all" will do fuzz-lean test for core-commmand, core-data, core-metadata, support-notifications, and support-scheduler,
```
lenny-intel, line 37:
```suggestion
# Example: CMD of "all" will do fuzz-lean test for core-commmand, core-data, core-metadata, support-notifications, and support-scheduler, 
```

Dockerfile.fuzz
```diff
@@ -0,0 +1,41 @@
...
+RUN python3 ./build-restler.py --dest_dir ./restler_bin/
+
+COPY fuzzing_docker.sh /restler-fuzzer/fuzzing.sh
+COPY /openapi/v3/* /restler-fuzzer/openapi/
+
+ENTRYPOINT ["/restler-fuzzer/fuzzing.sh"]
+# CMD core-data below only do fuzz-lean for core-data
+# This can take about 2 minutes to finish
+# You may swap out core-data to other server to perform fuzz-lean
+CMD ["core-data", "/restler-fuzzer/openapi/core-data.yaml"]
```
lenny-intel, line 36:
```suggestion
# docker run --net host --rm -v "$$PWD/fuzz_results:/fuzz_results" fuzz-edgex-go:latest core-data /restler-fuzzer/openapi/core-data.yaml"]
```

Dockerfile.fuzz
```diff
@@ -0,0 +1,41 @@
...
+# CMD of "all" will do fuzz-lean test for core-commmand, core-data, core-metadata, support-notifications, and support-scheduler, 
+# basically all services under openapi/v3/ directory
+# This can take more than 20 minutes to finish
+# Comment out above CMD and uncomment below CMD of "all" to fuzz all services
+# CMD ["all"]
```
lenny-intel, line 41:
```suggestion
# docker run --net host --rm -v "$$PWD/fuzz_results:/fuzz_results" fuzz-edgex-go:latest all
```

Makefile
```diff
@@ -5,7 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-.PHONY: build clean unittest hadolint lint test docker run sbom
+.PHONY: build clean unittest hadolint lint test docker run sbom docker-fuzz fuzz-test
```
lenny-intel, line 8:
```suggestion
.PHONY: build clean unittest hadolint lint test docker run sbom docker-fuzz fuzz-test-command
```

Makefile
```diff
@@ -367,3 +367,10 @@ sbom:
 	docker run -it --rm \
 		-v "$$PWD:/edgex-go" -v "$$PWD/sbom:/sbom" \
 		spdx/spdx-sbom-generator -p /edgex-go/ -o /sbom/ --include-license-text true
+
+docker-fuzz:
+	docker build -f Dockerfile.fuzz -t fuzz-edgex-go:latest .
+
+fuzz-test:
```
lenny-intel, line 374:
```suggestion
fuzz-test-command:
```

Makefile
```diff
@@ -367,3 +367,10 @@ sbom:
...
+# not joining the edgex-network due to swagger file url pointing to localhost for fuzz testing in the container
+	docker run --net host --rm -v "$$PWD/fuzz_results:/fuzz_results" fuzz-edgex-go:latest
```
lenny-intel, line 376: This should fail when requested changes are made to Dockerfile?
We need a target for each service, starting with `core-command`
Then as you work out the other services, you add targets for each of them. Once all are ready we added a fuzz-test-all target.
```suggestion
	docker run --net host --rm -v "$$PWD/fuzz_results:/fuzz_results" fuzz-edgex-go:latest core-command /restler-fuzzer/openapi/core-command.yaml
```

fuzzing_docker.sh
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env bash
```
lenny-intel, line 1: This script should be in `bin` folder with other scripts so we don't clutter top folder? Dockerfile also? Or should we have a fuzzing folder for these? @jim-wang-intel , @bnevis-i , thoughts?

lenny-intel (changes requested):

fuzzing_docker.sh
jim-wang-intel: yea, i like the idea to have a fuzzing folder and put all fuzz test related in that folder

vli11:
> > LGTM, except `make test` failed
> 
> @vli11 , Hado lint failing on your new docker file.

FROM mcr.microsoft.com/dotnet/sdk:6.0-alpine hadolint error: DL3026 error: Use only an allowed registry in the FROM image; but docker hub does have this official image https://hub.docker.com/_/microsoft-dotnet-sdk/https://hub.docker.com/_/microsoft-dotnet-sdk/. any suggestions? @lenny-intel @bnevis-i

bnevis-i:
> FROM mcr.microsoft.com/dotnet/sdk:6.0-alpine hadolint error: DL3026 error: Use only an allowed registry in the FROM image; but docker hub does have this official image https://hub.docker.com/_/microsoft-dotnet-sdk/https://hub.docker.com/_/microsoft-dotnet-sdk/. any suggestions? @lenny-intel @bnevis-i

I would ignore the error in .hadolint.yaml.  https://github.com/edgexfoundry/edgex-go/blob/main/.hadolint.yml

jim-wang-intel:
> > > LGTM, except `make test` failed
> > 
> > 
> > @vli11 , Hado lint failing on your new docker file.
> 
> FROM mcr.microsoft.com/dotnet/sdk:6.0-alpine hadolint error: DL3026 error: Use only an allowed registry in the FROM image; but docker hub does have this official image https://hub.docker.com/_/microsoft-dotnet-sdk/https://hub.docker.com/_/microsoft-dotnet-sdk/. any suggestions? @lenny-intel @bnevis-i

i wonder what is the reason that Microsoft doesn't publish the latest version to Docker hub. maybe the one in docker hub is stable version

bnevis-i:
> i wonder what is the reason that Microsoft doesn't publish the latest version to Docker hub. maybe the one in docker hub is stable version

Because Microsoft has the better registry?  (Probably true, it supports the latest OCI registry standards.)

bnevis-i (dismissed):
LGTM.  Third time lucky?

Makefile
```diff
@@ -367,3 +367,10 @@ sbom:
 	docker run -it --rm \
 		-v "$$PWD:/edgex-go" -v "$$PWD/sbom:/sbom" \
 		spdx/spdx-sbom-generator -p /edgex-go/ -o /sbom/ --include-license-text true
+
+docker-fuzz:
+	docker build -f Dockerfile.fuzz -t fuzz-edgex-go:latest -f fuzz_test/Dockerfile.fuzz .
```
jim-wang-intel, line 372:
```suggestion
	docker build -t fuzz-edgex-go:latest -f fuzz_test/Dockerfile.fuzz .
```

jim-wang-intel (changes requested):

.hadolint.yml
```diff
@@ -3,6 +3,7 @@ ignored:
   - DL3008
   - DL3018
   - DL3059
+  - DL3026
```
jim-wang-intel, line 6: instead of adding this for all global, i would suggest adding the ignore just in that dockerfile instead

jim-wang-intel (approved):
LGTM

bnevis-i (approved):

lenny-intel (approved):
LGTM

Issue #4615, opened 2023-07-24
Add better error handling when run in hybrid mode with common Config being missing for services


jim-wang-intel:
# 🚀 Tech Debt

### Relevant Package [**REQUIRED**]

 This feature request is for all edgex-go services

### Description [**REQUIRED**]
 A clear and concise description of the problem or missing capability...
Services from edgex-go running in hybrid mode, we wanna a better error messages or handling when the common config is not specified or missing.  One of example is for retrieving DB credentials:
```console
$ ./core-data 
level=INFO ts=2023-07-24T15:46:02.853853279Z app=core-data source=config.go:643 msg="Loading configuration file from res/configuration.yaml"
level=INFO ts=2023-07-24T15:46:02.853979139Z app=core-data source=config.go:227 msg="Private configuration loaded from file with 0 overrides applied"
level=WARN ts=2023-07-24T15:46:02.85419624Z app=core-data source=database.go:92 msg="couldn't retrieve database credentials: InsecureSecrets missing from configuration"
level=WARN ts=2023-07-24T15:46:03.854303098Z app=core-data source=database.go:92 msg="couldn't retrieve database credentials: InsecureSecrets missing from configuration"

```

### Describe the solution you'd like
 If you have a solution in mind, please describe it.
Add the similar error handling messages from `go-mod-bootstrap` to have a better or meaningful error messages in this case.
[See the details in this PR](https://github.com/edgexfoundry/go-mod-bootstrap/pull/566).

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?

PR #4617, opened 2023-07-25
feat: use a better error handling logic and messages when run in hybrid mode with no common config

jim-wang-intel:
Closes: #4615

## PR Checklist

  <link to docs PR>

## Testing Instructions

- git clone this PR
- build all services: `make build` 
- make sure your EdgeX stack is running insecure mode from edgex-compose/compose-builder: `make run no-secty`
- and then you would need to stop the service to test from the stack, like for example `docker stop edgex-core-data`
- to see the better error message and handling, you can start with `core-data`: change directory to `cmd/core-data`
- run the core-data in hybrid mode without any common config: `./core-data`, and you should see some better error messages with suggestion like `missing common conifg? Use -cc or -cp flags from the command line.`
- add the missing elements or properties in configuration.yaml /res to observe it going through the next issue and better error messages.   You would need to repeat this process until the service is starting up and running.
- the services been tested are:
  1. all the EdgeX core services: `core-command, core-data, core-metadata`
  2. EdgeX support services: `support-scheduler, support-notifications`
  3. security services: `security-proxy-auth, security-spiffe-token-provider`
  
changed configuration yaml example for core-data:
```yaml
MaxEventSize: 25000 # Defines the maximum event size in kilobytes
Writable:
  InsecureSecrets:
    DB:
      path: "redisdb"
      Secrets:
        username: ""
        password: ""
      SecretName: "redisdb"
      SecretData:
        username: ""
        password: ""
  LogLevel: "INFO"
  PersistData: true
  Telemetry:
    Metrics: # All service's metric names must be present in this list.
      EventsPersisted: false
      ReadingsPersisted: false
Service:
  Port: 59880
  Host: "localhost"
  StartupMsg: "This is the Core Data Microservice"
  RequestTimeout: "5s"

MessageBus:
  Optional:
    ClientId: "core-data"
  Protocol: "redis"
  Host: "localhost"
  Port: 6379
  Type: "redis"

Database:
  Name: "coredata"
  Host: "localhost"
  Port: 6379
  Timeout: "10s"
  Type: "redisdb"

```

### DIFF
--- a/internal/core/command/main.go
+++ b/internal/core/command/main.go
@@ -91,6 +91,11 @@ func MessagingBootstrapHandler(ctx context.Context, wg *sync.WaitGroup, startupT
 	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
 	configuration := container.ConfigurationFrom(dic.Get)
 
+	if len(configuration.Service.RequestTimeout) == 0 {
+		lc.Error("Service.RequestTimeout found empty in service's configuration, missing common config? Use -cp or -cc flags for common config")
+		return false
+	}
+
 	requestTimeout, err := time.ParseDuration(configuration.Service.RequestTimeout)
 	if err != nil {
 		lc.Errorf("Failed to parse Service.RequestTimeout configuration value: %v", err)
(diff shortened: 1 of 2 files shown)


lenny-intel (approved):
LGTM

Issue #4639, opened 2023-08-16
Fuzzing core-command: 500 errors when invalid input in body http://localhost:59882/api/v3/device/name/Random-Boolean-Device/Bool


vli11:
# 🐞 Bug Report

### Affected Services [**REQUIRED**]
core-command

### Is this a regression?

No

### Description and Minimal Reproduction [**REQUIRED**]

1. use compose-builder command `make run no-secty ds-virtual` to bring up EdgeX system
2. on edgex-go:
         - make docker-fuzz
         - make fuzz-test-command to run fuzzing test on core-command
You will see checker_500 error logs in folder "/edgex-go/fuzz_test/fuzz_results/core-command/FuzzLean/RestlerResults/experiment113/bug_buckets"

3. Or you can reproduce this problem using postman:
`PUT` `http://localhost:59882/api/v3/device/name/Random-Boolean-Device/Bool`
raw JSON "false" into the body

## 🔥 Exception or Error
http status code 500

Postman error:
{
    "apiVersion": "v3",
    "message": "failed to parse request body",
    "statusCode": 500
}

## 🌍  Your Environment
**Deployment Environment:**

**EdgeX Version [**REQUIRED**]:**
3.1.0 or v3

**Anything else relevant?**

PR #4640, opened 2023-08-16
fix(security): core-command fuzzing test http error 500

vli11:
fixed fuzzing core-command http error 500 when invalid input in body http://localhost:59882/api/v3/device/name/Random-Boolean-Device/Bool

Closes: #4639

## PR Checklist

- [x] I have added unit tests for the new feature or bug fix (if not, why?) just updated the unit test
  <link to docs PR>

## Testing Instructions
make docker-fuzz
make fuzz-test-command
1. after fuzz test is finished, open "/edgex-go/fuzz_test/fuzz_results/core-command/FuzzLean/RestlerResults/experiment113/bug_buckets", you should not see checker_500 error logs.
2. go back to edge-go, run `make test`, everything should pass

or you can verify it on postman:
PUT http://localhost:59882/api/v3/device/name/Random-Boolean-Device/Bool
raw JSON "false" into the body
output should not be 500 error, here is the expected output:
{
    "apiVersion": "v3",
    "message": "failed to parse request body",
    "statusCode": 400
}

### DIFF
--- a/internal/core/command/controller/http/command_test.go
+++ b/internal/core/command/controller/http/command_test.go
@@ -478,7 +478,7 @@ func TestIssueSetCommand(t *testing.T) {
 		{"Invalid - execute set command with invalid commandName", testDeviceName, nonExistName, testQueryStrings, testSettingsJsonStr, true, http.StatusBadRequest},
 		{"Invalid - empty device name", "", testCommandName, testQueryStrings, testSettingsJsonStr, true, http.StatusBadRequest},
 		{"Invalid - empty command name", testDeviceName, "", testQueryStrings, testSettingsJsonStr, true, http.StatusBadRequest},
-		{"Invalid - empty settings", testDeviceName, testCommandName, testQueryStrings, []byte{}, true, http.StatusInternalServerError},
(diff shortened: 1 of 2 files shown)


lenny-intel (approved):
LGTM

lenny-intel:
@cloudxxx8 , FYI, this fixes the status code return for invalid input which was triggering many fuzzing failures as it deems a Internal Server error (500) as a failure. In general we should never be returning Internal Server error (500) unless unavoidable, like expected element is missing, stability issues, etc.

jim-wang-intel (approved):

bnevis-i:
> @cloudxxx8 , FYI, this fixes the status code return for invalid input which was triggering many fuzzing failures as it deems a Internal Server error (500) as a failure. In general we should never be returning Internal Server error (500) unless unavoidable, like expected element is missing, stability issues, etc.

To be more specific, HTTP 500 means the web server was incapable of fulfilling the request.  Like out of memory, out of disk, 2+2=5, etc.  Everything else that is detectable should be in the 4xx error range.  For example, AWS returns HTTP 400 (Bad Request) with specific error codes in their JSON responses.

cloudxxx8 (approved):
LGTM

cloudxxx8:
You are right.  The original error code was wrong.

PR #4644, opened 2023-08-22
fix: Repair CORS support in Snap NGINX

bnevis-i:
## PR Checklist

- [X] I have opened a PR for the related docs change (if not, why?)
  <[link to docs PR](https://github.com/edgexfoundry/edgex-docs/pull/1204)>

## Testing Instructions
```
snapcraft
sudo snap install edgexfoundry*.snap --dangerous
sudo snap set edgexfoundry apps.security-bootstrapper-nginx.config.edgex-service-corsconfiguration-corsallowedorigin="*"
sudo snap set edgexfoundry apps.security-bootstrapper-nginx.config.edgex-service-corsconfiguration-enablecors=true
sudo systemctl restart snap.edgexfoundry.security-bootstrapper-nginx
sudo systemctl restart snap.edgexfoundry.nginx.service
curl -ki -X OPTIONS https://localhost:8443/core-data/api/v2/ping
HTTP/1.1 204 No Content
Server: nginx
Date: Wed, 23 Aug 2023 03:08:18 GMT
Connection: keep-alive
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET, POST, PUT, PATCH, DELETE
Access-Control-Allow-Headers: Authorization, Accept, Accept-Language, Content-Language, Content-Type, X-Correlation-ID
Access-Control-Max-Age: 3600
Vary: origin
Content-Type: text/plain; charset=utf-8
Content-Length: 0
```


snap/local/runtime-helpers/bin/setup-nginx.sh
```diff
@@ -14,3 +14,241 @@ if test ! -f "${keyfile}" ; then
...
+  echo "        return 204;" >> "${corssnippet}"
+  echo "      }" >> "${corssnippet}"
+  echo "      if (\$request_method != 'OPTIONS') {" >> "${corssnippet}"
+  # Always add headers regardless of response code.  Omit preflight-related headers (allow-methods, allow-headers, allow-credentials, max-age)
+  echo "        add_header 'Access-Control-Allow-Origin' '${EDGEX_SERVICE_CORSCONFIGURATION_CORSALLOWEDORIGIN}' always;" >> "${corssnippet}"
+  echo "        add_header 'Access-Control-Expose-Headers' '${EDGEX_SERVICE_CORSCONFIGURATION_CORSEXPOSEHEADERS}' always;" >> "${corssnippet}"
+  echo "        add_header 'Vary' 'origin' always;" >> "${corssnippet}"
+  echo "      }" >> "${corssnippet}"
+  echo "" >> "${corssnippet}"
+fi
```
farshidtz, line 59: Trying to understand why we extract from common config and set in the Nginx. Don't the services already set these headers? If yes, Nginx should just pass those headers to the caller.

farshidtz: Thanks for explaining.

3. Can't we pass through all preflight CORS requests to services?
4. Why not skip this step and forward the backend headers to the caller along with the body as in step 8?

bnevis-i: > 3. Can't we pass through all preflight CORS requests to services?
> 4. Why not skip this step and forward the backend headers to the caller along with the body as in step 8?

A combination of things:

a. To pass the CORS request to the service we would have to disable authentication at the proxy layer, which removes the catch-all case.  This would leave exposed any endpoints at the microservice level that didn't implement authentication.  For example, /api/v3/ping, any custom routes the user implemented without authentication, and the rules engine REST API if that is exposed.

b. The browser will withhold sending authentication information until it has performed the preflight request and confirmed that any authorization headers are allowed to be sent.

The combination of the above means we have to handle CORS as the proxy layer.

farshidtz: Not sure that I fully understand the problem. What I meant was to pass through OPTIONS requests to the services which already do respond correctly to them. This won't expose anything sensitive. 

IMO, the proxy could check the authentication header for requests others than OPTIONS and not deal with CORS at all if the backend services already do it correctly.

farshidtz (approved):
Thanks for the fix. It appears to work as expected.

PR #4687, opened 2023-09-20
ci: add automated release workflow on tag creation

ernestojeda:
New release workflow that will create GitHub releases when semantic version tags are pushed.


lenny-intel (approved):
LGTM

bill-mahoney (approved):

PR #4706, opened 2023-10-24
feat!: Remove snap packaging

farshidtz:
Removing snap packaging in accordance with the TSC vote to discontinue release of snaps as of v3.1.

## PR Checklist

  <link to docs PR>


README.md
```diff
@@ -128,7 +123,7 @@ If you want to **include** the delayed start feature in the builds for these ser
 INCLUDE_DELAYED_START_BUILD_CORE:="false"
 ```
 
-For support services, the delayed start feature is included by default as the default behavior of them are not started right away in Snap. Similarly, you can change the default and **exclude** it by modifying the boolean flag from `true` to `false` in the Makefile:
+For support services, the delayed start feature is included by default. Similarly, you can change the default and **exclude** it by modifying the boolean flag from `true` to `false` in the Makefile:
```
farshidtz, line 126: @jim-wang-intel 
Support services aren't disabled by default since v3.0 `edgexfoundry` snap. I suggest excluding the "delayed start" feature for support services to minimize the size of binary distributions.

lenny-intel (approved):
LGTM

PR #4755, opened 2024-02-01
feat: initial attempt at adding OpenZiti to edgex-go

dovholuknf:
WIP: I am only putting this PR up to illustrate how to configure the various services when running locally. There are changes here that likely need to be reverted/changed. This PR is for discussion on initial implemenation

## PR Checklist

  <link to docs PR>


cmd/core-command/res/configuration.yaml
```diff
@@ -13,11 +13,20 @@ Service:
   Host: localhost
   Port: 59882
   StartupMsg: "This is the Core Command Microservice"
+  SecurityOptions:
+    Mode: "zerotrust"
+    OpenZitiController: "openziti:1280"
+    OpenZitiServiceName: "edgex.core-command"
```
lenny-intel, line 19: Put these in the `all-services` section of core-common-config-bootstrapper
and remove them from each service's private config

cmd/core-data/res/configuration.yaml
```diff
@@ -9,16 +9,22 @@ Writable:
 #    Tags: # Contains the service level tags to be attached to all the service's metrics
     ##    Gateway="my-iot-gateway" # Tag must be added here or via Consul Env Override can only change existing value, not added new ones.
 Service:
-  Port: 59880
+  Port: 59810
```
lenny-intel, line 12: Why change the port number?

cmd/core-data/res/configuration.yaml
```diff
@@ -9,16 +9,22 @@ Writable:
...
+    OpenZitiServiceName: "edgex.core-data"
 
 MessageBus:
   Optional:
     ClientId: "core-data"
 
 Database:
   Name: "coredata"
+  Host: "127.0.0.1"
+  Port: 6379
```
lenny-intel, line 27: This is already set in common config. 
When running from command line in hybrid mode use `-cp -d -o` flags

cmd/security-file-token-provider/res/token-config.json
```diff
@@ -175,5 +175,8 @@
         }
       }
     }
+  },
+  "rules-engine": {
+    "edgex_use_defaults": true
```
lenny-intel, line 180: Why is this needed? 
should be `app-rules-engine` if really needed

cmd/support-scheduler/res/configuration.yaml
```diff
@@ -2,9 +2,14 @@ ScheduleIntervalTime: 500
 Writable:
     LogLevel: INFO
 Service:
-    Host: localhost
-    Port: 59861
+    Host: support-scheduler.edgex.ziti
+    Port: 80
```
lenny-intel, line 6: why port change?

internal/support/scheduler/application/scheduler/manager.go
```diff
@@ -6,6 +6,7 @@
 package scheduler
 
 import (
+	clientInterfaces "github.com/edgexfoundry/go-mod-core-contracts/v3/clients/interfaces"
```
lenny-intel, line 9: restore order of imports

lenny-intel (changes requested):

cmd/security-file-token-provider/res/token-config.json
dovholuknf: my understanding is that app-rules-engine and ekuiper are not the same, right? I used the name of the service as found in the docker-file which is `rules-engine:` for ekuiper. it also follows along with the REST requests which are performed by the UI which issues a GET to `http://localhost:4000/rules-engine/ping`

PR #4823, opened 2024-07-10
feat: allow rules-engine service to read its token

dovholuknf:
## PR Checklist

  <link to docs PR>

### DIFF
--- a/cmd/security-secretstore-setup/entrypoint.sh
+++ b/cmd/security-secretstore-setup/entrypoint.sh
@@ -44,6 +44,9 @@ fi
 echo "$(date) Changing ownership of ${EDGEX_SECRETS_ROOT} to ${EDGEX_USER}:${EDGEX_GROUP}"
 chown -Rh ${EDGEX_USER}:${EDGEX_GROUP} "${EDGEX_SECRETS_ROOT}"
 
+echo "$(date) Changing ownership of ${EDGEX_SECRETS_ROOT}/rules-engine to 1001:1001 to accommodate ekuiper"
+chown -Rh 1001:1001 "${EDGEX_SECRETS_ROOT}/rules-engine"
+
 # Signal tokens ready port for other services waiting on
 exec su-exec ${EDGEX_USER} /edgex-init/security-bootstrapper --configDir=/edgex-init/res listenTcp \
   --port="${STAGEGATE_SECRETSTORESETUP_TOKENS_READYPORT}" --host="${STAGEGATE_SECRETSTORESETUP_HOST}"
--- a/cmd/security-secretstore-setup/entrypoint.sh
+++ b/cmd/security-secretstore-setup/entrypoint.sh
@@ -44,11 +44,11 @@ fi
 echo "$(date) Changing ownership of ${EDGEX_SECRETS_ROOT} to ${EDGEX_USER}:${EDGEX_GROUP}"
 chown -Rh ${EDGEX_USER}:${EDGEX_GROUP} "${EDGEX_SECRETS_ROOT}"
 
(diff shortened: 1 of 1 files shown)


cmd/security-secretstore-setup/entrypoint.sh
```diff
@@ -44,6 +44,9 @@ fi
 echo "$(date) Changing ownership of ${EDGEX_SECRETS_ROOT} to ${EDGEX_USER}:${EDGEX_GROUP}"
 chown -Rh ${EDGEX_USER}:${EDGEX_GROUP} "${EDGEX_SECRETS_ROOT}"
 
+echo "$(date) Changing ownership of ${EDGEX_SECRETS_ROOT}/rules-engine to 1001:1001 to accommodate ekuiper"
+chown -Rh 1001:1001 "${EDGEX_SECRETS_ROOT}/rules-engine"
```
cloudxxx8, line 48: why shoud we provide 1001 user permission to it?

dovholuknf: The ekuiper docker container ships with a `kuiper` user assigned to id 1001. Before, kuiper was not secured using the JWT mechanism from vault, but now it needs access to a token in order to access the openziti overlay. 

#### example
```
$ docker run --rm lfedge/ekuiper:1.14 grep kuiper /etc/passwd
kuiper:x:1001:1001:Linux User,,,:/home/kuiper:/sbin/nologin
```

cloudxxx8: Could you claim the ekuiper users with variables as we did for the EdgeX users?  It can also provide the opportunity for users to overwrite via the environment variable.

```
# default User and Group in case never set
if [ -z "${EDGEX_USER}" ]; then
  EDGEX_USER="2002"
  EDGEX_GROUP="2001"
fi
```

cloudxxx8 (changes requested):

cmd/security-secretstore-setup/entrypoint.sh
```diff
@@ -44,6 +44,12 @@ fi
 echo "$(date) Changing ownership of ${EDGEX_SECRETS_ROOT} to ${EDGEX_USER}:${EDGEX_GROUP}"
 chown -Rh ${EDGEX_USER}:${EDGEX_GROUP} "${EDGEX_SECRETS_ROOT}"
 
+ekuiper_uid="${EDGEX_EKUIPER_USER_ID:-1001}"
+ekuiper_gid="${EDGEX_EKUIPER_GROUP_ID:-1001}"
+
+echo "$(date) Changing ownership of ${EDGEX_SECRETS_ROOT}/rules-engine to ${ekuiper_uid}:${ekuiper_gid} to accommodate ekuiper."
+chown -Rh ${ekuiper_uid}:${ekuiper_gid} "${EDGEX_SECRETS_ROOT}/rules-engine"
```
cloudxxx8, line 51: please follow our coding convention as other parts.  use upper case for these variables

cloudxxx8 (changes requested):

cmd/security-secretstore-setup/entrypoint.sh
dovholuknf: i left them lowercase in order to demonstrate they are local and not env vars, but i can capitalize them, sure.

dovholuknf:
all set (also fixed the DCO)

cloudxxx8 (approved):
LGTM

cmd/security-secretstore-setup/entrypoint.sh
cloudxxx8: I agree with your style. We may use this naming convention for the future new script, and apply to the existing script if possible

PR #292, opened 2018-06-09
#283 Remove Unused Schedule Property from domain/models/event.go

yanghua:
Signed-off-by: yanghua <yanghua1127@gmail.com>

cc @tsconn23

PR #4840, opened 2024-07-28
feat(core-data): Add message bus subscription log

M0hanrajp:

## PR Checklist

  <link to docs PR>

## Note

Please refer the discussion: [Subscribe/Publish Message bus topic - Need understanding on service communications #292](https://github.com/orgs/edgexfoundry/discussions/292) 
Conclusion: no logging for core-data subscription. 

## Testing Instructions

Please note that I'm using a custom Makefile from [edgex-native-build](https://github.com/M0hanrajp/edgex-native-build) (My custom configurations)
* I have gone through [Working in a Hybrid Environment](https://docs.edgexfoundry.org/3.1/getting-started/Ch-GettingStartedHybrid/) and at this time this setup is not working for me.
* So I have created a custom Makefile with configurations that help me in bringing up edgex with few commands.
* Forked & cloned the code from [edgex-go](https://github.com/edgexfoundry/edgex-go) & build and test it with my custom configurations.

Owners/maintainers can suggest me any other method of testing if it needs to be done.

And here is the result from my current configurations:
```bash
level=INFO ts=2024-07-28T06:27:08.812725069Z app=core-data source=subscriber.go:53 msg="core-data subscribing to topics: [{edgex/events/device/# 0xc0005e0180}]"
```

### DIFF
--- a/internal/core/data/controller/messaging/subscriber.go
+++ b/internal/core/data/controller/messaging/subscriber.go
@@ -54,6 +54,11 @@ func SubscribeEvents(ctx context.Context, dic *di.Container) errors.EdgeX {
 		return errors.NewCommonEdgeXWrapper(err)
 	}
 
+	// Log the topics to which core-data is subscribing
+	for _, t := range topics {
+		lc.Infof("Subscribed to topics: %s", t.Topic)
+	}
+
 	go func() {
 		for {
 			select {


internal/core/data/controller/messaging/subscriber.go
```diff
@@ -48,6 +48,9 @@ func SubscribeEvents(ctx context.Context, dic *di.Container) errors.EdgeX {
 			Messages: messages,
 		},
 	}
+	
+	// Log the topics to which core-data is subscribing
+	lc.Infof("core-data subscribing to topics: %v", topics)
```
cloudxxx8, line 53: the log should be put after the subscribe action
please move after line 58
since it's the core-data log, you don't need to add "core-data" into the log
also, it's not clear to print the channel, so you can use for loop ot print the topic only

```
	for _, t := range topics {
		lc.Infof("Subscribed to topics: %s", t.Topic)
	}
```

cloudxxx8 (changes requested):

cloudxxx8:
according to the guide line, you need to create an issue first to describe why you want to add the feature.
the commit message prefix should be `feat(core-data): `, no `refactor` because it's an enhancement not refactoring
`feat(core-data): Add msg bus subscription log`

M0hanrajp:
@cloudxxx8 , Need some time to make the changes. Sorry for not understanding a few terms.

Thanks

internal/core/data/controller/messaging/subscriber.go
M0hanrajp: Hello @cloudxxx8, please help to review I have pushed the updated changes. 

Thanks

M0hanrajp:
I have created and updated the Issue #4843

> according to the guide line, you need to create an issue first to describe why you want to add the feature. the commit message prefix should be `feat(core-data): `, no `refactor` because it's an enhancement not refactoring `feat(core-data): Add msg bus subscription log`

cloudxxx8 (approved):
LGTM

cloudxxx8:
@M0hanrajp sorry, that's my bad, I realezed your original commit message is correct
refactor prefix is the right way to go, because there is no behavior change here

cloudxxx8:
it's fine to keep it as is since it has been merged

M0hanrajp:
> @M0hanrajp sorry, that's my bad, I realezed your original commit message is correct refactor prefix is the right way to go, because there is no behavior change here

OK Thanks for letting me know.

Issue #4834, opened 2024-07-23
Create new support-cron-scheduler service


jackchenjc:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]

 This feature request is for creating the new support-cron-scheduler service.

### Description [**REQUIRED**]

This issue has been opened to create a new support-cron-scheduler service for EdgeX. Multiple pull requests will be submitted to facilitate easier review.

cloudxxx8:
close as implementation complete

PR #4844, opened 2024-07-31
feat: Implement Cron Scheduler Controller Package

jackchenjc:
This is the second PR of #4834.

- Implement the REST API controller package for Cron Scheduler service

## PR Checklist

https://github.com/edgexfoundry/edgex-go/pull/4831

## Testing Instructions

- Clone the fork [edgex-go](https://github.com/jackchenjc/edgex-go/tree/issue-4834-2nd) and checkout this branch
- Run the following command at root folder to build local images
```
make docker
```
- Clone this fork [edgex-compose](https://github.com/jackchenjc/edgex-compose/tree/test-cron-scheduler) and checkout  to `test-cron-scheduler` branch
- Head to the compose-builder folder
- Run the whole stack with local images in non-security mode with the following command
```
make run dev no-secty mqtt-bus
```
- Wait for the containers to start and use the following command to check the log of support-cron-scheduler service
```
docker logs edgex-support-cron-scheduler
```
- Use curl command or Postman to test the APIs, e.g. using the following command to check if the controller works as expected.
```
curl 'http://localhost:59863/api/v3/job/all'
```

### DIFF
--- a/internal/pkg/utils/http.go
+++ b/internal/pkg/utils/http.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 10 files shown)


internal/support/cronscheduler/controller/http/schedulejob.go
```diff
@@ -0,0 +1,210 @@
...
+
+	var reqDTOs []requestDTO.UpdateScheduleJobRequest
+	err := jc.reader.Read(r.Body, &reqDTOs)
+	if err != nil {
+		return utils.WriteErrorResponse(w, ctx, lc, err, "")
+	}
+
+	var responses []interface{}
+	for _, dto := range reqDTOs {
+		var response interface{}
```
cloudxxx8, line 174:
```suggestion
		var response any
```

cloudxxx8: replase all interface{} with any

internal/pkg/utils/http.go
```diff
@@ -139,17 +140,30 @@ func ParseTimeRangeOffsetLimit(c echo.Context, minOffset int, maxOffset int, min
...
 	if edgexErr != nil {
 		return start, end, offset, limit, edgexErr
 	}
-	limit, edgexErr = ParseQueryStringToInt(c, common.Limit, common.DefaultLimit, minLimit, maxLimit)
+
+	return start, end, offset, limit, nil
+}
+
+func ParseQueryStringTimeRangeOffsetLimit(c echo.Context, minOffset int, maxOffset int, minLimit int, maxLimit int) (start int, end int, offset int, limit int, edgexErr errors.EdgeX) {
+	// TODO: Create min and max constants for start and end in go-mod-core-contracts?
```
cloudxxx8, line 152: we don't need to do this

internal/pkg/utils/http.go
```diff
@@ -139,17 +140,30 @@ func ParseTimeRangeOffsetLimit(c echo.Context, minOffset int, maxOffset int, min
...
+	start, edgexErr = ParseQueryStringToInt(c, common.Start, 0, 0, math.MaxInt64)
```
cloudxxx8, line 153: since the function signature of ParseQueryStringToInt only accepts `int`, passing `math.MaxInt64` to the function will cause issue on arm32 machine.
On arm32 machine, int range equals to int32

cloudxxx8: use ParsePathParamToInt func

cloudxxx8 (changes requested):

internal/pkg/utils/http.go
cloudxxx8: I opened this issue https://github.com/edgexfoundry/edgex-go/issues/4845
we could handle the int and int64 in the later PR

jackchenjc: Since start, end, offset, and limit here are all URL parameters, we should use the `ParseQueryStringToInt` func, e.g. `/api/v3/job/all?start=1722479643&end=1722479650`.

Maybe we should make the design of this API consistent with the design of the reading and event APIs?
e.g. `/api/v3/event/start/:start/end/:end`

cloudxxx8: let's keep the same API, and handle the int64 in https://github.com/edgexfoundry/edgex-go/issues/4845

cloudxxx8 (approved):
LGTM

Issue #4852, opened 2024-08-07
Include Correlation ID in Logs


jackchenjc:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]

 This feature request is for support-cron-scheduler.

### Description [**REQUIRED**]
 Ensure correlation IDs are included in debug and error logs to simplify future debugging.

### Describe the solution you'd like
 If you have a solution in mind, please describe it.

### Describe alternatives you've considered

Add a new "FromContextOrNew" function in https://github.com/edgexfoundry/edgex-go/blob/main/internal/pkg/correlation/handler.go

PR #4857, opened 2024-08-08
feat: Include Correlation ID in Logs

jackchenjc:
closes: #4852

## PR Checklist

  <link to docs PR>

### DIFF
--- a/internal/pkg/correlation/handler.go
+++ b/internal/pkg/correlation/handler.go
@@ -2,6 +2,9 @@ package correlation
 
 import (
 	"context"
+
+	"github.com/google/uuid"
+
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/common"
 )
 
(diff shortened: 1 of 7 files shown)


internal/pkg/correlation/handler.go
```diff
@@ -12,3 +15,15 @@ func FromContext(ctx context.Context) string {
 	}
 	return hdr
 }
+
+// FromContextOrNew returns the correlation ID from the context if it exists, otherwise it generates a new one and put it back to the context.
+func FromContextOrNew(ctx context.Context) (context.Context, string) {
```
cloudxxx8, line 20: Context is an interface, so it's pointer
We don't have to return it

cloudxxx8 (changes requested):

internal/pkg/correlation/handler.go
cloudxxx8: yes, you are correct

cloudxxx8 (approved):
LGTM

internal/pkg/correlation/handler.go
```diff
@@ -12,3 +15,15 @@
...
 }
+
+// FromContextOrNew returns the correlation ID from the context if it exists, otherwise it generates a new one and put it back to the context.
+func FromContextOrNew(ctx context.Context) string {
+	hdr := FromContext(ctx)
+	if hdr == "" {
+		hdr = uuid.New().String()
+		// lint:ignore SA1029 legacy
+		// nolint:staticcheck // See golangci-lint #741
+		ctx = context.WithValue(ctx, common.CorrelationHeader, hdr)
```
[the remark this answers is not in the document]
cloudxxx8: this is resolved

Issue #4859, opened 2024-08-12
The default auto discovery interval is too short


cloudxxx8:
# 🐞 Bug Report

### Affected Services [**REQUIRED**]

 The issue is located in: common config

### Is this a regression?

No

### Description and Minimal Reproduction [**REQUIRED**]

The default auto discovery internal is 30s currently, and it is too short to most of the use cases.
Most of users don't need the continousely auto discovery schedule.
30s causes lots of computing or network resources waste in the most of cases.
It should be 0s (disabled) by default, and device service developers can modify it in the individual device service to fit their needs.

PR #4860, opened 2024-08-12
fix: Disable the default discovery scheduler

cloudxxx8:
The default auto discovery is disabled, but the default discovery is very short (30s).  When the user enables the auto discovery, it would trigger it every 30s by default.  This value doesn't fit the most of use cases.  By modifying it to 0s, the discovery scheduler will be disabled by default.

BREAKING CHANGE: If the user develops a Device Service with the auto
discovery mechanism and enable it, the discovery won't be automatically
triggered every 30s anymore.  Users have to define this
Device_Discovery_Interval: 30s in the Device Service to keep the
original behavior.

Close https://github.com/edgexfoundry/edgex-go/issues/4859

## PR Checklist

  <link to docs PR>

### DIFF
--- a/cmd/core-common-config-bootstrapper/res/configuration.yaml
+++ b/cmd/core-common-config-bootstrapper/res/configuration.yaml
@@ -136,4 +136,4 @@ device-services:
     Labels: []
     Discovery:
       Enabled: false
-      Interval: "30s"
+      Interval: "0s"


cloudxxx8:
recheck

weichou1229 (approved):
LGTM

PR #4867, opened 2024-08-16
feat: Generate Missed Schedule Action Records

jackchenjc:
This is the sixth PR of #4834

- Load and trigger the existing scheduled jobs from DB
- Calculate and generate the missed schedule action records
- Add missing dependency in Attribution.txt

## PR Checklist

  <link to docs PR>

### DIFF
--- a/internal/support/cronscheduler/application/scheduleactionrecord.go
+++ b/internal/support/cronscheduler/application/scheduleactionrecord.go
@@ -7,9 +7,15 @@ package application
 
 import (
 	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/robfig/cron/v3"
 
 	bootstrapContainer "github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/container"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/di"
+	"github.com/edgexfoundry/go-mod-core-contracts/v3/common"
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/dtos"
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/errors"
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/models"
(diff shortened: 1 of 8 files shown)


internal/support/cronscheduler/application/scheduleactionrecord.go
```diff
@@ -109,10 +115,136 @@ func DeleteScheduleActionRecordsByAge(ctx context.Context, age int64, dic *di.Co
...
 }
 
+// GenerateMissedScheduleActionRecords generates missed schedule action records
+func GenerateMissedScheduleActionRecords(ctx context.Context, dic *di.Container, job models.ScheduleJob, latestRecords []models.ScheduleActionRecord) errors.EdgeX {
+	dbClient := container.DBClientFrom(dic.Get)
+	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
+	correlationId := correlation.FromContext(ctx)
+
+	// All latest records should have the same creation time
+	lastRecordTimestamp := latestRecords[0].Created
```
cloudxxx8, line 125: don't access the harcode index dirctly
if the slice is empty, it will cause the service panic

use `ScheduledAt` instead of `Created`

cloudxxx8: Use for loop to check each record in the next PR (when action uuid is ready) to prevent the missing record happened in the middle of the schedule run.
For example, the scheduld is triggered with 100 actions, and the system crashed in the 70th action running.

internal/support/cronscheduler/application/scheduleactionrecord.go
```diff
@@ -109,10 +115,136 @@ func DeleteScheduleActionRecordsByAge(ctx context.Context, age int64, dic *di.Co
...
+			for _, action := range job.Actions {
+				actionRecord := models.ScheduleActionRecord{
+					JobName:     job.Name,
+					Action:      action,
+					Status:      models.Missed,
+					ScheduledAt: run.UnixMilli(),
+					Created:     missedRecordCreatedTime,
+				}
+
+				record, err := dbClient.AddScheduleActionRecord(ctx, actionRecord)
```
cloudxxx8, line 153: add a new function dbClient.AddScheduleActionRecords(ctx, []actionRecord) to improve the performance

internal/support/cronscheduler/application/schedulejob.go
```diff
@@ -183,3 +183,50 @@ func DeleteScheduleJobByName(ctx context.Context, name string, dic *di.Container
...
+	jobs, err := dbClient.AllScheduleJobs(context.Background(), 0, 999)
+	if err != nil {
+		return errors.NewCommonEdgeX(errors.KindDatabaseError, "failed to load all existing scheduled jobs", err)
+	}
+
+	// TODO: issue-4834-6th - Decide the limit value should be a hard-coded value or support the sql statement that does not limit the number of rows
+	allLatestRecords, err := dbClient.LatestScheduleActionRecords(ctx, 0, 999)
+	if err != nil {
+		return errors.NewCommonEdgeX(errors.KindDatabaseError, "failed to load latest schedule action records", err)
+	}
```
cloudxxx8, line 204: move the allLatestRecords into the `if job.AdminState == models.Unlocked {`
change the API to `getLatestRecordsByJobName(ctx, jobName)`, and you don't need to set any limit to this function

If the job is Locked or before the startTime or after the endTime, we don't need to query the latestRecords

internal/support/cronscheduler/application/scheduleactionrecord.go
```diff
@@ -109,10 +115,136 @@ func DeleteScheduleActionRecordsByAge(ctx context.Context, age int64, dic *di.Co
...
+// TODO: This can be moved to go-mod-core-contracts
 func fromScheduleActionRecordModelsToDTOs(records []models.ScheduleActionRecord) []dtos.ScheduleActionRecord {
 	scheduleActionRecordDTOs := make([]dtos.ScheduleActionRecord, len(records))
 	for i, record := range records {
 		scheduleActionRecordDTOs[i] = dtos.FromScheduleActionRecordModelToDTO(record)
 	}
 	return scheduleActionRecordDTOs
 }
+
+func getLatestRecordsByJobName(jobName string, allLatestRecords []models.ScheduleActionRecord) (latestRecords []models.ScheduleActionRecord) {
```
cloudxxx8, line 177: after we change to getLatestRecordsByJobName, we don't need this function anymore

cloudxxx8 (approved):

cloudxxx8:
merge this PR and fix those change requests in the next PR

PR #4898, opened 2024-09-09
feat: Automatically Trigger Missed Scheduled Jobs

### LINKED ISSUE #4897: Automatically trigger scheduled jobs when the service starts and detects missed schedule action records
# 🚀 Feature Request

 This feature request is for support-cron-scheduler

Automatically trigger scheduled jobs when the service starts and detects missed schedule action records, and support allowing users to configure whether to enable this feature for each job.

jackchenjc:
closes #4897

## PR Checklist

https://github.com/edgexfoundry/edgex-docs/pull/1373

### DIFF
--- a/internal/support/cronscheduler/application/scheduleactionrecord.go
+++ b/internal/support/cronscheduler/application/scheduleactionrecord.go
@@ -117,11 +117,12 @@ func DeleteScheduleActionRecordsByAge(ctx context.Context, age int64, dic *di.Co
 }
 
 // GenerateMissedScheduleActionRecords generates missed schedule action records
-func GenerateMissedScheduleActionRecords(ctx context.Context, dic *di.Container, job models.ScheduleJob, latestRecords []models.ScheduleActionRecord) errors.EdgeX {
+func GenerateMissedScheduleActionRecords(ctx context.Context, dic *di.Container, job models.ScheduleJob, latestRecords []models.ScheduleActionRecord) (errors.EdgeX, bool) {
 	dbClient := container.DBClientFrom(dic.Get)
 	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
 	correlationId := correlation.FromContext(ctx)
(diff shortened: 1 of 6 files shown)


internal/support/cronscheduler/config/config.go
```diff
@@ -25,6 +26,11 @@ type WritableInfo struct {
 	Telemetry       bootstrapConfig.TelemetryInfo
 }
 
+type Scheduler struct {
+	// AutoTrigger is a flag to indicate whether the scheduler should automatically trigger the missed scheduled jobs when the service starts
+	AutoTrigger bool
```
cloudxxx8, line 31: move this config to the job DTO/Model and renamed to `autoTriggerMissedRecords`
the default can be false

internal/support/cronscheduler/application/scheduleactionrecord.go
```diff
@@ -156,14 +158,14 @@ func GenerateMissedScheduleActionRecords(ctx context.Context, dic *di.Container,
...
-				return errors.NewCommonEdgeXWrapper(err)
+				return errors.NewCommonEdgeXWrapper(err), hasMissedAction
 			}
 
 			lc.Debugf("Missed schedule action records with action id: %s of job: %s have been created successfully. Correlation-ID: %s", actionId, job.Name, correlationId)
 		}
 	}
 
-	return nil
+	return nil, hasMissedAction
```
cloudxxx8, line 168:
```suggestion
	return nil, len(missedRecords) > 0
```
declaire `missedRecords` in the beginning

cloudxxx8: use this statement for all the return
try to avoid the individual boolean flag if possible

cloudxxx8 (changes requested):

cloudxxx8 (approved):
LGTM

PR #4924, opened 2024-09-27
fix: Use int64 instead of int for all the timestamp

### LINKED ISSUE #4845: Use int64 instead of int for all the timestamp
# 🐞 Bug Report

 The issue is located in: all

### Is this a regression?

No

We have some timestamp variable using `int` type instead of `int64`
On 32 bit machine the timestamp will be out of range in the future
Golang standard lib use int64 for all the timestamp
https://pkg.go.dev/time#Time.Unix

lindseysimple:
Fixes #4845. Use int64 instead of int for all the timestamp fields.

## PR Checklist

  <link to docs PR>


cloudxxx8 (approved):
LGTM

PR #4929, opened 2024-10-01
feat: Add the force query param to add device metadata API

### LINKED ISSUE #4694: [Core Metadata] Handle plugged devices moving from one node to another node
# 🚀 Feature Request

 This feature request is for Core Metadata

 Unable to successfully add device for USB camera when the camera is unplugged from Node A and plugged into Node B. Device USB Camera on Node B will get an error attempting to add the device for the camera when it discovers it. This is because a device with that name already exists in Core Metadata and is owned by Device USB Camera on Node A. i.e. the device doesn't get deleted when the camera is unplugged from Node A.

This issue is applicable to any devices that are plugged in and can be removed and plugged into a different node.

 One possible solution would be to add the `force` option on the Add Device Request and for the Device SDK to set this to `true` when adding devices that are auto provisioned.

 Have you considered any alternative solutions or workarounds?

lindseysimple:
Resolves #4694. Add the force query param to add device metadata API.

## PR Checklist

  <link to docs PR>

### DIFF
--- a/internal/core/metadata/application/device.go
+++ b/internal/core/metadata/application/device.go
@@ -42,7 +42,7 @@ const minAutoEventInterval = 1 * time.Millisecond
 
 // The AddDevice function accepts the new device model from the controller function
 // and then invokes AddDevice function of infrastructure layer to add new device
-func AddDevice(d models.Device, ctx context.Context, dic *di.Container, bypassValidation bool) (id string, edgeXerr errors.EdgeX) {
+func AddDevice(d models.Device, ctx context.Context, dic *di.Container, bypassValidation bool, force bool) (id string, edgeXerr errors.EdgeX) {
 	dbClient := container.DBClientFrom(dic.Get)
 	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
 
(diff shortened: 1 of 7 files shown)


internal/pkg/infrastructure/postgres/device.go
```diff
@@ -29,11 +29,6 @@ func (c *Client) AddDevice(d model.Device) (model.Device, errors.EdgeX) {
 		d.Id = uuid.New().String()
 	}
 
-	exists, _ := deviceNameExists(ctx, c.ConnPool, d.Name)
```
cloudxxx8, line 32: keep these checks in the infrastructure layer

internal/core/metadata/application/device.go
```diff
@@ -90,6 +104,49 @@ func AddDevice(d models.Device, ctx context.Context, dic *di.Container, bypassVa
 	return addedDevice.Id, nil
 }
 
+// forceAddDevice accepts the updated device model from AddDevice function if force flag is enabled
+// and then invokes UpdateDevice function of infrastructure layer to update the existing device
+// however, "delete device" and "add device" system  events will be published to the msg bus instead of "add device" which indicates a force add device action
+func forceAddDevice(d models.Device, ctx context.Context, dic *di.Container) (id string, edgeXerr errors.EdgeX) {
```
cloudxxx8, line 110: rename to updateDevice

internal/core/metadata/application/device.go
```diff
@@ -90,6 +104,49 @@ func AddDevice(d models.Device, ctx context.Context, dic *di.Container, bypassVa
...
+
+	// If device is successfully updated, check each AutoEvent interval value and display a warning if it's smaller than the suggested 10ms value
+	for _, autoEvent := range d.AutoEvents {
+		utils.CheckMinInterval(autoEvent.Interval, minAutoEventInterval, lc)
+	}
+
+	deviceDTO := dtos.FromDeviceModelToDTO(d)
+	// publish the "delete device" and "add device" system events continuously
+	publishSystemEvent(common.DeviceSystemEventType, common.SystemEventActionDelete, d.ServiceName, deviceDTO, ctx, dic)
+	go publishSystemEvent(common.DeviceSystemEventType, common.SystemEventActionAdd, d.ServiceName, deviceDTO, ctx, dic)
```
cloudxxx8, line 145: change to use update system event

cloudxxx8 (changes requested):

internal/pkg/infrastructure/postgres/device.go
lindseysimple: Done

internal/core/metadata/application/device.go
lindseysimple: Done

internal/core/metadata/application/device.go
lindseysimple: Done

cloudxxx8 (approved):
LGTM

PR #4938, opened 2024-10-04
fix: fix slice init length

cuishuang:
The intention here should be to initialize a slice with a capacity of  `len(scheduleActionRecords)`  rather than initializing the length of this slice. 

The online demo: https://go.dev/play/p/q1BcVCmvidW

## PR Checklist

  <link to docs PR>


cloudxxx8 (approved):
LGTM, thanks for the good catch

PR #4966, opened 2024-10-09
feat: Schedule action record retention implementation

### LINKED ISSUE #4912: Schedule Action Record Retention Implementation
# 🚀 Feature Request

 This feature request is for support-cron-scheduler

Since the schedule action record amount will keep increasing, we need to add a retention mechanism.

jackchenjc:
closes #4912

## PR Checklist

https://github.com/edgexfoundry/edgex-docs/pull/1377
  <link to docs PR>

### DIFF
--- a/cmd/support-cron-scheduler/res/configuration.yaml
+++ b/cmd/support-cron-scheduler/res/configuration.yaml
@@ -31,3 +31,8 @@ Database:
   Timeout: 5s
   Type: postgres
 
+Retention:
+  Enabled: false
+  Interval: 30s    # Purging interval defines when the database should be rid of records above the high watermark.
+  MaxCap: 10000    # The maximum capacity defines where the high watermark of records should be detected for purging the amount of the records to the minimum capacity.
+  MinCap: 8000     # The minimum capacity defines where the total count of records should be returned to during purging.
--- a/cmd/support-cron-scheduler/res/configuration.yaml
+++ b/cmd/support-cron-scheduler/res/configuration.yaml
@@ -32,7 +32,7 @@ Database:
   Type: postgres
 
 Retention:
-  Enabled: false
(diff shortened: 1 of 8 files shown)


cmd/support-cron-scheduler/res/configuration.yaml
```diff
@@ -31,3 +31,8 @@ Database:
   Timeout: 5s
   Type: postgres
 
+Retention:
+  Enabled: false
+  Interval: 30s    # Purging interval defines when the database should be rid of records above the high watermark.
```
cloudxxx8, line 36: Set the default Enabled to true and Interval to 24h
We set false in core-data and support-notification to prevent the breaking change, but there is no breaking change in this service.
The nature of this service is different from core-data and notifications.  The schedule records should not be generated so quick.  Interval 24h should be a fine setting.

internal/support/cronscheduler/application/scheduleactionrecord.go
```diff
@@ -166,6 +169,31 @@ func GenerateMissedScheduleActionRecords(ctx context.Context, dic *di.Container,
...
+			lc := bootstrapContainer.LoggingClientFrom(dic.Get)
+			timer := time.NewTimer(interval)
+			for {
+				timer.Reset(interval)
+				select {
+				case <-ctx.Done():
+					lc.Info("Exiting schedule action records retention")
+					return
+				case <-timer.C:
+					lc.Warn("Start to purge schedule action records")
```
cloudxxx8, line 185:
```suggestion
					lc.Info("Start checking the schedule action records and purge the outdated ones according to the retention settings")
```

cloudxxx8 (changes requested):

cloudxxx8 (approved):
LGTM

PR #4981, opened 2024-10-25
fix: Update default database to PostgreSQL in v4

### LINKED ISSUE #4980: Update default database to PostgreSQL in Common Config in V4
# 🚀 Feature Request

 This feature request is for all EdgeX service which uses database, config provider and registry services.

 The following changes will be applied starting from V4:
- The default database is changed from redis to PostgreSQL in Common Config.
- The default Message Bus is changed from redis to mqtt in Common Config.
- The default configuration provider and registry host are changed from Consul to Core Keeper.
- Remove support-scheduler in the code.

 Related configuration and Dockerfile will be updated.

lindseysimple:
Resolves #4980. Update default database, msg bus, config provider and registry host. Remove support-scheduler code.

## PR Checklist

  <link to docs PR>


cmd/support-notifications/res/configuration.yaml
```diff
@@ -36,10 +31,6 @@ MessageBus:
 
 Database:
   Name: notifications
```
cloudxxx8, line 33: do we still need the database name?

cloudxxx8 (changes requested):
modify the configuration file of support-cron-scheduler to use the commom config database setting.
it doesn't need to use its own anymore

cmd/support-notifications/res/configuration.yaml
lindseysimple: It's not needed by Postgres. Removed.

cloudxxx8 (approved):
LGTM

Issue #5009, opened 2024-11-21
Refactor core services to allow to build monolithic service easily


judehung:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]

 This feature request is for core services

### Description [**REQUIRED**]
 A clear and concise description of the problem or missing capability...
There is a user requirement that prefers to combine all core services into a single monolithic service. To easily achieve this based on current code structure, I'd like to propose some code refactoring:
1. Change the signature of main function for each core services to allow pass os arguments rather than directly parse `os.Args[1:]`, with this change, the single monolithic service can alter the behavior of each individual internal services by passing different arguments.
2. Update the common_config_bootstrapper to `return` instead of `os.Exit(0)`, which will cause the monolithic service to exit directly. 

### Describe the solution you'd like
 If you have a solution in mind, please describe it.
As described in the description above. 

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?
NA

PR #5010, opened 2024-11-21
feat: refactor services to build monolithic service easily

judehung:
fixes https://github.com/edgexfoundry/edgex-go/issues/5009

There is a user requirement that prefers to combine all core services into a single monolithic service. To easily achieve this based on current code structure, refactor following codes:
1. Change the signature of main function for each core services to allow pass os arguments rather than directly parse `os.Args[1:]`. With this change, the single monolithic service can alter the behavior of each individual internal services by passing different arguments.
2. Remove `os.Exit(0)` by end of a successful run of common_config_bootstrapper. This call is redundant and will force monolithic service to exit directly.

## PR Checklist

  <link to docs PR>

## Testing Instructions

make build
run core-keeper, core-common-config-boostrapper, core-command, core-data, core-metadata, and ensure all bootstraps as expected.

### DIFF
--- a/cmd/core-command/main.go
+++ b/cmd/core-command/main.go
@@ -17,6 +17,7 @@ package main
 
 import (
 	"context"
+	"os"
 
 	"github.com/edgexfoundry/edgex-go/internal/core/command"
 
@@ -25,5 +26,5 @@ import (
 
 func main() {
 	ctx, cancel := context.WithCancel(context.Background())
-	command.Main(ctx, cancel, echo.New())
+	command.Main(ctx, cancel, echo.New(), os.Args[1:])
 }
--- a/cmd/core-common-config-bootstrapper/main.go
+++ b/cmd/core-common-config-bootstrapper/main.go
@@ -15,10 +15,12 @@ package main
 
 import (
 	"context"
+	"os"
+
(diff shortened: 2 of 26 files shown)


cloudxxx8 (approved):
LGTM

cloudxxx8 (changes requested):
please make the same change to all other services
don't limit to core services
make them consistence

judehung:
> please make the same change to all other services don't limit to core services make them consistence

applied same changes to all edgex services in https://github.com/edgexfoundry/edgex-go/pull/5010/commits/9598f8521ec5dcf6ea01a1615061cc4ab782fe2b

cloudxxx8 (approved):
LGTM

sansmoraxz:
@judehung could you share a sample of how you have created the monolith service?

I am trying to do the same but the ports are conflicting viz. core-keeper and core-command.

judehung:
> @judehung could you share a sample of how you have created the monolith service?
> 
> I am trying to do the same but the ports are conflicting viz. core-keeper and core-command.

@sansmoraxz please refer to https://github.com/IOTechSystems/edgex-go-low-footprint/tree/main/low-footprint/core-combo-sample#reference-implementation for reference implementation.

PR #5045, opened 2025-01-10
feat: Update reading table schema

weichou1229:
Using id and origin as the index for TimescaleDB integration.

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?) not impact
- [ ] I have opened a PR for the related docs change (if not, why?) not impact
  <link to docs PR>

## Testing Instructions

Since we try to integrate the TimescaleDB as the time series db, we combine `id` and `origin` as the index. 
- replace postgres with timescaleDB
- run SQL to create the hypertable: `SELECT create_hypertable('reding', by_range('origin')); `
- run SQL to set the chunk interval: `SELECT set_chunk_time_interval('reading', 604800000000000);` -- 604800000000000 nanoseconds is for a chunk is 7 days
- check whether the chunk is created `select * from timescaledb_information.chunks;` and the interval is applied `select * from timescaledb_information.dimensions;`

### DIFF
--- a/cmd/core-data/res/db/sql/01-tables.sql
+++ b/cmd/core-data/res/db/sql/01-tables.sql
@@ -15,7 +15,7 @@ CREATE TABLE IF NOT EXISTS core_data.event (
 
 -- core_data.reading is used to store the reading information
 CREATE TABLE IF NOT EXISTS core_data.reading (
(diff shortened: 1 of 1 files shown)


cloudxxx8 (approved):
LGTM

PR #5070, opened 2025-02-06
feat!: Upgrade go-mod-messaging and remove event size check

jinlinGuan:
relate: https://github.com/edgexfoundry/go-mod-messaging/issues/381

Also, removed the Event size checking from Core Data message receiver.  We check the message size before the message publication, so there is no need to check again in the receiver.  Also, when it is received, the message has went through the network and been decoded, so the checking can't save the computing resources.

## PR Checklist

  <link to docs PR>


cloudxxx8 (approved):
LGTM

PR #5071, opened 2025-02-06
feat: Event retention enhancement

### LINKED ISSUE #5064: Core Data Event Retention Enhancement
# 🚀 Feature Request

 This feature request is for core data.

 Since some users want to keep old data for specific devices instead of removing all old data, we can try to improve the retention policy to do count-based or time-based retention.
ref: https://github.com/orgs/edgexfoundry/discussions/323

1. Since we want to retention data by specific device resource, add retention info in the `AutoEvent`
    ```
    type AutoEvent struct {
	  	Interval          string
	  	OnChange          bool
	  	OnChangeThreshold float64
	  	SourceName        string
	  	Retention *Retention
    }
  
    type Retention struct {
	  	MaxCap   int64
	  	MinCap   int64
	  	Duration string
    }
  
    ```
    - `Retention` contains the retention policy for the auto event. if `Retention` is nil, the default values will be applied from the `configuration.yaml`. 
    - `MaxCap` is the maximum capacity of the event, 
       - the high watermark of events should be detected for purging the amount of the event to the minimum capacity, 
       - you can disable the high watermark detetion by setting MaxCap = -1.
    - `MinCap` is the minimum capacity of the event, 
       - the total count of event should be kept in Core Data after purging.
       - you can purge all event if set MinCap = -1
    - `Duration` is the duration to keep the event, 
       - the expired events should be detected for purging, but the service will still keep the number of `MinCap`
       - you can disable the expired checking if set Duration to "0s"
       - Valid time units are "s", "m", "h", e.g., "1.5h" or "2h45m".

2. Using the original `Interval` setting to trigger the event purging process. 
    ```
    Retention: 
      Interval: "10m"
    ```
    - `Interval` is used to define the interval to trigger the event purging process. 
       - The default value is "10m", you can disable this process by using "0s". 
       - Valid time units are "s", "m", "h", e.g., "1.5h" or "2h45m".

3. The default values will be provided if the user doesn't define the Retention info in the AutoEvent.
    ```
    Retention: 
      Interval: "10m"
      DefaultMaxCap: -1
      DefaultMinCap: 1      
      DefaultDuration: "720h"
    ```
    - `DefaultMaxCap`  is the default maximum capacity  The default value is `-1`.
    - `DefaultMinCap` is the default minimum capacity. The default value is `1`. Be careful to use `minCap`, since the database uses `offset` to count the rows, the value becomes larger, and the database needs more time to count the rows.
    - `DefaultDuration` is the default duration to keep the event, the default value is `"720h"`.

For example, if the user defines two auto events as below, the default value will apply to the first auto event,
```
"device": {
  "name": "device_int_autoevents",
  ...
  "profileName": "profile_int_resources",
  "autoEvents": [
    { "interval": "1s", "onChange": false, "sourceName": "INT16_0" },  <= apply the default values from the configuration file
    { "interval": "1s", "onChange": false, "sourceName": "INT16_1", "retention": {"maxCap": 2000, "minCap": 1000, "duration": "30m"}}
  ]
}
```

---
### Use cases description

1. The count-based retention.

    The user can only define the `maxCap` and `minCap` to purge events like the original design for count-based retention.
    ```
         "device": {
             "name": "device_int_autoevents",
             ...
             "profileName": "profile_int_resources",
             "autoEvents": [
                 { 
                     "interval": "1s", "onChange": false, "sourceName": "INT16_1", 
                     "retention": {
                        "maxCap": 2000, 
                        "minCap": 1000, 
                        "duration": "0s"
                     }
                  }
               ]
            }
    ```
    In this case, the Core Data check whether the event count exceeds 2000, 
    if count >= 2000, 
         - then purge events to meet the minimum cap 1000. 
         - If minCap is -1, remove all old data. 

2. The time-based retention without minCap

	The user can only define the `duration` to purge events for time-based retention.
	  ```
		"device": {
      "name": "device_int_autoevents",
      ...
      "profileName": "profile_int_resources",
      "autoEvents": [
        { 
          "interval": "1s", "onChange": false, "sourceName": "INT16_1", 
          "retention": {
              "maxCap": -1, 
              "minCap": -1, 
              "duration": "24h"
          }
        }
       ]
     }
    ```
    In this case, the `minCap` is -1 which means we don't keep old data, then the Core Data purge events that the age(current time - event origin) is greater than "24h".
    
3. The time-based retention with minCap
	The user can only define the `duration` and `minCap` for time-based retention and also keep minimum expired events in DB. Note that the `minCap` is used to keep old data when purging events, the process does not count the new data.
	This is useful when the device unexpectedly shutdown, and we still want to keep some old data for tracing. 
  
	  ```
	  "device": {
	    "name": "device_int_autoevents",
	    ...
	    "profileName": "profile_int_resources",
	    "autoEvents": [
	        { 
	          "interval": "1s", "onChange": false, "sourceName": "INT16_1", 
	          "retention": {
	              "maxCap": -1, 
	              "minCap": 1000, 
	              "duration": "24h"
	          }
	        }
	    ]
	  }
	  ```
	  In this case, the Core Data purges events that the age(current time - event origin) is greater than "24h" and leaves 1000 events in DB.
	  The implementation might be,
      - find the 1000th record => SELECT * FROM edgex_db.event WHERE devicename=? and sourcename=? ORDER BY origin DESC OFFSET 1000 LIMIT 1
      - if the 1000th record doesn't exist, skip purging.
      - if the 1000th record exists, remove the expired data that is older than the 1000th record
        - DELETE FROM edgex_db.reading WHERE event_id=ANY( SELECT id FROM edgex_db.event WHERE devicename=? and sourcename=? and origin < 1000th.origin)
        - DELETE FROM edgex_db.event WHERE devicename=? and sourcename=? and origin < 1000th.origin
       
   
4. The time-based retention with `maxCap` and `minCap`
	The user can combine count-based and time-based retention by defining the `maxCap`, `minCap`, and `duration`.
	
	  ```
	  "device": {
	    "name": "device_int_autoevents",
	    ...
	    "profileName": "profile_int_resources",
	    "autoEvents": [
	        { 
	          "interval": "1s", "onChange": false, "sourceName": "INT16_1", 
	          "retention": {
	              "maxCap": 2000, 
	              "minCap": 1000, 
	              "duration": "24h"
	          }
	        }
	    ]
	  }
	  ```
	  In this case, the Core Data checks whether the event count exceeds 2000, 
	  if count >= 2000, do the time-based retention as in the previous section.

weichou1229:
Add retention policy on auto event and do time-based and count-based retention enhancement

Closes #5064 

## PR Checklist

- [x] I have opened a PR for the related docs change (if not, why?) https://github.com/edgexfoundry/edgex-docs/pull/1415
  <link to docs PR>


weichou1229:
This PR based on the core-contracts changes https://github.com/edgexfoundry/go-mod-core-contracts/pull/975

internal/core/data/config/config.go
```diff
@@ -40,10 +41,14 @@ type WritableInfo struct {
...
 
 type ReadingRetention struct {
-	Enabled  bool
+	Interval        string
+	DefaultMaxCap   int64
+	DefaultMinCap   int64
+	DefaultDuration string
+}
+
+type AutoEventRetention struct {
```
cloudxxx8, line 50:
```suggestion
type EventRetention struct {
```

internal/core/data/config/config.go
```diff
@@ -40,10 +41,14 @@ type WritableInfo struct {
 }
 
 type ReadingRetention struct {
```
cloudxxx8, line 43: remove this struct

internal/pkg/cache/device.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// RemoveAll removes all device out of the active device list.
+func (s *activeDeviceStore) RemoveAll() {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	if len(s.activeDeviceMap) == 0 {
+		return
+	}
+	s.activeDeviceMap = make(map[string]models.Device)
```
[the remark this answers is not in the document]
weichou1229: Updated.

internal/pkg/cache/device.go
```diff
@@ -0,0 +1,95 @@
...
+	bootstrapContainer "github.com/edgexfoundry/go-mod-bootstrap/v4/bootstrap/container"
+	"github.com/edgexfoundry/go-mod-bootstrap/v4/di"
+	"github.com/edgexfoundry/go-mod-core-contracts/v4/clients/logger"
+	"github.com/edgexfoundry/go-mod-core-contracts/v4/models"
+)
+
+var (
+	initStoreOnce sync.Once
+	store         *activeDeviceStore
+)
```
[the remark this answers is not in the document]
weichou1229: Removed the global var.

internal/core/data/config/config.go
```diff
@@ -21,14 +21,15 @@ import (
...
-	Retention    ReadingRetention
+	Writable           WritableInfo
+	Clients            bootstrapConfig.ClientsCollection
+	MessageBus         bootstrapConfig.MessageBusInfo
+	Database           bootstrapConfig.Database
+	Registry           bootstrapConfig.RegistryInfo
+	Service            bootstrapConfig.ServiceInfo
+	MaxEventSize       int64
+	Retention          ReadingRetention
+	AutoEventRetention AutoEventRetention
```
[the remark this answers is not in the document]
weichou1229: Removed the AutoEventRetention.

internal/core/data/config/config.go
weichou1229: Rename ReadingRetention to EventRetention.

internal/core/data/config/config.go
weichou1229: Removed the AutoEventRetention struct.

internal/core/data/application/reading.go
```diff
@@ -309,28 +320,146 @@ func AsyncPurgeReading(interval time.Duration, ctx context.Context, dic *di.Cont
...
+		autoEvent.Retention.MaxCap = config.Retention.DefaultMaxCap
+	}
+	if autoEvent.Retention.MinCap == 0 {
+		autoEvent.Retention.MinCap = config.Retention.DefaultMinCap
+	}
+	if autoEvent.Retention.Duration == "" {
+		autoEvent.Retention.Duration = config.Retention.DefaultDuration
+	}
+
+	if autoEvent.Retention.MaxCap >= 0 {
```
cloudxxx8, line 353:
```suggestion
	if autoEvent.Retention.MaxCap > 0 {
```

cloudxxx8 (changes requested):

internal/core/data/application/reading.go
weichou1229: Updated and upgraded dependency libs.

cloudxxx8 (approved):
LGTM

PR #5112, opened 2025-03-12
fix: fix error when empty scheme in Rest channel

judehung:
This is to fix an error for support-notifications with followng case: users add a subscription with Rest type channel but doesn't specify shceme will result in empty scheme when sending notifications, thus failing the sending operation.

This fix is to specify default http scheme when scheme is empty.

## PR Checklist

  <link to docs PR>


cloudxxx8 (approved):
LGTM

PR #5127, opened 2025-04-08
docs: Update CoreData API doc

weichou1229:
Update CoreData API doc and fix format

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?) not impact
- [ ] I have fully tested (add details below) this the new feature or bug fix (if not, why?) not impact
- [ ] I have opened a PR for the related docs change (if not, why?) not impact
  <link to docs PR>

### DIFF
--- a/openapi/core-data.yaml
+++ b/openapi/core-data.yaml
@@ -351,6 +351,15 @@ components:
         minimum: 0
         default: 0
       description: "The number of items to skip before starting to collect the result set."
+    readingOffsetParam:
+      in: query
+      name: offset
+      required: false
+      schema:
+        type: integer
+        minimum: -1
+        default: 0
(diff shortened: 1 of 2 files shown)


openapi/core-data.yaml
```diff
@@ -351,6 +351,15 @@ components:
         minimum: 0
         default: 0
       description: "The number of items to skip before starting to collect the result set."
+    readingOffsetParam:
+      in: query
+      name: offset
+      required: false
+      schema:
+        type: integer
+        minimum: 0
```
judehung, line 360: If API allows offset to be -1, should we specify the minimum to be -1 instead of 0 here?

weichou1229: You're right, updated.

judehung (approved):
LGTM

PR #5130, opened 2025-04-11
fix: Allow any value by setting additionalProperties: true

### LINKED ISSUE #5129: Potential mismatch between Swagger schema and actual protocols structure
# 📚 Docs or Wiki Bug Report

There may be a mismatch between the documented Swagger schema and the actual structure of the protocols object used in Device.

Based on the Swagger documentation for defining dictionaries: https://swagger.io/docs/specification/v3_0/data-models/dictionaries/

> To define a dictionary, use type: object and use the additionalProperties keyword to **specify the _type of values_ in key/value pairs**.

Swagger doc example:
```ts
{ "en": "English", "fr": "French" }
````
is defined using the following schema:
```yaml
type: object
additionalProperties:
  type: string
```
So my understanding is that the type in `additionalProperties` refers to the type of the value in a key/value pair - in swagger doc example, “English” and “French” are strings.

Now, looking at the definition in `protocol` and `ProtocolProperties`, The current definition is:
https://github.com/edgexfoundry/edgex-go/blob/285c4605f0d8a1204ede5df9efc844b172296372/openapi/core-metadata.yaml#L168-L172
https://github.com/edgexfoundry/edgex-go/blob/285c4605f0d8a1204ede5df9efc844b172296372/openapi/core-metadata.yaml#L978-L982
From this, we can interpret:

- `protocols` is an `object`, where each value is a `ProtocolProperties`
    ```ts
    protocols: {
      [key]: ProtocolProperties
    }
    ```
- `ProtocolProperties` is also an `object`, where each value is of type `object`
    ```ts
    ProtocolProperties: {
      [key]: object
    }
    ```

Which means the full type structure is:
```ts
protocols: {
  [key: string]: {
    [key: string]: object;
  };
}
```
However, in practice, the values under `ProtocolProperties` often include primitive like `strings`, `numbers`, and `booleans`.
Example `ProtocolProperties` value for a ModbusRTU device:
```ts
protocols: {
  "modbusRtu": {
    "Address": "/dev/ttyS2", // string
    "BaudRate": 1200, // number
    "LinkRecovery": false // boolean
  }
}
```
These values would be invalid under the current schema, which restricts them to only `objects`.

### What's the affected URL? [**REQUIRED**]
https://github.com/edgexfoundry/edgex-go/blob/285c4605f0d8a1204ede5df9efc844b172296372/openapi/core-metadata.yaml#L168-L172
https://github.com/edgexfoundry/edgex-go/blob/285c4605f0d8a1204ede5df9efc844b172296372/openapi/core-metadata.yaml#L978-L982

#### Expected:
The schema allows protocol property values to be primitive like `strings`, `numbers`, `booleans`, etc.

#### Actual:
The current schema restricts values to type: `object`, which makes primitive values invalid.

## 📷Screenshot 

N/A

## 🔥 Exception or Error
N/A

## 🌍  Your Environment
N/A

jumpingliu:
fix #5129 

## PR Checklist

  <link to docs PR>

### DIFF
--- a/openapi/core-metadata.yaml
+++ b/openapi/core-metadata.yaml
@@ -978,8 +978,7 @@ components:
     ProtocolProperties:
       description: A map of properties for the given protocol
       type: object
-      additionalProperties:
-        type: object
+      additionalProperties: true
     ResourceOperation:
       description: "Defines an operation of which a device is capable."
       type: object


jumpingliu:
@cloudxxx8 Hi Cloud, this is the draft PR, could you please review this PR first to see my understanding is correct? Allow any value by setting additionalProperties: true

cloudxxx8 (approved):
LGTM

Issue #5134, opened 2025-04-18
Allow offset = -1 to disable total count for the Core Data event query


weichou1229:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]

 This feature request is for Core Data

### Description [**REQUIRED**]
The total count was designed for pagination, so it's not useful if the client doesn't need the pagination for the query.
If the Offset = -1, it means the pagination is not required, and the total couut could be ignored (return 0).

- add default offset configurable for event and reading API
-  skip event total count if the offset is -1

PR #5135, opened 2025-04-18
feat: Query events without total count

weichou1229:
fixes https://github.com/edgexfoundry/edgex-go/issues/5134

Query events without total count to prevent large dataset query and count.

## PR Checklist

- [x] I have opened a PR for the related docs change (if not, why?) updated the swagger file
  <link to docs PR>

## Testing Instructions

Run core service and device service to test the event or reading count.

### DIFF
--- a/openapi/core-data.yaml
+++ b/openapi/core-data.yaml
@@ -348,18 +348,9 @@ components:
       required: false
       schema:
         type: integer
-        minimum: 0
-        default: 0
-      description: "The number of items to skip before starting to collect the result set."
-    readingOffsetParam:
-      in: query
-      name: offset
-      required: false
-      schema:
-        type: integer
         minimum: -1
         default: 0
-      description: "The number of items to skip before starting to collect the result set. Specify -1 will not count the total reading to improve the performance for the large dataset."
+      description: "The number of items to skip before starting to collect the result set. Setting this value to -1 will skip counting the total number of events or readings, which can improve performance when working with large datasets."
     limitParam:
       in: query
       name: limit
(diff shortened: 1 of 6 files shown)


cmd/core-data/res/configuration.yaml
```diff
@@ -9,6 +9,7 @@ Writable:
 #    Tags: # Contains the service level tags to be attached to all the service's metrics
     ##    Gateway="my-iot-gateway" # Tag must be added here or via Consul Env Override can only change existing value, not added new ones.
   EventPurge: false # Remove the related events and readings once received the device deletion system event
+  DefaultQueryOffset: 0
```
judehung, line 12: Why do we allow users to configure their preferable default query offset, but not other query parameters?

Moreover, the go-mod-core-contracts actually defines DefaultOffset. https://github.com/edgexfoundry/go-mod-core-contracts/blob/main/common/constants.go#L216

openapi/core-data.yaml
```diff
@@ -343,23 +343,14 @@ components:
...
-    readingOffsetParam:
       in: query
       name: offset
       required: false
       schema:
         type: integer
         minimum: -1
         default: 0
-      description: "The number of items to skip before starting to collect the result set. Specify -1 will not count the total reading to improve the performance for the large dataset."
+      description: "The number of items to skip before starting to collect the result set. Specify -1 will not count the total event or reading to improve the performance for the large dataset."
```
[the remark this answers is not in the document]
weichou1229: Updated the description.

cmd/core-data/res/configuration.yaml
weichou1229: If we allow users to configure the default query offset, the user can query the API without putting the parameter.

judehung (approved):
LGTM

Issue #5150, opened 2025-05-21
Bug: Windows Path Separator Compatibility Issue in getSortedSqlFileNames Function


2018yuli:
While developing on Windows with EdgeX v4.0.0, I encountered an issue where PostgreSQL (pg) does not initialize properly. This is due to the handling of file paths in the getSortedSqlFileNames function within pkg/db/postgres/utils.go.

The current implementation:
```go
sqlFiles = append(sqlFiles, sqlFileName{order, filepath.Join(sqlFilesDir, fileName)})
```
assumes the Unix-style path separator, which causes issues on Windows.

**Suggested Fix:**
To ensure compatibility with Windows file path separators, the path should be normalized using filepath.ToSlash:  
```go
sqlFiles = append(sqlFiles, sqlFileName{order, filepath.ToSlash(filepath.Join(sqlFilesDir, fileName))})
```

This change ensures that file paths are handled correctly across different platforms (Windows, Linux, etc.).

I understand that EdgeX is primarily designed for Linux environments, but this small fix would improve compatibility for Windows developers.

Thank you for considering this fix!

cloudxxx8:
thanks for reporting it.
since you have already had a solution, are you able to open a PR for it?

PR #5151, opened 2025-05-21
fix: Windows Path Separator Compatibility Issue in getSortedSqlFileNames Function (#5150)

2018yuli:
**Problem:**
While developing on Windows with EdgeX v4.0.0, I encountered an issue where PostgreSQL (pg) does not initialize properly. This issue is caused by the handling of file paths in the getSortedSqlFileNames function within pkg/db/postgres/utils.go.

The current implementation:

```go
sqlFiles = append(sqlFiles, sqlFileName{order, filepath.Join(sqlFilesDir, fileName)})
```
assumes the Unix-style path separator, which causes issues on Windows systems.

**Suggested Fix:**
To ensure compatibility with Windows file path separators, I propose normalizing the path using filepath.ToSlash, as shown below:

```go
func getSortedSqlFileNames(embedFiles embed.FS, sqlFilesDir string) ([]string, error) {
    ...
    sqlFiles = append(sqlFiles, sqlFileName{order, NormalizePath(sqlFilesDir, fileName)})
....
}

// NormalizePath normalizes file paths to be platform-independent.
func NormalizePath(baseDir, fileName string) string {
    return filepath.ToSlash(filepath.Join(baseDir, fileName))
}
```

**Rationale:**
Although EdgeX is primarily designed for Linux environments, this small fix will improve compatibility for developers working on Windows, making the project more accessible to a wider audience.

**Related Issue:**
This change is in response to the issue where PostgreSQL initialization fails on Windows due to path handling issues.

**Testing Instructions:**
* [x] I am not introducing a breaking change
* [x]  I am not introducing a new dependency
* [ ] I have added unit tests for the new feature or bug fix
* [x] I have fully tested this new feature or bug fix
```
level=INFO ts=2025-05-21T13:05:40.3388986+08:00 app=core-metadata source=message.go:58 msg="Service started in: 78.3767ms"
```


internal/pkg/db/postgres/utils.go
```diff
@@ -144,3 +144,8 @@ func WrapDBError(message string, err error) errors.EdgeX {
 	}
 	return errors.NewCommonEdgeX(errors.KindDatabaseError, message, err)
 }
+
+// NormalizePath normalizes file paths to be platform-independent.
+func NormalizePath(baseDir, fileName string) string {
+    return filepath.ToSlash(filepath.Join(baseDir, fileName))
+}
```
cloudxxx8, line 151: add an empty line

cloudxxx8:
you need to sign-off the commit to pass the DCO

cloudxxx8 (changes requested):

internal/pkg/db/postgres/utils.go
```diff
@@ -144,3 +144,8 @@ func WrapDBError(message string, err error) errors.EdgeX {
 	}
 	return errors.NewCommonEdgeX(errors.KindDatabaseError, message, err)
 }
+
+// NormalizePath normalizes file paths to be platform-independent.
+func NormalizePath(baseDir, fileName string) string {
+    return filepath.ToSlash(filepath.Join(baseDir, fileName))
```
cloudxxx8, line 150: according to the go docs, filepath.Join should use the OS separator.
https://pkg.go.dev/path/filepath#Join

Windows is `\` and Linux is `/`.  If you wrap it with filepath.ToSlash, will it become always `/`?
Could this really solve the issue?

cloudxxx8 (changes requested):

2018yuli:
Hi,
This issue is related to how Go's embed package works. On Windows, when you use go:embed, the file paths are normalized to use / as the separator, rather than \, which is the native separator on Windows.
For example, given the following directory structure:
/testdata/
    file1.txt

If we embed this directory in a Go program, the paths will be represented as /testdata/file1.txt instead of \testdata\file1.txt on Windows. This is due to Go's behavior with embed paths, which always use / as the path separator, regardless of the operating system.
To illustrate this, here's a test case that ensures the paths behave as expected:
// main.go
package main

import (
      "embed"
      "fmt"
      "github.com/stretchr/testify/assert"
      "io/fs"
      "testing"
)

//go:embed testdata/*
var content embed.FS

func TestEmbedPaths(t *testing.T) {
      expectedPaths := []string{
            "testdata",
            "testdata/file1.txt",
      }

      unexpectedPaths := []string{
            "testdata\\file1.txt",
      }

      // print all files in the "testdata" directory
      err := fs.WalkDir(content, "testdata", func(path string, d fs.DirEntry, err error) error {
            if err != nil {
                  return err
            }

            // print the path of each file in the "testdata" directory
            fmt.Println("Embedded file:", path)

            assert.Contains(t, expectedPaths, path, "path %s not found in expected paths", path)
            assert.NotContains(t, unexpectedPaths, path, "path %s found in unexpected paths", path)

            return nil
      })
      if err != nil {
            t.Fatal(err)
      }
}

=== RUN   TestEmbedPaths
Embedded file: testdata
Embedded file: testdata/file1.txt
--- PASS: TestEmbedPaths (0.00s)
PASS

Would it be helpful to add a comment in the code to clarify this behavior, or would you prefer adding a test case to ensure that paths are properly validated across different platforms?
I tested on my Windows machine, using filepath.ToSlash to normalize paths to / should be fine.

________________________________
发件人: Cloud Tsai ***@***.***>
发送时间: 2025年5月21日 13:59
收件人: edgexfoundry/edgex-go ***@***.***>
抄送: 2018yuli ***@***.***>; Author ***@***.***>
主题: Re: [edgexfoundry/edgex-go] fix: Windows Path Separator Compatibility Issue in getSortedSqlFileNames Function (#5150) (PR #5151)


@cloudxxx8 requested changes on this pull request.

________________________________

In internal/pkg/db/postgres/utils.go<https://github.com/edgexfoundry/edgex-go/pull/5151#discussion_r2099421675>:

> @@ -144,3 +144,8 @@ func WrapDBError(message string, err error) errors.EdgeX {
        }
        return errors.NewCommonEdgeX(errors.KindDatabaseError, message, err)
 }
+
+// NormalizePath normalizes file paths to be platform-independent.
+func NormalizePath(baseDir, fileName string) string {
+    return filepath.ToSlash(filepath.Join(baseDir, fileName))


according to the go docs, filepath.Join should use the OS separator.
https://pkg.go.dev/path/filepath#Join

Windows is \ and Linux is /. If you wrap it with filepath.ToSlash, will it become always /?
Could this really solve the issue?

cloudxxx8:
Yes, adding unit test could be very helpful.

Also, 

1. Please squash your commits with sign-off.  DCO can be passed if there is any commits in this PR is not valid.
2. Add an empty line to the end of the file.
3. Run go fmt before committing the code.  The Jenkins check is failed due to the gofmt check.

```
[2025-05-29T08:32:56.223Z] gofmt -l $(find . -type f -name '*.go'| grep -v "/vendor/")
[2025-05-29T08:32:56.223Z] ./internal/pkg/db/postgres/utils.go
[2025-05-29T08:32:56.223Z] [ "`gofmt -l $(find . -type f -name '*.go'| grep -v "/vendor/")`" = "" ]
```

cloudxxx8:
@2018yuli you need to rebase the main branch to resolve the conflict
I just opened another PR based on your change.  Could you review it?
https://github.com/edgexfoundry/edgex-go/pull/5170

2018yuli:
Since cloudxxx8 has resolved the conflict and made the changes in PR 5170, I believe this PR can be closed. Thank you!

PR #5161, opened 2025-05-28
docs: Correct schema definition for discoveredDevice

yichun-chou:
- **Fixed schema definition for discoveredDevice**
  The original schema incorrectly used `items` under `type: object`, which is only valid for `arrays` according to [OpenAPI spec](https://swagger.io/docs/specification/v3_0/data-models/data-types/?sbsearch=Items#arrays).
  Replaced with `allOf` to properly reference the `DiscoveredDevice` schema.
- **Corrected `ProvisionWatchersResponse` and `MultiProvisionWatchersResponse` property name to camelCase**
  Confirmed that both `/provisionwatcher/all` and `/provisionwatcher/name/{name}` endpoints return responses with camelCase property names.
  image
  image

## PR Checklist

  <link to docs PR>

### DIFF
--- a/openapi/core-metadata.yaml
+++ b/openapi/core-metadata.yaml
@@ -669,8 +669,8 @@ components:
             - UNLOCKED
         discoveredDevice:
           type: object
-          items:
-            $ref: '#/components/schemas/DiscoveredDevice'
+          allof:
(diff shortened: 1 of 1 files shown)


cloudxxx8 (approved):
LGTM

Issue #5242, opened 2025-09-01
Enhance Reading API to add a new aggregateFunc query parameter


lindseysimple:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]

 

This feature request is for Core Data service.

### Description [**REQUIRED**]

The current Reading APIs support only two aggregation operations: count (all readings) and countByDeviceName. However, users increasingly require more advanced aggregation capabilities on reading data. At present, clients must fetch all raw readings and perform these calculations themselves, which results in significant client-side processing and resource overhead.

#### Add a new aggregateFunc query parameters to the readings API

The `aggregateFunc` parameter allows clients to request aggregated readings directly from the API, using common SQL aggregate functions applied to numeric fields. When specified, the query returns aggregated values grouped by device name, resource name, and value type.

Supported values:

- MIN – Returns the smallest value within the selected column.
- MAX – Returns the largest value within the selected column.
- COUNT – Returns the number of rows in the result set.
- SUM – Returns the total sum of all values in a numeric column.
- AVG – Returns the average of all values in a numeric column.

Notes:

- Aggregation can only be performed on fields stored as numeric data types. For non-numeric value types (e.g., String, Bool), the aggregate result will be null.
- When `aggregateFunc` is defined, the `offset` and `limit` query parameters are ignored.

cloudxxx8:
RedisDB persistence won't support this feature

PR #5247, opened 2025-09-08
fix: Add offset/limit clauses to aggregate readings

lindseysimple:
Relates to #5242. Add offset/limit clauses when querying aggregate readings. Use numeric value for aggregated readings.

## PR Checklist

  <link to docs PR>


internal/core/data/application/readingagg.go
```diff
@@ -111,5 +113,23 @@ func getReadingAggregation(dic *di.Container, aggReadingDBFunc func(interfaces.D
...
+	if err != nil {
+		return readings, errors.NewCommonEdgeXWrapper(err)
+	}
+
+	convertAggregateReadingsToNumeric(readings)
+	return readings, nil
+}
+
+// convertAggregateReadingsToNumeric iterates the BaseReading DTOs and normalizes their Value field into a numeric representation
+func convertAggregateReadingsToNumeric(readings []dtos.BaseReading) {
```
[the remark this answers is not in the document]
cloudxxx8: let the aggregated value directly go into the numeric reading model and then numeric reading dto without any conversion
https://github.com/edgexfoundry/edgex-go/blob/5c25be3e1a794386de6cc01739111128939cdde3/internal/pkg/infrastructure/postgres/models/reading.go#L39

openapi/core-data.yaml
```diff
@@ -562,17 +562,17 @@ components:
...
             profileName: "Random-Float-Device"
             valueType: "Float32"
-            value: "112.2"
+            value: 112.2
           - deviceName: "Random-String-Device"
             resourceName: "String"
             profileName: "Random-String-Device"
             valueType: "String"
-            value: null
+            value: ""
```
lindseysimple, line 575: The aggregate value is changed to empty instead of null for non-numeric value types after using `NumericReading` model.

internal/pkg/infrastructure/postgres/readingagg.go
```diff
@@ -100,14 +151,98 @@ func (c *Client) ReadingsAggregationByDeviceNameAndResourceName(deviceName strin
...
+
+		// convert the BaseReading fields to BaseReading struct defined in contract
+		baseReading := readingDBModel.GetBaseReading()
+
+		reading := model.NumericReading{
+			BaseReading: baseReading,
+		}
+		if readingDBModel.NumericValue != nil {
+			// reading type is NumericReading
+			val, err := parseAggNumericReading(readingDBModel.ValueType, aggregateFunc, readingDBModel.NumericValue)
```
cloudxxx8, line 194: modify the valueType from the reading model as well

cloudxxx8 (changes requested):

internal/pkg/infrastructure/postgres/readingagg.go
lindseysimple: Done

internal/pkg/infrastructure/postgres/readingagg.go
```diff
@@ -100,14 +151,104 @@ func (c *Client) ReadingsAggregationByDeviceNameAndResourceName(deviceName strin
...
+	case common.AvgFunc:
+		// Always convert to Float64 for AVG function
+		val, err := numericValue.Float64Value()
+		if err != nil {
+			return pgClient.WrapDBError("failed to parse numeric data type to float64", err)
+		}
+		reading.ValueType = common.ValueTypeFloat64
+		reading.NumericValue = val.Float64
+	case common.SumFunc, common.MinFunc, common.MaxFunc:
+		if valueType == common.ValueTypeFloat64 || valueType == common.ValueTypeFloat32 {
```
cloudxxx8, line 229: use switch case instead of `if` for the valueType determination here

cloudxxx8 (changes requested):

cloudxxx8 (approved):
LGTM

Issue #5255, opened 2025-09-16
Refactor Core Data application layer code


lindseysimple:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]

 

This feature request is for Core Data code refactoring.

### Description [**REQUIRED**]
 A clear and concise description of the problem or missing capability...

1. Synchronized the  Core Data  application layer code to refactor the functions to be the [struct](https://github.com/edgexfoundry/edgex-go/blob/main/internal/core/data/application/app.go#L34-L40) methods of `CoreDataApp`.

```
// CoreDataApp encapsulates the Core Data Application functionality
// TODO: Extend this App usage beyond Events.
type CoreDataApp struct {
	lc                       logger.LoggingClient
	eventsPersistedCounter   gometrics.Counter
	readingsPersistedCounter gometrics.Counter
}
```

2. Currently the `application/reading.go` file has grown too large and difficult to maintain, extract the event purging related code into a separate file.

### Describe the solution you'd like

Same as described in the **_Description_** section.

PR #5256, opened 2025-09-16
fix: Refactor core data app layer code

lindseysimple:
Relates to #5255.
- Extract event purging code from reading.go
- Move device info cache to app layer
- Refactor reading functions to CoreDataApp methods.

## PR Checklist

  <link to docs PR>


internal/pkg/infrastructure/redis/event.go
```diff
@@ -12,6 +12,8 @@ import (
 	"time"
 
 	pkgCommon "github.com/edgexfoundry/edgex-go/internal/pkg/common"
+	pgModels "github.com/edgexfoundry/edgex-go/internal/pkg/infrastructure/postgres/models"
```
cloudxxx8, line 15: move `internal/pkg/infrastructure/postgres/models` to `internal/pkg/infrastructure/models`, because the DeviceInfo and Reading models could be opened to other database implementation.
Also, it doesn't look right to put the postgres specific model to DBClient interface.

internal/core/data/application/eventpurge.go
```diff
@@ -0,0 +1,308 @@
...
+	pgModels "github.com/edgexfoundry/edgex-go/internal/pkg/infrastructure/postgres/models"
+
+	bootstrapContainer "github.com/edgexfoundry/go-mod-bootstrap/v4/bootstrap/container"
+	"github.com/edgexfoundry/go-mod-bootstrap/v4/di"
+	"github.com/edgexfoundry/go-mod-core-contracts/v4/clients/logger"
+	"github.com/edgexfoundry/go-mod-core-contracts/v4/common"
+	"github.com/edgexfoundry/go-mod-core-contracts/v4/errors"
+)
+
+var asyncPurgeEventOnce sync.Once
```
cloudxxx8, line 25: since CoreDataApp is used by all the application layer, move the global variable to the CoreDataApp struct

lindseysimple: Done

Issue #5289, opened 2025-10-20
Improve insecure encryption key handling


FelixTing:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]

 This feature request is for proxy-auth

### Description [**REQUIRED**]
https://github.com/edgexfoundry/edgex-go/blob/676951d955e4c4c86fc34e64f6f39dfe192b487c/internal/pkg/utils/crypto/aes.go#L24
The default AES encryption key is hardcoded in the source code

### Describe the solution you'd like
Move the default AES key to a plaintext file and allow overriding via environment variable

PR #5290, opened 2025-10-20
feat: Improve insecure encryption key handling

FelixTing:
Move hardcoded encryption key to a file and allow overriding via environment variable

fix: #5289 

## PR Checklist

  <link to docs PR>

### DIFF
--- a/internal/security/proxyauth/cryptor.go
+++ b/internal/security/proxyauth/cryptor.go
@@ -6,7 +6,9 @@
 package proxyauth
 
 import (
+	"bytes"
 	"context"
+	"crypto/aes"
 	"os"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


internal/security/proxyauth/cryptor.go
```diff
@@ -15,12 +19,43 @@ import (
...
+
+func loadDefaultAESKey() ([]byte, error) {
+	if envKey := getenv(EnvAesKey); envKey != "" {
+		return []byte(envKey), nil
+	}
+
+	if keyData, err := readFile(defaultAESKeyFile); err == nil {
+		key := []byte(string(keyData))
+		for len(key) > 0 && (key[len(key)-1] == '\n' || key[len(key)-1] == '\r' || key[len(key)-1] == ' ') {
+			key = key[:len(key)-1]
```
judehung: You can use https://pkg.go.dev/bytes#TrimRight or https://pkg.go.dev/bytes#TrimSpace to robustly handle all types of whitespace.

PR #5395, opened 2026-04-16
core-command: prevent cascading MQTT command timeouts

### LINKED ISSUE #5394: core-command: Cascading MQTT Command Timeouts After Downstream Timeout
# 🐞 Bug Report

 The issue is located in: `core-command`

### Is this a regression?
No. This behavior is reproducible in earlier releases as well (not newly introduced in v4.0).

I am seeing a recurring failure pattern in `core-command` when handling external MQTT command requests:

1. System works normally at first.
2. One internal request times out.
3. Subsequent requests also begin timing out in sequence.
4. MQTT command/status reads become effectively unusable until the timeout wave clears.

From code-path analysis, this appears to be head-of-line blocking in the external MQTT handler path, where a synchronous request/reply wait is done inside the message callback.

#### Minimal Reproduction Steps
1. Send recurring external MQTT command requests, for example:
   - `edgex/command/request/PowerMeter-floor1/PowerReadings/get`
   - `edgex/command/request/device-group-classroom-1f/VRF-Controller-floor1/ACStatus/get`
   - `edgex/command/request/AirSensor-floor1/AirQualityReadings/get`
2. Make one downstream target intermittently non-responsive (in my case, `device-power-1f` path).
3. Observe:
   - first timeout,
   - then repeated timeouts for subsequent requests,
   - degraded MQTT command/status behavior.

### Expected Behavior
A timeout from one downstream device/service request should fail only that request. Other independent MQTT command requests should continue to be handled normally.

### Actual Behavior
After one timeout, subsequent requests are impacted and also time out repeatedly, causing command/status operations over MQTT to degrade significantly.

### Additional Context / Analysis
Potential blocking point in `core-command`:

```go
response, err := internalMessageBus.Request(requestEnvelope, deviceRequestTopic, deviceResponseTopicPrefix, requestTimeout)
```

Location:
- `internal/core/command/controller/messaging/external.go`

This call is synchronous.

`Request()` resolves to:
- `vendor/github.com/edgexfoundry/go-mod-messaging/v4/internal/pkg/request.go` (`DoRequest`)

Flow there:
1. Subscribe to `<responseTopicPrefix>/<request-id>`
2. Publish request
3. Block waiting for `response | error | timeout`
4. Unsubscribe

Timeout generated at:

```go
case <-timer.C:
    return nil, fmt.Errorf("timed out waiting for response on %s topic", responseTopicChan.Topic)
```

This blocking request/reply wait in the external MQTT callback path appears to create head-of-line blocking under partial downstream failure. One slow request can delay subsequent requests and trigger chained timeouts/backlog behavior.

## 🔥 Exception or Error
<pre><code>
level=DEBUG ts=2026-04-16T04:46:45.484743906Z app=core-command source=external.go:99 msg="Received command request from external message broker on topic 'edgex/command/request/PowerMeter-floor1/PowerReadings/get' with 354 bytes"
level=DEBUG ts=2026-04-16T04:46:45.488002302Z app=core-command source=external.go:164 msg="Sending Command request to internal MessageBus. Topic: edgex/device/command/request/device-power-1f/PowerMeter-floor1/PowerReadings/get, Request-id: aa0f2204-4fc0-45dd-b4ca-f3f436f606d1 Correlation-id: 1542f011-b1f8-47fe-a0dd-a856d8df61f2"
level=DEBUG ts=2026-04-16T04:46:45.488214912Z app=core-command source=external.go:165 msg="Expecting response on topic: edgex/response/device-power-1f/aa0f2204-4fc0-45dd-b4ca-f3f436f606d1"
level=ERROR ts=2026-04-16T04:47:15.491943317Z app=core-command source=external.go:187 msg="Failed to send DeviceCommand request with internal MessageBus: timed out waiting for response on edgex/response/device-power-1f/aa0f2204-4fc0-45dd-b4ca-f3f436f606d1 topic"
level=DEBUG ts=2026-04-16T04:47:15.492317693Z app=core-command source=external.go:195 msg="Published response message to external message broker on topic 'edgex/command/response/PowerMeter-floor1/PowerReadings/get' with 368 bytes"

(After first timeout, the same pattern repeats for subsequent requests with new request IDs.)
</code></pre>

## 🌍  Your Environment
**Deployment Environment:** `k3s` (Kubernetes)

**EdgeX Version [**REQUIRED**]:** `v4.0 (Odessa)`

**Anything else relevant?**
- MessageBus: `MQTT`
- OS: `Linux`
- Impact: cascading timeouts under partial downstream failure; unrelated command requests are affected until timeout wave clears.

### Request
Could maintainers confirm whether this callback-path blocking is expected in current design, and whether external handler refactoring should be considered (for example bounded worker dispatch / fail-fast behavior) so one slow downstream target does not degrade unrelated MQTT command requests?

utafrali:
The external message handler was blocking on internalMessageBus.Request() when downstream services timed out. This caused other MQTT messages to queue up and timeout as well, creating cascading failures. Moved the request handling to a goroutine so the handler returns immediately. Fixes #5394.

### DIFF
--- a/internal/core/command/controller/messaging/external.go
+++ b/internal/core/command/controller/messaging/external.go
@@ -166,19 +166,20 @@ func commandRequestHandler(requestTimeout time.Duration, dic *di.Container) mqtt
 
 		internalMessageBus := bootstrapContainer.MessagingClientFrom(dic.Get)
 
-		// Request waits for the response and returns it.
-		response, err := internalMessageBus.Request(requestEnvelope, deviceRequestTopic, deviceResponseTopicPrefix, requestTimeout)
-		if err != nil {
-			errorMessage := fmt.Sprintf("Failed to send DeviceCommand request with internal MessageBus: %v", err)
-			responseEnvelope := types.NewMessageEnvelopeWithError(requestEnvelope.RequestID, errorMessage)
-			publishMessage(client, externalResponseTopic, qos, retain, responseEnvelope, lc)
-			return
-		}
+		go func() {
+			response, err := internalMessageBus.Request(requestEnvelope, deviceRequestTopic, deviceResponseTopicPrefix, requestTimeout)
+			if err != nil {
+				errorMessage := fmt.Sprintf("Failed to send DeviceCommand request with internal MessageBus: %v", err)
+				responseEnvelope := types.NewMessageEnvelopeWithError(requestEnvelope.RequestID, errorMessage)
+				publishMessage(client, externalResponseTopic, qos, retain, responseEnvelope, lc)
+				return
+			}
 
-		lc.Debugf("Command response received from internal MessageBus. Topic: %s, Request-id: %s Correlation-id: %s", response.ReceivedTopic, response.RequestID, response.CorrelationID)
+			lc.Debugf("Command response received from internal MessageBus. Topic: %s, Request-id: %s Correlation-id: %s", response.ReceivedTopic, response.RequestID, response.CorrelationID)
 
-		response.ReceivedTopic = externalResponseTopic
-		publishMessage(client, externalResponseTopic, qos, retain, *response, lc)
+			response.ReceivedTopic = externalResponseTopic
+			publishMessage(client, externalResponseTopic, qos, retain, *response, lc)
+		}()
 	}
 }
 
--- a/internal/core/command/controller/messaging/external_test.go
+++ b/internal/core/command/controller/messaging/external_test.go
@@ -369,6 +369,9 @@ func Test_commandRequestHandler(t *testing.T) {
 
 			expectedInternalRequestTopic := common.BuildTopic(baseTopic, common.CoreCommandDeviceRequestPublishTopic, testDeviceServiceName, testDeviceName, testCommandName, testMethod)
 			expectedInternalResponseTopicPrefix := common.BuildTopic(baseTopic, common.ResponseTopic, testDeviceServiceName)
+			require.Eventually(t, func() bool {
+				return mqttClient.AssertNumberOfCalls(new(testing.T), "Publish", 1)
+			}, time.Second, time.Millisecond*10)
 			client.AssertCalled(t, "Request", tt.payload, expectedInternalRequestTopic, expectedInternalResponseTopicPrefix, mock.Anything)
 		})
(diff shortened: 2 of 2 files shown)


qs-wang:
We actually reported the same against v3, and had to create a fork to make a patched version. Would be good to have this fixed in V4 ASAP. This is really a performance blocker.

utafrali:
Yeah, definitely understand - this is a real blocker. The fix should be solid for v4.

qs-wang:
@Yangsta911 is working on a similar fix, but a little bit different.  This is from our experiences of real projects.

He will be creating a PR shortly.  I'd suggest to review both.  Thanks!

utafrali:
Sounds good, the maintainers can review both and pick the better approach.

judehung (commented):
Thanks @ugurtafrali — you identified the root cause correctly: the Paho callback was being held by the synchronous `internalMessageBus.Request`, serializing every subsequent external command behind a slow or offline device. Moving the call into a goroutine is the right starting point.

My one concern is that the goroutine is unbounded. Each inbound MQTT message spawns a new goroutine that lives for up to `requestTimeout`(default 30 s). In exactly the scenario this PR is trying to fix — devices timing out — goroutines accumulate at the rate the broker delivers messages, which under a burst (broker reconnect replay, a misbehaving publisher, etc.) can grow without limit. The original failure mode (cascading timeouts) gets replaced with a worse one (goroutine / memory exhaustion), and there's no backpressure signal to the client.

[#5428](https://github.com/edgexfoundry/edgex-go/pull/5428)  takes your approach and adds a bounded semaphore in front of the goroutine spawn:
  - Same fix for HOL blocking (callback returns immediately).
  - A configurable cap on in-flight requests (default 32).
  - At capacity, a "service busy" envelope is published inline so the
    client knows the request was shed rather than silently queued.

It's ~40 LOC and keeps the same simplicity as your patch — just with a ceiling so the failure mode is bounded.

Could you take a look at [#5428](https://github.com/edgexfoundry/edgex-go/pull/5428) and let me know whether it solves the issue as you've reproduced it?

cloudxxx8:
resolved by https://github.com/edgexfoundry/edgex-go/pull/5428

PR #5396, opened 2026-04-17
feat: Introduce external command queue configuration for MQTT commands

### LINKED ISSUE #5394: core-command: Cascading MQTT Command Timeouts After Downstream Timeout

Yangsta911:
Added a new configuration section for managing external command queue parameters, including maximum concurrent commands, queued commands, overload publish channel capacity, and shutdown timeout. Updated related code to utilize these settings in the MQTT command processing workflow.

- Updated configuration structure to include ExternalCommandQueue.
- Modified OnConnectHandler to initialize the external command processor with new limits.
- Enhanced command request handling to support non-blocking enqueueing of external commands.

This change aims to improve the handling of external MQTT commands by allowing better concurrency management and reducing potential blocking issues.

## PR Checklist

- [x] I have fully tested this new feature or bug fix (if not, why?)

## Testing Instructions
1. Configure small `ExternalCommandQueue` limits.
2. Send external MQTT command bursts above capacity.
3. Verify concurrency/queue stay within configured bounds.
4. Verify enqueue remains non-blocking when saturated.
5. Verify shutdown completes within configured timeout.

Closes #5394

### DIFF
--- a/internal/core/command/controller/messaging/external.go
+++ b/internal/core/command/controller/messaging/external.go
@@ -7,10 +7,12 @@
 package messaging
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 
 	mqtt "github.com/eclipse/paho.mqtt.golang"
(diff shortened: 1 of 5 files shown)


qs-wang:
Fixes https://github.com/edgexfoundry/edgex-go/issues/5394.

internal/core/command/controller/messaging/external.go
```diff
@@ -92,32 +321,34 @@ func commandQueryHandler(dic *di.Container) mqtt.MessageHandler {
...
+		cfg := container.ConfigurationFrom(p.dic.Get)
 		lc.Debugf("Received command request from external message broker on topic '%s' with %d bytes", message.Topic(), len(message.Payload()))
 
-		externalMQTTInfo := container.ConfigurationFrom(dic.Get).ExternalMQTT
+		externalMQTTInfo := cfg.ExternalMQTT
 		qos := externalMQTTInfo.QoS
 		retain := externalMQTTInfo.Retain
 
-		requestEnvelope, err := types.NewMessageEnvelopeFromJSON(message.Payload())
+		payload := append([]byte(nil), message.Payload()...)
```
judehung: What's the purpose of this append operation?

judehung (commented):
Thanks for taking this on, @Yangsta911 — the architectural direction here (bounded concurrency, non-blocking enqueue, backpressure response, graceful shutdown) is the right shape for #5394. After reviewing the implementation in detail, however, I think this PR carries more machinery than the problem warrants and has a few correctness issues that would need to be resolved before it could land. Summarizing the main concerns:

1. **Duplicate metadata lookup.** `commandRequestMQTTHandler` calls    `retrieveServiceNameByDevice` for validation and discards the result, then `processExternalMQTTCommand` calls it again to use the value. Since that helper makes two Metadata HTTP calls, every external command now incurs 4 Metadata calls instead of 2, and the two calls can return different results if Metadata changes between them.

2.**Send-on-closed-channel race at shutdown.** `shutdownWorkerPool` closes   `errPubCh` after `workerWg.Wait()`, but the handler path (`tryEnqueue` → `enqueueOverloadPublish`) is not gated by `jobsClosed` and can send on `errPubCh` after it's been closed, panicking the Paho callback goroutine.

3. **Validation regression.** `validateGetCommandQueryParameters` moved from the handler into the worker. When the queue is full, a client sending an invalid `ds-returnevent` value now receives "service busy" instead of the specific validation error. Validation is cheap and in-memory — it should stay in the handler.

4. **Silent drop on unescape error in the worker.** The worker re-parses the topic and silently returns on `url.PathUnescape` failure, even though the request has already been admitted to the queue and the caller is waiting on `externalResponseTopic`.

This PR adds ~600 LOC across 5 files: a worker pool struct, a separate overload-publisher goroutine, a shutdown watcher, four config fields, and lifecycle plumbing. The root cause — Paho's per-subscription callback serialization being held by a synchronous `internalMessageBus.Request` — can be addressed with a bounded semaphore and a goroutine in ~40 LOC, keeping the same correctness properties:

- Bounded concurrency: semaphore capacity
- Non-blocking enqueue: non-blocking `select` on the semaphore
- Backpressure response: inline busy envelope on overload
- Reconnect-safe: semaphore captured in the closure

The smaller surface area also avoids the lifecycle/state hazards above because there's no queue to close and no separate publisher to coordinate.

My preference is to go with the simpler approach proposed in [#5428](https://github.com/edgexfoundry/edgex-go/pull/5428) for fixing [#5394](https://github.com/edgexfoundry/edgex-go/issues/5394). Before we decide, though, could you please review [#5428](https://github.com/edgexfoundry/edgex-go/pull/5428) and confirm whether it actually resolves the issue as you've observed it in your real project? You have the load profile and reproduction context, so your validation would be more meaningful than ours. If [#5428](https://github.com/edgexfoundry/edgex-go/pull/5428) does solve the problem for you, I'd like to land that one and revisit the worker pool / dedicated publisher machinery only if measured evidence (load tests, ops feedback) later shows it's needed.

Thanks again for the work and the write-up — the analysis here was genuinely useful in clarifying what the minimum viable fix looks like.

cloudxxx8:
resolved by https://github.com/edgexfoundry/edgex-go/pull/5428

PR #5446, opened 2026-06-29
test: align min interval error expectation with parser

Old-Ding:
## Summary
- remove the manually constructed parse-duration error from `TestCheckMinInterval`
- derive the expected error from `time.ParseDuration`, matching the production path

## Why
The test expected a generic `errors.errorString`, while `CheckMinInterval` logs the concrete error returned by `time.ParseDuration`. On current Go this is a `*time.parseDurationError`, so the mock assertion fails even though the production behavior is correct.

## Testing
- `go test ./internal/pkg/utils -run TestCheckMinInterval -count=1`
- `go test ./internal/pkg/utils -count=1`

### DIFF
--- a/internal/pkg/utils/time_test.go
+++ b/internal/pkg/utils/time_test.go
@@ -6,11 +6,11 @@
 package utils
 
 import (
-	"errors"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cloudxxx8 (approved):
LGTM

PR #5447, opened 2026-06-29
fix(security): handle Windows file URLs in waitFor

Old-Ding:
## Summary
- normalize `file:` URL paths before passing them to `os.Stat`
- build a valid platform file URI in the waitFor test

## Why
On Windows, a valid `file:///D:/...` URL parses to a path with a leading slash. Passing that path directly to `os.Stat` fails, so waitFor could not detect existing files by URI on Windows.

## Testing
- `go test ./internal/security/bootstrapper/command/waitfor -run TestExecute -count=1`
- `go test ./internal/security/bootstrapper/command/waitfor -count=1`
- `go test ./internal/security/bootstrapper/command -count=1`

### DIFF
--- a/internal/security/bootstrapper/command/waitfor/command.go
+++ b/internal/security/bootstrapper/command/waitfor/command.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"runtime"
 	"strings"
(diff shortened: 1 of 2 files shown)


cloudxxx8 (approved):
LGTM

PR #5462, opened 2026-09-02
feat: Support existenceOnly and basicInfoOnly on device profile query

### LINKED ISSUE #5461: Support existenceOnly and basicInfoOnly on device profile query
# 🚀 Feature Request

 This feature request is for core-metadata

`GET /deviceprofile/name/{name}` always returns the whole device profile,
including every `deviceResources` and `deviceCommands` entry. Callers that only
need to know whether a name is taken, or only need the profile's basic
attributes, pay the full payload cost.

This is expensive for real profiles: a BACnet or OPC-UA profile
can carry hundreds of resources, so the response is orders of magnitude larger
than the information the caller actually wants.

Two optional boolean query parameters on the existing
`GET /deviceprofile/name/{name}` route:

- `existenceOnly=true` — respond `200` with just a `BaseResponse` (no `profile`
  field) when the profile exists, `404` when it does not. Core-metadata can
  answer this without reading the profile row content at all.
- `basicInfoOnly=true` — respond with the profile's basic info only, omitting
  `deviceResources` and `deviceCommands`.

`existenceOnly` takes precedence when both are supplied. A profile that does
not exist is still reported as `404` in every mode, so error handling is
unchanged.

weichou1229:
- Trim the GET deviceprofile/name/{name} response according to the query params
- Add CheckDeviceProfileExistsByName answering existence without reading the profile
- Update the core-metadata OpenAPI document
- Bump go-mod-core-contracts to v4.1.0-dev.46

Close #5461

## PR Checklist

- [x] I have opened a PR for the related docs change (if not, why?) update API doc
  <link to docs PR>


cloudxxx8 (approved):
LGTM

PR #5464, opened 2026-09-02
feat: Support deleting device resource/command tags

### LINKED ISSUE #5463: Support deleting device resource/command tags
# 🚀 Feature Request

 This feature request is for core-meadata

`PATCH /deviceprofile/name/{name}/tags` can only add or update tags on device
resources and device commands. There is currently no way to remove one.

`mergeTags` in `dtos/requests/deviceprofiletags.go` is a purely additive recursive
merge, so a `null` value is stored as a literal JSON null under the key instead of
removing it:

Request:
```json
{
  "apiVersion": "v3",
  "deviceResources": [{ "name": "Float32", "tags": { "tag1": null } }]
}
```
Result: "tags": { "tag1": null } — the key remains.

Treat a null tag value as a deletion of that key and needs no DTO or API shape change, since Tags is already map[string]any.

Suggested change in mergeTags, before the existing type assertions (a nil
value fails the assertion today and falls through to the plain overwrite):

if value == nil {
    delete(dest, key)
    continue
}

 Have you considered any alternative solutions or workarounds?

weichou1229:
- Document the null tag deletion semantics in the OpenAPI spec
- Add a controller test asserting the merged tags after a deletion

Close #5463

## PR Checklist

- [x] I have opened a PR for the related docs change (if not, why?) Update API doc
  <link to docs PR>


cloudxxx8 (approved):
LGTM

